PR #3, opened 2019-11-04
IconButton split focus working and test covered

laraharrow:


packages/components/src/Button/IconButton.tsx
```diff
@@ -102,29 +113,68 @@ export const IconButtonStyle = styled.button<IconButtonProps>`
```
mdodgelooker: I think you can get rid of `IconButtonStyle` now.

packages/components/src/Button/IconButton.tsx
```diff
@@ -102,29 +113,68 @@ export const IconButtonStyle = styled.button<IconButtonProps>`
...
   ${reset}
   ${space}
 
   padding: 3px;
 
-  ${iconButtonVariant}
+  ${props => props.outline && outlineCSS}
+
   ${pseudoClasses}
   ${({ shape }) => shape === 'round' && 'border-radius: 100%;'}
```
dbchristopher: @laraharrow looks like these changes broke the `round` button variant we're using for the sidebar toggle. We should probably always make sure that the width and height match for IconButton. 

Screen Shot 2019-11-06 at 3 31 01 PM

Also do you think we should still have a white background by default? Or should the background be an optional prop?

packages/components/src/Form/Inputs/InputSearch/InputSearch.tsx
```diff
@@ -53,6 +53,8 @@ export interface InputSearchProps
    */
   summary?: string
 
+  onClear?: () => void
+
   value?: string
```
mdodgelooker: I think you might've created this branch based on your other branch `lara/search-onClear/lens351` so it has the `InputSearch` changes in it. Since this one may get merged first, you'll should probably rebase onto master:
`git pull --rebase origin master`
`git push --force origin lara/iconbutton-focus`

packages/components/src/Button/IconButton.test.tsx
```diff
@@ -57,3 +60,28 @@ test('IconButton accepts events', () => {
   )
   assertSnapshot(<IconButton label="Test" icon="Favorite" onClick={noop} />)
 })
+
+test('Button Focus split - on tab not on click', () => {
```
dbchristopher: that thet description could probably be a little more clearer for future maintenance. Maybe this: "Button renders focus ring on tab input but not on click"

dbchristopher:
This is a question for @lukelooker or maybe the product manager that requested this change -- 

Looks like this only hides the box shadow, but still renders other focus styles when clicked on. Is that going to be sufficient? Or is the goal to have it look more neutral/default after moving your mouse away from it?

click-examples

packages/components/src/Button/IconButton.test.tsx
```diff
@@ -57,3 +60,28 @@ test('IconButton accepts events', () => {
...
+    <ThemeProvider theme={theme}>
+      <>
+        <IconButton label="Favorite" color="danger" icon="Favorite" />
+        <IconButton label="Trash" color="danger" icon="Trash" />
+      </>
+    </ThemeProvider>
+  )
+
+  fireEvent.click(getByTitle('Favorite'))
+  expect(getByTitle('Favorite')).not.toHaveStyle(`box-shadow: none)`)
```
dbchristopher: I'm not sure this assertion is actually doing anything? `box-shadow: none)` has that syntax error (extra parenthesis) for one, so it would never actually match that style. But also `box-shadow: none` actually would be a valid style after clicking, right? Sometimes it can be helpful to validate before and after states  for stuff like this, just to make sure what you're testing for is actually having a tangible effect.

That being said, a snapshot test would probably be better anyways to capture the specific desired style settings

packages/components/src/Button/IconButton.tsx
lukelooker: I'd actually like to preserve the change in this PR and make IconButton's background transparent and mitigate specifically for this case as I think it's fairly unusual.

lukelooker:
This PR has a bunch of extra commits. I'd recommend creating a new branch and just cherry-picking the 1 relevant commit over to it 

> This is a question for @lukelooker or maybe the product manager that requested this change --
> 
> Looks like this only hides the box shadow, but still renders other focus styles when clicked on. Is that going to be sufficient? Or is the goal to have it look more neutral/default after moving your mouse away from it?
> 
> click-examples

That's an interesting point – I think we should keep this PR focused bringing parity to the `visibleFocus` behavior for both `IconButton` & `Button` (preserving the existing `:focus` background color behavior. At least for now I'd like to keep the rest of the focus behavior to offer as much affordance as possible – @jhardy perhaps you want to ponder whether we should pull back further on the `transparent` Button's affordance? (maybe pulling off the background behaviors and just sticking to the color change – more inline with a link)

packages/components/src/Button/IconButton.tsx
dbchristopher: ok so accepting a background prop then? sounds good. @laraharrow can you also update the styleguide toggle button to reflect the new intended use here?

lukelooker: I'd actually say that case should be handled by extension (I actually did this when I . E.g.:

```
const SpecialButton = styled(IconButton)`
  background: ${props => props.theme.colors.palette.white};
`
```

PR #6, opened 2019-11-05
update onClear and tests

laraharrow:


.vscode/settings.json
```diff
@@ -23,6 +23,7 @@
     "labelledby",
     "nomarker",
     "polarbrightness",
+    "seperator",
```
mdodgelooker: Actually spell-check was right – can you update to "separator" in `InputSearchControls` and `MenuSearchControls`?

mdodgelooker:
After chatting with @laraharrow I realized I was totally wrong – `InputSearch` is already controllable, there's even an example of it in `www`. 🙂

.vscode/settings.json
lukelooker: This should get removed from this file so folks (like myself) that can't spell don't get caught by this mis-spelling.

lukelooker:
This branch has a bunch of unrelated commits in it – would be good to cut a new branch from latest master and then cherry-pick over the relevant commit:

```
git checkout master
git pull
git checkout -B new-branch-name
git cherry-pick 98ac23f53b3f87b301f18cbe450e48dff7f05006
```

PR #16, opened 2019-11-06
Fade-in sidebar toggle button

dbchristopher:
Still seeing the sidebar toggle button jump on production page loads. This PR hides the button until Javascript loads the position is finalized.


dbchristopher:
@laraharrow here's another small PR for you. hope this renders a more seamless loading experience

lukelooker:
I like this approach but as I was poking at the implementation it occurred to me that we could solve this by not requiring JS calculation to position the IconButton. `transform: translate...` syntax allows for positioning behavior based on the calculated size of the element. See PR #17

PR #18, opened 2019-11-07
Fix on prop bleeding to dom in ToggleSwitch

mdodgelooker:
`<div on={true}` causes a React warning. This solution isn't elegant but it works and I couldn't find any better inspiration in https://github.com/styled-components/styled-components/issues/135

The other props like `size` and `disabled` weren't causing breakage but I filtered those as well since they have no place on a `<div/>`.

### DIFF
--- a/packages/components/src/Form/Inputs/ToggleSwitch/ToggleSwitch.tsx
+++ b/packages/components/src/Form/Inputs/ToggleSwitch/ToggleSwitch.tsx
@@ -60,6 +60,10 @@ export interface KnobProps {
   on?: boolean
 }
 
+interface KnobContainerBaseProps extends KnobProps, PseudoProps {
+  children: ReactNode
+}
+
 export interface ToggleSwitchProps
   extends SpaceProps,
     Omit<InputProps, 'type'>,
@@ -99,14 +103,9 @@ const KnobContainer: FC<KnobProps> = ({ className, ...props }) => {
   )
 }
 
-const KnobContainerBase = styled(
-  ({
-    children,
-    className,
-  }: KnobProps & PseudoProps & { children: ReactNode }) => (
-    <div className={className}>{children}</div>
-  )
-)`
+const KnobContainerBase = styled(({ children, className }) => (
+  <div className={className}>{children}</div>
+))<KnobContainerBaseProps>`
   ${reset}
   ${pseudoClasses}
 
--- a/packages/components/src/Form/Inputs/ToggleSwitch/ToggleSwitch.tsx
+++ b/packages/components/src/Form/Inputs/ToggleSwitch/ToggleSwitch.tsx
@@ -25,7 +25,7 @@
  */
 
 import { rem, rgba } from 'polished'
-import React, { forwardRef, Ref } from 'react'
+import React, { forwardRef, Ref, ReactNode, FC } from 'react'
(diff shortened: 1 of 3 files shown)


packages/components/src/Form/Inputs/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -67,7 +67,7 @@ export interface ToggleSwitchProps
   size?: number
 }
 
-const Knob = styled.div<KnobProps>`
+const Knob = styled(({ className }) => <div className={className} />)`
```
dbchristopher: is this actually ever receiving the className? looks like on line 86 the className prop is plucked off the props and applied to `KnobContainerBase` instead

mdodgelooker: That was my initial thought too, but if I don't put that there, it breaks the styles – I guess b/c `styled` creates a new `className`.

dbchristopher:
Wait, disregard that error @mdodgelooker. Just restarted the server dev server and the error went awa

packages/components/src/Form/Inputs/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -83,26 +83,30 @@ const Knob = styled.div<KnobProps>`
...
 
-const KnobContainerBase = styled.div<KnobProps & PseudoProps>`
+const KnobContainerBase = styled(
+  ({
+    children,
+    className,
+  }: KnobProps & PseudoProps & { children: ReactNode }) => (
+    <div className={className}>{children}</div>
+  )
+)`
```
dbchristopher: I personally feel like it's cleaner to define a new TS interface rather than inlining the types. E.G.

```
interface KnobContainerBaseProps extends KnobProps, PseudoProps {
  children: ReactNode
}

const KnobContainerBase = styled(({ children, className }) => (
  <div className={className}>{children}</div>
))<KnobContainerBaseProps>`
  ...
`
```

packages/components/src/Form/Inputs/ToggleSwitch/ToggleSwitch.tsx
dbchristopher: ah interesting! yeah that makes sense

packages/components/src/Form/Inputs/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -83,26 +87,25 @@ const Knob = styled.div<KnobProps>`
...
-        size={props.size}
-        on={props.on}
-        ref={ref}
-      >
-        <Knob {...props} />
-      </KnobContainerBase>
-    )
-  }
-)
+const KnobContainer: FC<KnobProps> = ({ className, ...props }) => {
```
dbchristopher: I played with it a little more locally. I think you can revert the changes to `Knob` if you destructure the `on` prop here too. That way it won't be passed down into `<Knob {...props} />` in the first place

dbchristopher: ah, never mind! then that breaks the functionality altogether. Ok PR approved 😅

PR #20, opened 2019-11-07
IconButton and test working for split focus on tab and click

laraharrow:


packages/components/src/Button/IconButton.test.tsx
```diff
@@ -57,3 +60,27 @@ test('IconButton accepts events', () => {
...
+    <ThemeProvider theme={theme}>
+      <>
+        <IconButton label="Favorite" color="danger" icon="Favorite" />
+        <IconButton label="Trash" color="danger" icon="Trash" />
+      </>
+    </ThemeProvider>
+  )
+
+  fireEvent.click(getByTitle('Favorite'))
+  expect(getByTitle('Favorite')).not.toHaveStyle(`box-shadow: none)`)
```
dbchristopher: `box-shadow: none)` will never match a valid style because of that errant `)`

packages/components/src/Button/IconButton.test.tsx
```diff
@@ -57,3 +60,27 @@ test('IconButton accepts events', () => {
...
+  const button = getByTitle('Favorite').closest('button')
+
+  button &&
+    fireEvent.keyUp(button, {
+      charCode: 9,
+      code: 9,
+      key: 'Tab',
+    })
+
+  expect(button).toHaveStyle(`box-shadow: 0 0 0 0.15rem rgba(204,31,54,0.25)`)
```
dbchristopher: snapshot test for this

packages/components/src/Button/IconButton.tsx
```diff
@@ -52,13 +53,23 @@ export type IconButtonSizes =
...
 export interface IconButtonProps
-  extends Omit<CompatibleHTMLProps<HTMLButtonElement>, 'children' | 'type'>,
-    IconButtonVariantProps,
+  extends IconButtonFocusProps,
+    Omit<CompatibleHTMLProps<HTMLButtonElement>, 'children' | 'type'>,
     PseudoProps,
     SpaceProps {
   type?: 'button' | 'submit' | 'reset'
 
+  color?: ButtonColors
```
dbchristopher: you could probably just add the prop `focusVisible?: boolean` here, right? Is there any reason to keep `IconButtonFocusProps` as a separate interface? I'm not seeing it being referenced anywhere else

PR #22, opened 2019-11-08
field checkbox and radio align right & update tests

laraharrow:


laraharrow:
Here is a comparison with the old version on the right and the new version on the left:

Screen Shot 2019-11-08 at 5 03 53 PM

jhardy:
Overall I think this is ok, I think a bigger conversation is needed about how we style these for the variants with the label position and how the label and field width adjusts.

Ideally, if we can I'd like to update the markup rendered for this component, but it might be out of scope for this PR. This `label` here should wrap the `div` below so that the  whole area is clickable. 

image

Right now there is a few px space between the label and the checkbox that is not clickable ending up with a scenario where the checkbox can feel broken.

lukelooker:
Closing for now as @jhardy wants to take a more holistic look at this issue.

PR #28, opened 2019-11-11
Refactor z-index out of menu components

dbchristopher:
### :sparkles: Changes

- Removes use of z-index to enforce menu component stacking order. This is primarily accomplished through working around any need for `position` settings on MenuItem, so it no longer competes with the MenuGroupLabel `position: sticky;` behavior. 

- [x] PR is ideally < 400LOC

### :camera: Screenshots

sticky-header

### DIFF
--- a/packages/components/src/Menu/MenuItem/MenuItem.tsx
+++ b/packages/components/src/Menu/MenuItem/MenuItem.tsx
@@ -117,10 +117,6 @@ export interface MenuItemProps
    */
   current?: boolean
   /**
-   * Display a marker next to the MenuItem if it is current
-   */
-  currentMarker?: boolean
-  /**
    * Sets the correct accessible role for the MenuItem:
    * - Use **'link'** for items that navigation to another page
    * - Use **'button'** for items that trigger in page interactions, like displaying a modal
@@ -133,7 +129,6 @@ export interface MenuItemProps
 export const MenuItem: FC<MenuItemProps> = props => {
   const {
     current,
-    currentMarker,
     children,
     detail,
     icon,
@@ -165,7 +160,6 @@ export const MenuItem: FC<MenuItemProps> = props => {
   return (
     <MenuItemListItem
       current={current}
-      currentMarker={currentMarker}
       itemStyle={style}
       aria-current={current && 'page'}
       onClick={onClick}
--- a/packages/components/src/Menu/MenuItem/MenuItemButton.tsx
+++ b/packages/components/src/Menu/MenuItem/MenuItemButton.tsx
@@ -48,7 +48,6 @@ export const MenuItemButton = styled.button<MenuItemButtonProps>`
   font-size: inherit;
   flex: 1;
   outline: none;
-  padding-left: ${({ theme }) => theme.space.small};
   text-align: left;
   text-decoration: none;
 `
--- a/packages/components/src/Menu/MenuItem/MenuItemButton.tsx
+++ b/packages/components/src/Menu/MenuItem/MenuItemButton.tsx
@@ -48,6 +48,7 @@ export const MenuItemButton = styled.button<MenuItemButtonProps>`
   font-size: inherit;
   flex: 1;
   outline: none;
+  padding-left: ${({ theme }) => theme.space.small};
   text-align: left;
   text-decoration: none;
 `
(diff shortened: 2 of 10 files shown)


packages/components/src/Menu/MenuItem/MenuItem.tsx
```diff
@@ -116,10 +116,6 @@ export interface MenuItemProps
    * Indicates the MenuItem is checked
    */
   current?: boolean
-  /**
-   * Display a marker next to the MenuItem if it is current
-   */
-  currentMarker?: boolean
```
dbchristopher: I could not discern any behavior or style change when `currentMarker` is true. The best I can tell is that it might have been redundant with the `current` prop, so I went ahead and removed it

packages/components/src/Menu/__snapshots__/MenuGroup.test.tsx.snap
```diff
@@ -110,16 +109,18 @@ exports[`MenuGroup - JSX label 1`] = `
   -webkit-flex-wrap: wrap;
   -ms-flex-wrap: wrap;
   flex-wrap: wrap;
-  position: relative;
   -webkit-text-decoration: none;
   text-decoration: none;
   -webkit-transition: background 150ms cubic-bezier(0.86,0,0.07,1), color 150ms cubic-bezier(0.86,0,0.07,1);
   transition: background 150ms cubic-bezier(0.86,0,0.07,1), color 150ms cubic-bezier(0.86,0,0.07,1);
+  margin: 3px 0;
```
dbchristopher: I had to add a small margin between each menu item in order for the focus ring to appear correctly. As you can see in the gif, only the last DOM item would have the focus ring appear in its entirety. And what's worse, the first item would have it's focus ring obfuscated in entirety. 

focus-ring

This was the only solution I can think of that didn't involve tweaking z-index in the `focus` style. @lukelooker is this margin acceptable? I'm definitely open to other ideas here too!

dbchristopher:
@lukelooker this one is ready for you!

packages/components/src/Menu/MenuItem/MenuItem.tsx
lukelooker: It's used to show a border to the left of the item when it's current (replicates a behavior done in the Helltool UI for the content sidebar.

image


`currentMarker` assigned to `Cheddar`:

image

No `currentMarker` assigned:

image

lukelooker: Maybe @jhardy can add some context here (I forget why we separated these in the first place)

dbchristopher: oh I see -- before the refactor you would have needed both current AND currentMarker to be true in order for that left border to appear. that does seem weird and unnecessary (marking `current={true}` seems like it should be sufficient)

packages/components/src/Menu/MenuItem/MenuItemButton.tsx
```diff
@@ -48,6 +48,7 @@ export const MenuItemButton = styled.button<MenuItemButtonProps>`
   font-size: inherit;
   flex: 1;
   outline: none;
+  padding-left: ${({ theme }) => theme.space.small};
```
lukelooker: Why is this addition needed?

dbchristopher: that was to correct for the additional 4px border on the left which renders the `current` marker (it still renders a 4px transparent border when not marked `current={false}`). I'm happy to remove this line if we don't mind things shifting over a little!

dbchristopher:
looks like the top item still gets blocked by the group label, but it's definitely an improvement:
better-focus

dbchristopher:
@lukelooker ok I added a small margin below the MenuGroupLabel to fix the first item's focus ring, and then moved the spacing adjustment to make it more clear why that padding tweak was needed

packages/components/src/Menu/MenuItem/MenuItemListItem.tsx
```diff
@@ -92,6 +92,8 @@ export const MenuItemListItem = styled(Li)<MenuListItemProps>`
     border-left-style: solid;
     border-left-color: ${({ itemStyle, current }) =>
       current ? itemStyle.marker.color : 'transparent'};
+    padding-left: calc(${({ theme, itemStyle }) =>
+      `${theme.space.medium} - ${itemStyle.marker.size}px`});
```
dbchristopher: subtracting the marker size from the standard left padding also makes this tweak more precise

PR #35, opened 2019-11-13
Lara/avatar icon/lens 356

laraharrow:


.vscode/settings.json
```diff
@@ -15,7 +15,9 @@
     }
   ],
   "cSpell.words": [
+    "Bowerman",
```
lukelooker: Pretty sure my name shoudn't be here :)

dbchristopher: BUT how else would angry users find you 😜

packages/components/src/Avatar/AvatarIcon.tsx
```diff
@@ -0,0 +1,74 @@
...
+    </div>
+  )
+}
+
+const AvatarIconCircle = styled.div`
+  align-items: center;
+  display: flex;
+  justify-content: center;
+  border: 1px solid;
+  border-color: currentColor;
```
dbchristopher: I think `1px solid currentColor;` would work?

packages/components/src/Avatar/AvatarUser.tsx
```diff
@@ -0,0 +1,107 @@
...
+    last_name: string | null
+    avatar_url: string | null
+  }
+}
+
+const AvatarLayout: FC<AvatarUserProps> = ({ color, className, user }) => {
+  const initials =
+    user.first_name &&
+    user.last_name &&
+    `${user.first_name.substr(0, 1)}${user.last_name.substr(0, 1)}`
```
dbchristopher: a nice quirk of javascript is that strings can also be treated like arrays. So this could be condensed to `    `${user.first_name[0]}${user.last_name[0]}`

packages/components/src/Avatar/AvatarUser.tsx
```diff
@@ -0,0 +1,107 @@
...
+    <div className={className}>
+      <AvatarInitials color={color}>{initials}</AvatarInitials>
+      {user.avatar_url && <AvatarPhoto src={user.avatar_url} />}
+    </div>
+  )
+}
+
+const avatarCircle = css`
+  border: 1px solid;
+  border-color: currentColor;
```
dbchristopher: combine rules if that works

packages/components/src/Avatar/AvatarUser.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+import React, { FC } from 'react'
+import { color, reset, space, layout, SpaceProps } from '@looker/design-tokens'
+import styled, { css } from 'styled-components'
+
+export interface AvatarProps extends SpaceProps {
+  /**
+   *  @default `40px`
+   **/
+  size?: number
```
dbchristopher: not seeing this prop actually used anywhere?

dbchristopher: ah I get it, it's rolled into `layout` from the design tokens 👍

packages/components/src/Avatar/AvatarIcon.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const AvatarLayout: FC<AvatarIconProps> = ({
+  color,
+  className,
+  icon = 'Users',
+}) => {
+  return (
+    <div className={className}>
+      <AvatarIconCircle>
+        <Icon name={icon} size={24} color={color} />
```
dbchristopher: We should probably pass in a size variable here, or else it will only scale the border but not the icon inside:
Screen Shot 2019-11-15 at 10 17 46 AM

packages/components/src/Avatar/AvatarUser.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+const AvatarLayout: FC<AvatarUserProps> = ({ color, className, user }) => {
+  const initials =
+    user.first_name &&
+    user.last_name &&
+    `${user.first_name.substr(0, 1)}${user.last_name.substr(0, 1)}`
+
+  return (
+    <div className={className}>
+      <AvatarInitials color={color}>{initials}</AvatarInitials>
```
dbchristopher: We should probably scale the initials based on `size` prop too
Screen Shot 2019-11-15 at 10 20 27 AM

packages/components/src/Avatar/AvatarUser.tsx
```diff
@@ -0,0 +1,107 @@
...
+  color?: string
+
+  className?: string
+}
+
+interface AvatarUserProps extends AvatarProps {
+  user: {
+    first_name: string | null
+    last_name: string | null
+    avatar_url: string | null
```
dbchristopher: make sure to mark avatar_url optional so that typescript doesn't complain if we just want to render the initials

packages/playground/src/index.tsx
```diff
@@ -26,62 +26,32 @@
...
-
-        <Menu>
-          <MenuDisclosure>
-            <ButtonOutline>Pizza Menu Selection....</ButtonOutline>
-          </MenuDisclosure>
-          <MenuList minWidth="18rem">{contents}</MenuList>
-        </Menu>
-      </Box>
+        {/* <AvatarUser m="medium" user={data} /> */}
+        <AvatarIcon icon="Trash" m="medium"></AvatarIcon>
```
dbchristopher: I wonder if it makes sense to roll these into a single `Avatar` component? Seems like we could easily just decide what to render based on the props: 

1. If props.avatar_url exists, render avatar image
1. If props.avatar_icon exists, render icon
1. Otherwise render initials

I don't know, what do you think? I feel like it might be nice to have one less component for users to have to keep track of

packages/components/src/Avatar/AvatarIcon.tsx
laraharrow: yep fix that

packages/components/src/Avatar/AvatarUser.tsx
laraharrow: of course🤦🏽‍♀️

packages/components/src/Avatar/AvatarUser.tsx
laraharrow: I think it always returns something as its part of the structure but sometimes it returns null

PR #91, opened 2019-11-20
[LENS-414] Confirm modal style improvements

dbchristopher:
### :sparkles: Changes

- Update base Confirm text styling to match [figma](https://www.figma.com/file/sDUnKT8DfIroF2yNbXfrc2/Dialog?node-id=168%3A1667), including a more reasonable in and max width
- Add support for using a React component as the modal message, including documentation for rich content example.
-Update base theme to render more vibrant button colors

- [x] PR is ideally < 400LOC

### :camera: Screenshots

Screen Shot 2019-11-20 at 11 54 00 AM
Screen Shot 2019-11-20 at 11 53 53 AM

### DIFF
--- a/packages/components/src/Icon/Icon.tsx
+++ b/packages/components/src/Icon/Icon.tsx
@@ -70,6 +70,10 @@ const Styled = styled.div<Omit<IconProps, 'name'>>`
   height: ${props => props.height};
   align-items: center;
   display: inline-flex;
+
+  svg {
+    width: inherit;
+  }
 `
 
 Styled.defaultProps = { size: '1rem' }
--- a/packages/components/src/Icon/Icon.tsx
+++ b/packages/components/src/Icon/Icon.tsx
@@ -70,10 +70,6 @@ const Styled = styled.div<Omit<IconProps, 'name'>>`
   height: ${props => props.height};
   align-items: center;
   display: inline-flex;
-
-  svg {
-    width: inherit;
-  }
 `
 
 Styled.defaultProps = { size: '1rem' }
--- a/packages/components/src/Modal/Dialog/DialogSurface.tsx
+++ b/packages/components/src/Modal/Dialog/DialogSurface.tsx
@@ -3,17 +3,17 @@
  MIT License
 
  Copyright (c) 2019 Looker Data Sciences, Inc.
- 
+
  Permission is hereby granted, free of charge, to any person obtaining a copy
  of this software and associated documentation files (the "Software"), to deal
  in the Software without restriction, including without limitation the rights
  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
  copies of the Software, and to permit persons to whom the Software is
  furnished to do so, subject to the following conditions:
- 
+
  The above copyright notice and this permission notice shall be included in all
  copies or substantial portions of the Software.
- 
+
  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
@@ -39,8 +39,3 @@ export const DialogSurface = styled(ModalSurface)<ModalSurfaceProps>`
     transform: translateY(0%);
   }
 `
-
-DialogSurface.defaultProps = {
-  maxHeight: '90vh',
-  maxWidth: '100%',
-}
--- a/packages/components/src/Modal/Layout/ModalFooter.tsx
+++ b/packages/components/src/Modal/Layout/ModalFooter.tsx
@@ -53,22 +53,16 @@ const Layout = styled.footer`
   align-items: center;
   display: flex;
   flex-direction: row-reverse;
-  padding: ${props => props.theme.space.large}
-    calc(${p => p.theme.space.xlarge} - ${p => p.theme.space.xsmall});
-  /**
-    * Subtract margin placed on children from the horizontal padding applied to the component
-    * so that the content is properly aligned.
-    * padding is: vertical: large, horizontal: calc(xlarge - xsmall)
-    **/
+  padding: ${({ theme }) => `${theme.space.large} ${theme.space.xlarge}`};
 `
 
 const SpaceChildren = styled.div<SpaceProps>`
   ${space};
-  display: flex;
-  flex-direction: row-reverse;
-
-  & > * {
-    margin-right: ${props => props.theme.space.xsmall};
-    margin-left: ${props => props.theme.space.xsmall};
+  display: grid;
+  grid-gap: ${({ theme }) => theme.space.medium};
+  grid-auto-flow: column;
+  direction: rtl; /* Grid equivalent to flex-direction: row-reverse */
+  * {
+    direction: ltr;
   }
 `
--- a/packages/components/src/Modal/Layout/ModalHeader.tsx
+++ b/packages/components/src/Modal/Layout/ModalHeader.tsx
@@ -30,14 +30,14 @@ import {
   CompatibleHTMLProps,
   reset,
 } from '@looker/design-tokens'
+
 import { IconNames } from '@looker/icons'
 import styled from 'styled-components'
 import React, { FC, useContext } from 'react'
 
-import { Box } from '../../Layout/Box'
-import { ButtonTransparent } from '../../Button'
+import { IconButton } from '../../Button'
+import { Flex } from '../../Layout'
 import { Heading } from '../../Text'
-import { Icon } from '../../Icon'
 import { ModalContext } from '../ModalContext'
 
 export interface ModalHeaderProps
@@ -47,6 +47,7 @@ export interface ModalHeaderProps
    * Specify an icon to be used for close. Defaults to `Close`
    */
   closeIcon?: IconNames
+  children: string
 }
 
 export const ModalHeader: FC<ModalHeaderProps> = ({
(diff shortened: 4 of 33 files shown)


packages/components/src/Modal/Dialog/DialogSurface.tsx
```diff
@@ -40,7 +40,4 @@ export const DialogSurface = styled(ModalSurface)<ModalSurfaceProps>`
   }
 `
 
-DialogSurface.defaultProps = {
-  maxHeight: '90vh',
-  maxWidth: '100%',
-}
+DialogSurface.defaultProps = {}
```
dbchristopher: Moved these default props to ModalSurface

lukelooker: Since it's empty probably best to just go ahead and remove this altogether :)

packages/components/src/Modal/Layout/ModalFooter.tsx
```diff
@@ -53,13 +53,7 @@ const Layout = styled.footer`
...
   display: flex;
   flex-direction: row-reverse;
-  padding: ${props => props.theme.space.large}
-    calc(${p => p.theme.space.xlarge} - ${p => p.theme.space.xsmall});
-  /**
-    * Subtract margin placed on children from the horizontal padding applied to the component
-    * so that the content is properly aligned.
-    * padding is: vertical: large, horizontal: calc(xlarge - xsmall)
-    **/
+  padding: ${({ theme }) => `${theme.space.large} ${theme.space.xlarge}`};
```
lukelooker: This change causes a couple of issues:

1. There's margin on the left hand side of the secondary button(s)
2. The gap between the buttons is narrower (previously it was `xsmall + xsmall` whereas now it's just `xsmall`)

image

packages/components/src/Modal/Layout/ModalHeader.tsx
```diff
@@ -58,31 +59,27 @@ export const ModalHeader: FC<ModalHeaderProps> = ({
...
 
 const Header = styled.header<SpaceProps>`
   ${reset}
   ${space}
-
-  display: flex;
-  align-items: center;
-  flex-direction: row;
-
-  ${Heading} {
```
lukelooker: This needs to be restored (for now at least).

It reconciles a bug where we previously mis-calculated Heading if no `as` or `fontSize` was explicitly specified. Changing this is definitely something we should probably explore but would need an accomodating fix within Helltool to prevent changes to presentation and snapshots.

packages/components/src/Modal/Layout/ModalHeader.tsx
```diff
@@ -58,31 +59,27 @@ export const ModalHeader: FC<ModalHeaderProps> = ({
...
+        <IconButton
+          tabIndex={-1}
+          color="neutral"
+          size="medium"
+          onClick={closeModal}
+          label="Close"
+          icon={closeIcon}
+          mr="-10px"
+        />
+      </Flex>
```
lukelooker: This changes makes me very happy but there's a couple of minor nits.

A lot of existing implementations already have a `<Heading>` component within them so we'll generate a bunch of errors where React is producing `Heading > Heading`

IconButton - the size `medium` is a bit larger than the old version. `small` would produce a `20x20` SVG which would be inline with the previous version.

packages/www/src/documentation/components/modals/confirm.mdx
```diff
@@ -60,3 +60,78 @@ A render-props version, for use in class components.
...
+      onConfirm={close => {
+        alert('Now you know.')
+        close()
+      }}
+    >
+      {open => <Button onClick={open}>Do something fancy</Button>}
+    </Confirm>
+  )
+}
+```
```
lukelooker: I'd propose a somewhat simpler example (I'd prefer not so show `styled` in examples except where we're explicitly try to teach folks about that functionality):

```

```jsx
;() => {
  /*
   * import { Confirm, Flex, FlexItem, Icon, Paragraph, Link } from '@looker/components'
   * import styled from 'styled-components'
   */

  const RichMessage = () => (
    <Flex>
      <FlexItem>
        <Icon
          mr="large"
          name="CircleInfoOutline"
          size="80px"
          color="palette.purple400"
        />
      </FlexItem>
      <FlexItem>
        <Paragraph>
          Canadians say "sorry" so often that{' '}
          <strong>in 2009 a law was passed</strong> declaring that an apology
          cannot be used as an admission of guilt.
        </Paragraph>
        <Cite>
          SOURCE:{' '}
          <Link
            href="https://www.theloop.ca/canadians-love-to-say-sorry-so-much-we-had-to-make-this-law/"
            target="_blank"
          >
            the loop
          </Link>
        </Cite>
      </FlexItem>
    </Flex>
  )

  const Cite = styled.cite`
    color: ${({ theme }) => theme.colors.palette.charcoal300};
    margin-top: ${({ theme }) => theme.space.xsmall};
    display: block;
  `

  return (
    <Confirm
      title="Did you know?"
      message={<RichMessage />}
      onConfirm={close => {
        alert('Now you know.')
        close()
      }}
    >
      {open => <Button onClick={open}>Do something fancy</Button>}
    </Confirm>
  )
}
```

packages/components/src/Modal/Layout/ModalHeader.tsx
dbchristopher: Looks like there are only 6 instances of `<ModalHeader>` in Helltool at the moment -- I'm happy to take point on migrating those! I do feel like only accepting `string` for that specific value is the right move either way

Relatedly, there are exactly zero uses of `<Confirm>` in helltool at the moment 😖 Hopefully the improved style and more usable maxwidth will encourage adoption there

packages/components/src/Modal/Layout/ModalHeader.tsx
dbchristopher: So assuming I can update helltool to strip out now-redundant inline `<Heading>` components, no change is needed here right? As we'd now be providing the proper formatting out of the box I don't think we'd have a case where a heading was used without `as` or `fontSize`

packages/components/src/Modal/Layout/ModalHeader.tsx
dbchristopher: > IconButton - the size medium is a bit larger than the old version. small would produce a 20x20 SVG which would be inline with the previous version.

Yeah it is bigger than the old one for sure. But it also seems closer to the figma reference to my eyes. Although the figma layout certainly lands somewhere in the middle between those two sizes 😖 What do you think?

Screen Shot 2019-11-21 at 3 31 11 PM

dbchristopher: Also you'll notice I did allow the max-width to go a little wider than what's in figma, as I don't think the modal-footer example with four actions would fit in that narrower layout:

Screen Shot 2019-11-21 at 3 35 09 PM

Do you think the 700px max width is the right choice? Or should I also reduce that to conform closer to what's in figma?

packages/www/src/documentation/components/modals/confirm.mdx
dbchristopher: Great simplification 👍

packages/components/src/Modal/Layout/ModalFooter.tsx
dbchristopher: Ok I updated the footer layout to use `grid-gap` to handle that item spacing, rather than margins. That way we don't have to deal with any finicky `calc` subtraction to get footer items to line up with content

dbchristopher:
Ok @lukelooker I addressed all the feedback and this should be ready for another look! It still includes the theme updates, so once #113 is merged in the line diffs should be greatly reduced

packages/components/src/Modal/Layout/ModalHeader.tsx
jhardy: I am working on having a better spec in figma for some of these items. I would say for now we can go with what looks good as the current spec is a bit outdated and need some refinement based on some newer patterns we have in the codebase

packages/components/src/Modal/Layout/ModalFooter.tsx
lukelooker: Awesome

packages/components/src/Modal/Layout/ModalHeader.tsx
lukelooker: Yep - that'd be great. Let's make sure to note that in the CHANGELOG since it's a change in behavior (albeit essentially a bugfix)

packages/components/src/Modal/Layout/ModalHeader.tsx
lukelooker: Works for me. 

Just to clarify the example is slightly contrived since I've never seen a real-world implementation (thus-far) that has more than 1 "secondary" button :)

packages/components/src/Modal/Layout/ModalHeader.tsx
lukelooker: That works for me!

packages/components/src/Icon/Icon.tsx
```diff
@@ -70,6 +70,10 @@ const Styled = styled.div<Omit<IconProps, 'name'>>`
   height: ${props => props.height};
   align-items: center;
   display: inline-flex;
+
+  svg {
+    width: inherit;
+  }
```
dbchristopher: Fixes an issue where the icon gets rendered super tiny for some reason when used in www
Screen Shot 2019-11-25 at 11 45 48 AM

dbchristopher:
ok @lukelooker I think I addressed all the feedback, and fixed the bug in the Drawer layout. This should be ready for another look whenever you're around! No rush 😄

dbchristopher:
Awesome, thanks @lukelooker!

PR #113, opened 2019-11-21
Update theme colors, and copy old values to legacy variables

dbchristopher:
### :sparkles: Changes

- Implements the theme changes used in #91 
- Creates new `legacySemanticColors`, `legacyPalette`, and `legacyTheme` to provide the option of maintaining current color settings

- [x] PR is ideally < 400LOC

### DIFF
--- a/packages/design-tokens/src/theme.ts
+++ b/packages/design-tokens/src/theme.ts
@@ -50,8 +50,6 @@ import {
   fontFamilies,
   fontSizes,
   fontWeights,
-  classicSemanticColors,
-  classicPalette,
   lineHeights,
   palette,
   radii,
@@ -93,9 +91,4 @@ export const theme: Theme = {
   shadows,
   space,
   transitions,
-}
-
-export const classicTheme: Theme = {
-  ...theme,
-  colors: { palette: classicPalette, semanticColors: classicSemanticColors },
 }
--- a/packages/design-tokens/src/theme.ts
+++ b/packages/design-tokens/src/theme.ts
@@ -50,6 +50,8 @@ import {
   fontFamilies,
   fontSizes,
   fontWeights,
+  legacySemanticColors,
+  legacyPalette,
   lineHeights,
   palette,
   radii,
@@ -81,6 +83,21 @@ export interface Theme {
 export const theme: Theme = {
   breakpoints,
   colors: { palette, semanticColors },
+  easings,
+  fontSizes,
+  fontWeights,
+  fonts: fontFamilies,
+  lineHeights,
+  radii,
+  reset: defaultReset,
+  shadows,
+  space,
+  transitions,
+}
+
+export const legacyTheme: Theme = {
+  breakpoints,
+  colors: { palette: legacyPalette, semanticColors: legacySemanticColors },
   easings,
   fontSizes,
   fontWeights,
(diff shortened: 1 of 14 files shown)


packages/design-tokens/src/theme.ts
```diff
@@ -92,3 +94,18 @@ export const theme: Theme = {
...
+  fontSizes,
+  fontWeights,
+  fonts: fontFamilies,
+  lineHeights,
+  radii,
+  reset: defaultReset,
+  shadows,
+  space,
+  transitions,
+}
```
lukelooker: I think it'd be more ideal to do this as a spread & extend behavior so the difference between the two is more clear.

```
export const legacyTheme: Theme = {
  ...theme,
  colors: { palette: legacyPalette, semanticColors: legacySemanticColors }
}
```

packages/design-tokens/src/tokens/color/palette.ts
```diff
@@ -32,6 +32,82 @@ export const palette: Palette = {
...
+  red700: '#990F14',
+  red800: '#730B0F',
+  red900: '#52080B',
+}
+
+export const legacyPalette: Palette = {
+  textInverted: '#FFFFFF',
+  transparent: 'transparent',
+  white: '#FFFFFF',
+
```
lukelooker: I'd really prefer to see less duplication here by spreading the existing palette into a new object and just overriding the few values we're altering.

packages/design-tokens/src/tokens/color/semantic_colors.ts
```diff
@@ -51,6 +53,53 @@ const {
 } = palette
 
 export const semanticColors: SemanticColors = {
+  danger: {
+    altText: red400,
```
lukelooker: Same note as earlier - extend instead of duplicate.

dbchristopher:
@lukelooker I'm not actually sure if that is going to be the most maintainable approach going forward. Extending and overriding from the same base establishes a concretion between the classic theme and the up-to-date values. The new theme will only exist in relation to the old theme -- but it seems like we should plan for them to diverge more in the future. Especially considering we have a huge color refactor coming in the near future.

My intent was that the classic theme would be a time capsule of this moment, and we'd have more freedom to modify the default theme as needed. Does that make sense? Do you still feel like I should link the two objects in that way?

dbchristopher:
I can see it makes sense for theme vs classicTheme, as the colors are the only slice that will change. But I feel like it would be helpful to keep `palette` and `semanticColor` objects uncoupled from the classic versions

lukelooker:
If we're going to export (and therefore support) `classicTheme` then I think it's a "maintained abstraction" and I'd prefer to maintain something smaller where we can see the simple delta vs a full copy. The spread and extend model is basically what Styled Components recommends for overriding so it seems in keeping with generally held practice.

Alternatively we could simply _not_ offer classicTheme and do the extension behavior within classic implementations themselves (that had been my original thinking but I don't hate this approach either). The downside of exporting a `classicTheme` is that it's an implicit guarantee that it will continue to work vs pushing it out to the consumer to maintain said "extension"

dbchristopher:
> The downside of exporting a classicTheme is that it's an implicit guarantee that it will continue to work vs pushing it out to the consumer to maintain said "extension"

Yeah that's exactly why I preferred to duplicate rather than override. Less chance for unintentional changes

dbchristopher:
Ok @lukelooker I removed all classic theme objects from this PR, and went ahead and moved those theme overrides to our internal app. (PR #48505)

PR #115, opened 2019-11-22
Scroll lock & focus trap improvements

mdodgelooker:
### :sparkles: Changes

- `ModalContext`, `ModalPortal`, `getModalRoot`, and `CustomizableModalAttributes` renamed to `IntersitialContext`, `IntersitialPortal`, `getIntersitialRoot`, and `CustomizableIntersitialAttributes`, respectively.
- Focus trap and scroll lock methods added to `InterstitialContext` so that they can be disabled/re-enabled when nested `Popover`s open/close.
- The developer now has the ability to control the focus trap and scroll lock behaviors if they have a component that needs to modify them (e.g. a `ReactSelect` inside of a `Modal` or `Popover`).

- [x] PR is ideally < 400LOC

### :camera: Screenshots

nested-popovers1
nested-popovers2
nested-popovers3


packages/components/package.json
```diff
@@ -22,7 +22,7 @@
     "build:esm": "ttsc --project tsconfig.build.json --module es2015 --target es5 --outDir dist/esm",
     "build:cjs": "ttsc --project tsconfig.build.json --module commonjs --target es5 --outDir dist/cjs",
     "clean": "rm -Rf ./dist",
-    "develop": "yarn clean && ttsc --project tsconfig.build.json --watch --module commonjs --target es5 --outDir dist/cjs"
+    "develop": "ttsc --project tsconfig.build.json --watch --module commonjs --target es5 --outDir dist/cjs"
```
lukebowerman:
```suggestion
    "develop": "yarn clean && ttsc --project tsconfig.build.json --watch --module commonjs --target es5 --outDir dist/cjs"
     "develop-playground": "ttsc --project tsconfig.build.json --watch --module commonjs --target es5 --outDir dist/cjs"
```

For now (until we can rationalize this to a single command) perhaps we can do this as two separate commands rather than clobbering the Gatsby-compliant version?

packages/components/src/Interstitial/InterstitialPortal.tsx
```diff
@@ -24,44 +24,48 @@
...
-  useScrollLock(refToUse)
+export const InterstitialPortal = forwardRef(
+  ({ children }: { children: ReactNode }, ref: Ref<HTMLDivElement>) => {
+    const el = useRef(document.createElement('div'))
 
-  useEffect(() => {
-    const modalRoot = getModalRoot()
-    if (!modalRoot) return
+    useEffect(() => {
+      const modalRoot = getInterstitialRoot()
```
lukelooker: Still has references to modalRoot. Perhaps just root or domNode (something suitably generic) would make more sense?

packages/components/src/Popover/Popover.tsx
```diff
@@ -243,15 +258,18 @@ function usePopoverToggle(
       event.stopPropagation()
     }
     if (isOpen) {
+      document.addEventListener('mousedown', handleMouseDown, true)
```
lukelooker: Why do we need both `mousedown` and `click` listeners now?

packages/playground/package.json
```diff
@@ -18,9 +18,11 @@
     "@looker/design-tokens": "^0.7.4",
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
+    "react-select": "^3.0.8",
```
lukelooker: We'll need to pull this out before merging so they don't end up polluting the yarn.lock / inferring that we endorse `react-select`

CHANGELOG.md
```diff
@@ -32,7 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ColorWheel` - Optimize ColorWheel value change performance & restored `ColorWheel` documentation
 - `MenuGroup` - Fix bug where menu group headers render a drop shadow by default
 - `MenuSearch` - fixed keyboard shortcuts (up / down arrows to escape focus on input) and Improved documentation example.
-- `ModalPortal` - Fixed scrolling within a `Modal` (`Dialog` & `Drawer`)
+- `InterstitialPortal` - Fixed scrolling within a `Modal` (`Dialog` & `Drawer`)
```
lukelooker: Probably shouldn't update this entry since it was still called `ModalPortal` at this point in time.

packages/components/src/Interstitial/InterstitialContext.ts
```diff
@@ -24,42 +24,18 @@
...
-        >
-          {children}
-        </ButtonOutline>
-      )}
-    </Popover>
-  </Box>
-)
+const interstitialContext: InterstitialContextProps = {}
+
+export const InterstitialContext = createContext(interstitialContext)
```
lukelooker: We'd talked about aliasing this to `ModalContext` as well as providing a proxy of `close` to `closeModal` to ensure backwards compatibility.

packages/components/src/Popover/Popover.tsx
mdodgelooker: Adding `mousedown` was to fix an issue where `react-select` will automatically scroll down when the menu opens if there's not enough room to display it, and then `Popover` thinks you've clicked outside even though your click started inside. I tried removing `click` but then failing tests made me realize it's needed in order to cancel any other potential click handler (just close the `Popover` on first click).

However, I dug a little further into `react-select` and it turns out you can avoid the auto scroll if you add `menuPosition="fixed"`, so I'm gonna pull out `mousedown` and consider that the workaround.

mdodgelooker:
Interstitial is out and Modal is back in. Reverted all changes in `playground`. As for testing, it turns out `jsdom` does not have tabbing through focus and scrolling in the sense that a browser does so it doesn't make much sense to write a jest test for scroll lock and focus trap.

PR #213, opened 2019-12-06
Select (combobox)

mdodgelooker:
### :sparkles: Changes

This work draws heavily from https://github.com/reach/reach-ui (https://reacttraining.com/reach-ui/combobox/) with a few differences:
- Converted to TypeScript
- Added a childless / non-compositional version (using `options` prop, as well as `inputProps`, `listProps`, and `optionProps` for customizations of those parts)
- Accepts `data: any` to go along with rendered value (could be an ID, unformatted version of the value, etc)
- Better handling of options as they update (compositional version)
- Opens on every click

- [ ] PR is ideally < 400LOC

### :camera: Screenshots

select


packages/playground/src/index.tsx
```diff
@@ -24,86 +24,67 @@
...
+            onClick={handleClick}
+          >
+            Kiwis
+          </Button>
+          <Button mt="medium" data-fruit="Oranges" onClick={handleClick}>
+            Oranges
+          </Button>
+          <Divider my="xlarge" />
+          <InputText width={240} />
+        </Box>
```
dbchristopher: I think we want to be saving demo content for posterity, so this code should be moved to a subfolder

dbchristopher: e.g. for the slider stuff I moved it out to `playground/src/Slider/SliderDemo.tsx`

packages/components/src/Form/Inputs/Select/Select.tsx
```diff
@@ -24,180 +24,213 @@
...
+  if (options) {
+    if (children) {
+      // eslint-disable-next-line no-console
+      console.warn(`Warning: options and children cannot be used together.
+      If you wish to build your Select with the options prop, do not define any children.`)
+    }
+
+    content = (
+      <>
+        <SelectInput value={value} {...inputProps} />
```
dbchristopher: This is a weird bug. For some reason I can only get the select box to open the first time I try to interact with it:
select-bug

packages/components/src/Form/Inputs/InputProps.ts
```diff
@@ -34,7 +34,7 @@ export interface InputProps extends CompatibleHTMLProps<HTMLInputElement> {
 export const inputPropKeys = [
   'accept',
   'autoFocus',
```
dbchristopher: do you think this should also be all lowercase `autofocus` to match the html standard attribute?

packages/components/src/Form/Inputs/Select/Select.tsx
```diff
@@ -24,180 +24,213 @@
...
+export type OnSelect = (option: SelectOptionObject) => void
+
+export function useControlledSelect(initialValue = '') {
+  const [value, setOption] = useState(initialValue)
+  const handleSelect = useCallback((option: SelectOptionObject) => {
+    setOption(option.value)
+  }, [])
+  const handleChange = useCallback((e: React.FormEvent<HTMLInputElement>) => {
+    setOption(e.currentTarget.value)
+  }, [])
```
dbchristopher: re: use of useCallback, do we actually want to memoize event handlers? I feel like we'd prefer they fire fresh every time they're called. Considering they aren't firing any network requests, there shouldn't be any performance difference right? 🤔

packages/components/src/Form/Inputs/Select/SelectInput.tsx
```diff
@@ -0,0 +1,257 @@
...
+  // prop
+  const handleChange = useCallback(
+    (event: FormEvent<HTMLInputElement>) => {
+      isInputting.current = true
+      if (!isControlled) {
+        handleValueChange(event.currentTarget.value)
+      }
+      setTimeout(() => {
+        isInputting.current = false
+      }, 0)
```
dbchristopher: tiny nit: i believe `requestAnimationFrame` is the preferred way to delay a process until the next processor cycle

packages/components/src/Form/Inputs/Select/SelectInput.tsx
```diff
@@ -0,0 +1,257 @@
...
+</svg>`
+const indicatorSize = '1rem'
+const indicatorPadding = '.25rem'
+const indicatorPrefix = 'data:image/svg+xml;base64,'
+export const selectIndicatorBG = (color: string) =>
+  typeof window !== 'undefined' &&
+  `url('${indicatorPrefix}${window.btoa(
+    indicatorRaw.replace('#1C2125', color)
+  )}')`
+
```
dbchristopher: Can we use the bundled CaretDown icon for this?
Screen Shot 2019-12-06 at 11 29 27 AM

packages/components/src/Form/Inputs/Select/SelectInput.tsx
```diff
@@ -0,0 +1,257 @@
...
+const indicatorRaw = `<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M7.41 8L12 12.58L16.59 8L18 9.41L12 15.41L6 9.41L7.41 8Z" fill="#1C2125"/>
+</svg>`
+const indicatorSize = '1rem'
+const indicatorPadding = '.25rem'
+const indicatorPrefix = 'data:image/svg+xml;base64,'
+export const selectIndicatorBG = (color: string) =>
+  typeof window !== 'undefined' &&
+  `url('${indicatorPrefix}${window.btoa(
+    indicatorRaw.replace('#1C2125', color)
```
dbchristopher: if we do need to keep the SVG element hard-coded here, I'd prefer killing the hard-coded `fill` attribute altogether and styling that in the css

packages/components/src/Form/Inputs/Select/helpers.ts
```diff
@@ -0,0 +1,250 @@
...
+  return function handleKeyDown(event: KeyboardEvent<HTMLDivElement>) {
+    const options = optionsRef ? optionsRef.current : contextOptions || []
+    switch (event.key) {
+      case 'ArrowDown': {
+        // Don't scroll the page
+        event.preventDefault()
+
+        // If the developer didn't render any options, there's no point in
+        // trying to navigate--but seriously what the heck? Give us some
+        // options fam.
```
dbchristopher: 🤣

packages/components/src/Form/Inputs/Select/helpers.ts
```diff
@@ -0,0 +1,250 @@
...
+    })
+  }
+}
+
+// We don't want to track the active descendant with indexes because nothing is
+// more annoying in a select than having it change values RIGHT AS YOU HIT
+// ENTER. That only happens if you use the index as your data, rather than
+// *your data as your data*. We use this to generate a unique ID based on the
+// value of each item.  This function is short, sweet, and good enough™
+// https://stackoverflow.com/questions/6122571/simple-non-secure-hash-function-for-javascript
```
dbchristopher: nice 👍

mdodgelooker:
Thanks so much @dbchristopher ! I realized I messed up the rebase on this branch so there's a bunch of bogus changes showing (which I'll revert). Sorry about that!

packages/components/src/Form/Inputs/InputProps.ts
mdodgelooker: `autoComplete` was a mistake on my part – I thought `React` did the camelCase conversion there but they don't. The do with `autoFocus` though.

packages/components/src/Form/Inputs/Select/SelectInput.tsx
mdodgelooker: Indeed there is a way!

packages/components/src/Form/Fields/FieldText/__snapshots__/FieldText.test.tsx.snap
```diff
@@ -125,6 +125,7 @@ exports[`A FieldText with an error validation aligned to the left 1`] = `
   padding-top: 0rem;
   padding-bottom: 0rem;
   font-size: 0.875rem;
+  background: #FFF2F4;
```
mdodgelooker: The light-red background on `InputText` with `validationType="error"` was broken in an earlier commit.

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -95,6 +94,10 @@ export const InputText = styled(InputComponent).attrs(
   ${space}
   ${typography}
   ${pseudoClasses}
+  ${props =>
+    props.validationType === 'error'
+      ? `background: ${props.theme.colors.palette.red000};`
+      : ''}
```
mdodgelooker: Fixes light-red background issue mentioned above.

mdodgelooker:
Ok, ready for round 2!

packages/components/src/Form/Inputs/Select/Select.tsx
mdodgelooker: Oddly, I think that's how Reach UI originally designed it. But a native `select` opens on every click, and so does `ReactSelect`, so we should too. Fixed.

packages/components/src/Form/Inputs/Select/Select.tsx
mdodgelooker: You're right – I was thinking it might reduce re-renders down the line but it doesn't appear to and the re-renders (no dom change mostly) so far don't seem to affect performance. Removed `useCallback`.

packages/components/src/Form/Inputs/Select/helpers.ts
mdodgelooker: Sadly I cannot take credit for the excellent commenting. 😁@lukelooker do you think we need to remove the humor, keep it buttoned-down?

packages/components/src/Form/Inputs/Select/SelectInput.tsx
mdodgelooker: Aha! ✅

packages/components/src/Form/Inputs/Select/helpers.ts
lukelooker: Humor (as long as not too colorful) is fine by me :)

mdodgelooker:
FYI I removed the "highlight searched words" feature – it was buggy and was not in the design. Perhaps an upgrade for a later time.

packages/components/src/Form/Fields/Field.tsx
```diff
@@ -74,6 +74,10 @@ export interface FieldProps {
    * Defines the label for the field.
    */
   label?: string
+  /**
+   * id attribute for the label.
+   */
+  labelID?: string
```
mdodgelooker: Adding this specifically for `FieldSelect` – it really helps the `Select`'s accessibility if both the input and the menu can have `aria-labelledby={labelID}`.

packages/components/src/Form/Fields/FieldSelect/FieldSelect.tsx
```diff
@@ -26,29 +26,39 @@
 
 import React, { forwardRef, Ref } from 'react'
 import styled from 'styled-components'
-import uuid from 'uuid/v4'
+import { useID } from '../../../utils'
```
mdodgelooker: I ended up making a hook for the fallback uuidv4 generation feature, with `useMemo` included so it doesn't update the `id` dom attribute on every render.

dbchristopher:
@mdodgelooker didn't this originally have some sort of typeahead/filtering behavior? Looks like it won't do partial string matching now -- in this example I have to type out the full `Bananas` before it finds the match:

no-typeahead

Also I feel like it would be nice for the UX if I could hit `enter` after it matches an option to confirm my selection. At the moment I have to use the arrow keys to select an option before it will let me use the enter key

dbchristopher:
I assume that change is related to removing the "highlight searched words" feature? Even without the partial-word highlighting, I feel like it would be pretty important for mvp to at least filter available options as you type

dbchristopher:
Oh yes, and definitely make sure the string matching is not case sensitive! In this example, it should know `bananas` and `Bananas` are the same thing:
Screen Shot 2019-12-12 at 11 21 06 AM

packages/components/src/Form/Inputs/Select/Select.tsx
```diff
@@ -24,180 +24,222 @@
...
-    | GroupedOptionsType<SelectOptionProps>
+    Omit<
+      CompatibleHTMLProps<HTMLDivElement>,
+      'readOnly' | 'onSelect' | 'onChange'
+    > {
+  /**
+   * Called with the selection value when the user makes a selection from the
+   * list.
+   */
+  onSelect?: OnSelect
```
dbchristopher: How would you feel about making this relevant callback be `onChange` instead? We're using `onChange` for the rest of our form elements, so it might prevent some confusion if we stick to the same terminology

dbchristopher: I had to look this up too, but apparently `onselect` is the html callback for when you highlight some text, not for when you interact with a select input. So that would make this prop name doubly confusing. 

https://www.w3schools.com/jsref/event_onselect.asp

packages/playground/src/Select/SelectDemo.tsx
```diff
@@ -0,0 +1,88 @@
...
+  { data: { id: 1 }, value: 'Apples' },
+  { data: { id: 2 }, value: 'Bananas' },
+  { data: { id: 3 }, value: 'Oranges' },
+  { data: { id: 4 }, value: 'Pineapples' },
+  { data: { id: 5 }, value: 'Kiwis' },
+]
+const listProps = { persistSelection: true }
+
+export function SelectDemo() {
+  const { inputProps, ...selectProps } = useControlledSelect('Apples')
```
dbchristopher: This seems like it will be a big source of confusion to other developers. I feel like you shouldn't have to know about this hook in order to use the controlled component pattern on the Select component. Seems like you should just be able to pass in a change handler and a value and the rest of it should _just work_ as if it was a standard form input (I realize that's easier said than done 😖)

I feel like part of the complexity comes from the fact that the component allows you to pass in text input handlers separate from the select behavior. I kind of feel like you should just make the input text box untouchable from the outside, and that _should_ simplify the api surface area and mitigate some potential gotchas 🤞

packages/playground/src/Select/SelectDemo.tsx
```diff
@@ -0,0 +1,88 @@
...
+        openOnFocus={true}
+        width={240}
+        mb="medium"
+        inputProps={inputPropsMemo}
+        listProps={listProps}
+        options={options}
+        aria-label="Fruits"
+        {...selectProps}
+      />
+      <Paragraph>{selectProps.value}</Paragraph>
```
dbchristopher: This might be too much for V1, but do you feel like it's reasonable to enforce that the selected value is actually on the provided list? You can apparently type in anything you want and it will treat that like a standard selected option, which might not be helpful if this is used to collect form data

Screen Shot 2019-12-12 at 12 14 06 PM

packages/components/src/Form/Inputs/Select/Select.tsx
mdodgelooker: @dbchristopher very good point. My only concern is how to make it clear which callback happens when the user actively selects an option, and which happens as they type (the `onChange` on `SelectInput` or `inputProps.onChange` for the child-less version). Do you think having two distinct `onChange`s is ok, given enough documentation?

mdodgelooker:
@dbchristopher thanks for reviewing! There actually wasn't any filtering logic yet, just the (buggy) highlighting. Over on https://reacttraining.com/reach-ui/combobox/ the filtering in the examples is external to the `Combobox` components, probably because of the compositional nature. Since we offer a non-compositional approach as well, an internal filtering feature is probably in order.

But this PR is already big, and it has feature parity with our current `Select`, so I'd rather leave filtering for a future PR. @lukelooker thoughts?

I will see if I can fix the case-sensitivity and navigate-to-matching-option issues you brought up above though.

packages/components/src/Form/Inputs/Select/Select.tsx
dbchristopher: Yeah I can understand the confusion between having multiple `onChange`s to deal with! Per [my other comment about useControlledSelect](https://github.com/looker-open-source/components/pull/213#discussion_r357349214) it feels to me like exposing both the input box and select api adds unnecessary complexity. 

Does it seem viable to just make the input textbox basically untouchable? Given how massive and complex this component is, it seems like people will have a better experience if we just dial in all the useful defaults rather than keeping every aspect of it extensible. What do you think?

mdodgelooker:
Closing this in favor of https://github.com/looker-open-source/components/pull/316

PR #223, opened 2019-12-06
IconButton Default Tooltips

lukelooker:
### :sparkles: Changes

- IconButton now displays a Tooltip with the IconButton’s label by default
 - `tooltipDisabled` - don’t display a tooltip by default (use of this is strongly discouraged)
 - `tooltipPlacement` - allow the developer to explicitly specify placemnet of tooltip.

Small Fixes:
- Set base zindex value for Gatsby pages to `1` to get on top of react-live stacking layer
- Banner now offers better default tooltip on “Dismiss [Warning]” instead of just “Dismiss”
- Took MenuSearch “Clear” IconButton out of default tab index

- [x] PR is ideally < 400LOC


packages/components/src/Form/Inputs/InputSearch/InputSearch.tsx
```diff
@@ -112,10 +113,16 @@ const InputSearchComponent = forwardRef(
 
       ...inputProps
     } = props
+
+    const ref = forwardRef || useRef<null | HTMLInputElement>(null)
```
mdodgelooker: If someone uses a callback ref on `InputSearch` the `focusInput` functionality wouldn't work. You could do:
```
const internalRef = useRef<null | HTMLInputElement>(null)
const ref = useForkedRef<HTMLInputElement>(internalRef, forwardedRef)
```
and then below...

packages/components/src/Form/Inputs/InputSearch/InputSearch.tsx
```diff
@@ -112,10 +113,16 @@ const InputSearchComponent = forwardRef(
...
     const [inputValue, setValue] = useState(value)
 
+    const focusInput = () =>
+      ref && !isFunction(ref) && ref.current && ref.current.focus()
```
mdodgelooker: `internalRef.current && internalRef.current.focus()`

packages/components/src/Form/Inputs/InputSearch/InputSearch.tsx
lukelooker: Ah... didn;t even occur to me. :)

PR #239, opened 2019-12-10
Make Menu easier to use

mdodgelooker:
### :sparkles: Changes

- Replace `cloneElement` with `MenuContext` for menu state and `MenuItemStyleContext` for `MenuItem` style inheritance (`compact` and `customizationProps` props). 
  - Fixes the issue where any intermediary component (for functionality, styling, etc) between `Menu`, `MenuList`, `MenuGroup`, and `MenuItem` breaks the menu functionality if it doesn't pass down certain props.
  - Makes it easier to close the menu via `MenuContext`'s `setOpen`.
- Add a separate context `MenuItemStyleContext` for managing `MenuItem` style inheritance. Styles can be controlled via `compact` and `customizationProps` at both the `MenuList` and `MenuGroup` level, so we merge these props from the parent context (if it exists) with the nested context.
- Add `escapeWithReference` to both `Popover` and `MenuList` to allow better placement control (with corresponding docs update). 
- This refactor revealed some bugginess in `Tooltip` and `Popover` that inspired me to use my new `useCallbackRef` in both, which plays much more nicely with `useEffect` than `useRef`.
- **Still outstanding: `Menu` should support `Popover`'s `hoverDisclosureRef`.**

- [ ] PR is ideally < 400LOC


packages/components/src/Menu/Menu.tsx
```diff
@@ -24,48 +24,45 @@
...
   /**
-   * Initial state of Menu
+   * Initial state of Menu (or use for controlled menu)
    * @default false
    */
   isOpen?: boolean
+  /**
+   * Use for controlled menu
+   */
+  setOpen?: (isOpen: boolean) => void
```
mdodgelooker: Adding for controlled usage, since `MenuContext` makes `useMenu` moot (`Menu` is always required, for `<MenuContext.Provider>`.

packages/components/src/Menu/MenuContext.ts
```diff
@@ -0,0 +1,42 @@
...
+  setOpen?: (isOpen: boolean) => void
+  triggerElement?: HTMLElement | null
+  triggerCallbackRef?: (node: HTMLElement | null) => void
+}
+
+const menuContext: MenuContextProps = {}
+const menuItemStyleContext: MenuSharedProps = {}
+
+export const MenuContext = createContext(menuContext)
+export const MenuItemStyleContext = createContext(menuItemStyleContext)
```
mdodgelooker: This needs to be a separate context since it can be re-established at deeper levels (I.e. the props on a `MenuGroup` differ from those on the `MenuList`).

packages/components/src/Menu/Menu.tsx
```diff
@@ -24,48 +24,45 @@
...
-  const menu = useMenu(disabled, isOpen)
+export const Menu: FC<MenuProps> = ({
+  children,
+  disabled,
+  isOpen: controlledIsOpen = false,
+  setOpen: controlledSetOpen,
+}) => {
+  const isControlled = useRef(controlledSetOpen !== undefined)
+  const [isOpen, setOpen] = useState(controlledIsOpen)
+  const [triggerElement, triggerCallbackRef] = useCallbackRef()
```
mdodgelooker: Going this route for refs more and more for any element that needs to be "seen" in a `useEffect`. Since `triggerElement` above is React state, it will actually have an impact in `useEffect` dependency arrays, while a mutable object generated by `useRef` won't, even when it's `current` value has changed.

packages/components/src/Menu/MenuItem/MenuItem.tsx
```diff
@@ -52,6 +54,24 @@ export interface MenuSharedProps {
...
+  let customizationProps = parentCustomizations || propCustomizations
+  if (customizationProps && parentCustomizations) {
+    customizationProps = merge({}, parentCustomizations, propCustomizations)
+  }
+
+  const compact = compactProp === undefined ? contextCompact : compactProp
+
+  return { compact, customizationProps }
+}
+
```
mdodgelooker: Merging `compact` and `customizationProps` from props with those from context. Whee.

packages/components/src/Menu/MenuList.tsx
```diff
@@ -172,16 +143,16 @@ export const MenuList = styled(MenuListInternal)`
   ${maxHeight}
   ${height}
 
+  min-width: 12rem;
   ${minWidth}
   ${maxWidth}
   ${width}
 
   overflow: auto;
+  border-radius: inherit;
```
mdodgelooker: This, as well as a couple other changed lines were to solve for the `MenuItem`'s background color (white or grey usually) breaks through the nicely rounded corners of `Popover`. My first fix attempt was to just put `overflow: hidden` on `OverlaySurface` > `Inner` but that felt like too big of a hammer. Thoughts?

packages/components/src/utils/useCallbackRef.ts
```diff
@@ -1,12 +1,12 @@
...
 
-export function useCallbackRef(
-  element?: HTMLElement
-): [HTMLElement | null, (node: HTMLElement | null) => void] {
-  const [currentElement, setCurrentElement] = useState(element || null)
-  const callbackRef = useCallback(
-    (node: HTMLElement | null) => setCurrentElement(node),
-    []
-  )
+export function useCallbackRef(): [
```
mdodgelooker: I previously had an optional "initial element" in here which in this refactor I tried to actually use and found to be quite buggy, so I removed it.

PR #255, opened 2019-12-11
restructure of Tab  to accommodate focus when using  mouse or  keyboard

laraharrow:
✨ Changes
- update Tabs for focus
- Tabs do not show focus rings on mouse click. only when using the keyboard.

✅ Requirements
- [x] PR is ideally < 400LOC

📷 Screenshots
tabs


packages/components/src/Tabs/Tab.tsx
```diff
@@ -99,7 +89,15 @@ export const Tab = styled.button.attrs((props: TabProps) => ({
...
+    ${props =>
+      props.focusVisible
+        ? `box-shadow: 0 0 0 0.15rem ${rgba(
+            props.theme.colors.palette.purple300,
+            0.25
+          )};
+          outline: none;`
+        : `box-shadow: none;
+          outline: none;`};
   }
```
dbchristopher: I think we'd be better off if this wasn't wrapped in the `&:focus {}` pseudo-selector. Seems like it will be easier to debug if there is no chance of browser focus and the focusVisible prop trying to do two different things

dbchristopher: I'd just leave the `outline: none` as a static value too since we'll never want the default browser focus ring rendering anyways. Example:

```
&:focus {
  outline:none;
}

${({ focusVisible })=>
      focusVisible && `box-shadow: 0 0 0 0.15rem ${rgba(props.theme.colors.palette.purple300, 0.25)};`
}
```

packages/components/src/Tabs/Tab.tsx
```diff
@@ -112,4 +110,48 @@ export const Tab = styled.button.attrs((props: TabProps) => ({
...
-Tab.defaultProps = { fontWeight: 'semiBold', pb: 'small', pt: 'xsmall' }
+const TabJSX = forwardRef((props: TabProps, ref: Ref<HTMLButtonElement>) => {
+  const { children, onBlur, onKeyDown, ...restProps } = props
+
+  const [isFocusVisible, setFocusVisible] = useState(false)
+
+  const handleOnKeyUp = (event: React.KeyboardEvent<HTMLButtonElement>) => {
+    setFocusVisible(true)
+    onKeyDown && onKeyDown(event)
+  }
```
dbchristopher: It would be good to stay consistent about whether this is a keyUp or keyDown event handler!

packages/components/src/Tabs/Tab.tsx
```diff
@@ -112,4 +110,48 @@ export const Tab = styled.button.attrs((props: TabProps) => ({
...
+
+  const handleOnBlur = (event: React.FocusEvent<HTMLButtonElement>) => {
+    setFocusVisible(false)
+    onBlur && onBlur(event)
+  }
+
+  const onClick = () => {
+    if (!props.disabled && props.onSelect) {
+      props.onSelect()
+    }
```
dbchristopher: For consistency's sake I think it would be good to destructure `disabled` and `onSelect` from the props object too

packages/components/src/Tabs/Tab.tsx
```diff
@@ -112,4 +110,48 @@ export const Tab = styled.button.attrs((props: TabProps) => ({
...
+      ref={ref}
+    >
+      {children}
+    </TabStyle>
+  )
+})
+
+TabJSX.displayName = 'TabJSX'
+
+export const Tab = styled(TabJSX)``
```
dbchristopher: You can actually define all the styles here and kill the TabStyle component altogether. When you wrap a component in styled() all it does is take the associated style definitions and translates that to a `className` prop to get passed into `TabJSX`. 

```
const TabJSX = forwardRef((props: TabProps, ref: Ref<HTMLButtonElement>) => {
   ...etc

  return (
    <button
      focusVisible={isFocusVisible}
      onKeyUp={handleOnKeyUp}
      onBlur={handleOnBlur}
      onClick={onClick}
      {...restProps}
      ref={ref}
    >
      {children}
    </button>
  )
})

TabJSX.displayName = 'TabJSX'

export const Tab = styled(TabJSX)<TabProps>`
  ${reset}
  ${space}
  ${layout}
  ${border}
  ${typography}

   ...etc
`
```

packages/components/src/Tabs/Tab.tsx
dbchristopher: I also think it makes sense to run `setFocusVisible(false)` in the click handler too. It seems like clicking on an already focused element would clear the focus ring style:

click-focus

packages/components/src/Tabs/Tab.tsx
laraharrow: good point, I kept the focusVisible  a little different than your suggestion because I was getting a complaint on where was `props` coming from.

```
  &:focus {
    outline: none;
  }

  ${props =>
    props.focusVisible &&
    `box-shadow: 0 0 0 0.15rem ${rgba(
      props.theme.colors.palette.purple300,
      0.25
    )};`}
```

packages/components/src/Tabs/Tab.tsx
laraharrow: oops good point

packages/components/src/Tabs/Tab.tsx
jhardy: +1 this was my only piece of feedback from a design side.

laraharrow: Adding setFocusVisible(false) is a great idea!
As for the destructure can we chat about it I'm not sure it works the way I tried :/

packages/components/src/Tabs/Tab.tsx
laraharrow: oh cool

packages/components/src/Button/size.ts
```diff
@@ -65,6 +65,7 @@ export const buttonSize = variant({
   variants: {
     xxsmall: {
       fontSize: 'xsmall',
+      lineHeight: '8px',
```
dbchristopher: is this the fix for [LENS-453](https://looker.atlassian.net/browse/LENS-453)? Was it meant to be committed in here?

packages/components/src/Tabs/Tab.tsx
```diff
@@ -84,12 +75,104 @@ export const Tab = styled.button.attrs((props: TabProps) => ({
...
+/*
```
dbchristopher: you can kill this massive comment block now!

packages/components/src/Tabs/Tabs.test.tsx
```diff
@@ -192,3 +192,28 @@ test('hooks working', () => {
...
+            <TabPanel>this is tab1 content</TabPanel>
+            <TabPanel>this is tab2 content</TabPanel>
+          </TabPanels>
+        </Tabs>
+      </>
+    </ThemeProvider>
+  )
+
+  fireEvent.click(getByText('tab1'))
+  expect(getByText('tab2')).toMatchSnapshot()
```
dbchristopher: should these both be selecting `tab1`?

packages/components/src/Tabs/Tabs.test.tsx
```diff
@@ -192,3 +192,28 @@ test('hooks working', () => {
...
+
+  fireEvent.keyUp(getByText('tab2'), { charCode: 9, code: 9, key: 'Tab' })
+  expect(getByText('tab2')).toMatchSnapshot()
```
dbchristopher: So I was just checking out the rendered snapshot code, and it appears that none of the compiled classes actually include the relevant box shadow 😱

This is pure speculation on my part, but I _think_ maybe you can only run one styled component snapshot per test. I tried splitting this assertion into two separate test cases and the snapshots looked more accurate to what I'd expect to see:

```
describe('focus behavior', () => {
  const TabTest = () => (
    <ThemeProvider theme={theme}>
      <>
        <Tabs>
          <TabList>
            <Tab>tab1</Tab>
            <Tab>tab2</Tab>
          </TabList>
          <TabPanels>
            <TabPanel>this is tab1 content</TabPanel>
            <TabPanel>this is tab2 content</TabPanel>
          </TabPanels>
        </Tabs>
      </>
    </ThemeProvider>
  )

  test('Tab Focus: does not render focus ring after click', () => {
    const { getByText } = render(<TabTest />)

    fireEvent.click(getByText('tab1'))
    expect(getByText('tab1')).toMatchSnapshot()

    fireEvent.keyUp(getByText('tab2'), { charCode: 9, code: 9, key: 'Tab' })
    expect(getByText('tab2')).toMatchSnapshot()
  })

  test('Tab Focus: renders focus ring for keyboard navigation', () => {
    const { getByText } = render(<TabTest />)

    fireEvent.keyUp(getByText('tab2'), { charCode: 9, code: 9, key: 'Tab' })
    expect(getByText('tab2')).toMatchSnapshot()
  })
})
```

packages/components/src/Button/size.ts
laraharrow: oops no not at all

dbchristopher:
oh yeah @laraharrow don't forget to update the changelog here too!

packages/components/src/Tabs/Tabs.test.tsx
```diff
@@ -192,3 +192,39 @@ test('hooks working', () => {
...
+  )
+
+  test('Tab Focus: does not render focus ring after click', () => {
+    const { getByText } = render(<TabTest />)
+
+    fireEvent.click(getByText('tab1'))
+    expect(getByText('tab1')).toMatchSnapshot()
+
+    fireEvent.keyUp(getByText('tab2'), { charCode: 9, code: 9, key: 'Tab' })
+    expect(getByText('tab2')).toMatchSnapshot()
```
dbchristopher: oh woops @laraharrow these lines should be deleted since we're testing this in the next assertion block. sorry that was my bad in the example snippet I pasted in!

PR #276, opened 2019-12-13
Local Dev Proxy

dbchristopher:
Draft PR for setting up data fetching without needing to query against a local [primary backend repo] instance.

## Setup Instructions: 
1. rename `packages/server/.env-sample` -> `packages/server/.env`
1. rename `packages/server/looker-sample.ini` -> `packages/server/looker.ini`
    * fill in `base_url`, `client_id`, and `client_secret` values in looker.ini
1. `yarn && yarn workspace server start`

Then start playground and www like normal. Data query demos can be viewed at these urls: 
* **WWW:** http://localhost:8000/data/get-me 
* **Playground:** http://localhost:3000

## Next Steps

- [ ] Figure out what we're actually able to deploy to production (need a live nodejs server to navigate CORS issues)
- [ ] Figure out which parts of the SDK documentation is relevant to Looker/Components users
- [ ] set up `useSpace` -- get a list of looks and render in cards
- [ ] Finalize setup documentation

## Screenshots

www-demo


packages/server/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig.json",
+  "include": ["src", "types/*.d.ts"],
+  "compilerOptions": {
+    "module": "commonjs",
```
dbchristopher: commonjs modules allow the typescript express server to be able to parse `import` statements rather than needing to use `require`

packages/server/src/index.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const app = express()
+app.use(bodyParser.urlencoded({ extended: false }))
+app.use(pino())
+
+app.get('/api/me', async (_: any, res: any) => {
+  const me = await fetchUserData()
+  res.setHeader('Content-Type', 'application/json')
+  res.send(JSON.stringify({ me }))
+})
```
dbchristopher: Once the express server is running, this is how you configure a new REST endpoint (`/api/me`) in this case

lukelooker: Would it be possible to just pass everything through, basically glob match `/api/*` and then just return the response that come from the server?

packages/server/src/fetchUserData.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Initialized SDK object
+ * @type {LookerSDK} SDK object configured for use with Node
+ */
+const sdk = new LookerSDK(session)
+
+const fetchUserData = async () => {
+  const userFields =
+    'id, first_name, last_name, display_name, email, personal_space_id, home_space_id, group_ids, role_ids'
+  // retrieve your user account to verify correct credentials
+  const me = await sdk.ok(sdk.me(userFields))
```
lukelooker: Simply skipping the fields here (just calling `sdk.me()`) will return everything - seems like that might be simpler?

packages/server/src/forwardRequest.ts
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Initialized SDK object
+ * @type {LookerSDK} SDK object configured for use with Node
+ */
+const sdk = new LookerSDK(session)
+
+const forwardRequest = async (req: any) => {
+  const endpoint = req.url.replace('/api/', '')
+  const apiRoot = (sdk.stream.authSession.transport as any).apiPath
```
dbchristopher: I had to do this blunt type coercion because the ITransport interface appears to be out of date. If I try to get `transport.apiPath` directly, TS complains that "Property 'apiPath' does not exist on type 'ITransport'". When I run `console.log(sdk.stream.authSession.transport)` this is the output:

NodeTransport {
  options: NodeSettingsIniFile {
    base_url: 'https://app.dev.looker.com:19999',
    api_version: '3.1',
    verify_ssl: true,
    timeout: 120,
    fileName: 'looker.ini'
  },
  apiPath: 'https://app.dev.looker.com:19999/api/3.1'
}

packages/server/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/looker-open-source/components",
+    "directory": "packages/server"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "start": "node-env-run src --exec nodemon | pino-colada"
```
dbchristopher: should this command be `develop` to be more symmetrical with the rest of the monorepo commands?

lukelooker: Yeah - I think that's probably a good call. Also helps infer that this code is probably not appropriate for a "production" environment :)

packages/playground/src/data/GetMe.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+import React, { useState } from 'react'
+import { Button, List, ListItem, Heading } from '@looker/components'
+import styled from 'styled-components'
+
+export const GetMe = () => {
+  const [user, setUser] = useState<any>({})
+
+  function fetchUser() {
+    fetch('/api/user?fields=id,display_name,email,home_space_id,role_ids')
```
lukelooker: What do you think of using the terse-r version:

```fetch('/api/user')```

File: packages/www/src/documentation/data/demos/GetMe.tsx [hunk 7]
```diff
@@ -0,0 +1,84 @@
...
+
+import React, { useState } from 'react'
+import { Button, List, ListItem, Heading } from '@looker/components'
+import styled from 'styled-components'
+
+export const GetMe = () => {
+  const [user, setUser] = useState<any>({})
+
+  function fetchUser() {
+    fetch('/api/user?fields=id,display_name,email,home_space_id,role_ids')
```
lukelooker: Again, I think I'd prefer: 

```    fetch('/api/user')```

File: packages/www/src/documentation/data/get-me.mdx [hunk 8]
```diff
@@ -0,0 +1,9 @@
+---
+title: Get Me
+---
+
+import { GetMe } from './demos'
+
+Fetch the logged in user
```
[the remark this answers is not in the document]
dbchristopher: Oh I'm actually not really sure this will be the best or most interesting example to start out with for. I think I might delete it altogether and build that List of Spaces idea you initially suggested. 

Also I really hope that we're can at least figure out a general idea for the production proxy this week if possible. Ideally I'd prefer not releasing this until it can run both locally and on www. Does that seem totally unrealistic to you?

packages/server/README.md
```diff
@@ -0,0 +1,42 @@
+# Looker Components Server
+
+A local proxy server for requesting Looker data in development. You must have account authorization to a live/production Looker instance to be able to run this server.
+
+## Setup Instructions
+
+### Configure server
+
+1. Rename `packages/server/.env-sample` -> `packages/server/.env`
```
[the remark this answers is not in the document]
dbchristopher: I guess it was just a more defensive defensive -- in general sensitive information is kept in `.env` files, so it would be safer not to check it in to the repo. But since we only have plans to keep the client secrets in looker.ini, I can go ahead and add the stub .env file

packages/playground/src/data/GetMe.tsx
dbchristopher: sure that's fine! The intent was just to show that you can still pass arguments to the standard api endpoints, but that's probably not relevant for what lens users need to get started

packages/server/src/forwardRequest.ts
jkaster: We're still discussing whether or not it's needed, but we can discuss it in `sdk` instead if you want.

packages/server/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+ */
+
+import express, { Request, Response } from 'express'
+import bodyParser from 'body-parser'
+import pino from 'express-pino-logger'
```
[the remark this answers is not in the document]
dbchristopher: yep! that's right, just helpful for debugging

dbchristopher:
> I couldn't quite figure out how to get node-fetch to stop complaining about hitting at "self-signed certificate" on my local instance but it worked fine with a *.looker.com instance.

Oh wow yeah I've never actually tried this against a local instance. I have it set up to hit https://app.dev.looker.com/ at the moment, and I hope this will mean less local setup is needed in general 😄 🙏

jkaster:
`verify_ssl=false` will cause the NodeTransport to ignore self-signed cert errors, but I don't try it with `localhost`.

packages/www/src/documentation/data/get-me.mdx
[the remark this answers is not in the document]
lukelooker: I don't think this is a terrible "getting started" example honestly since it doesn't require any iteration and I don't think there's any case where you won't have sufficient rights to get that API response :)

lukelooker:
> > I couldn't quite figure out how to get node-fetch to stop complaining about hitting at "self-signed certificate" on my local instance but it worked fine with a *.looker.com instance.
> 
> Oh wow yeah I've never actually tried this against a local instance. I have it set up to hit https://app.dev.looker.com/ at the moment, and I hope this will mean less local setup is needed in general 😄 🙏

Agreed, totally, looking forward to not _having_ to run a local instance :)

lukelooker:
> `verify_ssl=false` will cause the NodeTransport to ignore self-signed cert errors, but I don't try it with `localhost`.

Yeah apparently `self-signed` is getting caught by something else. No need to worry about it for now - was just a note if others run into it.

packages/www/src/documentation/data/get-me.mdx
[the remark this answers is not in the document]
dbchristopher: Yeah that's true, it is nice in it's simplicity! But I guess it brings up another question -- what do you see as the value of this example in the first place? That context would help me figure out more about what I need to write for this documentation.

For me, this example just proved that _data fetching works_. Do lens consumers need that proof? I guess if this is meant to be a getting started guide it would be helpful to double check that they got the looker.ini values filled in correctly. But I thought that getting started stuff would more likely live in the package README?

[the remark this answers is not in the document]
dbchristopher: On further reflection, I don't see much value in fetching the list of Spaces or Looks either 😆 I went ahead and moved this user fetching example to `/getting-started/data-fetching`. 

Beyond that I think what makes the most sense is digging into the planned visualization components, which would live in under the `Components` section. I'll wrap up this PR and then sync up with Justin on what it would take to port over a basic Single Value vis or something

package.json
```diff
@@ -21,6 +21,11 @@
   "scripts": {
     "build": "yarn lerna run 'build:all'",
     "postbuild": "run-p -c 'lint'",
+    "develop": "npm-run-all develop-components sleep --parallel develop-server develop-playground develop-www",
```
dbchristopher: I'm getting real tired of having to boot up like four different terminal windows to develop lens. I added this convenient script powered by `npm-run-all` that condenses everything to a one line command. 

Now you can run `yarn develop` and it will do a fresh asset build, start watching for component changes, and then after as short delay boot up playground, www, and the proxy server. @lukelooker want to give it a shot and let me know what you think?

package.json
```diff
@@ -21,6 +21,11 @@
...
+    "develop-components": "yarn workspace @looker/components develop &",
```
dbchristopher: I had to kick this command to the background by appending the `&` argument. 

I seem to find that if I want to actively develop components I need the watch command fully booted up before starting www or playground. Since the watch command will never "complete", npm-run-all never gets to go on to the next steps in the startup sequence.  So my best solution was to run the components watch command in the background, then after a short artificial delay (the `sleep` command) allow it to boot up the rest of the packages in parallel.

package.json
```diff
@@ -21,6 +21,11 @@
   "scripts": {
     "build": "yarn lerna run 'build:all'",
     "postbuild": "run-p -c 'lint'",
+    "develop": "npm-run-all build develop-components sleep --parallel develop-server develop-playground develop-www",
```
dbchristopher: I'm getting real tired of having to boot up like four different terminal windows to develop lens. I added this convenient script powered by `npm-run-all` that condenses everything to a one line command. 

Now you can run `yarn develop` and it will do a fresh asset build, start watching for component changes, and then after as short delay boot up playground, www, and the proxy server. @lukelooker want to give it a shot and let me know what you think? If you approve I'll add a note about it to the README

dbchristopher:
Ok @lukelooker I kicked this PR out of `Draft` state and it should be ready for final review!

PR #314, opened 2019-12-19
updated documentation for better user experience

laraharrow:
✨ Changes

- update documentation and chagelog on useConfirm for better user experience

✅ Requirements
- [x] PR is ideally < 400LOC

📷 Screenshots

Screen Shot 2019-12-19 at 12 35 06 PM

### DIFF
--- a/packages/www/src/documentation/components/modals/confirm.mdx
+++ b/packages/www/src/documentation/components/modals/confirm.mdx
@@ -48,7 +48,6 @@ A custom hook that returns the opener function and rendered confirmation dialog.
 
 ```jsx
 ;() => {
-  const [keyPressed, setKeypressed] = React.useState('none')
   function handleConfirm(close) {
     alert('You deleted it')
     close()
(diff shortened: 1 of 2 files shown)


packages/www/src/documentation/components/modals/confirm.mdx
```diff
@@ -4,63 +4,71 @@ propsOf: Confirm
...
-    }
-  }, [confirm])
+
   return (
-    <Paragraph>
+    <>
       {confirmation}
-      To trigger a confirmation dialog, hit the delete or backspace key.
-    </Paragraph>
+      <ButtonTransparent onClick={confirm}>useConfirm</ButtonTransparent>
```
dbchristopher: Any reason to use the transparent button here instead of the normal button? It just looks less clickable to me, so it might make the example a little more confusing to t he end user

laraharrow: I'll check with Luke on that, it was his idea :)  I agree with you!

laraharrow: I updated it to be `ButtonOutline`

PR #372, opened 2020-01-10
Lara/chip/lens 510

laraharrow:
✨ Changes

- create Chip component

✅ Requirements
- [x] PR is ideally < 400LOC

📷 Screenshots

Screen Shot 2020-01-13 at 3 17 14 PM


packages/components/src/Chip/Chip.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  fireEvent.click(getByRole('button'))
+  expect(onDeleteTrigger).toHaveBeenCalled()
+
+  fireEvent.keyDown(getByRole('button'), {
+    charCode: 8,
+    code: 8,
+    key: 'Backspace',
+  })
+  expect(onDeleteTrigger).toHaveBeenCalled()
```
mdodgelooker: To make sure this was called again after the backspace keydown, use `toHaveBeenCalledTimes(2)` here instead of `toHaveBeenCalled()`.

laraharrow: oh make sense!
Thanks

PR #382, opened 2020-01-10
[LENS-502] Protect modal input changes from being accidentally discarded

dbchristopher:
### :sparkles: Changes

- Remove generic `close` callback in favor of more specific `onChange` and `onCancel` callbacks. We need to be more specific about which action is closing the modal in order to know whether to render the confirmation dialog.
- Update Confirm to update state based on change events
- Create a new DiscardChangesDialog to render as fallback if changes are about to be lost
- Update modal title to allow an icon to be rendered along with title text ]

- [ ] PR is ideally < 400LOC

### :camera: Screenshots
confirm-close-dialog

### DIFF
--- a/packages/components/src/Modal/Dialog/Confirm.tsx
+++ b/packages/components/src/Modal/Dialog/Confirm.tsx
@@ -24,7 +24,7 @@
 
  */
 
-import React, { FC, ReactNode, useState, SyntheticEvent } from 'react'
+import React, { FC, ReactNode, useState } from 'react'
 import isFunction from 'lodash/isFunction'
 import { useToggle } from '../../utils/useToggle'
 import { ConfirmationDialog, ConfirmationProps } from './ConfirmationDialog'
--- a/packages/components/src/Modal/Dialog/Confirm.tsx
+++ b/packages/components/src/Modal/Dialog/Confirm.tsx
@@ -3,17 +3,17 @@
  MIT License
 
  Copyright (c) 2019 Looker Data Sciences, Inc.
- 
+
  Permission is hereby granted, free of charge, to any person obtaining a copy
  of this software and associated documentation files (the "Software"), to deal
  in the Software without restriction, including without limitation the rights
  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
  copies of the Software, and to permit persons to whom the Software is
  furnished to do so, subject to the following conditions:
- 
+
  The above copyright notice and this permission notice shall be included in all
  copies or substantial portions of the Software.
- 
+
  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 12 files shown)


dbchristopher:
@lukelooker I have a bit more work to do on this: primarily I still need to catch up DialogManager to use utilize the new functionality in `<Confirm />`. But I've had my hands in this for a while now, and I thought it would be good to get some eyes on where I'm at so far.

packages/components/src/Modal/Dialog/Confirm.tsx
```diff
@@ -24,34 +24,111 @@
...
+export interface ConfirmProps extends Omit<ConfirmationProps, 'onCancel'> {
   /**
    * Render prop is passed the confirmation opener
    */
   children: (open: () => void) => ReactNode
+  /**
+   * If true, watch for child input changes. Prompt user to confirm cancel action if there are unsaved changes.
+   * @default false
+   */
+  protectChanges?: boolean
```
lukelooker: It seems like these features should actually go on `DialogManager` (`ModalManager` I guess). In my mind `Confirm` is just supposed to be a replacement for classic `window.confirm` behavior (https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm).

packages/components/src/Modal/Dialog/DiscardChangesDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <ModalFooter>
+        <ButtonTransparent onClick={closeDialog} color="danger">
+          Discard Changes
+        </ButtonTransparent>
+        <ButtonTransparent onClick={resetDialog} color="neutral">
+          Go Back
+        </ButtonTransparent>
+      </ModalFooter>
+    </Dialog>
+  )
```
lukelooker: All of these strings will need to be i18n'able so we should make them variables that can be sent in via the interface I guess?

packages/components/src/Modal/Dialog/DiscardChangesDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+}
+
+const MessageGrid = styled.div`
+  display: grid;
+  grid-gap: 1rem;
+  grid-template-columns: auto 1fr;
+`
```
lukelooker: It's not clear to me why this `MessageGrid` component is needed.

packages/components/src/Modal/Layout/ModalHeader.tsx
```diff
@@ -35,50 +35,76 @@ import { IconNames } from '@looker/icons'
...
   closeIcon?: IconNames
-  children: string
+  /**
+   * Render an icon next to the title text
+   */
+  headerIcon?: IconNames
+  /**
+   * Specify color of header icon
+   */
+  headerIconColor?: string
```
lukelooker: Neat! Might be good to run this by @jhardy to see if / how to fit this into the Lens style and if any matching changes are needed in Figma if he agrees with this behavior.

packages/components/src/Modal/Layout/ModalHeader.tsx
```diff
@@ -35,50 +35,76 @@ import { IconNames } from '@looker/icons'
...
+      />
     </Header>
   )
 }
 
 const Header = styled.header<SpaceProps>`
   ${reset}
   ${space}
+  display: grid;
+  grid-template-columns: [icon] auto [text] 1fr [close] auto;
```
lukelooker: Clever ;)

packages/components/src/Modal/ModalSurface.tsx
```diff
@@ -51,62 +51,65 @@ export interface ModalSurfaceProps
...
-      <Style
-        className={`surface-overflow ${className}`}
-        tabIndex={-1}
-        surfaceStyle={style as CSSObject}
-        {...props}
-      />
-    </HotKeys>
-  )
-}
+export const ModalSurface = forwardRef(
```
lukelooker: Why does this need to be forwarded?

lukelooker: I get this now reading through the code more (needed to do field-change auto-discovery)

packages/components/src/Modal/Dialog/DiscardChangesDialog.tsx
dbchristopher: Oh yeah, that's a good catch! Have we actually done any translation service integration in Lens?  Do you have an example of how that works?

packages/components/src/Modal/Dialog/DiscardChangesDialog.tsx
dbchristopher: `MessageGrid` handles the layout that groups the Warning Icon and Title, while right aligning the close button:

Screen Shot 2020-01-13 at 11 35 32 AM

The Flow component was evenly spacing the the three elements which of course isn't the right layout. And this switch to CSS grid was less code than trying to have a flow within a flow to achieve that same layout

packages/components/src/Modal/Layout/ModalHeader.tsx
dbchristopher: This new modal header + icon layout was one of Jared's designs in figma... but now I'm not seeing it in the the figma file anymore. @jhardy is this still a design-approved dialog layout?

Screen Shot 2020-01-13 at 11 40 52 AM

packages/components/src/Modal/Dialog/DiscardChangesDialog.tsx
dbchristopher: OH WAIT! Sorry I was thinking of a totally different grid. Sorry, I think this was actually just from an earlier design I was playing around with and then didn't end up shipping. I'll remove this

packages/components/src/Modal/ModalSurface.tsx
dbchristopher: yep that's right! but considering your feedback that this functionality should not be provided  out-of-the-box, I'll go ahead and remove all the ref forwarding changes

dbchristopher:
I'm realize with this new set of assumptions (not tracking input changes, not managing modal fallbacks in Confirm) not much of of this work is actually salvageable. I'm going to close this PR and copy `DiscardChangesDialog` to a new branch and start over from there

packages/components/src/Modal/Dialog/DiscardChangesDialog.tsx
dbchristopher: It seems like internationalization is just a feature of [INTERNAL REPO NAME], not necessarily looker components. I wonder if the only relevant package change is actually just the modal header icon support. This fallback dialog is probably just something we can use in documentation examples

lukelooker: Yep, it's really just a matter of making sure that all the strings: "Discard Changes", "Are you sure...", etc. can passed in via the component interface so the consumer can specify the correct language strings if need be.

dbchristopher: Yeah that makes sense. Just for the sake of over communicating here, the whole project is shaping up to be more of a "recipe" than an actual feature to ship. I'm basically just working on an example that can be posted in www rather than any new abstractions on DialogManager. 

I'm happy to keep pushing forward on it if you think people would actually make use of that example. But I just wanted to call that out in case you're underwhelmed by the end result and wind up feeling it wasn't worth the time.

dbchristopher:
> Some minor nits and adjustments but it's in good shape overall.

I disagree 😆 

I think I'm having trouble illustrating how DialogManager as an abstraction is blocking me here. It will be more clear when I post my followup PR. But maybe when that's live you can help me parse through the example and see what if anything makes sense to provide out-of-the-box for this functionality

PR #400, opened 2020-01-14
[LENS-502] Confirm Close Dialog Example

dbchristopher:
### :sparkles: Changes

- Create new `ConfirmationDialogContent` component to aid in laying out the common user confirmation dialog pattern
- `ModalHeader` supports an optional icon to render next to the title
- Set up an example of how it would look to confirm closing a modal with DialogManager
- Edit all modal documentation for clarity
- Add "advanced use" documentation for instructions on rendering a fallback modal

### :camera: Screenshots

can-close-modal


packages/playground/src/Confirm/CanCloseDialogManagerDemo.tsx
```diff
@@ -0,0 +1,232 @@
...
+      </ModalContent>
+      <ModalFooter>
+        <ButtonTransparent onClick={closeDialog} color="danger">
+          Discard Changes
+        </ButtonTransparent>
+        <ButtonTransparent onClick={resetDialog} color="neutral">
+          Go Back
+        </ButtonTransparent>
+      </ModalFooter>
+    </>
```
dbchristopher: This seems to be a pretty common modal layout: Title, Message, Confirm Button, Cancel Button. That could be a good candidate for abstraction --- something like `<ConfirmationDialogContent />` maybe. As all string content needs to be generated from Helltool, there isn't really a way to pre-fill this modal content like I attempted in the previous PR.

dbchristopher:
@lukelooker here's round two with the CanClose modal example. Given the nature of DialogManager and what it does/doesn't provide, I'm not seeing much opportunity to extract much of the logic onto DialogManager itself. I think I'd have to fundamentally rethink what DialogManager even does for you, which would create a huge mess in [INTERNAL REPO].

For now I think it makes the most sense to just treat this as an advanced usage example, and copying a lot of this over to the DialogManager documentation as-is. But before I go down that path I'd love to get your eyes on it to see if you can spot any opportunities to own some of the logic and simplify the use case for external users.

packages/playground/src/Confirm/CanCloseDialogManagerDemo.tsx
```diff
@@ -0,0 +1,229 @@
...
+        content={
+          <DiscardChangesDialog
+            closeDialog={handleConfirmClose}
+            resetDialog={handleDialogReset}
+          />
+        }
+        width="500px"
+        isOpen={isOpen && cancellingInput}
+        onClose={handleConfirmClose}
+      />
```
dbchristopher: given how much state lives external to the DialogManager, it seems to make the most sense to just treat them as two entirely independent dialogs. Then I need to manually decide which one to render based on form state

packages/playground/src/Confirm/CanCloseDialogManagerDemo.tsx
dbchristopher: Update: went ahead and abstracted this layout as a new component: https://github.com/looker-open-source/components/pull/400/commits/aca8d0188c2ff1b677fd931f9acdd1de92b471b4

 I named it `ConfirmationDialogContent` as it's basically what we were doing with the older `ConfirmationDialog`, but without being wrapped in `Dialog`. Definitely open to other naming suggestions though.

packages/components/src/Modal/ModalSurface.tsx
```diff
@@ -46,7 +46,7 @@ export interface ModalSurfaceProps
     BoxShadowProps,
     BackgroundColorProps,
     LayoutProps {
-  surfaceStyle?: CSSObject
+  surfaceStyles?: CSSObject
```
dbchristopher: changed this to `surfaceStyles` in order to match the prop name used in DialogManager

packages/components/src/Modal/ModalManager.tsx
```diff
@@ -35,7 +35,7 @@ import React, {
 import { ManagedModalProps } from '../Modal'
 
 export interface ModalManagerProps extends ManagedModalProps {
-  children: (onClick: () => void, ref: RefObject<any>) => ReactNode
+  children?: (onClick: () => void, ref: RefObject<any>) => ReactNode
```
dbchristopher: I made the children optional so that you have the option to control isOpen state externally, rather than being required to use the render prop

packages/playground/src/Confirm/CanCloseDialogManagerDemo.tsx
```diff
@@ -0,0 +1,210 @@
...
+            }
+          />
+        }
+        width="500px"
+        isOpen={isOpen && cancellingInput}
+        onClose={handleConfirmClose}
+      />
+    </>
+  )
+}
```
dbchristopher: Realistically it seems to me that all this markup, state, and callbacks need to be customizable. I'm really at a loss for how to create an abstraction here without creating a completely unusable mess of props. But I'm definitely open to suggestions @lukelooker! When you have a chance to take a look, I'd love to know if you think any of this should be codified onto the DialogManager

packages/components/src/Modal/Dialog/ConfirmationDialogContent.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { FC, ReactElement } from 'react'
```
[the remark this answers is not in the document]
dbchristopher: Yeah that sounds good to me! `ConfirmLayout` is much easier to type, and I like the idea of keeping `*Layout*  as the standard convention

dbchristopher:
ok published those changes!

PR #409, opened 2020-01-16
Fix focus trap issue with Select and Dialog

mdodgelooker:
### :sparkles: Changes

- Fixes an issue where clicking on an option of a `Select` inside a `Dialog` breaks the `Dialog`'s focus trap and blurs the `Select`'s, which prevents the click from registering a change.

- [x] PR is ideally < 400LOC

### DIFF
--- a/packages/components/src/Modal/Modal.tsx
+++ b/packages/components/src/Modal/Modal.tsx
@@ -95,11 +95,14 @@ export function Modal({
     callbackRef: focusRef,
     disable: disableFocusTrap,
     enable: enableFocusTrap,
+    isEnabled: focusTrapEnabled,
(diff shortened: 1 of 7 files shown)


packages/components/src/Modal/Modal.tsx
```diff
@@ -110,6 +113,9 @@ export function Modal({
         disableScrollLock,
         enableFocusTrap,
         enableScrollLock,
+        focusTrapEnabled,
+        focusTrapRef,
+        scrollLockEnabled,
```
mdodgelooker: Adding `focusTrapEnabled` and `scrollLockEnabled` state values to `ModalContext` here wasn't specifically related to this bug, just correcting an oversight.

PR #419, opened 2020-01-16
[LENS-438] Make ModalSurface maxWidth overwrite-able

dbchristopher:
### :sparkles: Changes

-  Adds responsive maxWidth prop support for `ModalSurface` and all consumers

- [x] PR is ideally < 400LOC

### DIFF
--- a/packages/www/src/documentation/components/modals/dialog.mdx
+++ b/packages/www/src/documentation/components/modals/dialog.mdx
@@ -14,7 +14,7 @@ import { Banner } from '@looker/components'
 
 Dialogs break out of the standard application flow and UI to present new information or required actions.
 
-# Standard Use
+## Standard Use
 
 `Dialog` requires that the developer manages state by assigning `true` or `false` to the `isOpen` prop.
 
@@ -35,7 +35,7 @@ Dialogs break out of the standard application flow and UI to present new informa
 }
 ```
 
-# DialogManager
+## DialogManager
 
 DialogManager gives an easy way to compose a Dialog without the need to manage open/close state.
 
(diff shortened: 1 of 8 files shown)


dbchristopher:
@mdodgelooker Here's my fix for https://looker.atlassian.net/browse/LENS-483 

I added maxWidth prop overrides. Let me know if you think that will be sufficient for the DashboardScene modal that ran into issues last time. No rush of course; happy to sit on this and ship it next week or whatever if you need to focus on summit stuff!

packages/www/src/documentation/components/modals/dialog.mdx
```diff
@@ -49,3 +49,29 @@ DialogManager gives an easy way to compose a Dialog without the need to manage s
   {onClick => <Button onClick={onClick}>Open Dialog </Button>}
 </DialogManager>
 ```
+
+## Props: width and maxWidth
+
+All variants of Dialog allow you to override content `width` and `max-width` style to suit your content. We support responsive width arrays an well as static strings.
```
lukelooker: Ok - I had to poke at this to remind myself why we support both of these so I think we should explain them. Here's a stab at it (please feel free to edit my terrible writing):

>
> By default width is unassigned as acts as `auto` sizing the Dialog surface to be wide enough to contain its content. At the same time `maxWidth` constrains the Dialog surface's width to no larger than the specified value.
> 
> `maxWidth` defaults to one of three sizes depending on the responsive breakpoint of the page (`['90vw', '90vw', '600px']`).
> 
> With those constraints in mind if you want a very width Dialog that has complex content you should use `maxWidth` and if have a `Dialog` that must be a specific width assign values to both.

dbchristopher:
ok I updated the docs with Luke's suggestion! I need another ✅whenever you have time

dbchristopher:
@lukelooker had to fix some merge conflicts here, so I'll need another ✅ when you have a moment

PR #430, opened 2020-01-17
Change children type from "string" to "string | string[]" so you can interpolate values

treshugart:

- [x] PR is ideally < 400LOC
- [x] Link(s) to related Github issues (my PR is the issue)

For small changes, I generally opt to propose them via a PR as it's less process, so here it is!

I'm proposing we change the type for `children` on `ModalHeader` to `string | string[]` so we can interpolate values.

```tsx
// You can do this:
<ModalHeader>text</ModalHeader>

// But not this because it's string[]
<ModalHeader>text {'some more text'}</ModalHeader>

// Current workaround
<ModalHeader children={`text ${'some more text'}`} />
```


lukelooker:
Tagging @dbchristopher on the review here as he worked on these recently. :)

treshugart:
> We intended that it would only accept strings as we're keeping the styling consistent by wrapping the child content in an H3.

Ok, awesome. That makes a lot of sense. I wasn't quite sure because the general convention seemed to favor passing components, so I erred on that side just in case.

Changes made and re-review requested. Thanks!

PR #452, opened 2020-01-22
DateTime components

laraharrow:
✨ Changes

- Add components Dates DateTime Times
- create basic tests
- updated documentation

✅ Requirements
- [x] PR is ideally < 400LOC

📷 Screenshots
Screen Shot 2020-01-28 at 12 03 06 PM


packages/components/src/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,83 @@
...
+  hour?: string
+  minute?: string
+  second?: string
+  timeZoneName?: string
+}
+
+export interface DateTimeProps
+  extends SpaceProps,
+    CompatibleHTMLProps<HTMLSpanElement> {
+  moment?: any
```
lukelooker: Rename to `date` - `any` as interface isn't allowed (this basically turns off Typescript)

`date?: Date`

packages/components/src/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,83 @@
...
+  location?: string
+  options?: DateTimeFormate
```
lukelooker: See ticket description... should be `style?: 'short', 'medium', 'long', 'full' "

packages/components/src/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,83 @@
...
+    minute: 'numeric',
+    month: 'long',
+    second: 'numeric',
+    timeZoneName: 'short',
+    weekday: 'long',
+    year: 'numeric',
+  },
+  ...props
+}) => {
+  return <span {...props}>{moment.toLocaleDateString(location, options)}</span>
```
lukelooker: Should not have a `<span>` wrapping - use a fragment `<>` instead

packages/components/src/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,83 @@
...
+}
+
+export const DateTime = styled(DateTimeLayout)`
+  ${reset}
+
+  ${space}
+  ${typography}
+
+  display: inline-flex;
+`
```
lukelooker: This is not needed.

packages/components/src/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+DateTime.defaultProps = {}
```
lukelooker: Should have `style: 'medium'` as a defaultProp

packages/components/src/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+export interface DateTimeProps
+  extends SpaceProps,
+    CompatibleHTMLProps<HTMLSpanElement> {
+  moment?: any
+  location?: string
+  options?: DateTimeFormate
+}
+
+const DateTimeLayout: FC<DateTimeProps> = ({
```
lukelooker: Rename to `DateTime` (doesn't need Layout)

packages/components/src/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,60 @@
...
+  children = new Date(Date.now()),
+  hideDate,
+  hideTime,
+  locale,
+  timeZone,
+  style,
+}) => {
+  const timeStyle = hideTime ? undefined : style
+  const dateStyle = hideDate ? undefined : style
+  // const TimeZone = timeZone ?
```
lukelooker: Don't need this line

packages/components/src/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,79 @@
...
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+import React, { FC } from 'react'
+
+type FormatStyles = 'short' | 'medium' | 'long' | 'full'
+
+export interface DateTimeProps extends Intl.DateTimeFormatOptions {
```
dbchristopher: I don't think you want to extend `DateTimeFormatOptions` here, do you? Did you mean for end users to pass in `toLocaleDateString` config options as component props?

packages/components/src/DateTime/DateTime.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+  assertSnapshot(<DateTime style="short" />)
+})
+
+test('DateTime renders if style is passed', () => {
+  assertSnapshot(<DateTime style="long" />)
+})
+
+test('DateTime renders if style is passed', () => {
+  assertSnapshot(<DateTime style="full" />)
+})
```
dbchristopher: I already wrote you on slack, but just leaving a note here that a lot of these need a static date passed in as the child

packages/components/src/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,79 @@
...
+  timeZone,
+}) => {
+  const options: ExtendedDateTimeFormatOptions = {
+    dateStyle,
+    timeStyle,
+    timeZone,
+  }
+
+  options.dateStyle = hideDate ? undefined : style
+  options.timeStyle = hideTime ? undefined : style
```
dbchristopher: You're overwriting the values of `dateStyle` and `timeStyle` with the `style` prop. Is that intentional?

Without these lines it looks like the `style` prop is unused altogether. So it seems like you probably don't need all three props

packages/components/src/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  if (timeZone) {
+    options.dateStyle = 'long'
+    options.timeStyle = 'long'
+  }
```
dbchristopher: This seems like it will produce confusing behavior -- the user defined style props will be overwritten if they pass in a timezone prop. Is that actually what the developers would expect when setting a timezone?

packages/components/src/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,79 @@
...
+  locale,
+  style,
+  timeStyle,
+  timeZone,
+}) => {
+  const options: ExtendedDateTimeFormatOptions = {
+    dateStyle,
+    timeStyle,
+    timeZone,
+  }
```
dbchristopher: Rather than defining this object and then mutating it, we can conditionally set the relevant object keys:

```ts
  const options: ExtendedDateTimeFormatOptions = {
    ...(!hideDate && { dateStyle }),
    ...(!hideTime && { timeStyle }),
    timeZone,
  }
```

packages/components/src/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,79 @@
...
+    options.timeStyle = 'long'
+  }
+
+  const format = children.toLocaleDateString(locale, options)
+
+  return <>{format}</>
+}
+
+DateTime.defaultProps = {
+  style: 'medium',
```
dbchristopher: I'm guessing we'd want default props for dateStyle and timeStyle here too

packages/components/src/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+import React, { FC } from 'react'
+
+type FormatStyles = 'short' | 'medium' | 'long' | 'full'
+
+export interface DateTimeProps extends Intl.DateTimeFormatOptions {
+  children?: Date
+  dateStyle?: FormatStyles
+  hideDate?: boolean
+  hideTime?: boolean
```
dbchristopher: After seeing this in use, and all the confusion about hiding/showing and different style options, I feel like would be better to deliver two separate `Date` and `Time` components. 

Rather than juggling a bunch of different props, something like this approach seems a lot easier to use:

```
<>
  <Date style="long">${timestamp}</Date>
  <Time style="short">${timestamp}</Time>
<>
```

jhardy: Do we think `style` is the right name for this prop? I feel something like `<Date format="long">` would be more semantic and doesn't remind me of applying inline styles.

dbchristopher: Oh that’s a good call! I like `format` a lot better 👍

CHANGELOG.md
```diff
@@ -5,12 +5,24 @@ All notable changes to this project will be documented in this file.
...
 
+
+## [UNRELEASED]
+
+### Added
+
+<<<<<<< HEAD
+- New components `Date` `DateTime` `Time` plus documentation and tests
+=======
+- New component `DateTime` plus documentation and tests
```
lukelooker: Remove merge conflict info here.

CHANGELOG.md
```diff
@@ -5,12 +5,24 @@ All notable changes to this project will be documented in this file.
...
+
 ## [0.7.16] - 2020-01-30
 
 ### Fixed
 
 - `Select` - Addressed integration test issue where events happen fast enough to reveal state change issue
 
+>>>>>>> 9c73767625a47f5fbe1b347ff105f32e2fd99412
```
lukelooker: And here...

packages/components/src/DateTime/DateTime.test.tsx
```diff
@@ -0,0 +1,63 @@
+import { assertSnapshot } from '@looker/components-test-utils'
+import React from 'react'
+import { DateFormat } from './DateFormat'
+import { DateTimeFormat } from './DateTimeFormat'
+import { TimeFormat } from './TimeFormat'
+
+test('DateTime renders', () => {
+  assertSnapshot(
+    <DateTimeFormat>{new Date('January 25, 1988 11:58:03')}</DateTimeFormat>
```
lukelooker: Seems like it'd be ideal to assign date value once and use it across tests:

```
const date = new Date('January 25, 1988 11:58:03')

test('DateTime renders', () => assertSnapshot(<DateTimeFormat>date</DateTimeFormat>)
```

packages/components/src/index.ts
```diff
@@ -30,6 +30,7 @@ export * from './Banner'
 export * from './Button'
 export * from './Card'
 export * from './Chip'
+export * from './DateTime'
```
lukelooker: Rename to `DateTimeFormat` since that's the common ancestor for all

packages/www/src/documentation/components/content/date-time.mdx
```diff
@@ -0,0 +1,148 @@
...
+
+`Full` displayes the day of the week and the timeZone
+
+```jsx
+<Flex justifyContent="space-around" alignItems="center">
+  <Text>
+    <DateTime format="short" dateStyle="long" />
+  </Text>
+</Flex>
+``` -->
```
lukelooker: Remove dead code

packages/www/src/documentation/components/navigation.ts
```diff
@@ -145,6 +145,7 @@ const components: NavigationSection = {
         { title: 'Card', path: 'card' },
         { title: 'Chip', path: 'chip' },
         { title: 'ColorWheel', path: 'color-wheel' },
+        { title: 'DateTime', path: 'date-time' },
```
dbchristopher: Maybe rename the title to `Date & Time` since this documentation now covers three different components

packages/www/src/documentation/components/content/date-time.mdx
```diff
@@ -0,0 +1,148 @@
...
+      </TableDataCell>
+    </TableRow>
+    <TableRow>
+      <TableDataCell>Full</TableDataCell>
+      <TableDataCell>
+        <DateTimeFormat format="full" />
+      </TableDataCell>
+    </TableRow>
+  </TableBody>
+</Table>
```
dbchristopher: This is a nice way to describe the props! I would just copy it out to a separate jsx file in `documentation/components/layout/demos` since the actual react markup isn't relevant to the demo. You can see Box.mdx for an example of another time we did that

laraharrow: thats great!!!

PR #539, opened 2020-02-06
[LENS-522] Mixed State Checkbox

dbchristopher:
### :sparkles: Changes

- Updates  `<Checkbox />` to support new prop value `checked="mixed"`
- Creates new useMixedStateCheckbox hook to sync state between parent and child checkboxes

- [ ] PR is ideally < 400LOC

### :camera: Screenshots

mixed-checkbox


dbchristopher:
@lukelooker @mdodgelooker I added support for the requested `mixed` state checkbox feature. While the rendering the new prop value was straightforward enough, actually syncing the state between parent/child checkboxes is a little tricky. I created a new custom hook to try and codify that relationship and hopefully help other devs manage that state. 

I'd love to get both of your eyes on this to see if you think the custom hook is helpful, and I'd love to hear any suggestions you might have for simplifying this pattern for other devs.

You can see the demo running in the playground.

packages/components/src/Form/Inputs/Checkbox/Checkbox.tsx
```diff
@@ -89,13 +114,18 @@ const CheckboxContainer = styled.div<CheckboxRadioContainerProps>`
     z-index: 1;
   }
   ${FauxCheckbox} {
-    background: ${props => props.theme.colors.palette.white};
-  }
-  input[type='checkbox']:checked {
-    & + ${FauxCheckbox} {
-      color: #fff;
-      ${inputColor}
-    }
```
dbchristopher: I couldn't rely on the native `:checked` value for this anymore, since html spec hasn't caught up to this newer use case for checkboxes

packages/playground/src/Checkbox/MixedStateCheckboxDemo.tsx
```diff
@@ -0,0 +1,117 @@
...
+    parent: {
+      setState: setParentState,
+      state: parentState,
+    },
+    children: [
+      { setState: setAppleState, state: appleState },
+      { setState: setBananaState, state: bananaState },
+      { setState: setAvocadoState, state: avocadoState },
+    ],
+  }
```
dbchristopher: This is still kind of a lot that our users will have to manually set up in order for this behavior to work. I'd love any suggestions on whether more of this could be moved to the hook

packages/components/src/utils/useMixedStateCheckbox.ts
```diff
@@ -0,0 +1,35 @@
...
+export type MixedBoolean = true | false | 'mixed'
+
+export interface CheckboxMetaState {
+  state: MixedBoolean
+  setState: Dispatch<SetStateAction<MixedBoolean>>
+}
+
+export interface CheckboxTree {
+  parent: CheckboxMetaState
+  children: CheckboxMetaState[]
```
dbchristopher: I am not sure if using the term `children` will lead to confusion, considering it's a common prop name in react. Still, `parent`/`children` seems more clear than trying to double down on the tree metaphor with something like `root`/`branches` or something

packages/components/src/Form/Fields/FieldCheckbox/__snapshots__/FieldCheckbox.test.tsx.snap
```diff
@@ -49,12 +49,8 @@ exports[`A FieldCheckbox 1`] = `
 
 .c4 .c5 {
   background: #FFFFFF;
-}
-
-.c4 input[type='checkbox']:checked + .c5 {
-  color: #fff;
-  background: #0087e1;
-  border-color: #0087e1;
+  border-color: #DEE1E5;
```
lukelooker: This _seems_ to be indicating that the checkbox:checked is getting a grey border (charcoal200) rather than the previous blue value However I'm not sure we even show the checkbox so maybe this doesn't matter anyway?

lukelooker: Ignore me... this entire selector is gone... jez i hate snapshots sometimes.

packages/components/src/utils/useMixedStateCheckbox.ts
```diff
@@ -0,0 +1,35 @@
+import { useEffect, Dispatch, SetStateAction } from 'react'
```
lukelooker: I think this Hook should stay in the `Checkbox` folder rather than in general utils since it's related to that component specifically (staying with our pattern with `useTooltip`, `usePopover`, etc.

packages/components/src/utils/useMixedStateCheckbox.ts
lukelooker: Yeah - I kinda like using `children` here since it has "equity" in React-land :)

dbchristopher:
Ok thanks for the feedback @lukelooker! I'll make those changes. I also added tests for the useMixedStateCheckbox hook; I'll add documentation next and then it should be ready to ship!

mdodgelooker:
This is awesome @dbchristopher ! Your hook and demo got me geeking out on how `useReducer` might replace `useState` and `useEffect` to reduce re-renders: https://github.com/looker-open-source/components/compare/dc/mixed-state-checkbox...mdodge/useMixedStateCheckboxx?expand=1

If you feel like that approach makes sense, feel free to use/improve it, or stick with what you've got since re-renders usually don't cost much.

dbchristopher:
Awesome thanks for digging into that @mdodgelooker! I always love the idea of useReducer, and it's a bummer that it doesn't come up very often lol. 

There are parts that I really love about your solution, but I do feel pretty hesitant to go that route. 

On one hand I **love** that it drastically reduces the amount of boilerplate that other developers will have to set up. But on the other hand, relying on array position seems like it will be harder to debug and maintain in the long term. Especially if you think about rendering default values from a server response, there will be an awkward transition to convert named objects to a flat array.

I also find the reducer a lot harder to read and reason about in this case. Unless there are horrible render side effects (like dispatching a network request for example), rerenders shouldn't have a performance impact. I vastly prioritize readability for simple ui components like this.

Anyways, thanks so much for taking a look though! It's definitely cool to see how it could look to leverage a reducer to manage this state. We should still keep your branch as a reference -- it might be easier to make the call which way to go once people actually try integrating this into existing form logic

dbchristopher:
Awesome, thanks @mdodgelooker! @lukelooker I need to get your final approval on those requested changes before it will let me merge. I think this should be all ready to go now 🙏

PR #540, opened 2020-02-07
update Fields to support description and detail props - update inputText to have width 100% to accommodate the changes

laraharrow:
✨ Changes

- update Fields to support description and details
- update inputText to have a width of 100%

✅ Requirements
- [x] PR is ideally < 400LOC

📷 Screenshots

Screen Shot 2020-02-13 at 4 54 36 PM


laraharrow:
I have not updated the snapshot but when I changed the width on InputText inside field it created some snapshots conflicts.

packages/components/src/Form/Fields/FieldText/FieldText.tsx
```diff
@@ -49,6 +49,7 @@ const FieldTextComponent = forwardRef(
           id={id}
           validationType={validationMessage && validationMessage.type}
           ref={ref}
+          width="100%"
```
lukelooker: This change will have pretty broad effects on existing uses. Was this called out in the designs?

laraharrow: this was something suggested by Jared.

lukelooker: k. I think this change will introduce too much breakage in existing use-cases. 

If we need 100% width `InputText` for `InlineEdit` let's apply it more specifically.

packages/components/src/Form/Fields/Field.tsx
```diff
@@ -66,6 +67,14 @@ export interface FieldProps {
...
    */
   alignValidationMessage?: FormControlDirections
+  /*
+   * optional extra description
+   */
+  description?: ReactNode
+  /**
+   * notes and details added to the top right corner of the field
+   */
+  detail?: ReactNode
```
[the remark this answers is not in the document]
laraharrow: good call!

PR #543, opened 2020-02-10
Fix links to documentation, remove a few 'experimental' flags

lukelooker:
### :sparkles: Changes

- Fix links to documentation
- Remove a few 'experimental' flags

- [ ] PR is ideally < 400LOC

### DIFF
--- a/packages/www/src/documentation/components/actions/button-group.mdx
+++ b/packages/www/src/documentation/components/actions/button-group.mdx
@@ -1,7 +1,7 @@
 ---
 title: ButtonGroup & ButtonToggle
 figma: file/Ut0DDG4hiuRG1HiZrC5Zyf/Dashboards-Next-Components?node-id=529%3A776
-status: experimental
+github: Button
 ---
 
 ## ButtonGroup
--- a/packages/www/src/documentation/components/actions/button-group.mdx
+++ b/packages/www/src/documentation/components/actions/button-group.mdx
@@ -1,7 +1,7 @@
 ---
 title: ButtonGroup & ButtonToggle
 figma: file/Ut0DDG4hiuRG1HiZrC5Zyf/Dashboards-Next-Components?node-id=529%3A776
-github: Button
+github: Button/ButtonGroup.mdx
 ---
 
 ## ButtonGroup
--- a/packages/www/src/documentation/components/actions/button.mdx
+++ b/packages/www/src/documentation/components/actions/button.mdx
@@ -1,7 +1,7 @@
 ---
 title: Button
 status: stable
-github: Button/Button
+github: Button/Button.mdx
 propsOf: Button
 figma: file/Dirjx0RKbOOrLacqHq61By/Button-%26-Links?node-id=914%3A981&viewport=1741%2C951%2C1
 ---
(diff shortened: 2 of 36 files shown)


packages/www/src/documentation/components/actions/button-group.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 title: ButtonGroup & ButtonToggle
 figma: file/Ut0DDG4hiuRG1HiZrC5Zyf/Dashboards-Next-Components?node-id=529%3A776
-status: experimental
+github: Button
```
dbchristopher: should this be `Button/ButtonGroup`?

packages/www/src/documentation/components/actions/button.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 title: Button
 status: stable
-github: Button
+github: Button/Button
```
dbchristopher: the GH link 404s without the `.tsx` extension.

packages/www/src/documentation/components/content/avatar.mdx
```diff
@@ -1,6 +1,7 @@
 ---
 title: Avatar
 propsOf: Avatar
+github: Avatar
```
dbchristopher: Is there any logic for when we link directly to the tsx file vs the module root directory?

packages/www/src/documentation/components/content/color-wheel.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: ColorWheel
-github: ColorWheel
+github: Forms/Fields/FieldColor
```
dbchristopher: 404 -- should be `Forms/Fields/FieldColor/ColorWheel` 

But also, should we move ColorWheel to the components root directory? Seems possible it could be used outside of the `FieldColor` component

packages/www/src/documentation/components/actions/button-group.mdx
lukelooker: Yep. 🤦‍♂

packages/www/src/documentation/components/actions/button.mdx
lukelooker: Fixed. 🤦‍♂

packages/www/src/documentation/components/content/avatar.mdx
lukelooker: My logic was that if a page pointed to more than one Component I tried to point to place in the source that would give the narrowest view of everything described on the page so in this case Avatar covers a small collection of components.

packages/www/src/documentation/components/content/color-wheel.mdx
lukelooker: For scope of this PR (fixing links) I'm just going to correct the link but that it might be worth looking at promoting the ColorWheel in the future.

packages/www/src/documentation/components/actions/button-group.mdx
lukelooker: Actually, no, the latter one `Button` is incorrect but this is correct as this page points at both ButtonGroup & ButtonToggle so to see both components' code you need to go to the /Button path.

PR #559, opened 2020-02-13
Fix controlled FieldColor

mdodgelooker:
### :sparkles: Changes

- `FieldColor` was not truly controllable – any change to the value prop would not register.
- With these changes, a developer takes control of the state using the `onChange` and `value` props
- Added support for `defaultValue` for when an initial value is needed but the field can still be uncontrolled.
- A demo of the issue: https://codesandbox.io/s/reverent-curran-inmig

- [x] PR is ideally < 400LOC

### :camera: Screenshots
**Issue:**
issue

**Fix:**
fix

### DIFF
--- a/packages/components/src/Form/Fields/FieldColor/FieldColor.test.tsx
+++ b/packages/components/src/Form/Fields/FieldColor/FieldColor.test.tsx
@@ -28,7 +28,10 @@ import 'jest-styled-components'
 import '@testing-library/jest-dom/extend-expect'
 import React from 'react'
 import { fireEvent } from '@testing-library/react'
-import { renderWithTheme } from '@looker/components-test-utils'
+import {
+  assertSnapshotShallow,
+  renderWithTheme,
+} from '@looker/components-test-utils'
 import { theme } from '@looker/design-tokens'
 import { ThemeProvider } from 'styled-components'
 
@@ -36,11 +39,15 @@ import { Button } from '../../../Button'
 import { FieldColor } from './FieldColor'
 
 describe('FieldColor', () => {
+  test('Default render', () => {
+    assertSnapshotShallow(<FieldColor />)
+  })
+
   test('with hidden input', () => {
-    const { queryByRole } = renderWithTheme(
+    const { queryByDisplayValue } = renderWithTheme(
       <FieldColor value="yellow" hideInput />
     )
-    expect(queryByRole('input')).not.toBeInTheDocument()
+    expect(queryByDisplayValue('yellow')).not.toBeInTheDocument()
   })
 
   test('starts with a named color value', () => {
@@ -48,22 +55,21 @@ describe('FieldColor', () => {
     expect(getByDisplayValue('green')).toBeInTheDocument()
   })
 
-  test('starts with a named color value', () => {
+  test('responds to input value change', () => {
     const { getByDisplayValue } = renderWithTheme(<FieldColor value="green" />)
     fireEvent.change(getByDisplayValue('green'), { target: { value: 'blue' } })
     expect(getByDisplayValue('blue')).toBeInTheDocument()
   })
 
+  const FieldColorValidationMessage = () => {
+    return (
+      <ThemeProvider theme={theme}>
+        <FieldColor validationMessage={{ message: 'Error!', type: 'error' }} />
+      </ThemeProvider>
+    )
+  }
+
   test('with a validation message', () => {
-    const FieldColorValidationMessage = () => {
-      return (
-        <ThemeProvider theme={theme}>
-          <FieldColor
-            validationMessage={{ message: 'Error!', type: 'error' }}
-          />
-        </ThemeProvider>
-      )
-    }
     const { queryByText } = renderWithTheme(<FieldColorValidationMessage />)
     expect(queryByText('Error!')).toBeInTheDocument()
   })
--- a/packages/components/src/Form/Fields/FieldColor/FieldColor.test.tsx
+++ b/packages/components/src/Form/Fields/FieldColor/FieldColor.test.tsx
@@ -57,7 +57,9 @@ describe('FieldColor', () => {
 
   test('responds to input value change', () => {
     const { getByDisplayValue } = renderWithTheme(<FieldColor value="green" />)
-    fireEvent.change(getByDisplayValue('green'), { target: { value: 'blue' } })
+    const input = getByDisplayValue('green')
+    input.focus()
+    fireEvent.change(input, { target: { value: 'blue' } })
     expect(getByDisplayValue('blue')).toBeInTheDocument()
   })
 
@@ -82,8 +84,8 @@ describe('FieldColor', () => {
     const input = getByLabelText('Background Color')
     fireEvent.change(input, { target: { value: '#FFFF00' } })
     expect(onChangeMock).toHaveBeenCalledWith({
-      currentTarget: { value: '#ffff00' },
-      target: { value: '#ffff00' },
+      currentTarget: { value: '#FFFF00' },
+      target: { value: '#FFFF00' },
     })
   })
 
--- a/packages/components/src/Form/Fields/FieldColor/FieldColor.test.tsx
+++ b/packages/components/src/Form/Fields/FieldColor/FieldColor.test.tsx
@@ -28,10 +28,7 @@ import 'jest-styled-components'
 import '@testing-library/jest-dom/extend-expect'
 import React from 'react'
 import { fireEvent } from '@testing-library/react'
-import {
-  assertSnapshotShallow,
-  renderWithTheme,
-} from '@looker/components-test-utils'
+import { renderWithTheme } from '@looker/components-test-utils'
 import { theme } from '@looker/design-tokens'
 import { ThemeProvider } from 'styled-components'
 
@@ -39,10 +36,6 @@ import { Button } from '../../../Button'
 import { FieldColor } from './FieldColor'
 
 describe('FieldColor', () => {
-  test('Default render', () => {
-    assertSnapshotShallow(<FieldColor />)
-  })
-
   test('with hidden input', () => {
     const { queryByRole } = renderWithTheme(
       <FieldColor value="yellow" hideInput />
@@ -61,15 +54,16 @@ describe('FieldColor', () => {
     expect(getByDisplayValue('blue')).toBeInTheDocument()
   })
 
-  const FieldColorValidationMessage = () => {
-    return (
-      <ThemeProvider theme={theme}>
-        <FieldColor validationMessage={{ message: 'Error!', type: 'error' }} />
-      </ThemeProvider>
-    )
-  }
-
   test('with a validation message', () => {
+    const FieldColorValidationMessage = () => {
+      return (
+        <ThemeProvider theme={theme}>
+          <FieldColor
+            validationMessage={{ message: 'Error!', type: 'error' }}
+          />
+        </ThemeProvider>
+      )
+    }
     const { queryByText } = renderWithTheme(<FieldColorValidationMessage />)
     expect(queryByText('Error!')).toBeInTheDocument()
   })
--- a/packages/components/src/Form/Fields/FieldColor/FieldColor.test.tsx
+++ b/packages/components/src/Form/Fields/FieldColor/FieldColor.test.tsx
@@ -35,52 +35,94 @@ import {
 import { theme } from '@looker/design-tokens'
 import { ThemeProvider } from 'styled-components'
 
+import { Button } from '../../../Button'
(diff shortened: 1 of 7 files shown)


packages/components/src/Form/Fields/FieldColor/FieldColor.test.tsx
```diff
@@ -28,59 +28,95 @@ import 'jest-styled-components'
...
+import { renderWithTheme } from '@looker/components-test-utils'
 import { theme } from '@looker/design-tokens'
 import { ThemeProvider } from 'styled-components'
 
+import { Button } from '../../../Button'
 import { FieldColor } from './FieldColor'
 
-test('Default FieldColor', () => {
-  assertSnapshotShallow(<FieldColor />)
-})
```
mdodgelooker: This snapshot was already rendering an empty `ShallowWrapper {}` so I removed it.

packages/components/src/Form/Fields/FieldColor/FieldColor.tsx
```diff
@@ -74,70 +81,106 @@ const createEventWithHSVValue = (
...
+function getColorFromText(text?: string) {
+  return text && isValidColor(text) ? str2simpleHsv(text) : undefined
+}
+
 export const FieldColorComponent = forwardRef(
   (
     {
       alignValidationMessage,
       hideInput,
-      id = uuid(),
```
mdodgelooker: Just noticed that this would change the id on every render, so I replaced it with the `useID` hook that exists for this purpose.

packages/components/src/Form/Fields/FieldColor/FieldColor.test.tsx
```diff
@@ -28,59 +28,95 @@ import 'jest-styled-components'
...
+describe('FieldColor', () => {
+  test('with hidden input', () => {
+    const { queryByRole } = renderWithTheme(
+      <FieldColor value="yellow" hideInput />
+    )
+    expect(queryByRole('input')).not.toBeInTheDocument()
```
dbchristopher: this passes even when I delete the `hideInput` prop. I don't think there ever is an element with `aria-role="input"` attribute, so you'll probably need a different selector to validate the prop behavior

packages/components/src/Form/Fields/FieldColor/FieldColor.test.tsx
```diff
@@ -28,59 +28,95 @@ import 'jest-styled-components'
...
+  test('starts with a named color value', () => {
+    const { getByDisplayValue } = renderWithTheme(<FieldColor value="green" />)
+    expect(getByDisplayValue('green')).toBeInTheDocument()
+  })
 
-test('FieldColor starts with a named color value', () => {
-  const { getByDisplayValue } = renderWithTheme(<FieldColor value="green" />)
-  expect(getByDisplayValue('green')).toBeInTheDocument()
-})
+  test('starts with a named color value', () => {
```
dbchristopher: this test description is duplicated from above

packages/components/src/Form/Fields/FieldColor/FieldColor.test.tsx
```diff
@@ -28,59 +28,95 @@ import 'jest-styled-components'
...
-  expect(onChangeMock).toHaveBeenCalledWith({
-    currentTarget: { value: '#ffff00' },
-    target: { value: '#ffff00' },
+    const button = getByText('Turn yellow')
+    const input = getByPlaceholderText('Select a color')
+    expect(input).toHaveValue('')
+    fireEvent.click(button)
+    expect(input).toHaveValue('yellow')
+    fireEvent.change(input, { target: { value: 'purple' } })
+    expect(input).toHaveValue('purple')
```
dbchristopher: Very thorough! 👍

packages/components/src/Form/Fields/FieldColor/FieldColor.tsx
```diff
@@ -74,70 +81,106 @@ const createEventWithHSVValue = (
...
+
+    // If there's been an external change, update the input text value
+    const isInputting = useRef(false)
+    if (controlledValue && value !== controlledValue && !isInputting.current) {
+      setValue(controlledValue)
+    }
 
-    const [color, setColor] = useState<SimpleHSV>(initialValue)
-    const [inputTextValue, setInputTextValue] = useState(props.value || '')
+    const colorToUse = isControlled ? colorFromProps || initialWhite : color
```
dbchristopher: Woah this is **a lot** of logic to to render a color value. I wonder if it could be simplified at all? Might be worth giving it a once-over with fresh eyes tomorrow to see if there are any opportunities there

packages/components/src/Form/Fields/FieldColor/FieldColor.tsx
```diff
@@ -74,70 +81,106 @@ const createEventWithHSVValue = (
...
     const validationMessage = useFormContext(props)
+
+    const isControlled = useControlWarn({
+      controllingProps: ['onChange', 'value'],
+      isControlledCheck: () => onChange !== undefined,
+      name: 'ButtonGroup',
+    })
+
+    const colorFromProps = getColorFromText(controlledValue)
+    const defaultColorFromProps = getColorFromText(defaultValue)
```
dbchristopher: is it fair to say that `value` takes priority over `defaultValue`? we could probably do a fallback here so we don't have to juggle both variables. maybe this?

```js
const whiteHSV = polarbrightness2hsv(white())
const colorFromProps = getColorFromText(controlledValue || defaultValue || whiteHSV)
```

then I feel like you might be safe to delete logic around setting `initialValue` and `initialColor` values, and just set `useState(colorFromProps)`, and maybe there will be opportunities to simplify from there

packages/components/src/Form/Fields/FieldColor/FieldColor.tsx
```diff
@@ -74,70 +81,106 @@ const createEventWithHSVValue = (
...
+    const initialColor = colorFromProps || defaultColorFromProps || initialWhite
+
+    const [color, setColor] = useState<SimpleHSV>(initialColor)
+    const [value, setValue] = useState(initialValue)
+
+    // If there's been an external change, update the input text value
+    const isInputting = useRef(false)
+    if (controlledValue && value !== controlledValue && !isInputting.current) {
+      setValue(controlledValue)
+    }
```
dbchristopher: I'm not sure if I understand the point in tracking `isInputting` state. Is there worry that an external value could change while they're typing and then lose their work or something?

packages/components/src/Form/Fields/FieldColor/FieldColor.test.tsx
mdodgelooker: Ah good catch! Fixed it.

packages/components/src/Form/Fields/FieldColor/FieldColor.test.tsx
mdodgelooker: Fixed.

packages/components/src/Form/Fields/FieldColor/FieldColor.test.tsx
mdodgelooker: Whoops, hold on, I did something wrong while rebasing...

mdodgelooker: Ok _now_ it's fixed. :)

packages/components/src/Form/Fields/FieldColor/FieldColor.tsx
mdodgelooker: Good call – I incorporated the above and it did simplify things a bit. The complexity of offering both controlled and uncontrolled on our inputs is becoming a theme these days. Perhaps something to discuss as a team at some point.

packages/components/src/Form/Fields/FieldColor/FieldColor.tsx
mdodgelooker: It's for when state is controlled externally, and the user is typing but what they've got so far isn't a valid color string. The `onChange` is called with fallback white until the input string becomes a valid color, so `controlledValue` will be `'#ffffff'` while `value` might be `'#'`. In this case we don't want to sync up the `value` with `controlledValue` but otherwise we do.
color

packages/components/src/Form/Fields/FieldColor/FieldColor.tsx
dbchristopher: ugh yeah it really gets to be a lot juggling multiple ways an input component can be used 😫

packages/components/src/Form/Fields/FieldColor/FieldColor.tsx
dbchristopher: oof yeah that's a lot of state to juggle 😖

packages/components/src/Form/Fields/FieldColor/FieldColor.tsx
```diff
@@ -74,70 +81,101 @@ const createEventWithHSVValue = (
...
         />
         <LuminositySlider
           min={0}
           max={100}
           step={1}
-          value={color.v * 100}
+          value={colorToUse.v * 100}
           width={colorWheelSize}
           onChange={handleSliderChange}
         />
```
dbchristopher: @mdodgelooker I just found this bug with the color wheel/slider: Looks like the luminosity gets reset as soon as I interact with the wheel, so I can never select anything less than full brightness:

value-slider

I just verified it's not an issue in production, so it does seem to have something to do with these controlled variables 😖

packages/components/src/Form/Fields/FieldColor/FieldColor.tsx
```diff
@@ -74,70 +81,101 @@ const createEventWithHSVValue = (
...
+    const [value, setValue] = useState(defaultValue || '')
+
+    // If there's been an external change in the color, update the input value
+    // except when the user is manually typing a color string (isInputting.current === true)
+    // since onChange will have been called with #ffffff until the typed value is a valid color
+    // and updating the input text with that would interfere with typing
+    const isInputting = useRef(false)
+    if (controlledValue && value !== controlledValue && !isInputting.current) {
+      setValue(controlledValue)
+    }
```
dbchristopher: @mdodgelooker this block of code reminded me basically of what useEffect is for -- just updating the values at an appropriate component lifecycle. I did a super quick stab at setting that up in a new branch off of master, and the FieldColorDemo seems to work fine:

https://github.com/looker-open-source/components/pull/608

I'm sure there are other considerations this didn't account for, but I just wanted to submit it to you for consideration. Considering the new luminosity slider bug that got introduced here, I thought it worth exploring an alternate approach real quick

mdodgelooker: Thanks @dbchristopher , I love how clean that approach is! I replaced the `isInputting` logic with your focus state code from https://github.com/looker-open-source/components/pull/608 with a few changes to fix an issue on the uncontrolled version where the input is cleared on blur. And then from some breakages with the new tests I added, it revealed an issue where if the user types a valid color, it'll always get replaced on blur with the hex equivalent (which I could see being confusing to some users).

mdodgelooker: Oh and also I wrapped the focus and blur handlers so if the developer wants to add their own handlers they can do so.

packages/components/src/Form/Fields/FieldColor/FieldColor.tsx
```diff
@@ -115,13 +151,12 @@ export const FieldColorComponent = forwardRef(
...
-      setInputTextValue(value)
+      const newValue = event.currentTarget.value
+      setInputTextValue(newValue)
 
-      const newColor =
-        !value || !isValidColor(value) ? initialWhite : str2simpleHsv(value)
-      setColor(newColor)
-      callOnChange(newColor)
+      const isValid = isValidColor(newValue)
+      callOnChange(isValid ? newValue : initialWhite)
```
mdodgelooker: If the typed text is a valid color, we can call `onChange` with that instead of converting to hex.

packages/components/src/Form/Fields/FieldColor/FieldColor.tsx
```diff
@@ -57,46 +64,75 @@ export interface FieldColorProps
...
    */
   hideInput?: boolean
-
   value?: string
+  defaultValue?: string
 }
 
 const createEventWithHSVValue = (
-  color: SimpleHSV
+  color: SimpleHSV | string
```
mdodgelooker: Accept string so the user can type a valid color and that can get passed directly to the `onChange`.

mdodgelooker:
@dbchristopher one more 👍after resolving conflicts w/ master 🙏

PR #578, opened 2020-02-22
InputDateRange

dbchristopher:
date-range-picker


packages/components/src/utils/i18n.ts
```diff
@@ -74,3 +78,42 @@ export const dateFnLocaleMap = {
...
+): Date | false => {
+  // Date format 'P' represents localized dates in date-fns
+  const parsedValue = parse(value, 'P', new Date(), {
+    locale: dateFnLocaleMap[locale],
+  })
+
+  parsedValue.setFullYear(formatYear(parsedValue))
+
+  return isValid(parsedValue) && parsedValue
+}
```
dbchristopher: Moved these here from the `InputDate` component file

dbchristopher:
@mdodgelooker this is finally ready for review if you have any time to take a look this week 😅 

There's a ton here for sure and it was a lot to get all the edge cases reconciled. Definitely let me know if anything doesn't make sense here!

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
```diff
@@ -0,0 +1,366 @@
...
+  }
+
+  ${InputText} {
+    border: none;
+    border-radius: 0;
+    padding: 0;
+    margin: 0;
+    box-shadow: none;
+    background: transparent;
+  }
```
mdodgelooker: `inputTextValidation` is missing here, if you want to support the red border styles, as well as `inputTextDisabled` (though I think this input doesn't currently support disabled). @lukelooker actually had a cool idea for the iconBefore/iconAfter/prefix/suffix feature of InputText where you later the text on top of the input, set pointer-events to none, and then you move the input styles to the wrapper. It would look something like this: https://codesandbox.io/s/zen-rgb-m4t6n

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
```diff
@@ -0,0 +1,366 @@
...
+import {
+  LocaleCodes,
+  Locales,
+  formatDateString,
+  parseDateFromString,
+} from '../../../utils'
+import { InputProps } from '../InputProps'
+
+interface InputDateRangeProps
+  extends Omit<InputProps, 'defaultValue' | 'onChange'>,
```
mdodgelooker: Unless you're planning to spread props to the inputs (which seems like a hassle in such a complex component) you could remove this line and just add `validationType?: ValidationType`.
Currently props like `disabled` are accepted but don't do anything.

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
```diff
@@ -0,0 +1,366 @@
...
+          <InputText
+            placeholder={`Date (${formatDateString(
+              new Date(Date.now()),
+              locale
+            )})`}
+            value={inputs.from.value}
+            onChange={handleTextInputChange}
+            onBlur={handleValidation}
+            data-testid="date-from-text-input"
+            id="date-from"
```
mdodgelooker: This could lead to multiple elements with the same ID if `<InputDateRange/>` is used more than once on a page. `utils/useID` might be helpful here.

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.test.tsx
```diff
@@ -0,0 +1,180 @@
...
+    onChange: jest.fn(),
+  }
+  const { getAllByText, getByTestId } = renderWithTheme(
+    <InputDateRange {...mockProps} />
+  )
+
+  const fromInput = getByTestId('date-from-text-input') as HTMLInputElement
+  const toInput = getByTestId('date-to-text-input') as HTMLInputElement
+  const startDate = getAllByText('2')[0] // the 2nd day of the month
+  const endDate = getAllByText('15')[0] // the 15th day of the month
```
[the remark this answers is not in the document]
dbchristopher: oh yeah that's a good call! I'll update this test so the endDate is from the right-hand calendar, which I think should be sufficient to verify everything works as expected

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.test.tsx
```diff
@@ -0,0 +1,180 @@
...
+  const mockProps = {
+    defaultValue: {
+      from: new Date('June 3, 2019'),
+      to: new Date('June 9, 2019'),
+    },
+    onChange: jest.fn(),
+  }
+  const { getByTestId } = renderWithTheme(<InputDateRange {...mockProps} />)
+  const fromInput = getByTestId('date-from-text-input') as HTMLInputElement
+  const toInput = getByTestId('date-to-text-input') as HTMLInputElement
```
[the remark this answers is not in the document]
dbchristopher: good call. added a new test for that here: https://github.com/looker-open-source/components/pull/578/commits/5cfc23b31271a836e7fabc0d8dbc26d72cc66bd9

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
dbchristopher: 🤔 hmmm, I don't love that the approach relies on hard-coding the padding value in order for the layout to line up correctly. It's probably fine for icons. But whenever we figure out i18n and start translating these hardcoded strings, the layout is going to break down. 

I just dropped `FROM` into google translate and pasted the results into that code sandbox. Here is where the cursor ended up for a few different locales:

Screen Shot 2020-02-28 at 2 01 20 PM
Screen Shot 2020-02-28 at 2 02 19 PM
Screen Shot 2020-02-28 at 2 03 25 PM 1
Screen Shot 2020-02-28 at 2 04 29 PM

dbchristopher: I added support for rendering validation errors with the current layout approach. Let me know what you think!

Screen Shot 2020-02-28 at 2 45 14 PM

dbchristopher:
Ok @mdodgelooker I think I got everything! Implemented the controlled component functionality, so it can now be updated by an externally defined `value` prop. Ready for another look. Thanks for all the feedback!

controlled-date-range

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
```diff
@@ -0,0 +1,416 @@
...
+  useEffect(() => {
+    // controlled component: update state when value changes externally
+    if (value && !isEqual(value, dateRange)) {
+      setDateRange(value)
+      value.from && inputs.from.setValue(formatDateString(value.from, locale))
+      value.to && inputs.to.setValue(formatDateString(value.to, locale))
+      value.from &&
+        !isDateRangeInView(value, viewMonth) &&
+        setViewMonth(value.from)
+    }
```
dbchristopher: I used the same `useEffect` pattern from #608 here for updating state on prop change. Seems to work well!

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
```diff
@@ -0,0 +1,416 @@
...
+
+    // render a console warning if developers pass in a value without a change listener
+    if (value && !onChange) {
+      // eslint-disable-next-line no-console
+      console.error(
+        'Warning: Failed prop type: You provided a `value` prop to <InputDateRange /> without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue` instead. Otherwise, please provide an `onChange` callback.'
+      )
+    }
```
dbchristopher: I initially attempted to use the `useControlWarn` hook that I saw you added in #559 @mdodgelooker. For some reason I could never get the configuration dialed in to where it would actually render the warning for some reason 🤷‍♂ Since I didn't have any use for the returned `isControlled` variable anyways I just went with a one-off here.

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
mdodgelooker: Nice! I hear you that the other approach really isn't flexible.

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
mdodgelooker: `useControlWarn` was inspired by https://github.com/reach/reach-ui/blob/master/packages/tabs/src/index.tsx as a way to mimic the "uncontrolled to controlled" warning that React gives on form inputs, but TBH I'm starting to feel it's annoying to use and I _think_ that React warning is just an attempt to be helpful in case the developer made a mistake, not a sign of any deeper issue. So yeah, what you have looks awesome. 👍

PR #581, opened 2020-02-24
Action List: Prep Work

ElliotTheHuman:

- [x] PR is ideally < 400LOC


ElliotTheHuman:
We opted to go with a different method of hiding / showing the row options button since using `hoverDisclosureRef` on the row encountered problems for keyboard navigation. 

More specifically, if there were focusable elements within ActionList cells, there would be no way for a keyboard user to reach the row options button (since focus needs to either be on the row or the button itself for the options button to appear).

PR #582, opened 2020-02-24
Remove requestAnimationFrame dependencies

dbchristopher:
A super rough stab at using pure css to hide/show the action list options menu. This should hopefully simplify the lifecycle and remove the need to coordinate disparate requestAnimationFrame calls.

tab-focus-behavior


ElliotTheHuman:
LGTM!!

PR #604, opened 2020-02-27
Fix Popover mousedown issue

mdodgelooker:
### :sparkles: Changes

Background info:
1. `Select` opens (via `usePopopver`) on mousedown
2. This revealed a scroll-lock bug fixed in https://github.com/looker-open-source/components/pull/562
3. The change above caused the following issue, fixed in the current PR:
  When an element (modal, another popover, etc) is stacking on top of a popover, a mousedown on this element closes the popover (usually thereby removing the other element as well)

### DIFF
--- a/packages/components/src/Popover/Popover.tsx
+++ b/packages/components/src/Popover/Popover.tsx
@@ -220,6 +220,15 @@ function useOpenWithoutElement(isOpen: boolean, element: HTMLElement | null) {
   return openWithoutElem
 }
 
+function isNodeInOrAfter(nodeA: Node, nodeB: Node) {
+  const relationship = nodeA.compareDocumentPosition(nodeB)
+  return (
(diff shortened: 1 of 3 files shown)


packages/components/src/Popover/Popover.tsx
```diff
@@ -301,6 +305,7 @@ function usePopoverToggle(
       }
 
       event.stopPropagation()
+      event.preventDefault()
```
mdodgelooker: This line isn't as crucial as the other changes, but it fixes the odd behavior where clicking on the another `Select` (or menu disclosure, etc) while you have one open will close the current one and focus the next one but not open it. It also makes it so `<a>` links clicked while a popover is open are not followed – which seems desirable to me, but let me know if you disagree.

selects

PR #609, opened 2020-02-29
ComboboxMulti

mdodgelooker:
### :sparkles: Changes

- Added -Multi versions for all Combobox components
- Moved many many lines of code to hooks in /Combobox/utils for use in both the -Multi and original components
- Made most of the types generic so that the -Multi versions could be created with just an argument or 2
- Layout changes in `InputSearch` and `InputChipsBase` to support `MultiSelect` design spec
- Added `ReplaceText` – a new component that supports search term highlighting in `ComboboxOption` but could be good for other uses as well
- TODO (next PR): don't change input text to option text when highlighting already-selected options; instead highlight the corresponding chip to indicate that clicking/enter/delete will remove that value.


packages/components/src/Form/Inputs/Checkbox/Checkbox.tsx
```diff
@@ -45,7 +45,7 @@ export interface CheckboxProps
   extends Omit<InputProps, 'type' | 'checked'>,
     CheckboxContainerProps {}
 
-const CheckMark = () => {
+export const CheckMark = () => {
```
mdodgelooker: Adding the visual elements of `Checkbox` to `ComboboxMultiOption`.

packages/components/src/Form/Inputs/Combobox/Combobox.test.tsx
```diff
@@ -136,8 +136,8 @@ describe('Keyboard navigation', () => {
 
     const items = getAllByRole('option')
     expect(input).toHaveValue('')
-    expect(items[0]).not.toHaveAttribute('aria-selected')
-    expect(items[1]).not.toHaveAttribute('aria-selected')
+    expect(items[0]).toHaveAttribute('aria-selected', 'false')
+    expect(items[1]).toHaveAttribute('aria-selected', 'false')
```
mdodgelooker: Caused by refactoring, harmless change.

packages/components/src/Form/Inputs/Combobox/Combobox.tsx
```diff
@@ -34,195 +34,149 @@ import {
...
-  }
-  function handleInputChange(e: React.FormEvent<HTMLInputElement>) {
-    setOption(e.currentTarget.value)
-  }
-  return {
-    inputProps: { onChange: handleInputChange },
-    onChange: handleChange,
-    value,
-  }
-}
```
mdodgelooker: This wasn't being used anywhere. I think I wrote it with the idea that it would make using `Combobox` easier but now it's more of an under-the-hood component and `Select` is easy to use.

packages/components/src/Form/Inputs/Combobox/ComboboxInput.tsx
```diff
@@ -303,16 +197,16 @@ const indicatorRaw = `
...
   />
 </svg>`
-const indicatorSize = '1rem'
-const indicatorPadding = '.25rem'
+export const indicatorSize = '1rem'
+export const indicatorPadding = '.25rem'
 const indicatorPrefix = 'data:image/svg+xml;base64,'
 export const selectIndicatorBG = (color: string) =>
   typeof window !== 'undefined' &&
   `url('${indicatorPrefix}${window.btoa(
```
mdodgelooker: Turns out `window.btoa` is actually why gatsby prod doesn't show the caret (no `window` b/c it uses server-side rendering). I could fix this by installing the `btoa` package – does that seem reasonable?

packages/components/src/Form/Inputs/Combobox/utils/index.ts
```diff
@@ -0,0 +1,27 @@
...
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+export * from './getComboboxText'
```
mdodgelooker: So far this is the only util I can see a need to export.

packages/components/src/Form/Inputs/InputChips/InputChips.tsx
```diff
@@ -183,52 +157,22 @@ export const InputChipsInternal = forwardRef(
...
-      return (
-        <Chip onDelete={onChipDelete} key={value} mb={1} mt={1} ml="xxsmall">
-          {value}
-        </Chip>
-      )
-    })
-
     return (
-      <InputSearch
+      <InputChipsBase
```
mdodgelooker: Added this component because it doesn't work well to allow the user to input chips free-form via comma/enter, so `InputChipsBase` takes care of accepting values, rendering the chips, and allowing their removal, and `InputChips` keeps the inputting values logic.

packages/components/src/Form/Inputs/InputSearch/InputSearch.tsx
```diff
@@ -197,7 +197,6 @@ InputSearchComponent.displayName = 'InputSearchComponent'
 export const InputSearch = styled(InputSearchComponent)`
   align-items: center;
   display: flex;
-  position: relative;
```
mdodgelooker: This was causing one `Combobox` to stack on top of the previous one's toolitp for the X (clear) icon, and I don't think it's actually needed for anything?

packages/components/src/Form/Inputs/InputSearch/InputSearch.tsx
```diff
@@ -182,7 +182,7 @@ const InputSearchComponent = forwardRef(
           onChange={handleChange}
           value={inputValue}
           focusStyle={{ outline: 'none' }}
-          px="0"
+          pr="0"
```
mdodgelooker: Slight adjustments to padding, gotta let the chips get super close to the left-hand border.

packages/components/src/Form/Inputs/Combobox/Combobox.test.tsx
```diff
@@ -62,15 +71,31 @@ describe('<Combobox/> with children', () => {
     expect(handleChange).toHaveBeenCalledWith({ label: 'Foo', value: '101' })
   })
 
-  test('Opens and closes on click', () => {
-    const { getByText, getByPlaceholderText, queryByText } = renderWithTheme(
-      <Combobox>
+  // Testing same behavior on both Combobox and ComboboxMulti
+  test.each([
```
mdodgelooker: This was my solution to test behaviors that are the same across `Combobox` & `ComboboxMulti` – though they are driven by the same hooks, that's not a guarantee that one won't break while the other does.

packages/components/src/Form/Inputs/InputSearch/InputSearch.tsx
```diff
@@ -131,7 +131,13 @@ const InputSearchComponent = forwardRef(
...
     const ref = useForkedRef<HTMLInputElement>(internalRef, forwardedRef)
 
-    const focusInput = () => internalRef.current && internalRef.current.focus()
+    function handleMouseDown() {
+      // set focus to input on mousedown of container
+      // need requestAnimationFrame here due to browser updating focus _after_ mousedown is called
+      window.requestAnimationFrame(() => {
+        internalRef.current && internalRef.current.focus()
+      })
+    }
```
mdodgelooker: This fixed a bug where if you click on a `Select`'s caret 1x, it opens, 2x it closes, and 3x it _should_ open but does not, because the browser is triggering a blur event on the input just after mousedown, and then `Combobox` checks for where focus landed, which in that tiny window happens to be the `body`, until the click event on `InputSearch` registered and put focus back on the input. But by that time it was too late – `Combobox` decided not to re-open the menu.

packages/components/src/Form/Inputs/InputSearch/InputSearch.tsx
```diff
@@ -171,23 +178,33 @@ const InputSearchComponent = forwardRef(
...
-          ref={ref}
-        />
+        {children ? (
+          <Flex alignItems="flex-start" flexWrap="wrap">
+            {children}
+            {input}
+          </Flex>
+        ) : (
+          input
+        )}
```
mdodgelooker: Layout changes to support scrolling in `InputChips` and `ComboboxMultiInput`.

packages/components/src/Text/ReplaceText.test.tsx
```diff
@@ -0,0 +1,45 @@
+/*
```
mdodgelooker: A new component that supports search term highlighting in `ComboboxOption` but could be good for other uses as well.

mdodgelooker:
Thanks @lukelooker ! I definitely will circle back to your points above in the next PR, focusing on `MultiSelect` as the external-facing consumer of `ComboboxMulti`.

PR #658, opened 2020-03-10
TabPanels: Added FlexboxProps and LayoutProps

ElliotTheHuman:
### :sparkles: Changes
- Added FlexboxProps and Layout Props to `TabPanels` component
  - **Use case:** There are times where you want the contents of TabPanel to take up whatever space is left in a surrounding flex container. Adding these props means that `TabPanels` will take up all remaining space in a flex container using "height: 100%" or "flex: 1".
- **Note:** Also added TabsDemo to demonstrate a "flex container surrounding a Tabs element scenario".

- [x] PR is ideally < 400LOC

### DIFF
--- a/packages/components/src/Tabs/TabPanels.tsx
+++ b/packages/components/src/Tabs/TabPanels.tsx
@@ -50,7 +50,7 @@ export const TabPanels: FC<TabPanelsProps> = ({
   selectedIndex,
   ...props
 }) => {
-  const styleProps = omit(props, 'onSelectTab')
+  const tabPanelLayoutProps = omit(props, 'onSelectTab')
 
   const clonedChildren = Children.map(
     children,
@@ -61,7 +61,7 @@ export const TabPanels: FC<TabPanelsProps> = ({
     }
   )
 
-  return <Layout {...styleProps}>{clonedChildren}</Layout>
+  return <Layout {...tabPanelLayoutProps}>{clonedChildren}</Layout>
 }
 
 const Layout = styled.div<TabPanelLayoutProps>`
--- a/packages/components/src/Tabs/TabPanels.tsx
+++ b/packages/components/src/Tabs/TabPanels.tsx
@@ -26,21 +26,31 @@
 
 import React, { Children, cloneElement, FC } from 'react'
 import styled from 'styled-components'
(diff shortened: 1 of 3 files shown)


lukelooker:
NOTE: Don't include Slack / JIRA links here since repo is public and the general public can't follow those links :)

packages/components/src/Tabs/TabPanels.tsx
```diff
@@ -51,11 +61,13 @@ export const TabPanels: FC<TabPanelsProps> = ({
...
+  return <Layout {...tabPanelsLayoutProps}>{clonedChildren}</Layout>
 }
 
-const Layout = styled.div<SpaceProps>`
+const Layout = styled.div<TabPanelsLayoutProps>`
   ${reset}
+  ${flexbox}
+  ${layout}
   ${space}
 `
```
lukelooker: I don't like the two different (encumbered) interfaces here. If you flip the component definition a bit you can land with just 1 interface and actually simplify the code a little bit as well:

```js

export interface TabPanelsProps extends FlexboxProps, LayoutProps, SpaceProps {
  children: JSX.Element[]
  className?: string
  selectedIndex?: number
  onSelectTab?: (index: number) => void
}

const Layout: FC<TabPanelsProps> = ({
  children,
  className,
  selectedIndex,
  ...props
}) => {
  const tabPanelsLayoutProps = omit(props, 'onSelectTab')

  const clonedChildren = Children.map(
    children,
    (child: JSX.Element, index: number) => {
      return cloneElement(child, {
        selected: index === selectedIndex,
      })
    }
  )

  return (
    <div className={className} {...tabPanelsLayoutProps}>
      {clonedChildren}
    </div>
  )
}

export const TabPanels = styled(Layout)`
  ${reset}
  ${flexbox}
  ${layout}
  ${space}
`

TabPanels.defaultProps = { pt: 'large' }
```

ElliotTheHuman: 👍 Went ahead and implemented the suggested changes above in https://github.com/looker-open-source/components/pull/658/commits/3ba1413f58d8a992e59aaf89e526254098c89217

PR #661, opened 2020-03-11
Fix duplicate tooltips on IconButton

mdodgelooker:
### :sparkles: Changes

- As of https://github.com/looker-open-source/components/pull/576 we are no longer clobbering `IconButton` event handlers coming from props, which resulted in duplicate tooltips in the core product.
- This update checks if any of the props event handlers matches (via `toString` ??) the event handlers for built-in tooltip to detect an outer tooltip.

- [x] PR is ideally < 400LOC

### :camera: Screenshots
**Issue:**
issue

**Fix:**
fix

### DIFF
--- a/packages/components/src/Button/IconButton.tsx
+++ b/packages/components/src/Button/IconButton.tsx
@@ -136,12 +136,27 @@ const IconButtonComponent = forwardRef(
       placement: tooltipPlacement,
     })
 
-    const eventHandlers = {
-      onBlur: useWrapEvent(onBlur, propsOnBlur),
-      onFocus: useWrapEvent(onFocus, propsOnFocus),
-      onMouseOut: useWrapEvent(onMouseOut, propsOnMouseOut),
-      onMouseOver: useWrapEvent(onMouseOver, propsOnMouseOver),
-    }
+    const hasOuterTooltip =
+      (propsOnFocus && onFocus.toString() === propsOnFocus.toString()) ||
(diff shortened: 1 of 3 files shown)


packages/components/src/Button/IconButton.tsx
```diff
@@ -136,12 +136,27 @@ const IconButtonComponent = forwardRef(
...
-      onMouseOut: useWrapEvent(onMouseOut, propsOnMouseOut),
-      onMouseOver: useWrapEvent(onMouseOver, propsOnMouseOver),
-    }
+    const hasOuterTooltip =
+      (propsOnFocus && onFocus.toString() === propsOnFocus.toString()) ||
+      (propsOnBlur && onBlur.toString() === propsOnBlur.toString()) ||
+      (propsOnMouseOut &&
+        onMouseOut.toString() === propsOnMouseOut.toString()) ||
+      (propsOnMouseOver &&
+        onMouseOver.toString() === propsOnMouseOver.toString())
```
mdodgelooker: This is pretty ridiculous – to check if any of the props handlers match the internal tooltip handlers character for character. Would it be better to just clean up any instance of an `IconButton` wrapped in a `Tooltip` (either just rely on `label` or add `tooltipDisabled`)?

My only other idea would be to add a context to `Tooltip` and look for that context here. It works well, but only for the render prop version, not the hook version.

mdodgelooker:
I like the looser check for `hasOuterTooltip` better. Ready for review.

PR #667, opened 2020-03-16
Upgrade acorn to address security vulnerability

lukelooker:


lukelooker:
This is a better version of #666 (use package.json resolution to preserve change consistently)

PR #687, opened 2020-03-17
InputTimeSelect

dbchristopher:
# Examples

## Convert shorthand input to closest option:
time-input

## Converts a 24 hour time input to appropriate formatting
time-input-24

## Accepts time inputs that do not show up on interval list
non-time-option

## While keyboard navigating, list scrolls to keep highlighted option visible
keyboard-nav


dbchristopher:
@mdodgelooker this is ready for review whenever you have time!! Thanks so much again for helping me get past some of those UX issues; it was such a lifesaver.

packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
```diff
@@ -123,7 +132,13 @@ const ComboboxListInternal = forwardRef(
         ref={popoverRef}
         p="none"
       >
-        <ul {...props} ref={forwardedRef} role="listbox" tabIndex={-1} />
+        <ul
+          {...props}
+          ref={forwardedRef}
+          role="listbox"
+          tabIndex={-1}
+          data-testid="combobox-list"
```
dbchristopher: added the testid to support tests which validate how the list content changes with `interval` and `format` props

packages/components/src/Form/Inputs/Combobox/ComboboxOption.tsx
```diff
@@ -68,6 +69,10 @@ export interface ComboboxOptionObject {
    * The value to match against when suggesting.
    */
   value: string
+  /**
+   * Make sure this option is visible in a long menu.
+   */
+  scrollIntoView?: boolean
```
dbchristopher: Since this prop causes it to scroll and highlight, I wonder if scrollIntoView isn't descriptive enough. Maybe `focus` or `highlight` would be a better prop name?

packages/www/src/Shared/ComponentStatus/ComponentStatus.tsx
```diff
@@ -81,6 +81,8 @@ const statusBackground = (props: StatusProps) => {
   }
 }
 
+const StyledBox = styled(Box)``
```
dbchristopher: Had to wrap this in `styled` because otherwise it didn't like styling the Reffable Box component on line 93.

> Argument of type 'ForwardRefExoticComponent<BoxProps & RefAttributes<HTMLDivElement>>' is not assignable to parameter of type 'Interpolation<ThemedStyledProps<Pick<DetailedHTMLProps<HTMLAttributes<HTMLSpanElement>, HTMLSpanElement>, "style" | "title" | ... 252 more ... | "is"> & ... 6 more ... & StatusProps, any>>'.

packages/components/src/utils/useReadOnlyWarn.ts
```diff
@@ -0,0 +1,13 @@
+import { useEffect } from 'react'
+
+// render a console warning if developers pass in a value without a change listener
+export function useReadOnlyWarn<V, CB>(name: string, value: V, onChange: CB) {
```
mdodgelooker: Nice!!

packages/components/src/Popover/Popover.tsx
```diff
@@ -489,6 +499,7 @@ export function usePopover({
     </ModalContext.Provider>
   )
   return {
+    contentContainerRef: { current: containerElement },
```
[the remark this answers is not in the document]
dbchristopher: ah thank you! that makes more sense

packages/components/src/Layout/Box/Box.tsx
```diff
@@ -44,12 +45,20 @@ export interface BoxProps
     FlexboxProps,
     PseudoProps,
     CursorProps,
-    UserSelectProps {}
+    UserSelectProps {
+  as?: ElementType
```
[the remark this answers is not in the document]
dbchristopher: I guess we were supporting it implicitly! And there is even a [Box unit test to validate that prop behavior](https://github.com/looker-open-source/components/blob/master/packages/components/src/Layout/Box/Box.test.tsx#L134-L149). 

But the weird part for me is that for some reason wrapping the component in forwardRef broke the styled component types. Once I made that change I got a bunch of lint errors like this:

```
packages/components/src/Layout/Box/Box.test.tsx:143:14 - error TS2322: Type '{ type?: string | undefined; children?: ReactNode; as: string; }' is not assignable to type 'IntrinsicAttributes & BoxProps & RefAttributes<HTMLDivElement>'.
  Property 'as' does not exist on type 'IntrinsicAttributes & BoxProps & RefAttributes<HTMLDivElement>'.

143         <Box as="input" {...props} />
                 ~~
```

I agree that it would be good to deprecate using Box in that way, but seemed out of scope for this PR. What do you think?

packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
```diff
@@ -155,7 +198,8 @@ const comboboxListStyles = css<ComboboxListInternalProps>`
   ${space}
   list-style-type: none;
   margin: 0;
-  padding: ${props => (props.isMulti ? props.theme.space.xsmall : 0)} 0;
+  padding: ${({ isMulti, theme }) => (isMulti ? theme.space.xsmall : 0)} 0;
```
dbchristopher: I had to make this change to get around the conflicting (`arrow-parens`)[https://eslint.org/docs/rules/arrow-parens] rule. Should be fine to land this PR I think. Then I'll open a PR into our shared eslint-config package to get the setting more in line with prettier preferences

dbchristopher:
@mdodgelooker I implemented the suggested changes and this should be ready for another review! Really glad you liked the ux updates to Combobox components 😄

packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
mdodgelooker: 👍

packages/components/src/Popover/Popover.tsx
```diff
@@ -141,7 +141,7 @@ export interface PopoverProps extends UsePopoverProps {
     /**
      * Used by popper.js to position the OverlaySurface relative to the trigger
      */
-    ref: Ref<any>,
+    ref: Ref<HTMLElement>,
```
mdodgelooker: The `any` wasn't great, but TS isn't liking `HTMLElement`. I can't think of a type that would be better here... 😕

packages/components/src/Layout/Box/Box.tsx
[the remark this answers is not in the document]
dbchristopher: 🤔 For some popover.tsx was initially mad at me for trying to add a box ref. Seems fine now! I'll revert.

dbchristopher:
Ok third time's the charm! Should be good to go now @mdodgelooker 🤞

PR #714, opened 2020-03-30
Prettier 2.0: Latest @looker/prettier-config & eslint-config

lukelooker:
### :sparkles: Changes

- Added `lint:prettier` task to verify all files are passing
- Added .prettierignore to keep prettier out of build artifacts
- Applied prettier to all files in repository
- Lots of minor changes to support new Prettier 2.0 defaults

- [ ] PR is ideally < 400LOC

### DIFF
--- a/packages/components/src/ActionList/ActionList.tsx
+++ b/packages/components/src/ActionList/ActionList.tsx
@@ -119,17 +119,17 @@ export const ActionListLayout: FC<ActionListProps> = ({
 export const ActionList = styled(ActionListLayout)<ActionListProps>`
   ${ActionListRowColumns} {
     display: grid;
-    grid-template-columns: ${props =>
-      props.columns.map(column => `${column.widthPercent}%`).join(' ')};
+    grid-template-columns: ${(props) =>
+      props.columns.map((column) => `${column.widthPercent}%`).join(' ')};
     align-items: center;
   }
 
   ${/* sc-selector */ ActionListItemColumn},
   ${/* sc-selector */ ActionListHeaderColumn} {
     display: flex;
-    padding: ${props => props.theme.space.small};
+    padding: ${(props) => props.theme.space.small};
   }
 
-  ${props => numericColumnCSS(getNumericColumnIndices(props.columns))}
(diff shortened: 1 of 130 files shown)


packages/components/src/ActionList/ActionList.tsx
```diff
@@ -119,17 +119,17 @@ export const ActionListLayout: FC<ActionListProps> = ({
 export const ActionList = styled(ActionListLayout)<ActionListProps>`
   ${ActionListRowColumns} {
     display: grid;
-    grid-template-columns: ${props =>
-      props.columns.map(column => `${column.widthPercent}%`).join(' ')};
+    grid-template-columns: ${(props) =>
+      props.columns.map((column) => `${column.widthPercent}%`).join(' ')};
```
dbchristopher: @lukelooker so we do want parens everywhere now? We should probably revert that last change I made to eslint-config 🤦‍♂

dbchristopher: Seeing these conflicting rules play out on save in vscode -- 
save-conflict

dbchristopher: Yeah, looks like they changed the default for 2.0

Probably the easiest thing at this point is to revert to the old style, unless there is some practical reason why parentheses are better in general

Screen Shot 2020-03-31 at 10 13 50 AM

dbchristopher:
@lukelooker looks like there is also a conflicting rule with trailing commas. I didn't initially notice this also changed on save: 
Screen Shot 2020-03-31 at 10 25 24 AM

packages/components/src/ActionList/ActionList.tsx
lukelooker: Yeah - I already did so :)

lukelooker: https://github.com/looker-open-source/eslint-config/commit/43b139df4efed4cfd98c199f36e73437cb1d2715

dbchristopher: oh! so is that just an issue with my VScode? 🤔

PR #717, opened 2020-03-30
Added CodeBlock component

lukelooker:
### :sparkles: Changes

- New `CodeBlock` component

- [x] PR is ideally < 400LOC

### :camera: Screenshots

image


packages/components/src/Text/CodeBlock.tsx
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+import { BorderProps } from '@looker/design-tokens'
+import styled from 'styled-components'
+import { TextBase, TextBaseProps } from './TextBase'
+
+export interface CodeBlockProps extends TextBaseProps, BorderProps {}
+
+export const CodeBlock = styled(TextBase).attrs({
+  as: 'pre',
```
jhardy, line 28: I think semantically this component should render a `<pre>` wrapping a `<code>` instead of just a `<pre>` 

Output would be `<pre><code>codeblock text</code></pre>`

packages/www/src/documentation/components/text/code.mdx
```diff
@@ -5,8 +5,28 @@ figma: file/SmIM3VUQUgo8xb4o7UUxflkT/Typography?node-id=304%3A2
...
 ```jsx
-<Code>const noop = () => null</Code>
+<Paragraph>
+  Some text with code: <Code>const noop = () => null</Code>
+</Paragraph>
+```
+
+## CodeBlock
+
+A `CodeBlock` is a dedicated container for multiple lines of source code.
```
jhardy, line 18: one question that came to mind is are people going to expect any type of syntax highlight for this component? I know that is out of scope for this PR. I just wonder if people will be reaching for this component and expect it to do more since it is positioned for displaying source code. Might be worth adding a note how it is unstyled.

packages/components/src/Text/CodeBlock.tsx
lukelooker: 👍

packages/www/src/documentation/components/text/code.mdx
lukelooker: Great point. Added more details.

PR #735, opened 2020-04-01
Scale icon size in Button

jhardy:
### :sparkles: Changes

This PR updates how icons are sized in `Button` before the icon was the same at all `Button` sizes. With this PR we scale the icon size based on the size of the `Button`

**Before**
Here is before where all the icons render at 16px
image

**After**
Now the icon scales with the button, most noticeable in `xsmall` and `large` sizes
image

- [ ] PR is ideally < 400LOC

### DIFF
--- a/packages/components/src/Button/icon.ts
+++ b/packages/components/src/Button/icon.ts
@@ -76,6 +76,7 @@ export const iconSizes = (props: ButtonProps) => {
     case 'large':
       iconSize = 18
       break
+    case 'medium':
     default:
       iconSize = 16
   }
--- a/packages/components/src/Button/icon.ts
+++ b/packages/components/src/Button/icon.ts
@@ -64,10 +64,37 @@ export const iconMargins = (props: ButtonProps) => {
   }
 }
 
+export const iconSizes = (props: ButtonProps) => {
+  let iconSize = 16
+  switch (props.size) {
+    case 'xsmall':
+      iconSize = 12
+      break
+    case 'small':
+      iconSize = 14
+      break
+    case 'large':
+      iconSize = 18
+      break
+    default:
+      iconSize = 16
+  }
+
+  if (props.iconBefore || props.iconAfter) {
+    return css`
+      width: ${iconSize}px;
+      height: ${iconSize}px;
+    `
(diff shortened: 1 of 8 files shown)


packages/components/src/Button/icon.ts
```diff
@@ -64,10 +64,37 @@ export const iconMargins = (props: ButtonProps) => {
...
+    case 'xsmall':
+      iconSize = 12
+      break
+    case 'small':
+      iconSize = 14
+      break
+    case 'large':
+      iconSize = 18
+      break
+    default:
```
lukelooker, line 79: I prefer seeing all cases handled in the switch even if there functionally equivalent – makes it easier to catch issues where the default value might get changed in one place in the code but not another.

```suggestion
    case 'medium':
    default:
```

packages/components/src/Button/icon.ts
```diff
@@ -64,10 +64,37 @@ export const iconMargins = (props: ButtonProps) => {
...
+  }
+
+  if (props.iconBefore || props.iconAfter) {
+    return css`
+      width: ${iconSize}px;
+      height: ${iconSize}px;
+    `
+  } else {
+    return false
+  }
```
lukelooker, line 90: Is there a reason to not just _always_ set this? If it's non-conditional than the fewer classes need to be generated within the SC stack so it's worth considering the simplification from that perspective as well as just easier to understand.

```suggestion
  return css`
    width: ${iconSize}px;
    height: ${iconSize}px;
  `
```

jhardy: @lukelooker without the conditional it always returns this css for every button. I tested this and it blew up 30 or so snapshots from anywhere a button was used. which scared me a bit. Testing them visually the snapshot failures were just class name changes and everything visually looked fine, but I did not go through each failure.

I think it will be fine since I believe its only generating a selector that targets an icon in the button, but I will defer to your judgment here.

lukelooker: Arg - silly snapshots. I think we should go for the pain of reviewing the snapshot changes now since I don't want someone down the road to think this logic check is required.

PR #737, opened 2020-04-02
[LENS-626] InputTime Component

dbchristopher:
## Only fires a change event when all three sub-inputs are filled in:

change-event

## Converts 24-hour time to 12-hour time (type 23 and it sets `11`:-- `PM`):
auto-convert

## Cycle through values with the up/down arrow keys
arrow-keys


packages/components/src/Form/Inputs/InputTime/InputTime.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+    warn: jest.fn(),
+  } as unknown) as Console
+})
+
+afterEach(() => {
+  jest.resetAllMocks()
+  global.console = globalConsole
+})
+
+const selectSubInputs = (mockProps: any) => {
```
dbchristopher, line 46: I feel like it's not the end of the world to have the `any` type used in a throwaway test snippet, but I'm incredibly perplexed as to why it breaks when I do `(mockProps: InputTimeProps)`. This is the blocking typescript error:

```bash
      packages/components/src/Form/Inputs/InputTime/InputTime.test.tsx:47:44 - error TS2769: No overload matches this call.
      Overload 1 of 2, '(props: Pick<Pick<InputTimeProps & { children?: ReactNode; } & { borderRadius?: "large" | "medium" | "small" | "none" | "xsmall" | undefined; color?: string | undefined; ... 6 more ...; borderColor: string; }, "p" | ... 56 more ... | "children"> & Partial<...>, "p" | ... 65 more ... | "height"> & { ...; } & { ...; }): ReactElement<...>', gave the following error.
        Type '{ format?: "24h" | "12h" | undefined; defaultValue?: string | undefined; value?: string | undefined; onChange?: ((time?: string | undefined) => void) | undefined; validationType?: "error" | undefined; ... 56 more ...; borderLeft?: string | ... 4 more ... | undefined; }' is not assignable to type 'Pick<Pick<InputTimeProps & { children?: ReactNode; } & { borderRadius?: "large" | "medium" | "small" | "none" | "xsmall" | undefined; color?: string | undefined; ... 6 more ...; borderColor: string; }, "p" | ... 56 more ... | "children"> & Partial<...>, "p" | ... 65 more ... | "height">'.
          Types of property 'px' are incompatible.
            Type 'string | number | symbol | (string | number | symbol | null)[] | { [x: string]: string | number | symbol | undefined; [x: number]: string | number | symbol | undefined; } | null | undefined' is not assignable to type '"large" | "medium" | "small" | "none" | "xsmall" | "xxsmall" | "xlarge" | "xxlarge" | "xxxlarge" | "xxxxlarge" | ((string | number | symbol | null)[] & "large") | ((string | number | symbol | null)[] & "medium") | ... 18 more ... | undefined'.
              Type 'null' is not assignable to type '"large" | "medium" | "small" | "none" | "xsmall" | "xxsmall" | "xlarge" | "xxlarge" | "xxxlarge" | "xxxxlarge" | ((string | number | symbol | null)[] & "large") | ((string | number | symbol | null)[] & "medium") | ... 18 more ... | undefined'.
      Overload 2 of 2, '(props: StyledComponentPropsWithAs<FC<InputTimeProps>, any, { borderRadius?: "large" | "medium" | "small" | "none" | "xsmall" | undefined; color?: string | undefined; fontSize?: "large" | "medium" | ... 10 more ... | undefined; ... 5 more ...; borderColor: string; }, "px" | ... 7 more ... | "height">): ReactElement<...>', gave the following error.
        Type '{ format?: "24h" | "12h" | undefined; defaultValue?: string | undefined; value?: string | undefined; onChange?: ((time?: string | undefined) => void) | undefined; validationType?: "error" | undefined; ... 56 more ...; borderLeft?: string | ... 4 more ... | undefined; }' is not assignable to type 'Pick<Pick<InputTimeProps & { children?: ReactNode; } & { borderRadius?: "large" | "medium" | "small" | "none" | "xsmall" | undefined; color?: string | undefined; ... 6 more ...; borderColor: string; }, "p" | ... 56 more ... | "children"> & Partial<...>, "p" | ... 65 more ... | "height">'.
          Types of property 'px' are incompatible.
            Type 'string | number | symbol | (string | number | symbol | null)[] | { [x: string]: string | number | symbol | undefined; [x: number]: string | number | symbol | undefined; } | null | undefined' is not assignable to type '"large" | "medium" | "small" | "none" | "xsmall" | "xxsmall" | "xlarge" | "xxlarge" | "xxxlarge" | "xxxxlarge" | ((string | number | symbol | null)[] & "large") | ((string | number | symbol | null)[] & "medium") | ... 18 more ... | undefined'.
              Type 'null' is not assignable to type '"large" | "medium" | "small" | "none" | "xsmall" | "xxsmall" | "xlarge" | "xxlarge" | "xxxlarge" | "xxxxlarge" | ((string | number | symbol | null)[] & "large") | ((string | number | symbol | null)[] & "medium") | ... 18 more ... | undefined'.

    47   const { getByTestId } = renderWithTheme(<InputTime {...mockProps} />)
```

Basically `InputTime` somehow isn't compatible with `InputTimeProps`???

a81b4a67303548708aa8694b3e0724ae

dbchristopher:
@mdodgelooker @lukelooker I tagged you both for review; I'm not sure who will have more time to take a look in the next few days. But whoever decides to take it on, this is kind of a beast! There is definitely a lot going on in `InputTime.tsx`, so please let me know if any part of it doesn't make sense and I'm happy to take another stab at simplifying or clarifying where possible.

mdodgelooker:
Diving in now :)

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
```diff
@@ -0,0 +1,553 @@
...
+  ${inputTextValidation}
+`
+
+const InputTimeWrapper = styled.div<{
+  hasInputValues: boolean
+}>`
+  color: ${({ theme, hasInputValues }) =>
+    hasInputValues
+      ? theme.colors.palette.charcoal600
+      : theme.colors.palette.charcoal300};
```
mdodgelooker, line 525: Love this subtle effect 👌

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
```diff
@@ -0,0 +1,553 @@
...
+      const newValue =
+        format === '12h'
+          ? convert12To24HrString(`${hour}:${minute} ${period}`)
+          : `${hour}:${minute}`
+      if (newValue !== value) {
+        onChange && onChange(`${newValue}`)
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isComplete, hour, minute, period])
```
mdodgelooker, line 443: Nice!

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
```diff
@@ -0,0 +1,553 @@
...
+
+import { ValidationType } from '../../ValidationMessage'
+
+export interface InputTimeProps extends SpaceProps, BorderProps {
+  format?: TimeFormats
+  defaultValue?: string
+  value?: string
+  onChange?: (time?: string) => void
+  validationType?: ValidationType
+  onValidationFail?: (value: string) => void
```
mdodgelooker, line 73: `InputChips` uses `onInvalid` as the prop name for the same feature – wanna pick one and make both consistent? I have no preference. :)

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
```diff
@@ -0,0 +1,553 @@
...
+  className?: string
+}
```
mdodgelooker, line 75: I think you'll want to include some more input props here, like `disabled` for sure and maybe `readOnly`. I'm not sure how `id`, `name`, `required`, or any aria attributes would work since there are 3 inputs, but worth considering too. And do we think folks will want callbacks for stuff like `onClick`, `onFocus`, etc? Might be too complex in this PR...

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
```diff
@@ -0,0 +1,553 @@
...
+      const valueProp = value || defaultValue
+      // Controlled Component: update state on every value change
+      if (isValidTime(valueProp)) {
+        const [newHour, newMinute, newPeriod] = parseValue(format, valueProp)
+        hour !== newHour &&
+          dispatch({ payload: newHour, type: 'SET_HOUR_VALUE' })
+        minute !== newMinute &&
+          dispatch({ payload: newMinute, type: 'SET_MINUTE_VALUE' })
+        period !== newPeriod &&
+          dispatch({ payload: newPeriod, type: 'SET_PERIOD_VALUE' })
```
mdodgelooker, line 419: I can't think of a use case but right now you can't update the `format` on the component after it's been mounted. Not sure if that's something it needs to support.

packages/components/src/Form/Inputs/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+    return true
+  }
+  const [hour = 0, minute = 0] = map(value.split(':'), parseBase10Int)
+
+  if (hour < 24 && minute <= 60) {
+    return true
+  }
+
+  return false
+}
```
mdodgelooker, line 66: Awesome that this was all reusable from `InputTimeSelect` 👍

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
```diff
@@ -0,0 +1,553 @@
...
+  }
+  &:hover {
+    ${inputTextHover}
+  }
+
+  &:disabled {
+    ${inputTextDisabled}
+  }
+
+  ${inputTextValidation}
```
mdodgelooker, line 516: Should it have a white background?
image

dbchristopher:
> Out of curiosity, was it just the 24h/12h formatting that took the native <input type="time"/> off the table?

It was mainly to have more control over the style and experience. It's handled pretty inconsistently across browsers. Edge renders this monstrosity for example:

edge_time

And then of course IE wouldn't support it at all, so it would just appear as a text input.

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
dbchristopher: Oh good call on making those consistent, thanks for the heads up there! I might be biased, but I think I lean towards `onValidationFail`. The term "validation fail" sounds more like an event, whereas "invalid" is more like a state. But not a huge deal to me either if you end up wanting to go with `onInvalid` on the next round of review

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
dbchristopher: oh dang that's a good catch! I'll definitely add in disabled and readOnly, and look into the rest of those suggestions

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
dbchristopher: Yeah I can't imagine people wanting to change format preferences after mount, right? 🤔

dbchristopher:
Thanks for the all the feedback and helping push me to fill in those gaps with the extra form/event props props. This should be ready for another look whenever you're available @mdodgelooker. Don't stress about any delay if you can't get to it this week!

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
dbchristopher: I didn't do anything `name` because I think that would only be used for vanilla html form submissions right? I feel like every react app should be tracking state externally (not even the react docs show any use of `name` attribute), so it didn't feel necessary.

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
```diff
@@ -0,0 +1,578 @@
...
+  return (
+    <div
+      className={`${className} ${disabled && 'disabled'}`}
+      onFocus={onFocus}
+      onBlur={onBlur}
+    >
+      <InputTimeWrapper hasInputValues={hasInputValues}>
+        <InputTimeLayout>
+          <InputText
+            id={id}
```
dbchristopher, line 470: The use case for `id` prop would be to associate it with a Label, to support the behavior where you click the label and the input automatically focuses. So with that in mind I thought it made sense to just pass the `id` prop to the first sub input here

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
```diff
@@ -0,0 +1,578 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isComplete, hour, minute, period])
+
+  const hasInputValues = some([hour, minute, period], 'length')
+
+  return (
+    <div
+      className={`${className} ${disabled && 'disabled'}`}
+      onFocus={onFocus}
+      onBlur={onBlur}
```
dbchristopher, line 465: Luckily react automatically supports `focus within` and `blur within` type behavior on these callbacks. It was super straightforward to add onFocus and onBlur props by just attaching them to the parent element

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
```diff
@@ -0,0 +1,582 @@
...
+              value={period}
+              onKeyDown={readOnly ? noop : handlePeriodKeyDown}
+              onFocus={handlePeriodFocus}
+              onBlur={handleBlur}
+              onChange={noop} // suppress controlled component warning
+              ref={inputRefs.PERIOD}
+              data-testid="input-period"
+              disabled={disabled}
+              readOnly={readOnly}
+              required={required}
```
dbchristopher, line 513: For `required` I thought it made the most sense to just pass that through to all three sub inputs. But I also can't imagine people using this as they'll probably do manual form validation?

dbchristopher: Same with `disabled`, and `readOnly` props too

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
```diff
@@ -0,0 +1,584 @@
...
+      if (isValidTime(valueProp)) {
+        const [newHour, newMinute, newPeriod] = parseValue(format, valueProp)
+        hour !== newHour &&
+          dispatch({ payload: newHour, type: 'SET_HOUR_VALUE' })
+        minute !== newMinute &&
+          dispatch({ payload: newMinute, type: 'SET_MINUTE_VALUE' })
+        period !== newPeriod &&
+          dispatch({ payload: newPeriod, type: 'SET_PERIOD_VALUE' })
+      } else {
+        onValidationFail && onValidationFail(valueProp)
```
dbchristopher, line 437: Oops -- I forgot to actually fire the `onValidationFail` callback at the appropriate time. Added this in as well as associated documentation and tests for that behavior

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
mdodgelooker: Nice!

PR #752, opened 2020-04-07
Fix usePopper top placement with window "offset parent"

mdodgelooker:
### :sparkles: Changes

- PopperJS 2.0 isn't behaving well with `placement: 'top'` (whether set or as a result of "flipping") and an anchor element whose "offset parent" is the window itself.
- I couldn't quite pin down why this was happening but I think it has something to do with PopperJS expecting the `bottom:0` to be the bottom of the page but it's actually the bottom of the window.
- Setting the `adaptive` option to false in `computeStyles` fixes the issue because then it uses only `top` and `left` properties which are more reliable.
- You won't see this issue in components.looker.com because the scrolling content section prevents it from happening, but you can see it in the playground (on master) as demonstrated by this screen capture:
popper-bug

- [x] PR is ideally < 400LOC

### :camera: Screenshots
**Fixed:**
popper-bug-fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `usePopper` (used in `Popover` and `Tooltip`) positioning issue with `placement="top"` and offset parent is the window
 - `Select` runtime error with no options, defaultValue or placeholder
 
 ## [0.7.25] - 2020-03-23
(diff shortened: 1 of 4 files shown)


FezVrasta:
@mdodgelooker have you reported this issue on the Popper repository already? I'd like to look into it.

mdodgelooker:
> @mdodgelooker have you reported this issue on the Popper repository already? I'd like to look into it.

@FezVrasta I thought about it, but I was having trouble creating a minimal repro. I'll give it another whirl. Thanks for asking!

mdodgelooker:
@FezVrasta major facepalm moment – the issue is due to our playground's `template.html` missing `<!DOCTYPE html>`:

https://codesandbox.io/s/eloquent-noether-ftdg0

I don't feel this qualifies as an issue on @popperjs/core – do you?

FezVrasta:
Popper is meant to run on recent-ish browsers, so yeah I suppose this is expected

PR #755, opened 2020-04-08
ActionList: ActionListManager Component

ElliotTheHuman:
### :sparkles: Changes
- Added ActionListManager Component to handle when an `ActionList` has no results / `ActionListItem`s and when an `ActionList` is in a loading state

- [x] PR is ideally < 400LOC

### :camera: Screenshots
action-list-manager

### DIFF
--- a/packages/components/src/ActionList/ActionListManager.tsx
+++ b/packages/components/src/ActionList/ActionListManager.tsx
@@ -1,0 +1,77 @@
+/*
+
+ MIT License
+
+ Copyright (c) 2020 Looker Data Sciences, Inc.
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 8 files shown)


packages/components/src/ActionList/ActionListManager.tsx
```diff
@@ -0,0 +1,77 @@
...
+  noResultsText?: string
+}
+
+const NoResults = styled(Text)``
+
+NoResults.defaultProps = {
+  fontSize: 'xlarge',
+  fontWeight: 'light',
+  variant: 'subdued',
+}
```
lukelooker, line 50: This should be a separate component / file. :)

ElliotTheHuman: Addressed in 444fb08!

PR #756, opened 2020-04-09
Establish and utilize ComponentsProvider

lukelooker:
### :sparkles: Changes

- Establish and utilize Components provider
- Added as a new package so that test-utils doesn't depend on components (creates circular dependency problems)

- [X] Includes test coverage for all changes
- [x] PR is ideally < 400LOC

### DIFF
--- a/packages/design-tokens/README.md
+++ b/packages/design-tokens/README.md
@@ -27,7 +27,7 @@ const StyledHeading = styled(Heading).attrs((props: HeadingProps) => {
 
 ### Building your own theme
 
-Once you've created your own theme object modeled after the values in this repo, you can override in production by wrapping components in via the `ComponentsProvider` (a Styled Components `ThemeProvider` is used under the hood).
+Once you've created your own theme object modeled after the values in this repo, you can override in production by wrapping components the `ComponentsProvider` (a Styled Components `ThemeProvider` is used under the hood).
 
 ```jsx
 import { ComponentsProvider } from '@looker/components'
--- a/packages/design-tokens/README.md
+++ b/packages/design-tokens/README.md
@@ -27,10 +27,10 @@ const StyledHeading = styled(Heading).attrs((props: HeadingProps) => {
 
 ### Building your own theme
 
-Once you've created your own theme object modeled after the values in this repo, you can override in production by wrapping components in a ThemeProvider.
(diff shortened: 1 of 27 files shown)


packages/design-tokens/README.md
```diff
@@ -27,20 +27,20 @@ const StyledHeading = styled(Heading).attrs((props: HeadingProps) => {
 
 ### Building your own theme
 
-Once you've created your own theme object modeled after the values in this repo, you can override in production by wrapping components in a ThemeProvider.
+Once you've created your own theme object modeled after the values in this repo, you can override in production by wrapping components in via the `ComponentsProvider` (a Styled Components `ThemeProvider` is used under the hood).
```
dbchristopher, line 30: should that say `wrapping components in the ComponentsProvider` rather than `...in via the ComponentsProvider`?

packages/design-tokens/README.md
```diff
@@ -27,20 +27,20 @@ const StyledHeading = styled(Heading).attrs((props: HeadingProps) => {
...
+import { ComponentsProvider } from '@looker/components'
 
 const myTheme = {
   ///... theme values here
 }
 
 const MyApp = () => {
   return (
-    <ThemeProvider theme={myTheme}>
+    <ComponentsProvider theme={myTheme}>
```
dbchristopher, line 41: Is this actually correct? It doesn't look like the ComponentsProvider accepts a `theme` prop

dbchristopher: Oh nevermind -- it accepts it by extending the ThemeProviderProps. But then it doesn't actually use the theme prop does it?

packages/www/src/documentation/getting-started/theme.mdx
```diff
@@ -69,24 +72,24 @@ You can extend the theme provided by Looker Components by creating a copy and th
   `
 
   return (
-    <ThemeProvider theme={myTheme}>
+    <ComponentsProvider theme={myTheme}>
```
dbchristopher, line 75: Same comment as above. Is `myTheme` actually going to get used in this case?

packages/design-tokens/README.md
lukelooker: Yes, fixing :)

packages/design-tokens/README.md
lukelooker: Yeah - I messed up the spread within the ComponentsProvider. 🤦

lukelooker:
@dbchristopher This should be ready for re-review now :)

PR #793, opened 2020-04-15
removed labelFontWeight from FieldInline and update test for it

laraharrow:
### :sparkles: Changes

- removed labelFontWeight from FieldInline 
- update test for it

- [ ] PR is ideally < 400LOC

### DIFF
--- a/packages/components/src/Form/Fields/FieldInline.tsx
+++ b/packages/components/src/Form/Fields/FieldInline.tsx
@@ -24,7 +24,7 @@
 
  */
 
-import React, { FunctionComponent } from 'react'
+import React, { FC } from 'react'
 import styled from 'styled-components'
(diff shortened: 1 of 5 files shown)


packages/components/src/Form/Fields/FieldInline.tsx
```diff
@@ -47,13 +47,12 @@ const FieldInlineLayout: FunctionComponent<FieldBaseProps> = ({
   children,
   label,
   labelFontSize,
-  labelFontWeight,
```
dbchristopher, line 50: we probably need to omit this from the props interface too?

laraharrow: oh of course!

laraharrow: strange I see that I did that locally but it is not matching on git hub, I'll pull it and try to get it right this time

PR #798, opened 2020-04-17
Include GlobalStyle within ComponentsProvider

lukelooker:
### :sparkles: Changes

- While building out a sandbox prototype it occurred to me that we could encapsulate GlobalStyles within our default provider and simplify the setup process by one more step :)

- [x] PR is ideally < 400LOC


packages/providers/src/ComponentsProvider.tsx
```diff
@@ -33,8 +35,20 @@ import { ThemeProvider, ThemeProviderProps } from './ThemeProvider'
...
-export interface ComponentsProviderProps extends ThemeProviderProps {}
-
-export const ComponentsProvider: FC<ComponentsProviderProps> = (props) => (
-  <ThemeProvider {...props} />
+export interface ComponentsProviderProps extends ThemeProviderProps {
+  /**
+   * Prevent automatic injection of a basic CSS-reset into the DOM
+   * @default false
+   */
+  noGlobalStyles?: boolean
```
dbchristopher, line 43: Now just need to conditionally checking this before rendering GlobalStyle, right?

lukelooker: Doh! 🤦 

Fixed now.

PR #804, opened 2020-04-20
Refactor for Fields & Inputs to improve aria (a11y) support

lukelooker:
### :sparkles: Changes

- `Field*` components have improved  properly specify `aria-describedby`
- `Checkbox`, `InputText`, `TextArea` and `ToggleSwitch` now properly set `aria-invalid`

Work is based on a some tips from this article => https://www.ovl.design/text/inclusive-inputs

- - Update documentation
- [x] Updated CHANGELOG
- - Checked for i18n impacts
- [ ] PR is ideally < ~400LOC


packages/components/src/Form/Fields/FieldInline.tsx
```diff
@@ -42,11 +43,19 @@ export interface CustomizableFieldAttributesInterface
  * The label will always be placed on the right side of the input.
  */
 
-const FieldInlineLayout: FC<Omit<FieldBaseProps, 'labelFontWeight'>> = ({
+interface FieldInlinePropsInternal extends FieldBaseProps {
+  id: string
+}
```
lukelooker, line 48: Making `id` a required prop so that folks don't forget to send the id to `Field` as well as the input

packages/components/src/Form/Fields/FieldTextArea/FieldTextArea.test.tsx
```diff
@@ -49,19 +53,41 @@ test('A FieldTextArea disabled', () => {
   const wrapper = mountWithTheme(
     <FieldTextArea disabled id="FieldTextAreaID" label="👍" />
   )
-  wrapper.find('input:disabled')
+  wrapper.find('textarea').html().includes('disabled=""')
```
lukelooker, line 56: Not entirely sure how this was passing before (what `input:disabled` was this finding). Anyway, fixed it.

packages/components/src/Form/Fields/RequiredStar.tsx
```diff
@@ -0,0 +1,40 @@
...
+const RequiredStarLayout: FC<{ className?: string }> = () => (
+  <>
+    <span aria-hidden="true"> * </span>
+    <VisuallyHidden>required</VisuallyHidden>
+  </>
+)
+
+export const RequiredStar = styled(RequiredStarLayout)`
+  color: ${({ theme }) => theme.colors.palette.red500};
+`
```
lukelooker, line 40: This was previously in a funny place (`FieldBaseProps.tsx`) and was getting a little more complex so I figured it deserved its own file.

packages/components/src/Form/Fields/FieldCheckbox/FieldCheckbox.tsx
```diff
@@ -41,11 +41,13 @@ const FieldCheckboxLayout = forwardRef(
...
     return (
       <FieldInline
-        validationMessage={validationMessage}
         {...pickFieldProps(props)}
+        validationMessage={validationMessage}
+        id={id}
       >
         <Checkbox
           {...omitFieldProps(props)}
+          aria-describedby={`${id}-describedby`}
```
jhardy, line 50: Any reason we are reaching for `aria-describedby` instead of just using the `for={id}` on the `<Label>` and the `id={id}` on `Checkbox`?

packages/components/src/Form/Fields/Field.tsx
```diff
@@ -130,9 +140,9 @@ const FieldLayout: FunctionComponent<FieldProps> = ({
       </Label>
       {detail && <FieldDetail>{detail}</FieldDetail>}
       <InputArea>{children}</InputArea>
-      <MessageArea>
-        {validationMessage && <ValidationMessage {...validationMessage} />}
+      <MessageArea id={`${id}-describedby`}>
```
jhardy, line 143: I would need to test this, but do we need to handle the case here if they do not provide a description the field is described by its label?

packages/components/src/Form/Fields/FieldSelect/FieldSelect.tsx
```diff
@@ -39,12 +39,13 @@ const FieldSelectComponent = forwardRef(
...
     return (
       <Field
+        {...pickFieldProps(props)}
         id={id}
         validationMessage={validationMessage}
-        {...pickFieldProps(props)}
       >
         <Select
           {...omitFieldProps(props)}
+          aria-describedby={`${id}-describedby`}
```
jhardy, line 48: Same question here, seems like adding `aria-desrcibedy` seems excessive when connection can be made with label and input.

packages/components/src/Form/Fields/FieldCheckbox/FieldCheckbox.tsx
lukelooker: So `describedby` is in _addition_ to the label. The article referenced above provides a better summary than I'll be able to do but the TLDR is that it offers additional context beyond the label behavior.

packages/components/src/Form/Fields/Field.tsx
lukelooker: Again it's supplementary, `label` is required, `description` is optional, this just helps a screenreader understand the relationship of the `description` and/or `validationMessage` to the actual input.

packages/components/src/Form/Fields/FieldSelect/FieldSelect.tsx
lukelooker: See above :)

packages/components/src/Form/Fields/FieldCheckbox/__snapshots__/FieldCheckbox.test.tsx.snap
```diff
@@ -312,6 +315,7 @@ exports[`A FieldCheckbox with checked value 1`] = `
   </div>
   <div
     className="c7 "
+    id="FieldCheckboxID-describedby"
```
jhardy, line 318: Here input is described by an empty div instead of its label

PR #835, opened 2020-04-24
Remove readOnly prop on Radio

lukelooker:
### :sparkles: Changes

- Remove readOnly prop on Radio

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/Form/Fields/FieldBase.tsx
+++ b/packages/components/src/Form/Fields/FieldBase.tsx
@@ -24,7 +24,6 @@
 
  */
 
-import { ReactNode } from 'react'
 import { FontSizes, FontWeights } from '@looker/design-tokens'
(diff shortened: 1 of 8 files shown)


packages/components/src/Form/Fields/FieldBase.tsx
```diff
@@ -33,7 +34,7 @@ export interface FieldBaseProps {
   /**
    * Defines the label for the field.
    */
-  label?: string
+  label?: ReactNode
```
mdodgelooker, line 37: I think TS isn't happy with this.

lukelooker: Doh! Was poking at things for somebody and actually left that in.

PR #836, opened 2020-04-24
WWW: Remove tabIndex from live code editor

dbchristopher:
### :sparkles: Changes

UX Improvement to make it easier to use keyboard navigation in the www components website. Tabbing will now skip over live code editors.

- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots

keyboard-nav-www


dbchristopher:
@mdodgelooker did the tab-index change! Had to modify that in a javascript callback after all, which is a little bit hacky. But seems worth it for the better www ux

PR #837, opened 2020-04-24
ActionList: Select All

ElliotTheHuman:
### :sparkles: Changes

- Added `Select All` functionality to `ActionList`
- Added `Select All` tests and docs
- Fixed minor bug where having `canSelect` but no `canSelectAll` prop on `ActionList` led to header and items being misaligned

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots
Select All


packages/components/src/ActionList/ActionList.tsx
```diff
@@ -84,21 +85,34 @@ export interface ActionListProps {
...
+   */
+  onSelectAll?: () => void
+  /**
+   * The ids of all ActionListItems which should be displayed as "selected"
    */
   itemsSelected?: string[]
+  /**
+   * The ids of all selectable ActionListItems (i.e. all non-disabled items)
+   */
+  areAllItemsSelected?: MixedBoolean
```
lukelooker, line 115: Could this be dynamically determined? Determine the IDs of all the rows and see if it equals `itemsSelected`? 

If that's not feasible I think I'd prefer a more terse prop name:

```suggestion
  /**
   * Control the state of the "Select All" checkbox
   */
  allSelected?: MixedBoolean
```

packages/components/src/ActionList/ActionList.tsx
```diff
@@ -170,8 +190,8 @@ export const ActionList = styled(ActionListLayout)<ActionListProps>`
   }
 
   ${ActionListHeader} {
-    padding-left: ${({ canSelect }) =>
-      canSelect ? actionListCheckboxWidth : undefined};
+    padding-left: ${({ canSelect, canSelectAll }) =>
+      canSelect && !canSelectAll ? actionListCheckboxWidth : undefined};
```
lukelooker, line 194: I think a guard should be put in place to prevent this combination check – for `canSelectAll` to be `false` if `canSelect=false` so the logic doesn't need to be held in more than one place.

```suggestion
    padding-left: ${({ canSelect, canSelectAll }) =>
      !canSelectAll && actionListCheckboxWidth};
```

packages/components/src/ActionList/ActionListContext.tsx
```diff
@@ -25,19 +25,24 @@
  */
 
 import { createContext } from 'react'
+import { MixedBoolean } from '../Form'
 import { ActionListColumns } from './ActionList'
 
 export interface ActionListContextProps {
+  areAllItemsSelected?: MixedBoolean
```
lukelooker, line 32:
```suggestion
  allSelected?: MixedBoolean
```

packages/components/src/Form/Inputs/Checkbox/CheckMarkMixed.tsx
```diff
@@ -34,6 +34,7 @@ export const CheckMarkMixed = () => (
     fill="none"
     xmlns="http://www.w3.org/2000/svg"
   >
+    <title>Check Mark Mixed</title>
```
lukelooker, line 37: 👍

packages/components/src/ActionList/ActionList.tsx
ElliotTheHuman: That makes sense! I ended up changing the lines where the context object is created. The context's `canSelectAll` property is now reliant on both the `canSelect` and `canSelectAll` props, so any other component using the context will not have to performa an additional check for `canSelect`. See 6379e2d.

However, for this chunk of styling in particular, I _think_ we do need the combination check  (it's meant to handle the case where you have `canSelect` true but `canSelectAll` false). We want to have the header have a chunk of padding-left only in the case where the row's have checkboxes, and the header doesn't.

Let me know if I'm misunderstanding something here though with the suggested commit or the suggestion in general.

packages/components/src/ActionList/ActionList.tsx
ElliotTheHuman: Commit 12fa82a includes a proposed idea for dynamically determining all selected ids (that involves adding `allItems` and `setAllItems` properties to the context)

packages/components/src/ActionList/ActionListContext.tsx
ElliotTheHuman: Commit 12fa82a includes a proposed idea for dynamically determining all selected ids (that involves adding allItems and setAllItems properties to the context)

packages/components/src/ActionList/ActionList.tsx
ElliotTheHuman: A number of commits worked towards getting the `select-all` logic baked into the `ActionList`, but e44b2ff covers the final structure!

packages/components/src/ActionList/ActionList.tsx
```diff
@@ -112,11 +116,42 @@ export const ActionListLayout: FC<ActionListProps> = ({
...
   columns,
-  itemsSelected,
-  onClickRowSelect,
+  itemsSelected = [],
+  onClickRowSelect = false,
   onSelect,
+  onSelectAll,
   onSort,
 }) => {
+  const [allItems, setAllItems] = useState([] as string[])
```
lukelooker, line 125:
```suggestion
  const [allItems, setAllItems] = useState<string[]>([])
```

Using `as` is dangerous because Typescript will just believe whatever you tell it:

```js
  const [allItems, setAllItems] = useState([] as WhateverYouSayHereIsTheTypeNow)
```

packages/components/src/ActionList/ActionListItem.tsx
```diff
@@ -48,9 +48,17 @@ const ActionListItemInternal: FC<ActionListItemProps> = ({
...
+    addItemToAllItems,
+    canSelect,
+    itemsSelected,
+    onSelect,
+    onClickRowSelect,
+  } = useContext(ActionListContext)
+
+  useEffect(() => {
+    !disabled && addItemToAllItems(id)
+  }, [addItemToAllItems, disabled, id])
```
lukelooker, line 61: Add all items regardless of their disabled state - I think it feels funny otherwise.

```suggestion
  useEffect(() => addItemToAllItems(id), [addItemToAllItems, id])
```

I wonder if perhaps the final fix here (in a future PR) is an option to omit the checkbox for those items that are disabled?

packages/components/src/ActionList/ActionList.tsx
ElliotTheHuman: Resolved in 76dfb2d!

packages/components/src/ActionList/ActionListItem.tsx
ElliotTheHuman: Resolved in https://github.com/looker-open-source/components/pull/837/commits/0c8b5a8b1cfd4c930a5891ba1a42a93a77412fa7!

packages/components/src/ActionList/ActionList.tsx
```diff
@@ -84,21 +85,34 @@ export interface ActionListProps {
...
+   * Callback performed when user makes selects the header checkbox
+   */
+  onSelectAll?: () => void
+  /**
+   * The ids of all ActionListItems which should be displayed as "selected"
    */
   itemsSelected?: string[]
+  /**
+   * The ids of all selectable ActionListItems (i.e. all non-disabled items)
+   */
```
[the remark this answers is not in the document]
ElliotTheHuman: Not sure about this, but I _think_ this comment is on an outdated version of `ActionList.tsx`! This prop (and its comment from 112 to 114) were deleted in commit e251b33b56ffc1f0ef192252f61797d3059b46c1. The latest version of `ActionList.tsx` should have this doc string completely removed.

packages/components/src/ActionList/ActionListItem.tsx
ElliotTheHuman: And I actually like the look of omitted checkboxes for disabled rows. I added it in since it was a minor change, and thought it was worth including in this PR. (change included in commit bbd4a3010).

Screen Shot 2020-04-29 at 4 36 40 PM

ElliotTheHuman: Let me know if you'd rather keep this change out of the current PR!

lukelooker: I want to run that by design before we make that change so let's pull it out for now :)

ElliotTheHuman: 🖖

packages/components/src/ActionList/ActionListContext.tsx
```diff
@@ -28,16 +28,22 @@ import { createContext } from 'react'
 import { ActionListColumns } from './ActionList'
 
 export interface ActionListContextProps {
+  allItems: string[]
   canSelect: boolean
+  canSelectAll: boolean
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in https://github.com/looker-open-source/components/pull/837/commits/7dbfeb19bbe12968fb159a20886e1ee3303bb34e!

PR #870, opened 2020-05-03
Windowing for Select & SelectMulti

mdodgelooker:
### :sparkles: Changes

- Filtering in the core product often requires having up to 1,000 options in a dropdown list
- With this PR, if options >= 100, only the ones currently visible in the scroll window will render
- `windowedOptions` prop overrides the above (either `true` for < 100 or `false` for >= 100)
- This impacted various `Combobox` behaviors including keyboard navigating through options via the arrow keys and scrolling to the current value on open
  - `Select/useWindowedOptions.tsx` manages these behaviors
  - `utils/getWindowedListBoundaries.ts` is generic and may be used for windowing in other list components
- Grouped options not supported

- [x] Update documentation
- [x] Updated CHANGELOG

#### With windowing:
windowed

#### Without windowing:
unwindowed


jest.setup.js
```diff
@@ -42,6 +42,9 @@ const observeMock = function(cb, config) {
 const globalAny = global
 globalAny.IntersectionObserver = observeMock
 
+// js-dom doesn't do scrollIntoView
+Element.prototype.scrollIntoView = jest.fn()
```
mdodgelooker, line 46: Got an error on this guy in one of the new tests. https://github.com/jsdom/jsdom/issues/1695

packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
```diff
@@ -181,20 +193,24 @@ const ComboboxListInternal = forwardRef(
...
 
-      const scrollListener = () => {
+      const scrollListener = throttle(() => {
         if (contentContainer) {
           setListClientRectOnce(contentContainer)
           setListScrollPosition &&
             setListScrollPosition(contentContainer.scrollTop)
         }
-      }
+      }, 50)
```
mdodgelooker, line 202: Taking any performance boost I can get.

packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
```diff
@@ -181,20 +193,24 @@ const ComboboxListInternal = forwardRef(
...
 
-      contentContainer &&
+      if (contentContainer) {
         contentContainer.addEventListener('scroll', scrollListener)
+        scrollListener()
```
mdodgelooker, line 206: We now need the height and scroll position right away on initial render.

packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
```diff
@@ -181,20 +193,24 @@ const ComboboxListInternal = forwardRef(
...
+      }
 
       return () => {
         contentContainer &&
           contentContainer.removeEventListener('scroll', scrollListener)
+        setListScrollPosition && setListScrollPosition(0)
+        setListClientRect && setListClientRect(undefined)
```
mdodgelooker, line 213: Since these values are going into the context, we have to clear them when the list unmounts, because the provider is still mounted and otherwise persists those previous values.

packages/components/src/Form/Inputs/Combobox/utils/useAddOptionToContext.ts
```diff
@@ -38,13 +38,15 @@ export function useAddOptionToContext<
...
+  const { optionsRef, windowedOptionsPropRef } = useContext(context)
   const indexRef = useRef<number>(-1)
 
   useEffect(() => {
     const option = { label, scrollIntoView, value }
     const optionsRefCurrent = optionsRef && optionsRef.current
-    if (optionsRefCurrent) {
+    const windowedOptions =
+      windowedOptionsPropRef && windowedOptionsPropRef.current
+    if (optionsRefCurrent && !windowedOptions) {
```
mdodgelooker, line 49: This hook is for letting each option "register" itself to the context when it mounts, but since any option outside the scroll window will not be mounted, we disable this behavior here and add it a different way in `useWindowedOptions`.

packages/components/src/Form/Inputs/Combobox/utils/useKeyDown.ts
```diff
@@ -80,7 +81,7 @@ export function useKeyDown() {
     }
   }
 
-  return function handleKeyDown(event: KeyboardEvent<HTMLDivElement>) {
+  return throttle(function handleKeyDown(event: KeyboardEvent<HTMLDivElement>) {
```
mdodgelooker, line 84: Helps performance when the user clicks and holds an arrow key.

packages/components/src/Form/Inputs/Combobox/utils/useOptionEvents.ts
```diff
@@ -62,10 +67,11 @@ export function useOptionEvents<
     }
   }
 
-  function handleMouseEnter() {
+  const handleMouseEnter = throttle(() => {
```
mdodgelooker, line 70: Another small performance boost.

packages/components/src/Form/Inputs/Combobox/utils/useOptionEvents.ts
```diff
@@ -62,10 +67,11 @@ export function useOptionEvents<
...
+    if (isScrollingRef && isScrollingRef.current) return
```
mdodgelooker, line 71: Without this, if you keyboard navigate down to options below the scroll window, causing a small auto-scroll, and your mouse is resting somewhere on the list, a `mouseenter` will be triggered where your mouse is. (Existing issue.)

packages/components/src/Form/Inputs/Select/Select.test.tsx
```diff
@@ -228,19 +224,110 @@ describe('Select / SelectMulti', () => {
...
+      ],
+      [
+        'SelectMulti',
+        <SelectMulti
+          placeholder="Search"
+          noOptionsLabel={label}
+          key="select-multi"
+        />,
+      ],
+    ])('custom label text (%s)', (_, jsx) => {
```
mdodgelooker, line 319: Git diff is being funny here. This was an existing test, but the 2 above are now.

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
```diff
@@ -50,21 +51,39 @@ export interface SelectOptionGroupProps {
...
     )
   }
   return <ComboboxOption {...option} key={index} />
 }
 
-const renderMultiOption = (option: SelectOptionObject, index: number) => {
+const renderMultiOption = (
+  option: SelectOptionObject,
+  index: number,
+  scrollIntoView?: boolean
```
mdodgelooker, line 77: Enables scrolling to the bottom of the list if the user arrows up from the first option or to the top of the list if the user arrows down from the last option.

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
```diff
@@ -174,7 +199,42 @@ export function SelectOptions({
...
+        // If the windowedOptions prop is true but there are groups, give a warning
+        /* eslint-disable-next-line no-console */
+        console.warn(
+          'The `windowedOptions` prop does not support grouped options.'
+        )
+      }
+      return false
+    }
+    return true
+  }, [options, propsWindowedOptions])
```
mdodgelooker, line 224: When to window the options – option length is 100 or more, by default, with prop override.

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
```diff
@@ -174,7 +199,42 @@ export function SelectOptions({
...
+  const {
+    start,
+    end,
+    before,
+    after,
+    scrollToFirst,
+    scrollToLast,
+  } = useWindowedOptions(windowedOptions, options, isMulti)
+
+  const optionsToRender = options && options.slice(start, end + 1)
```
mdodgelooker, line 235: Will include the entire list if `windowedOptions` is false (see `getWindowedListBoundaries.ts`).

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
```diff
@@ -192,25 +252,41 @@ export function SelectOptions({
 
   return (
     <>
-      {options && options.length > 0
+      {options && scrollToFirst
+        ? renderToUse(options[0] as SelectOptionObject, 0, true)
```
mdodgelooker, line 256: If the user arrows up from the first option.

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
```diff
@@ -192,25 +252,41 @@ export function SelectOptions({
...
             createOption,
           ]
         : createOption || noOptions}
+      {after}
+      {options && scrollToLast
+        ? renderToUse(
+            options[options.length - 1] as SelectOptionObject,
+            0,
+            true
+          )
```
mdodgelooker, line 286: If the user arrows up from the first option.

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
```diff
@@ -174,7 +199,42 @@ export function SelectOptions({
...
+    )
+    if (groupedOptions) {
+      if (propsWindowedOptions) {
+        // If the windowedOptions prop is true but there are groups, give a warning
+        /* eslint-disable-next-line no-console */
+        console.warn(
+          'The `windowedOptions` prop does not support grouped options.'
+        )
+      }
+      return false
```
mdodgelooker, line 221: Not able to window with grouped options at this time.

packages/components/src/Form/Inputs/Combobox/utils/useOptionEvents.ts
dbchristopher: ah that's awesome!

packages/components/src/Form/Inputs/Select/Select.test.tsx
```diff
@@ -228,19 +224,110 @@ describe('Select / SelectMulti', () => {
...
+        getJSX(longOptions)
+      )
+
+      const input = getByPlaceholderText('Search')
+
+      fireEvent.mouseDown(input)
+
+      expect(queryByText('0')).toBeInTheDocument()
+      expect(queryByText('99')).not.toBeInTheDocument()
+
```
dbchristopher, line 300: this doesn't seem to validate that `99 options all render`. should you also add that `queryByText('99')` **is** in the document?

dbchristopher: Also I'm not sure that I can see any difference between the component rendered here and in the previous test. Is the `windowedOptions` prop toggled somewhere?

packages/components/src/Form/Inputs/Select/utils/useWindowedOptions.tsx
```diff
@@ -0,0 +1,129 @@
...
+        navigationOption.value,
+      ])
+      start = selectedIndex
+      end = selectedIndex
+    }
+    isFirstRender.current = false
+  }
+
+  // If the user keyboard navigates "down" from the last option or "up" from the first option
+  // we need to render the top or bottom of the list (which are outside our "window") and scroll there
```
dbchristopher, line 100: This doesn't seem to be working as expected. I'm continually pressing `up` arrow in this screencast:
up-nav

packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
```diff
@@ -72,6 +73,13 @@ export interface ComboboxListProps
    * @default true
    */
   closeOnSelect?: boolean
+  /**
+   * Render only the options visible in the scroll window
+   * Requires manually updating ComboboxContext.optionsRef with complete
+   * list of options in order for keyboard navigation to work properly
+   * @default false
+   */
+  windowedOptions?: boolean
```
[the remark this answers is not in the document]
mdodgelooker: I think convention calls for either "virtualize" or "window" https://reactjs.org/docs/optimizing-performance.html#virtualize-long-lists

I went for "window" because it's shorter but maybe "virtualize" is a more specific word... thoughts? I wonder if @lukelooker has an opinion... 😁

packages/components/src/Form/Inputs/Select/utils/useWindowedOptions.tsx
mdodgelooker: Oops, looking into it now.

packages/components/src/Form/Inputs/Combobox/utils/useOptionScroll.ts
```diff
@@ -58,12 +58,14 @@ export function useOptionScroll<
...
     listScrollPosition = 0,
     listClientRect = { height: 0 },
+    isScrollingRef,
   } = useContext(context)
   /* scroll menu list to specified element on mount */
   const [newTriggerElement, callbackRef] = useCallbackRef()
   useEffect(() => {
     if (scrollIntoView) {
       if (newTriggerElement) {
+        // if (isScrollingRef) isScrollingRef.current = true
```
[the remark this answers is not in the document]
mdodgelooker: ✅

packages/components/src/Form/Inputs/Select/Select.test.tsx
mdodgelooker: Oh whoops, good catch – bad copy/pasting on my part. 🤪Now it's checking that "98" (the 99th option) _is_ rendered.

The default behavior is to turn on windowing at 100, which I've read is where performance starts to go downhill. Then the `windowedOptions` prop can be used to override the default (either on or off, depending on the length of options).

packages/components/src/Form/Inputs/Select/utils/useWindowedOptions.tsx
mdodgelooker: This should be fixed now. I had moved the logic that automatically turns on windowing at 100 options to the wrong spot. Now it's a hook under `Select/utils` so it can be used by both `Select` and `SelectMulti`.

mdodgelooker:
Ready for another round of review @dbchristopher 🙂

PR #873, opened 2020-05-04
Lara/inline text area/lens591

laraharrow:
✨ Changes

- implemented InputTextArea

✅ Requirements
- [x] PR is ideally < 400LOC

📷 Screenshots

Screen Shot 2020-05-05 at 5 17 48 PM


packages/components/src/Form/Inputs/InlineTextArea/InlineTextArea.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+    return (
+      <div className={className} data-testid="inline-text-area">
+        <Input
+          onChange={handleChange}
+          underlineOnlyOnHover={underlineOnlyOnHover}
+          value={displayValue}
+          ref={ref}
+          {...pick(props, inputPropKeys)}
+        />
```
dbchristopher, line 71: Would this work with a textarea instead? the text that i'm typing/selecting doesn't actually line up with the visible text content with this combination
Screen Shot 2020-05-04 at 5 12 35 PM

packages/www/src/documentation/components/forms/inline-text-area.mdx
```diff
@@ -0,0 +1,22 @@
...
+  <InlineTextArea
+    mb="xlarge"
+    placeholder="Type here and keep typing to see multiple lines effect of this component it will wrap to a next line..."
+  />
+  <InlineTextArea
+    mt="xlarge"
+    value="Type here and keep typing to see multiple lines effect of this component it will wrap to a next line..."
+  />
+</Grid>
+```
```
dbchristopher, line 22: It would be good to document the `underlineOnlyOnHover` prop too

packages/components/src/Form/Inputs/InlineTextArea/InlineTextArea.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+ */
+
+import { assertSnapshot } from '@looker/components-test-utils'
+import React from 'react'
+import { InlineTextArea } from './InlineTextArea'
+
+test('InlineTextArea renders an input with a placeholder', () => {
+  assertSnapshot(<InlineTextArea placeholder="this is the placeholder" />)
+})
```
dbchristopher, line 33: it would also be good to test what happens when there is a value prop too. Also you should try to do it without snapshots!

packages/components/src/Form/Inputs/InlineTextArea/InlineTextArea.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { inputPropKeys, InputProps } from '../InputProps'
+
+interface InlineTextAreaProps
+  extends TypographyProps,
+    Omit<InputProps, 'type'> {
+  underlineOnlyOnHover?: boolean
+  value?: string
+}
+
+export const InlineTextAreaInternal = forwardRef(
```
dbchristopher, line 41: Has Jared had a chance to do design QA on this yet? I would assume it would shrink to fit the content just like `InlineInputText`, so just wanted to double check this is intended behavior
Screen Shot 2020-05-04 at 5 20 10 PM

laraharrow: Screen Shot 2020-05-05 at 12 14 41 PM
it should be the whole line based on the figma file!

PR #897, opened 2020-05-07
useTooltip: WAI-ARIA attributes

ElliotTheHuman:
### :sparkles: Changes

- Added `role=tooltip` to the `<TooltipContent />` element in `useTooltip`
- Updated return object in `useTooltip` to include a generated (or passed-in) tooltipId

**Note:** These changes were made to fulfill the two asks from Quality Eng.
```
The element that serves as the Tooltip container has role=tooltip.

The element that triggers the Tooltip references the Tooltip element with aria-describedby=xxx.
```

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


ElliotTheHuman:
Almost there. Current blocker is that snapshot tests for components that utilize `Tooltip` or `useTooltip` fail repeatedly. This is do to changing id's caused by `useID` in `useTooltip`.

lukelooker:
> Almost there. Current blocker is that snapshot tests for components that utilize `Tooltip` or `useTooltip` fail repeatedly. This is do to changing id's caused by `useID` in `useTooltip`.

The workaround for this is that the developer should be allowed to explicitly specify an `id` if they so choose - then in this case they it can be explicitly specified as a static value so that they don't change from test-run to test-run. This is the pattern we use for `Field*`

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -117,6 +124,7 @@ export function useTooltip({
   textAlign,
   disabled,
   surfaceStyles,
+  tooltipId,
```
lukelooker, line 127:
```suggestion
  id,
```

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -82,6 +77,11 @@ export interface UseTooltipProps {
    */
   textAlign?: TextAlignProperty
 
+  /**
+   * The id of the span containing the tooltip text (if absent, a random id will be generated)
+   */
+  tooltipId?: string
```
lukelooker, line 83:
```suggestion
  id?: string
```

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -190,6 +198,8 @@ export function useTooltip({
 
   const ref = useForkedRef(targetRef, surfaceCallbackRef)
 
+  const tooltipContentId = useID(tooltipId)
```
lukelooker, line 201: `id`
```suggestion
  const guaranteedId = useID(id)
```

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -207,27 +217,40 @@ export function useTooltip({
           color="palette.charcoal000"
           {...surfaceStyles}
         >
-          <TooltipContent width={width} textAlign={textAlign}>
+          <TooltipContent
+            role="tooltip"
+            id={tooltipContentId}
```
lukelooker, line 222:
```suggestion
            id={guaranteedId}
```

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -207,27 +217,40 @@ export function useTooltip({
...
+          >
             {content}
           </TooltipContent>
         </OverlaySurface>
       </ModalContext.Provider>
     ) : null
 
   return {
-    eventHandlers,
+    ...eventHandlers,
```
lukelooker, line 233:
```suggestion
    onBlur: handleClose,
    onFocus: handleOpen,
    onMouseOut: handleMouseOut,
    onMouseOver: handleOpen,
```

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -207,27 +217,40 @@ export function useTooltip({
...
     popperInstanceRef,
     ref: callbackRef,
     tooltip: popper,
+    tooltipId: tooltipContentId,
```
lukelooker, line 237:
```suggestion
    'aria-describedby': guaranteedId,
```

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -207,27 +217,40 @@ export function useTooltip({
...
 
 export const Tooltip: FC<TooltipProps> = ({ children, ...props }) => {
-  const { eventHandlers, tooltip, ref } = useTooltip(props)
+  const tooltipProps = useTooltip(props)
+
+  // Take tooltipId property and set it to aria-describedby property instead
+  const tooltipPropsLabeled = {
+    'aria-describedby': tooltipProps.tooltipId,
+    ...omit(tooltipProps, ['tooltip', 'tooltipId', 'popperInstanceRef']),
+  }
```
lukelooker, line 248:
```suggestion
  const tooltipPropsLabeled = {
    ...omit(tooltipProps, ['tooltip', 'popperInstanceRef']),
  }
```

packages/components/src/Tooltip/Tooltip.tsx
ElliotTheHuman: Resolved in e0a2e4fc1!

packages/components/src/Tooltip/Tooltip.tsx
ElliotTheHuman: Resolved in e0a2e4fc1!

packages/components/src/Tooltip/Tooltip.tsx
ElliotTheHuman: Resolved in e0a2e4fc1!

packages/components/src/Tooltip/Tooltip.tsx
ElliotTheHuman: Resolved in e0a2e4fc1!

packages/components/src/Tooltip/Tooltip.tsx
ElliotTheHuman: Resolved in e0a2e4fc1!

packages/components/src/Tooltip/Tooltip.tsx
ElliotTheHuman: Resolved in e0a2e4fc1!

ElliotTheHuman:
The suggestion to pass along id props from the parents of `Tooltip` / users of `useTooltip` to create consistent ids during testing worked great!

One note is that in order to achieve the `create a random id if no id was passed in as a prop` effect in `IconButton` and `Banner` (as well as with `ModalHeader`) I had to create statements like the one below:

```
<IconButton
```

CHANGELOG.md
```diff
@@ -5,7 +5,14 @@ All notable changes to this project will be documented in this file.
...
 
-## [UNRELEASED]
+## [Unreleased]
+
+### Changed
+
+- `useTooltip` includes a generated id (or passed-in prop value) for the resulting tooltip in the return object
+  - Doing this means that tooltip trigger elements can now have an `aria-describedby` property with said id as the value
+- `MenuDisclosure`, `Banner`, `IconButton`, `ModalHeader` explicitly use their id props to either provide `useTooltip` with an id or to provide another component that uses `useTooltip` with an id to generate the tooltip's id
+=======
```
lukelooker, line 15:
```suggestion
=======
```

ElliotTheHuman:
Re-requested review after CHANGELOG.md merge conflict resolution!

PR #908, opened 2020-05-12
Button size adjustments

jhardy:
### :sparkles: Changes

This PR increases the horizontal padding on our `Button` and `ButtonOutline` components which we are updating on the design systems side. It also reduces the horizontal padding on our `ButtonTransparent` for better visual alignment with other UI elements.

image

image

- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


packages/components/src/Button/size.ts
```diff
@@ -71,27 +71,27 @@ export const buttonSize = variant({
...
       fontSize: 'small',
       height: `${buttonSizeMap.small}px`,
-      px: 'small',
+      px: 'large',
     },
     medium: {
       fontSize: 'medium',
       height: `${buttonSizeMap.medium}px`,
-      px: 'medium',
+      px: '24px',
```
dbchristopher, line 89: any reason this is a hard-coded rather than named?

jhardy: We don't have this value in the theme, so using a hard coded value instead. I can add a comment here if that would help.

packages/components/src/Button/ButtonTransparent.tsx
```diff
@@ -25,13 +25,38 @@
...
 
 import styled from 'styled-components'
-import { ButtonBase } from './ButtonBase'
+import { ButtonBase, ButtonBaseProps } from './ButtonBase'
+
+export const transparentButtonPadding = (props: ButtonBaseProps) => {
+  let padding = '20px'
+  switch (props.size) {
+    case 'xsmall':
+      padding = '8px'
```
lukelooker, line 34:
```suggestion
      padding = theme.space.xxsmall
```

packages/components/src/Button/ButtonTransparent.tsx
```diff
@@ -25,13 +25,38 @@
...
+      break
+    case 'large':
+      padding = '20px'
+      break
+    case 'medium':
+    default:
+      padding = '20px'
+  }
+
+  return padding
```
lukelooker, line 47:
```suggestion
  switch (props.size) {
    case 'xsmall':
      return theme.padding.xsmall
    case 'small':
      return theme.padding.medium
    case 'large':
    case 'medium':
    default:
      return theme.padding.large
  }
```

packages/components/src/Button/icon.ts
```diff
@@ -39,13 +39,16 @@ export const iconMargins = (props: ButtonProps) => {
...
       break
     case 'small':
+      spacing.small = 8
+      spacing.large = 8
+      break
     case 'large':
     default:
-      spacing.small = 4
+      spacing.small = 12
       spacing.large = 8
```
lukelooker, line 52: We probably need to take holistic look at getting Icons on a common size-system like fonts, buttons, etc... (a problem for later but probably worth creating a ticket to get that nailed down)

packages/components/src/Button/size.ts
lukelooker: Perhaps we should add it, if even just to track the oddity? `materialLargeish`?

packages/components/src/Button/icon.ts
jhardy: Added this to our design systems todo

packages/components/src/Button/ButtonTransparent.tsx
jhardy: I took a pass at refactoring this to use styled-systems variant function. Could you give it another look?

packages/components/src/Button/size.ts
jhardy: I toyed around with the idea of adding this, my preference would be to have the outlier here and not name it so it doesn't get used anywhere else we have our named size properties. This way we can spend some time working on a name. 

If you still feel it's best to add it my vote would be `mediumLarge` for now.

packages/components/src/Button/ButtonTransparent.tsx
```diff
@@ -25,13 +25,40 @@
  */
 
 import styled from 'styled-components'
+import { variant } from 'styled-system'
 import { ButtonBase } from './ButtonBase'
 
+/* eslint-disable sort-keys-fix/sort-keys-fix */
+export const transparentButtonPadding = variant({
```
lukelooker, line 32: Super minor but this shouldn't be exported unless it's used elsewhere.

```suggestion
const transparentButtonPadding = variant({
```

packages/components/src/Button/size.ts
lukelooker: I'm cool with leaving a "special" value here for now - as we discussed a more holistic re-thinking of theme.space will come with Material stuff later.

PR #919, opened 2020-05-14
Fix Space issue w/ IconButton and add around, between & evenly support

lukelooker:
### :sparkles: Changes

- Fix Space issue w/ IconButton 
- Add around, between & evenly support to Space
- Use `Space` for IconButton and Button documentation (POC of fixes)

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/Layout/Space/Space.tsx
+++ b/packages/components/src/Layout/Space/Space.tsx
@@ -25,6 +25,7 @@
  */
 
 import styled, { css } from 'styled-components'
+import { variant } from 'styled-system'
 import { SpacingSizes, flexbox, FlexboxProps } from '@looker/design-tokens'
 import { simpleLayoutCSS, SimpleLayoutProps } from '../utils/simple'
 
@@ -66,6 +67,14 @@ export interface SpaceHelperProps extends SimpleLayoutProps, FlexboxProps {
   reverse?: boolean
 }
 
+interface SpaceProps extends SpaceHelperProps {
+  /**
+   * Align items vertically within `Space`
+   * @default 'center'
+   */
+  verticalAlign?: 'start' | 'center' | 'end'
+}
+
 export const defaultSpaceSize = 'medium'
 
 export const spaceCSS = css`
@@ -75,7 +84,7 @@ export const spaceCSS = css`
   display: flex;
 `
 
-const fauxFlexGap = (space: SpaceHelperProps) => css`
+const fauxFlexGap = (space: SpaceProps) => css`
   && > * {
     margin-left: ${({ theme }) => theme.space[space.gap || defaultSpaceSize]};
   }
@@ -86,8 +95,25 @@ const fauxFlexGap = (space: SpaceHelperProps) => css`
       : `&& > *:first-child { margin-left: ${theme.space.none}; }`}
 `
 
-export const Space = styled.div<SpaceHelperProps>`
+/* eslint-disable sort-keys-fix/sort-keys-fix */
+const verticalAlign = variant({
+  prop: 'verticalAlign',
+  variants: {
+    start: {
+      alignItems: 'flex-start',
+    },
+    center: {
+      alignItems: 'center',
+    },
+    end: {
+      alignItems: 'flex-end',
+    },
+  },
+})
+
+export const Space = styled.div<SpaceProps>`
   ${spaceCSS}
+  ${verticalAlign}
   flex-direction: ${({ reverse }) => (reverse ? 'row-reverse' : 'row')};
 
   ${({ around }) => around && 'justify-content: space-around;'}
@@ -97,4 +123,4 @@ export const Space = styled.div<SpaceHelperProps>`
     !around && !between && !evenly && fauxFlexGap}
 `
 
-Space.defaultProps = { alignItems: 'center', width: '100%' }
+Space.defaultProps = { alignItems: 'center' }
--- a/packages/components/src/Layout/Space/Space.tsx
+++ b/packages/components/src/Layout/Space/Space.tsx
@@ -34,7 +34,7 @@ export interface SpaceHelperProps extends SimpleLayoutProps, FlexboxProps {
    * Amount of space between grid cells
    * @default 'medium'
    */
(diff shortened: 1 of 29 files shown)


packages/www/src/documentation/components/actions/button.mdx
```diff
@@ -36,11 +38,13 @@ Use an outline button alongside a primary button to provide alternative actions
 #### Outline states
 
 ```jsx
-<ButtonOutline m="xsmall">Outline Idle</ButtonOutline>
-<ButtonOutline m="xsmall" className="hover">Outline Hover</ButtonOutline>
-<ButtonOutline m="xsmall" className="active">Outline Active</ButtonOutline>
-<ButtonOutline m="xsmall" color="danger">Outline Danger</ButtonOutline>
-<ButtonOutline disabled>Outline Disabled</ButtonOutline>
+<Space wrap>
```
jhardy, line 41: `wrap` seems to not be working. 
image

Thoughts on `wrap` being the default? I feel like its more useful to auto wrap and then specify `noWrap` or `wrap={false}`

packages/components/src/Layout/Space/Space.tsx
```diff
@@ -25,15 +25,39 @@
...
 import { simpleLayoutCSS, SimpleLayoutProps } from '../utils/simple'
 
-export interface SpaceHelperProps extends SimpleLayoutProps {
+export interface SpaceHelperProps extends SimpleLayoutProps, FlexboxProps {
   /**
    * Amount of space between grid cells
    * @default 'medium'
    */
-  gap?: SpacingSizes
+  gap?: SpacingSizes | 'between' | 'evenly'
```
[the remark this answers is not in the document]
lukelooker: Doh!

packages/components/src/Layout/Space/Space.tsx
```diff
@@ -46,23 +70,32 @@ export const defaultSpaceSize = 'medium'
 
 export const spaceCSS = css`
   ${simpleLayoutCSS}
+  ${flexbox}
 
   display: flex;
-  align-items: flex-start;
+`
+
+const fauxFlexGap = (space: SpaceHelperProps) => css`
```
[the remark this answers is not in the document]
lukelooker: 🎉  That's awesome! This definitely is an awkward workaround.

packages/components/src/Layout/Space/Space.tsx
```diff
@@ -46,23 +70,32 @@ export const defaultSpaceSize = 'medium'
...
-      : `&& > :first-child { margin-left: ${theme.space.none}; }`}
+  ${({ around }) => around && 'justify-content: space-around;'}
+  ${({ between }) => between && 'justify-content: space-between;'}
+  ${({ evenly }) => evenly && 'justify-content: space-evenly;'}
+  ${({ around, between, evenly }) =>
+    !around && !between && !evenly && fauxFlexGap}
 `
 
-Space.defaultProps = { width: '100%' }
+Space.defaultProps = { alignItems: 'center', width: '100%' }
```
jhardy, line 101: I am leaning towards saying we should default to `flex-start` here and that is because I think `Space`'s primary purpose is layout. In some cases like aligning text and ui elements you want to align them center. But when I think about layout for a full page and content I generally want them to align at the top. Here is a small image of what I am trying to express :D 

image

packages/components/src/Layout/Space/SpaceVertical.tsx
```diff
@@ -34,6 +34,7 @@ export const SpaceVertical = styled.div<SpaceHelperProps>`
   flex-shrink: 1;
 
   && > * {
+    width: 100%;
```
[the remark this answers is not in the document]
lukelooker: This is a really good point. I think I was depending on it in earlier versions because the you couldn't specify `width` explicitly on `Space` when you needed it but that's since been resolve so it seems like we should leave that decision to the developer.

packages/www/src/documentation/components/actions/button.mdx
lukelooker: I actually remove `wrap` for now. I think we might want a more specific component for cases where you want to wrap around (I'm also thinking we could do some really clever stuff around truncation if we _know_ we're dealing with overflow cases)

packages/components/src/Layout/Space/SpaceVertical.tsx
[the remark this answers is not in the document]
lukelooker: Removed :) Added a `stretch` prop that may be useful for some cases.

packages/components/src/Layout/Space/Space.tsx
lukelooker: I added some more explicit `align` / `verticalAlign` props.

Within our docs `SpaceVertical center` is actually the most desirable version in all cases put perhaps our use cases are weird?

packages/www/src/documentation/components/actions/button.mdx
jhardy: Should we remove `wrap` from the examples here?

packages/components/src/Layout/Space/Space.tsx
```diff
@@ -25,44 +25,104 @@
...
 import { simpleLayoutCSS, SimpleLayoutProps } from '../utils/simple'
 
-export interface SpaceHelperProps extends SimpleLayoutProps {
+export interface SpaceHelperProps extends SimpleLayoutProps, FlexboxProps {
   /**
    * Amount of space between grid cells
    * @default 'medium'
    */
-  gap?: SpacingSizes
+  gap?: SpacingSizes | 'between' | 'evenly'
```
jhardy, line 37: `around` still missing :)

lukelooker: Doh! These shouldn't have been here in the first place. Fixed!

packages/www/src/documentation/components/actions/button.mdx
lukelooker: Yes, done!

PR #922, opened 2020-05-15
Support wrapping in ButtonGroup/Toggle

mdodgelooker:
### :sparkles: Changes

- Minor changes to css in `ButtonGroup`, `ButtonToggle` and `ButtonSet` (the common component they share) to allow wrapping.
- To see it in action, go to http://localhost:8000/components/actions/button-group/ and copy/paste the last item a bunch of times.

- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots
image

### DIFF
--- a/packages/components/src/Banner/__snapshots__/Banner.test.tsx.snap
+++ b/packages/components/src/Banner/__snapshots__/Banner.test.tsx.snap
@@ -67,31 +67,6 @@ exports[`Banner can be dismissed 1`] = `
   opacity: 0.25;
 }
 
-.c3 {
-  -webkit-flex: auto;
-  -ms-flex: auto;
-  flex: auto;
-}
-
-.c0 {
-  border-radius: 0.25rem;
-  background-color: #e0f2ff;
-  width: 100%;
-  padding-left: 0.75rem;
-  padding-right: 0.75rem;
-  padding-top: 0.5rem;
-  padding-bottom: 0.5rem;
-  font-size: 0.875rem;
-  -webkit-align-items: center;
-  -webkit-box-align: center;
-  -ms-flex-align: center;
-  align-items: center;
-  display: -webkit-box;
-  display: -webkit-flex;
-  display: -ms-flexbox;
-  display: flex;
-}
-
 .c5 {
   background: transparent;
   border: 1px solid transparent;
@@ -123,6 +98,31 @@ exports[`Banner can be dismissed 1`] = `
 .c5[disabled]:active,
 .c5[disabled]:focus {
   color: #939BA5;
+}
+
+.c3 {
+  -webkit-flex: auto;
+  -ms-flex: auto;
+  flex: auto;
+}
+
+.c0 {
+  border-radius: 0.25rem;
+  background-color: #e0f2ff;
+  width: 100%;
+  padding-left: 0.75rem;
+  padding-right: 0.75rem;
+  padding-top: 0.5rem;
+  padding-bottom: 0.5rem;
+  font-size: 0.875rem;
+  -webkit-align-items: center;
+  -webkit-box-align: center;
+  -ms-flex-align: center;
+  align-items: center;
+  display: -webkit-box;
+  display: -webkit-flex;
+  display: -ms-flexbox;
+  display: flex;
 }
 
 .c2 {
--- a/packages/components/src/Button/ButtonGroup.tsx
+++ b/packages/components/src/Button/ButtonGroup.tsx
@@ -24,28 +24,40 @@
 
  */
 
-import uniqueId from 'lodash/uniqueId'
 import xor from 'lodash/xor'
 import React, { ChangeEvent, forwardRef, Ref } from 'react'
 import styled from 'styled-components'
-import { useControlWarn } from '../utils'
+import { useControlWarn, useID } from '../utils'
 import { ButtonItem, ButtonItemLabel } from './ButtonItem'
-import { ButtonGroupOrToggleProps, ButtonSet } from './ButtonSet'
+import { ButtonGroupOrToggleBaseProps, ButtonSet } from './ButtonSet'
+
+export interface ButtonGroupProps
+  extends Omit<ButtonGroupOrToggleBaseProps<string[]>, 'nullable'> {
+  /**
+   * Change callback for controlling the component
+   */
+  onChange?: (value: string[]) => void
+}
 
 const ButtonGroupFactory = forwardRef(
   (
-    { onChange, value, ...props }: ButtonGroupOrToggleProps<string[]>,
+    { onChange, value, ...props }: ButtonGroupProps,
     ref: Ref<HTMLDivElement>
   ) => {
+    const name = useID(props.name)
+
     const isControlled = useControlWarn({
       controllingProps: ['onChange', 'value'],
       isControlledCheck: () => onChange !== undefined,
       name: 'ButtonGroup',
     })
 
-    function handleChange(e: ChangeEvent<HTMLInputElement>) {
-      const newValue = xor(value, [e.target.value])
-      onChange && onChange(newValue)
+    function handleChange(e?: ChangeEvent<HTMLInputElement>) {
+      // event is only ever missing for nullable ButtonToggle
+      if (e) {
+        const newValue = xor(value, [e.target.value])
+        onChange && onChange(newValue)
+      }
     }
 
     return (
@@ -57,29 +69,21 @@ const ButtonGroupFactory = forwardRef(
               onChange: handleChange,
               value,
             }
-          : { name: props.name || uniqueId() })}
+          : { name })}
       />
     )
   }
 )
 
 export const ButtonGroup = styled(ButtonGroupFactory)`
-  margin: ${({
-    theme: {
-      space: { xxsmall },
-    },
-  }) => `0 -${xxsmall} -${xxsmall} 0`};
+  --spacing: ${({ theme }) => theme.space.xxsmall};
+  margin-bottom: calc(var(--spacing) * -1);
+  margin-right: calc(var(--spacing) * -1);
   ${ButtonItem} {
     height: 36px;
     border-radius: ${({ theme }) => theme.radii.medium};
-    margin: ${({
-      theme: {
-        space: { xxsmall },
-      },
-    }) => `0 ${xxsmall} ${xxsmall} 0`};
-    &:last-of-type {
-      margin-right: 0;
-    }
+    margin-bottom: var(--spacing);
+    margin-right: var(--spacing);
   }
 
   ${ButtonItemLabel} {
--- a/packages/components/src/Button/ButtonGroup.tsx
+++ b/packages/components/src/Button/ButtonGroup.tsx
@@ -31,17 +31,13 @@ import { useControlWarn, useID } from '../utils'
 import { ButtonItem, ButtonItemLabel } from './ButtonItem'
 import { ButtonGroupOrToggleBaseProps, ButtonSet } from './ButtonSet'
 
-export interface ButtonGroupProps
-  extends Omit<ButtonGroupOrToggleBaseProps<string[]>, 'nullable'> {
-  /**
-   * Change callback for controlling the component
-   */
-  onChange?: (value: string[]) => void
-}
-
 const ButtonGroupLayout = forwardRef(
   (
-    { onChange, value, ...props }: ButtonGroupProps,
+    {
+      onChange,
+      value,
+      ...props
+    }: Omit<ButtonGroupOrToggleBaseProps<string[]>, 'nullable'>,
     ref: Ref<HTMLDivElement>
   ) => {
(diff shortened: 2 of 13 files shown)


packages/components/src/Banner/__snapshots__/Banner.test.tsx.snap
```diff
@@ -67,31 +67,6 @@ exports[`Banner can be dismissed 1`] = `
...
-  -webkit-align-items: center;
-  -webkit-box-align: center;
-  -ms-flex-align: center;
-  align-items: center;
-  display: -webkit-box;
-  display: -webkit-flex;
-  display: -ms-flexbox;
-  display: flex;
-}
-
```
mdodgelooker, line 94: Why did this move position? No idea. :-/

packages/components/src/Button/ButtonGroup.tsx
```diff
@@ -66,12 +66,17 @@ const ButtonGroupFactory = forwardRef(
 export const ButtonGroup = styled<FC<ButtonGroupOrToggleProps<string[]>>>(
   ButtonGroupFactory
 )`
+  margin-bottom: -${(props) => props.theme.space.xxsmall};
```
mdodgelooker, line 69: Negative bottom margin on the container offsets the individual bottom margin on the last row of buttons (wrapping or not).

packages/components/src/Button/ButtonSet.tsx
```diff
@@ -125,18 +128,22 @@ export const ButtonSet = forwardRef(
...
-        alignItems="center"
-        display="inline-flex"
-        textAlign="center"
-        fontSize="small"
-        ref={ref}
-        {...props}
-      >
+      <div ref={ref} {...props}>
         {clonedChildren}
-      </Box>
```
mdodgelooker, line 137: Took the opportunity to get rid of `Box` which is an ongoing goal for the package.

packages/components/src/Button/ButtonToggle.tsx
```diff
@@ -78,32 +77,37 @@ const ButtonToggleFactory = forwardRef(
 export const ButtonToggle = styled<FC<ButtonGroupOrToggleProps<string>>>(
   ButtonToggleFactory
 )`
-  border: solid 1px ${(props) => props.theme.colors.palette.charcoal200};
+  border-radius: 4px;
```
mdodgelooker, line 80: Moved `border-radius` here instead of in the JSX (where it was bleeding into the dom, causing a React warning). Also moved the border itself down to the `ButtonItemLabel`.

packages/components/src/Button/ButtonGroup.tsx
lukelooker: FYI - I'm really leery of adding negative margins like this. Is it possible to get the parent to be smarter perhaps?

packages/components/src/Button/ButtonSet.tsx
```diff
@@ -34,13 +34,16 @@ import React, {
...
   Ref,
 } from 'react'
+import styled from 'styled-components'
 import {
   CompatibleHTMLProps,
+  border,
   BorderProps,
+  position,
   PositionProps,
+  space,
```
lukelooker, line 44: `Layout/utils/simple.ts` offers `SimpleLayoutProps` that might fit this well.

packages/components/src/Button/ButtonSet.tsx
```diff
@@ -78,7 +81,7 @@ export type ButtonSetType<
   T extends string | string[] = string[]
 > = ForwardRefExoticComponent<ButtonSetProps<T> & { ref: Ref<HTMLDivElement> }>
 
-export const ButtonSet = forwardRef(
+export const ButtonSetInternal = forwardRef(
```
lukelooker, line 84:
```suggestion
export const ButtonSeLayout = forwardRef(
```

`Layout` seems to be winning the war against `Internal` within our code base as-of late (I prefer it since it's more about the the DOM "layout" rather than some secret internal interface. :)

packages/components/src/Button/ButtonToggle.tsx
lukelooker:
```suggestion
  border-radius: ${({ theme }) => theme.radii.medium };
```

packages/components/src/Button/ButtonGroup.tsx
mdodgelooker: I know, it's gross. I couldn't find anything else though and this was on MDN https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Mastering_Wrapping_of_Flex_Items

Apparently Firefox supports `row-gap` for flex layouts but no other browser: https://developer.mozilla.org/en-US/docs/Web/CSS/row-gap#Browser_compatibility

mdodgelooker: And now I'm realizing this would need to be applied to the right margin b/c the `&:last-of-type` would only work if there's just 1 row.

mdodgelooker: Actually, let me see if I could use `grid` here instead of `flex`...

mdodgelooker: Whoops, `grid` doesn't work. Here's a css-variable based example, perhaps slightly less brittle: https://medium.com/@schofeld/mind-the-flex-gap-c9cd1b4b35d8

mdodgelooker: @lukelooker I wasn't able to come up with anything better than negative margin but with a css var to make it less brittle. Feel free to replace it with something better if you can find it. :)

packages/components/src/Button/ButtonGroup.tsx
```diff
@@ -24,28 +24,40 @@
...
-import { ButtonGroupOrToggleProps, ButtonSet } from './ButtonSet'
+import { ButtonGroupOrToggleBaseProps, ButtonSet } from './ButtonSet'
+
+export interface ButtonGroupProps
+  extends Omit<ButtonGroupOrToggleBaseProps<string[]>, 'nullable'> {
+  /**
+   * Change callback for controlling the component
+   */
+  onChange?: (value: string[]) => void
+}
```
mdodgelooker, line 40: Shuffled the interfaces a bit to support `nullable` on `ButtonToggle` and the underlying `ButtonSet` but not on `ButtonGroup` since it's nullable by nature.

packages/components/src/Button/ButtonGroup.tsx
```diff
@@ -24,28 +24,40 @@
...
 
 const ButtonGroupFactory = forwardRef(
   (
-    { onChange, value, ...props }: ButtonGroupOrToggleProps<string[]>,
+    { onChange, value, ...props }: ButtonGroupProps,
     ref: Ref<HTMLDivElement>
   ) => {
+    const name = useID(props.name)
```
mdodgelooker, line 47: If we're generating the name, gotta keep it consistent across renders.

packages/components/src/Button/ButtonGroup.tsx
```diff
@@ -24,28 +24,40 @@
...
       isControlledCheck: () => onChange !== undefined,
       name: 'ButtonGroup',
     })
 
-    function handleChange(e: ChangeEvent<HTMLInputElement>) {
-      const newValue = xor(value, [e.target.value])
-      onChange && onChange(newValue)
+    function handleChange(e?: ChangeEvent<HTMLInputElement>) {
+      // event is only ever missing for nullable ButtonToggle
+      if (e) {
```
mdodgelooker, line 57: Have to check for `e` since the `ButtonSet` interface now has it as optional.

packages/components/src/Button/ButtonGroup.tsx
```diff
@@ -57,22 +69,21 @@ const ButtonGroupFactory = forwardRef(
...
               value,
             }
-          : { name: props.name || uniqueId() })}
+          : { name })}
       />
     )
   }
 )
 
-export const ButtonGroup = styled<FC<ButtonGroupOrToggleProps<string[]>>>(
```
mdodgelooker, line 66: Unnecessary type arguments. 😜

packages/components/src/Button/ButtonItem.tsx
```diff
@@ -52,9 +52,10 @@ export const ButtonItemLabel = styled.label<ButtonItemProps>`
   cursor: pointer;
   display: inline-flex;
   align-items: center;
-  padding: 0 ${(props) => props.theme.space.small};
+  text-align: center;
```
mdodgelooker, line 55: When `ButtonToggle` wraps, we slightly stretch the `ButtonItem`s in complete rows to take up any leftover space, so now we need to center-align the text.

packages/components/src/Button/ButtonSet.tsx
```diff
@@ -62,29 +51,32 @@ export interface ButtonGroupOrToggleProps<
...
 
-interface ButtonSetProps<ValueType extends string | string[] = string[]>
-  extends Omit<ButtonGroupOrToggleProps<ValueType>, 'onChange'> {
-  onChange?: (e: ChangeEvent<HTMLInputElement>) => void
-  isToggle?: boolean
+export interface ButtonGroupOrToggleBaseProps<
+  ValueType extends string | string[] = string[]
+> extends Omit<ButtonSetProps<ValueType>, 'isToggle' | 'onChange'> {
+  onChange?: (value: ValueType) => void
 }
```
mdodgelooker, line 66: Attempted to make the interfaces slightly less confusing, while adding `nullable`.

packages/components/src/Button/ButtonToggle.test.tsx
```diff
@@ -24,55 +24,75 @@
...
-  Simulate.change(apples)
-  Simulate.change(bananas)
-  Simulate.change(oranges)
-  expect(onChangeMock).toHaveBeenCalledTimes(3)
-  expect(onChangeMock.mock.calls[0][0]).toEqual('Apples')
-  expect(onChangeMock.mock.calls[1][0]).toEqual('Bananas')
-  expect(onChangeMock.mock.calls[2][0]).toEqual('Oranges')
+    fireEvent.click(orangesInput)
+    expect(orangesInput).not.toBeChecked()
+  })
```
mdodgelooker, line 97: Only this bit is new, the rest is basically whitespace diff.

packages/components/src/Button/ButtonToggle.tsx
```diff
@@ -24,24 +24,30 @@
...
     {
       onChange,
       value: controlledValue,
       ...props
-    }: ButtonGroupOrToggleProps<string>,
+    }: ButtonGroupOrToggleBaseProps<string>,
     ref: Ref<HTMLDivElement>
   ) => {
+    //
+    const name = useID(props.name)
```
mdodgelooker, line 49: Keep a generated `name` consistent across renders.

packages/components/src/Button/ButtonToggle.tsx
```diff
@@ -50,63 +56,79 @@ const ButtonToggleFactory = forwardRef(
 
     const [value, setValue] = useState<string>()
 
-    function handleChange(e: ChangeEvent<HTMLInputElement>) {
+    function handleChange(e?: ChangeEvent<HTMLInputElement>) {
+      const newValue = e ? e.target.value : ''
```
mdodgelooker, line 60: If `e` is missing, it indicates the de-selection, which we have to use `onClick` on the `label` for.

packages/components/src/Button/ButtonToggle.tsx
```diff
@@ -50,63 +56,79 @@ const ButtonToggleFactory = forwardRef(
...
+      display: block;
+      background: ${(props) => props.theme.colors.palette.charcoal200};
+      position: absolute;
+      z-index: 1;
+    }
+    &::before {
+      height: 1px;
+      width: 100%;
+      left: 0;
+      bottom: -1px;
```
mdodgelooker, line 125: Adding this to get the border between rows if there's wrapping.

mdodgelooker:
@jhardy and @lukelooker I've updated the wrapping styling of `ButtonToggle` and also the `nullable` prop – ready for the next round of review!

packages/components/src/Button/ButtonGroup.tsx
lukelooker: Ok - I banged on this for longer than I want to admit and couldn't come up with a better solution - weird problem... Perhaps the future flex-gap feature will help with this.

I'm leery of the CSS variable because it can "leak" out of the styled-component – I think there are is an argument that you could use that as a feature but in this case I don't think it offers much benefit given the potential risk.

I did slightly simplify the implementation and added a `xxxsmall` size to negate the need for the `calc` behaviors (seemed funny that we didn't already have the size for these kinds of scenarios).

mdodgelooker:
@lukelooker the TS error should be fixed now and ready for one more approval.

PR #927, opened 2020-05-18
Tree and TreeItem

ElliotTheHuman:
### 🛠️ To-Do

- Add documentation
- Try and replace margin usage in `Tree` with padding on the content container instead (as per Jared's request)

### :sparkles: Changes

- Added `Tree` and `TreeItem` components

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC (excluding header comments)


packages/components/src/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+  className?: string
+  detail?: ReactNode
+  detailStopPropagation?: boolean
+  /**
+   * Gap size of the internal Space component
+   */
+  gapSize?: SpacingSizes
+  icon?: IconNames
+  iconSize?: SpacingSizes
+  onClick?: () => void
```
lukelooker, line 44: Not needed for this PR but we should work to fill out JSDocs for each externally-exposed property. Should include `@default` if one exists.

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { Icon, IconNames } from '../Icon'
+
+export interface TreeItemProps {
+  children: ReactNode
+  className?: string
+  detail?: ReactNode
+  detailStopPropagation?: boolean
+  /**
+   * Gap size of the internal Space component
+   */
```
lukelooker, line 40:
```suggestion
  /**
   * Gap size of the internal Space component
   * @default 'xsmall'
   */
```

packages/components/src/Tree/TreeItem.tsx
ElliotTheHuman: Will include JSDocs updates with updated test suite in separate PR.

PR #984, opened 2020-05-29
Remove SemanticColors, introduce new Theme.colors shape

lukelooker:
### :sparkles: Changes

- Removes semanticColors and all dependencies on it.
- Implement new theme.colors shape
  - Deprecate `danger` (now known as `critical`)
  - Deprecate `primary` (now known as `key`)
  - Provides a hard-coded version of the theme object to provide internal consistency
- Leverage new theme shape to replace all existing uses of `semanticColor*`
- Fix bug where `transparent` wasn't actually being derived in ButtonTransparent:disabled
  - Update snapshots for above fix and minor class-name recalculations

- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### :camera: Screenshots

image


packages/design-tokens/src/system/color/utils.ts
```diff
@@ -24,9 +24,19 @@
...
+// export const generateDerivedColors = (key: string, background: string) => {
+//   const derivedSwatches = {
+//     actionText: background,
+//     surface: tint(0.05, background),
+//     primaryInteractive: tintOrShadeUiColor(10, key),
+//     primaryPressed: tintOrShadeUiColor(18, key),
+//   }
+
+//   return derivedSwatches
+// }
```
dbchristopher, line 42: want to delete this file altogether?

packages/design-tokens/src/tokens/color/colors.ts
```diff
@@ -0,0 +1,80 @@
...
+  background: white,
+  text: charcoal800,
+  key: purple400,
+}
+
+const generateSurfaceColors = (coreColors: CoreColors): SurfaceColors => {
+  return {
+    field: coreColors.background,
+    inverse: coreColors.text,
+    onInverse: coreColors.background,
```
dbchristopher, line 57: maybe this should just be called `inverseText`? whenever I see `on...` I assume it's associated with a specific event or action

packages/design-tokens/src/tokens/color/colors.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+}
+
+const defaultIntentColors: IntentColors = {
+  link: blue600,
+  critical: red500,
+  warn: yellow500,
+  neutral: charcoal400,
+  positive: green500,
+  inform: blue500,
```
dbchristopher, line 67: will this be too close to link color? not sure if the end result will be text that looks clickable but isn't

packages/design-tokens/src/tokens/color/colors.ts
```diff
@@ -0,0 +1,80 @@
...
+
+/* eslint-disable sort-keys-fix/sort-keys-fix */
+
+const defaultCoreColors: CoreColors = {
+  background: white,
+  text: charcoal800,
+  key: purple400,
+}
+
+const generateSurfaceColors = (coreColors: CoreColors): SurfaceColors => {
```
dbchristopher, line 53: any reason this needs to be a function?

packages/components/src/Button/Button.tsx
```diff
@@ -28,36 +28,32 @@ import styled from 'styled-components'
...
+  color: ${({ theme, color = 'key' }) => theme.colors[`${color}Text`]};
 
   &:active,
   &.active {
-    background: ${({ theme, color = 'primary' }) =>
-      theme.colors.semanticColors[color].darker};
-    border-color: ${({ theme, color = 'primary' }) =>
-      theme.colors.semanticColors[color].darker};
+    background: ${({ theme, color = 'key' }) =>
+      theme.colors[`${color}Pressed`]};
```
dbchristopher, line 38: can these values be nested instead of doing string interpolation to find the key? I feel like `theme.colors[color].pressed` is a better api than `theme.colors[`${color}Pressed`]`

packages/design-tokens/src/system/color/stateful.ts
```diff
@@ -0,0 +1,108 @@
...
+
+export type StatefulColor = 'key' | 'critical' | 'neutral'
+
+export interface StatefulColorChoices {
+  subtle: string
+  accent: string
+  border: string
+  main: string
+  interactive: string
+  pressed: string
```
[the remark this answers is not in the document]
lukelooker: I can't totally disagree but Jared and discussed this one a lot and we went with `pressed` because there are cases where this get used but doesn't directly map to the `:active` pseudo state.

packages/design-tokens/src/system/color/stateful.ts
```diff
@@ -0,0 +1,108 @@
...
+  neutralInteractive: string
+  neutralPressed: string
+  neutralText: string
+  neutralTextAlt: string
+}
+
+interface KeyStatefulColors {
+  /**
+   * TBD
+   * Used for TBD
```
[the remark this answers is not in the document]
lukelooker: Yep. I need design's input on these as well :)

packages/design-tokens/src/system/color/utils.ts
lukelooker: Done :)

packages/design-tokens/src/tokens/color/colors.ts
lukelooker: This is just a reflection of our current situation. It actually works out in-practice as `link` is used in very narrow cases.

packages/design-tokens/src/tokens/color/colors.ts
lukelooker: Agreed. Changed :)

packages/design-tokens/src/tokens/color/colors.ts
lukelooker: Yes, the `surfaceColors` generally should be directly specified (the goal is that all the end-consumer will need to specify is the `CoreColors` so eventually all of the theme will be generated/derived like this.

This one was straightforward so I just went ahead with it :)

packages/components/src/Button/Button.tsx
lukelooker: The nesting actually causes a bunch of different issues and one of the goals with this new shape was to flatten to something like this. 

I have will eventually have a function that will parse the theme.colors object and return an object with the `StatefulColorChoices` shape but doing so will require a fairly complex refactor of `Button*` so in the interest of getting this first pass done I went with this simpler approach for now.

packages/components/src/Button/Button.tsx
```diff
@@ -28,36 +28,32 @@ import styled from 'styled-components'
...
   &:focus,
   &.hover {
-    background: ${({ theme, color = 'primary' }) =>
-      theme.colors.semanticColors[color].dark};
-    border-color: ${({ theme, color = 'primary' }) =>
-      theme.colors.semanticColors[color].dark};
+    background: ${({ theme, color = 'key' }) =>
+      theme.colors[`${color}Interactive`]};
+    border-color: ${({ theme, color = 'key' }) =>
+      theme.colors[`${color}Interactive`]};
```
mdodgelooker, line 48: Totally unrelated to this PR but while I was checking it in gatsby I noticed that the `:hover` style takes priority over the `:active` style since it's below – usually you're doing both if you're using a mouse. I'll make a separate ticket.

packages/design-tokens/src/system/color/blends.ts
```diff
@@ -0,0 +1,93 @@
...
+  text3: string
+  /**
+   * Reduced emphasis text
+   * Used for: Help text, meta information
+   * @default charcoal500
+   */
+  text4: string
+  /**
+   * TODO - Should be defined
+   * Used for: TBD
```
jhardy, line 89: disabled text, placeholder text

PR #985, opened 2020-05-29
Status component and add icon Error

laraharrow:
### :sparkles: Changes

- Create new component Status
- Add icon Error

- [x] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### :camera: Screenshots

Screen Shot 2020-05-29 at 9 48 01 AM


packages/components/src/Status/Status.tsx
```diff
@@ -0,0 +1,117 @@
...
+  accessibilityLabel?: string
+  icon?: JSX.Element
+}
+
+const getStatusIntentStyling = (intent: StatusIntent) => {
+  const statusTypeStyling: StatusTypeStyling = {}
+  const iconProps = {
+    mr: 'small',
+    size: 20,
+    style: { flexBasis: '20px', flexShrink: 0 },
```
jhardy:
```suggestion
    size: 24,
```
Status should not have margin on it by default here. I don't think we need the flex sizing here either. Setting the size should be enough

packages/components/src/Status/Status.tsx
```diff
@@ -0,0 +1,117 @@
...
+      statusTypeStyling.accessibilityLabel = 'Neutral'
+      break
+
+    case 'warning':
+      statusTypeStyling.icon = (
+        <Icon {...iconProps} name="Warning" color="palette.yellow500" />
+      )
+      statusTypeStyling.accessibilityLabel = 'Warning'
+      break
+    default:
```
jhardy, line 95: You can collapse neutral and default here
```
case 'neutral':
default:
... neutral styles
```

packages/components/src/Status/Status.tsx
```diff
@@ -0,0 +1,117 @@
...
+    TypographyProps {
+  /**
+   * @default: 'inform'
+   */
+  intent?: StatusIntent
+}
+
+interface StatusTypeStyling {
+  accessibilityLabel?: string
+  icon?: JSX.Element
```
jhardy, line 52: I this should be `IconNames` instead of a JSX element. That way we ensure we have a correct icons value here.

packages/components/src/Status/Status.tsx
jhardy: I would try returning an icon name, color and label here instead
```
statusTypeStyling.icon = "Warning"
statusTypeStyling.color = "palette.yellow500"
statusTypeStyling.accessibilityLabel = 'Warning'
```
then down below 
```
{ icon, color, accessibilityLabel } = getStatusIntentStyling(intent)
```

packages/components/src/Status/Status.tsx
```diff
@@ -0,0 +1,117 @@
...
+      <VisuallyHidden>{accessibilityLabel}</VisuallyHidden>
+    </>
+  )
+}
+
+Status.defaultProps = {
+  fontSize: 'small',
+  px: 'small',
+  py: 'xsmall',
+  width: '100%',
```
jhardy: I don't think you need any of these as default props on Status.

PR #1012, opened 2020-06-02
Dynamic theme generation

lukelooker:
### :sparkles: Changes

- Dynamic theme generation
- ComponentsProvider now accepts `coreColors` for dynamic theme specification
  - NOTE: If explicit key color is not specified we'll simply pass through the `fallback` values. This means that existing users of `@looker/components` will see no effective changes when this feature is added unless they use the `coreColors` prop on `ComponentsProvider`

image

- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


.vscode/settings.json
```diff
@@ -41,7 +41,6 @@
     "xxxxlarge"
   ],
   "prettier.arrowParens": "always",
-  "extensions.ignoreRecommendations": true,
```
lukelooker, line 44: Not sure why this got introduced. I missed having these when it appeared :)

dbchristopher:
this is so cool!

PR #1015, opened 2020-06-03
ButtonGroup & ButtonToggle: add options, use buttons

mdodgelooker:
### :sparkles: Changes

**This ended up being pretty much a complete rewrite.**

- Using hidden inputs to dictate the behavior of `ButtonGroup` and `ButtonToggle` was problematic for both accessibility (unable to focus) and CI in the core product
- I replaced the `<label><input>text</label>` structure with `<button>text</button>`
- In doing so, the uncontrolled version no longer worked, nor could I see a reason for supporting it (with inputs, you could use them in an old-school form) so now the components must be controlled
- Added `options` as an alternative to `ButtonItem` children
- Styling polish (especially for the wrapping scenario)

- [x] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### :camera: Screenshots
button-set


packages/components/src/Button/ButtonGroup.tsx
```diff
@@ -25,63 +25,55 @@
...
+import { ButtonItem } from './ButtonItem'
 import { ButtonGroupOrToggleBaseProps, ButtonSet } from './ButtonSet'
 
 const ButtonGroupLayout = forwardRef(
   (
-    {
-      onChange,
-      value,
-      ...props
-    }: Omit<ButtonGroupOrToggleBaseProps<string[]>, 'nullable'>,
```
mdodgelooker, line 40: Moved `nullable` (interface and logic) to where it belongs in `ButtonToggle.tsx` instead of the underlying component `ButtonSet.tsx`.

packages/components/src/Button/ButtonGroup.tsx
```diff
@@ -25,63 +25,55 @@
...
+    { onChange, value = [], ...props }: ButtonGroupOrToggleBaseProps,
     ref: Ref<HTMLDivElement>
   ) => {
-    const name = useID(props.name)
-
-    const isControlled = useControlWarn({
-      controllingProps: ['onChange', 'value'],
-      isControlledCheck: () => onChange !== undefined,
-      name: 'ButtonGroup',
-    })
```
mdodgelooker, line 49: No more uncontrolled version, since there are no longer `input`s involved.

packages/components/src/Button/ButtonGroup.tsx
```diff
@@ -25,63 +25,55 @@
...
 )
 
-export const ButtonGroup = styled(ButtonGroupLayout)`
-  margin: -${({ theme }) => theme.space.xxxsmall};
+ButtonGroupLayout.displayName = 'ButtonGroupLayout'
 
+export const ButtonGroup = styled(ButtonGroupLayout)`
   ${ButtonItem} {
-    border-radius: ${({ theme }) => theme.radii.medium};
-    height: 36px;
```
mdodgelooker, line 79: Moved `height` to `ButtonItem.tsx` since the value is common to both `ButtonGroup` and `ButtonToggle`.

packages/components/src/Button/ButtonGroup.tsx
```diff
@@ -25,63 +25,55 @@
...
-          : { name })}
+        value={value}
+        onItemClick={handleItemClick}
       />
     )
   }
 )
 
-export const ButtonGroup = styled(ButtonGroupLayout)`
-  margin: -${({ theme }) => theme.space.xxxsmall};
```
mdodgelooker, line 75: Moved this ugly negative margin behavior inside `.wrapping` below. Due to some nagging issues in the styles of the wrapping vs non-wrapping `ButtonToggle`, I decided to add the `wrapping` className via a callback ref in `ButtonSet.tsx`. Of course I would rather not have to use that, but at least it allows for less convoluted css.

packages/components/src/Button/ButtonItem.tsx
```diff
@@ -34,61 +34,119 @@ import {
...
+const ButtonLayout = forwardRef(
+  (
+    { children, onClick, value, ...props }: ButtonItemProps,
+    ref: Ref<HTMLButtonElement>
+  ) => {
+    const { value: contextValue, onItemClick } = useContext(ButtonSetContext)
+
+    function handleClick(e: MouseEvent<HTMLButtonElement>) {
+      onClick && onClick(e)
+      if (!e.defaultPrevented) {
```
mdodgelooker, line 55: Allows the parent handler to be cancelled.

packages/components/src/Button/ButtonSet.tsx
```diff
@@ -25,117 +25,120 @@
  */
 
 import React, {
-  cloneElement,
```
mdodgelooker, line 28: My original usage of `cloneElement` here was overly convoluted, so I changed to using `Context` instead.

packages/components/src/Button/ButtonSet.tsx
```diff
@@ -25,117 +25,120 @@
...
+export interface ButtonSetOption {
+  value: string
+  label?: string
+  disabled?: boolean
+}
 
 interface ButtonSetProps<ValueType extends string | string[] = string[]>
   extends SimpleLayoutProps,
-    Omit<CompatibleHTMLProps<HTMLDivElement>, 'value' | 'onChange'> {
+    Omit<CompatibleHTMLProps<HTMLDivElement>, 'value' | 'defaultValue'> {
```
mdodgelooker, line 51: Not sure why divs support `defaultValue` but this component definitely does not, so I wanted to make that clear.

packages/components/src/Button/ButtonSet.tsx
```diff
@@ -25,117 +25,120 @@
...
-  onChange?: (e?: ChangeEvent<HTMLInputElement>) => void
-  isToggle?: boolean
+  options?: ButtonSetOption[]
   /**
-   * Value can be unset by clicking selected button (ButtonToggle only)
+   * Value for controlling the component
    */
-  nullable?: boolean
+  value?: ValueType
+  onItemClick?: (e: MouseEvent<HTMLButtonElement>) => void
```
mdodgelooker, line 64: This is only used internally and is omitted from the external interface below.

packages/components/src/Button/ButtonSet.tsx
```diff
@@ -25,117 +25,120 @@
...
-> extends Omit<ButtonSetProps<ValueType>, 'isToggle' | 'onChange'> {
-  onChange?: (value: ValueType) => void
+> extends Omit<ButtonSetProps<ValueType>, 'onChange' | 'onItemClick'> {
+  onChange?: ButtonSetCallback<ValueType>
 }
 
 export type ButtonSetType<
-  T extends string | string[] = string[]
-> = ForwardRefExoticComponent<ButtonSetProps<T> & { ref: Ref<HTMLDivElement> }>
+  TValue extends string | string[] = string[]
```
mdodgelooker, line 74: Just keeping the value type arg naming consistent.

packages/components/src/Button/ButtonSet.tsx
```diff
@@ -25,117 +25,120 @@
...
+        const rowHeight = firstItem
+          ? firstItem.getBoundingClientRect().height
+          : buttonItemHeight
+        if (height > rowHeight * 2) {
+          setIsWrapping(true)
+        } else {
+          setIsWrapping(false)
+        }
+      }
+    }, [])
```
mdodgelooker, line 117: Couldn't get the css water tight for both wrapping and non-wrapping `ButtonToggle` so I opted for this solution to separate some of the styles.

packages/components/src/Button/ButtonSet.tsx
```diff
@@ -25,117 +25,120 @@
...
       })
-    })
 
     return (
-      <div ref={ref} {...props}>
-        {clonedChildren}
-      </div>
+      <ButtonSetContext.Provider value={context}>
+        <div
+          role="group"
```
mdodgelooker, line 134: This is needed for accessibility now that the items are `button`s instead of `input`s that share a common `name` attribute.

packages/components/src/Button/ButtonGroup.test.tsx
```diff
@@ -25,91 +25,84 @@
...
+        <ButtonGroup value={value} onChange={setValue}>
+          <ButtonItem>Apples</ButtonItem>
+          <ButtonItem>Oranges</ButtonItem>
+          <ButtonItem>Bananas</ButtonItem>
+        </ButtonGroup>
+      )
+    }
+    renderWithTheme(<TestComponent />)
+
+    expect(screen.getByRole('group')).toMatchSnapshot()
```
mdodgelooker, line 49: As much as I don't love snapshots, these components are so heavy with CSS. Thoughts?

packages/components/src/Button/ButtonItem.tsx
```diff
@@ -34,61 +34,114 @@ import {
...
-  transition: background ${(props) => props.theme.transitions.durationQuick}
-    ease;
-
-  &:hover {
-    background: ${(props) =>
-      !props.selected
-        ? `hsla(${parseToHsl(props.theme.colors.key).hue}, 25%, 97%, 0.7)`
-        : false};
+  border: none;
+  background: transparent;
```
mdodgelooker, line 96: Just resetting the default browser button border and background. `ButtonGroup` and `ButtonToggle` have pretty different border and background styles, so I moved most of them to those individual files.

packages/components/src/Button/ButtonItem.tsx
```diff
@@ -34,61 +34,119 @@ import {
...
+    border-color: hsla(
+      ${({ theme }) => parseToHsl(theme.colors.key).hue},
+      100%,
+      98%,
+      1
+    );
+
+    color: ${({ theme }) => theme.colors.key};
+    &[disabled] {
+      color: ${({ theme }) => theme.colors.keyFocus};
```
mdodgelooker, line 148: This does not seem like the right name of the color for selected & disabled, but it looks good to me... @jhardy, thoughts?
image

packages/components/src/Button/ButtonGroup.test.tsx
lukelooker: I think it's a logical place to use one - nice to get the "signal" that something changed. When we have visual snapshots this (probably most snapshots) could likely be replaced with those.

packages/components/src/Button/ButtonItem.tsx
```diff
@@ -34,61 +34,119 @@ import {
...
 
   &:active {
-    background: ${(props) =>
-      `hsla(${parseToHsl(props.theme.colors.key).hue}, 50%, 96%, 0.9)`};
+    background: hsla(
+      ${({ theme }) => parseToHsl(theme.colors.key).hue},
+      50%,
+      96%,
+      0.9
+    );
```
lukelooker, line 118: Could we use `keyAccent` or `keySubtle` here instead? Might need a @jhardy consult.

packages/components/src/Button/ButtonItem.tsx
```diff
@@ -34,61 +34,119 @@ import {
...
+  transition: background ${({ theme }) => theme.transitions.durationQuick} ease;
+  color: ${({ theme }) => theme.colors.text3};
+
+  &[aria-pressed='false']:not([disabled]):hover {
+    background: hsla(
+      ${({ theme }) => parseToHsl(theme.colors.key).hue},
+      25%,
+      97%,
+      0.7
+    );
```
lukelooker, line 109: Could we use `keyAccent` or `keySubtle` here instead? Might need a @jhardy consult.

packages/components/src/Button/ButtonItem.tsx
```diff
@@ -34,61 +34,119 @@ import {
...
+      98%,
+      1
+    );
+
+    border-color: hsla(
+      ${({ theme }) => parseToHsl(theme.colors.key).hue},
+      100%,
+      98%,
+      1
+    );
```
lukelooker, line 144: Few more. If we can't use those slots let's look at "naming" this behavior and moving the calculation functionality to `design-tokens/utils` - I want to consolidate and color math into that package.

packages/components/src/Button/ButtonItem.tsx
lukelooker: Generally `disabled` states are grey but admittedly this case is a little different than some. It's basically `:disabled + :active`

packages/components/src/Button/ButtonSet.tsx
lukelooker: Weird :p

packages/components/src/Button/ButtonSet.tsx
```diff
@@ -25,117 +25,120 @@
...
+    Omit<CompatibleHTMLProps<HTMLDivElement>, 'value' | 'defaultValue'> {
   /**
-   * One or more ButtonItem
+   * One or more ButtonItem (do not use if using options)
    */
-  children: ReactNode
+  children?: ReactNode
   /**
-   * Value for controlling the component
+   * Available options (do not use if using ButtonItem children)
```
lukelooker, line 57: Can we put a warning in if they supply both / neither? I guess there's a chance this might get rendered with no options and then updated?

packages/components/src/Button/ButtonItem.tsx
jhardy: +1 to using keyAccent and keySubtle here.

packages/components/src/Button/ButtonItem.tsx
```diff
@@ -34,61 +34,119 @@ import {
...
+    )
+  }
+)
+
+ButtonLayout.displayName = 'ButtonLayout'
+
+export const buttonItemHeight = 36
 
-export const ButtonItemLabel = styled.label<ButtonItemProps>`
-  cursor: pointer;
```
jhardy, line 52: Can we add `cursor: pointer;` back in? I'd like to keep the cursors consistent with our other buttons

packages/components/src/Button/ButtonSet.tsx
mdodgelooker: I have a warning down further in the file for both, but yeah I could see neither happening briefly if the options are being loaded asynchronously.

mdodgelooker:
@jhardy and @lukelooker I replaced the mathy colors with `keyAccent` and `keySubtle`. There wasn't a good option for the `:active` state (the next darker, `keyFocus` is too dark) so hopefully it's ok without for now?
(I also added a few things I realized were missing in the docs page.)

mdodgelooker:
Oops, snapsots...

PR #841, opened 2020-04-27
Monorepo Build & Development Improvements (Babel FTW!)

lukelooker:
### :sparkles: Changes

This PR switches the monorepo to use Babel (rather than `tsc` / `ttsc` to build the JS artifacts. Additionally it enables for internal use of the packages to no longer require pre-compilation meaning that the playground and Gatsby can show live-edits to any package leveraged within those.

| Task  | Before | After |  |
|-------|-------:|------:|---------|
| `prepare` | 160.5s | 15.1s | 🎉  assuming that `node_modules` is already in normalized state (no calls to NPM req'd) |
| `build` | 57.7s | 51.3s | 🤷‍♂️ |
| `lint:es` | 102.1s | 80.7s | Slower than I hoped due to cache-writing and addition of `*.mdx` files. However no longer requires build before run. This means lint & build can be run concurrently for faster CI runs. | 
| `lint:es` (cache-hit) | n/a |1.5s | 🎉🎉🎉 ESLint leverages caching so that secondary runs only run against files that have changed. |
| `lint:mdx` | 8.8s | n/a | 🎉  In new build system `*.mdx` is linted as part of larger ESLint run |  
| `lint:ts` | 47.5s | 28.6s | 🎉 |
| `test` | 195.6s | 343.4s | Needs further investigation | 

#### Reviewer notes

`packages/playground/*` & `packages/www/*` weren't being fully linted before this PR so there's a lot of noise in this PR to straighten that out (mostly auto-fixed object order issues)

#### Follow-ups

This items will be addressed in follow-up tickets. This list should be checked-off once those tickets have been created.

- [x] Correct & re-enable tests on Confirm & Prompt
- [x] Fix former usage of `moment.js`
- [x] Pre/post-build steps to rearrange source for publishing (remove `/lib` from artifact paths)
- [x] Investigate `jest` slow-down
- [ ] Research why `lint:es - "import/default": "off"` is so problematic now (perhaps it was never _actually_ running)

### :white_check_mark: Requirements

- [x] Includes test coverage for all changes
- [x] Build and tests are passing
- [x] Update documentation
- [x] Updated CHANGELOG
- [x] Checked for i18n impacts
- [x] Checked for a11y impacts
- [ ] PR is ideally < ~400LOC

PR #1027, opened 2020-06-07
Babel-jest FTW

lukelooker:
### :sparkles: Changes

- Jest no longer requires `yarn build` to be run beforehand. Instead, Jest now directly compiles typescript as part of the test process. 
- Switch to using dynamic package paths for `tsconfig` & Jest package resolution
- Rename `test-utils` and `providers` folders to match package names to allow for dynamic resolution
- Travis CI change: Build & test can now be run in parallel since Jest no longer uses build artifacts
- Snapshots changed shape slightly as they’re now benefitting from being generated via the Babel pipeline and the associated `babel-plugin-styled-components` plugin optimizations.

#### Benchmarks

**NOTE:** Before this change the developer was required to run `yarn build` before running `jest` to ensure that the build artifacts were in place and consistent with the source.

| Task  | Before | After |  |
|-------|-------:|------:|---------|
| `test` | ~50s + 33.8s | 34.1s | YMMV depending on cache status |
| `test --no-cache` |  ~50s + 428.9s | 81.1s |  |

Compare to: https://github.com/looker-open-source/components/pull/841

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


packages/components/src/Button/__snapshots__/Button.test.tsx.snap
```diff
@@ -76,7 +76,7 @@ exports[`Button Focus: renders outline when tabbing into focus, but not when cli
 
 exports[`Button Focus: renders outline when tabbing into focus, but not when clicking 2`] = `
 <button
-  class="sc-bxivhb keNUCd sc-ifAKCX sc-EHOje dGzJFz"
+  class="ButtonBase__ButtonOuter-sc-1bpio6j-0 dAphVM ButtonBase-sc-1bpio6j-1 Button-sc-18euc9m-0 gkFqSX"
```
mdodgelooker, line 79: This snapshot and the one below for `ButtonOutline` are interesting. It seems like the 2nd snapshot in a given test doesn't get the `jest-styled-components`-ized output. Unrelated to this PR though – a question for another time.

PR #1029, opened 2020-06-08
artwork is now available as prop for Icon

laraharrow:
### :sparkles: Changes

- Icon accepts artwork as props.

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### :camera: Screenshots

Screen Shot 2020-06-08 at 12 15 19 PM

### DIFF
--- a/packages/components/src/Icon/Icon.tsx
+++ b/packages/components/src/Icon/Icon.tsx
@@ -33,10 +33,11 @@ import {
   SizeMedium,
   SizeLarge,
 } from '@looker/design-tokens'
-import React, { forwardRef, Ref } from 'react'
+import React, { forwardRef, Ref, ReactNode } from 'react'
 import styled from 'styled-components'
 /* eslint import/namespace: ['error', { allowComputed: true }] */
 import { Glyphs, IconNames } from '@looker/icons'
+import omit from 'lodash/omit'
 import { simpleLayoutCSS, SimpleLayoutProps } from '../Layout/utils/simple'
 
 export type IconSize =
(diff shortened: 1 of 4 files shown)


playground/src/index.tsx
```diff
@@ -26,13 +26,42 @@
...
+                fill="#1C2125"
+              />
+            </svg>
+          }
+          name="GearOutline"
+          size="large"
+        />
+        <Icon name="GearOutline" size="xxsmall" />
+        <Icon size="xxsmall" />
+      </Flex>
```
lukelooker: Move this example to storybook and restore this file as it was. Playground shouldn't really be used anymore ;)

playground/src/index.tsx
```diff
@@ -26,13 +26,70 @@
...
+              <path
+                d="M50,59c0,0,4,3,10,0M56,66l2,12l-2,12M25,50c0,0,10,12,23,12c13,0,24,0,35-15"
+                fill="none"
+                stroke="#000"
+                strokeWidth="0.5"
+              />
+            </svg>
+          }
+          size="large"
+        />
```
lukelooker, line 72: I added this example to show a failure case that is fairly likely - if the artwork doesn't have a specified height & width (which is common) it won't show up.

I think we need to add a CSS selector to the `Icon.tsx` to do something like:

```
svg { 
  height: 100%;
  width: 100%;
}
```

packages/www/src/documentation/components/content/icon.mdx
```diff
@@ -19,6 +19,30 @@ import { IconList } from './demos'
...
+### artwork
+
+`artwork` alows for extarnal patterns that is not already available in our Icon list. Please pass ether `name` or `artwork` not both but yes at least one of them.
+
+```jsx
+<Icon
+  artwork={
+    <svg
+      width="24"
+      height="24"
```
lukelooker, line 31: Let's show an example that doesn't have height & width specified.

packages/components/src/Icon/Icon.tsx
```diff
@@ -50,19 +50,31 @@ export type IconSize =
...
-    const Glyph = Glyphs[name]
-
+  (
+    { artwork = undefined, name, ...props }: IconProps,
+    ref: Ref<HTMLDivElement>
+  ) => {
+    if ((artwork && name) || (!artwork && !name)) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        "Please pass ether name or artwork as Icon prop. If both are passed the default will be artwork. If non is passed Icon won't be displayed"
```
lukelooker, line 68:
```suggestion
        "You may only specify name OR artwork, not both."
```

packages/components/src/Icon/Icon.tsx
```diff
@@ -50,19 +50,31 @@ export type IconSize =
 export interface IconProps
   extends Omit<CompatibleHTMLProps<HTMLDivElement>, 'onClick'>,
     SimpleLayoutProps {
+  artwork?: ReactNode
```
lukelooker, line 53: Can you annotate (JSDoc) what these are so folks can understand how to use them?

playground/src/index.tsx
laraharrow: I'm not sure I know how to use storybook🤦🏽‍♀️

packages/components/src/Icon/Icon.tsx
laraharrow: not sure I understand what you want here

PR #1053, opened 2020-06-09
Fix ButtonGroup row spacing when wrapping

mdodgelooker:
### :sparkles: Changes

- A last-minute change on my previous `ButtonGroup` PR broke the row spacing

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

#### Issue:
image

#### Fixed:
image

### DIFF
--- a/packages/components/src/Button/ButtonSet.tsx
+++ b/packages/components/src/Button/ButtonSet.tsx
@@ -108,7 +108,7 @@ export const ButtonSetLayout = forwardRef(
         const rowHeight = firstItem
           ? firstItem.getBoundingClientRect().height
           : buttonItemHeight
-        if (height > rowHeight * 2) {
+        if (height >= rowHeight * 2) {
           setIsWrapping(true)
         } else {
           setIsWrapping(false)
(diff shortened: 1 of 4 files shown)


packages/components/src/Button/ButtonSet.tsx
```diff
@@ -108,7 +108,7 @@ export const ButtonSetLayout = forwardRef(
         const rowHeight = firstItem
           ? firstItem.getBoundingClientRect().height
           : buttonItemHeight
-        if (height > rowHeight * 2) {
+        if (height >= rowHeight * 2) {
```
mdodgelooker, line 111: At the last minute I had moved the bottom margin on `ButtonGroup` > `ButtonItem` to only apply when `.wrapping` is present, which made this check invalid if the group was wrapping into just 2 rows.

packages/components/src/Button/__snapshots__/ButtonGroup.test.tsx.snap
```diff
@@ -92,7 +92,7 @@ exports[`ButtonGroup controlled 1`] = `
 }
 
 <div
-  class="c0 c1"
+  class="wrapping c0 c1"
```
mdodgelooker, line 95: Since everything has a height of 0 in jsdom, the "wrapping" scenario is now invoked for the snapshots because 0 >= 0 * 2 is true. :)

PR #1059, opened 2020-06-10
FieldCollapseLabel component

laraharrow:
### :sparkles: Changes

- FieldCollapseLabel component

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### :camera: Screenshots
...


storybook/src/Forms/FieldCollapseLabel.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <FieldText label="Second Label" />
+        <FieldText
+          label="Third Label"
+          validationMessage={{
+            message: 'This is an error',
+            type: 'error',
+          }}
+        />
+      </Fieldset>
+    </FieldCollapseLabel>
```
dbchristopher, line 44: Hmm, I'm not quite sure about this api design -- it doesn't seem quite right to import an entirely new component just to hide the labels. It seems a lot more straightforward to just be a prop on the individual Field components:

```
<FieldText label="First Label" collapseLabel />
```

How would you feel about taking that approach instead?

laraharrow: The idea was that it would be leveraged on different components. @lukelooker can you way in here?

lukelooker: The intent here was to build the component separately but also expose it "underneath" `Fieldset` via a `collapseLabels` property. 

That's the expected use case today but I thought it'd be worth offering the wrapper component to allow for testing in exclusion as well as future re-use. A likely use case in the future will be when several Fields are joined together tightly - something like the "Advanced Filter":

image

dbchristopher: That Advanced Filter example would be a bunch `Input` components though, right? Not `Field` components. 

I really feel like we should avoid relying on external side effects to configure our layouts. It will be a lot more error prone and harder to maintain

packages/components/src/Form/Fields/Field.tsx
```diff
@@ -122,17 +134,27 @@ const FieldLayout: FunctionComponent<FieldPropsInternal> = ({
...
-        {label}
-        {required && <RequiredStar />}
-      </Label>
+      {labelCollapse ? (
+        <VisuallyHidden>{labelComponent}</VisuallyHidden>
+      ) : labelsCollapse ? (
+        <VisuallyHidden>{labelComponent}</VisuallyHidden>
+      ) : (
+        labelComponent
+      )}
```
dbchristopher, line 157: any reason why this ternary operator is double-nested? It appears to set up two cases where `<VisuallyHidden>{labelComponent}</VisuallyHidden>` gets rendered, but I'm not following what the difference is

packages/components/src/Form/Fieldset/Fieldset.tsx
```diff
@@ -68,6 +73,12 @@ const accordionIndicatorDefaults: AccordionIndicatorProps = {
   indicatorSize: 'medium',
 }
 
+export interface FieldsetContext {
+  labelsCollapse?: boolean
+}
+
+export const FieldsetContext = createContext<FieldsetContext>({})
```
dbchristopher, line 80: Nice. Context is a good solution if we want to avoid having to set `labelsCollapse` on every sub field. Is it safe to assume that every `Field` component will be rendered inside a `Fieldset`? Or should we also allow passing `labelsCollapse` in to the child components directly as a prop?

dbchristopher: I also wonder if maybe `hideLabels` would be a more straightforward name than `labelsCollapse`?

packages/components/src/Form/Fields/FieldCollapseLabel.tsx
```diff
@@ -0,0 +1,36 @@
...
+import styled from 'styled-components'
+import { Label } from '../Label'
+
+export const FieldCollapseLabel = styled.div`
+  ${Label} {
+    visibility: hidden;
+    height: 0;
+    padding: 0;
+  }
+`
```
dbchristopher, line 36: do we still need this component?

www/src/documentation/components/forms/fields.mdx
```diff
@@ -155,6 +155,26 @@ There are two "inline" variations:
...
+      label="Third Label"
+      validationMessage={{
+        message: 'This is an error',
+        type: 'error',
+      }}
+    />
+  </Fieldset>
+</FieldCollapseLabel>
+```
+
```
dbchristopher, line 177: so this is probably out of date now that we're controlling this through Fieldset context?

packages/components/src/Form/Fields/Field.tsx
dbchristopher: oh I get it -- `labelCollapse` vs `labelsCollapse`. Easy to miss! You could condense the ternary to this:

```
{labelCollapse || labelsCollapse ? (
  <VisuallyHidden>{labelComponent}</VisuallyHidden>
) : (
  labelComponent
)
```

In general I find nested ternary operators really hard to read, so try to avoid that when possible.

packages/components/src/Form/Fieldset/Fieldset.tsx
dbchristopher: > Or should we also allow passing labelsCollapse in to the child components directly as a prop?

I see now that you are in fact doing that with the `labelCollapse` prop. It would be good to add documentation for the two different way to hide the label

PR #1060, opened 2020-06-11
Form Improvements

lukelooker:
### :sparkles: Changes

- `Badge` now supports `ref` assignment (Tooltip badge is now possible)
- `IconButton` now supports `tooltipTextAlign`

- `Field`
  - No longer supports `labelWidth` property (unused feature)
  - No longer supports `labelFontSize` and `labelFontWeight` properties (all your fonts belong to us)

- `FieldInline` (`FieldCheckbox`, `FieldRadio` & `FieldToggleSwitch`)
  - Now supports `detail` property
  - `:disabled` state slightly more discernable from default state now

- `Radio` & `Checkbox` disabled states corrected
- `OptionsGroup` (`CheckboxGroup`, `RadioGroup`)
  - Now have tighter vertical spacing to feel more connected
  - Supports `detail` on Options array

- `Status` now allows DOM attributes (`aria-*` and the like)

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/Form/Fields/Field.tsx
+++ b/packages/components/src/Form/Fields/Field.tsx
@@ -67,8 +67,6 @@ export const fieldPropKeys = [
   'id',
   'inline',
   'label',
-  'labelFontSize',
-  'labelFontWeight',
   'labelWidth',
   'validationMessage',
   'width',
--- a/packages/components/src/Form/Fields/Field.tsx
+++ b/packages/components/src/Form/Fields/Field.tsx
@@ -29,10 +29,9 @@ import styled, { css } from 'styled-components'
 import { ResponsiveValue, TLengthStyledSystem } from 'styled-system'
 import omit from 'lodash/omit'
 import pick from 'lodash/pick'
+import { Paragraph, Text } from '../../Text'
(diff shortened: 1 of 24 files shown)


packages/components/src/Form/Fields/Field.tsx
```diff
@@ -177,8 +157,7 @@ export const Field = styled(FieldLayout)<FieldPropsInternal>`
     inline
       ? '"label input detail" ". messages messages"'
       : '"label detail" "input input" "messages messages"'};
-  grid-template-columns: ${({ inline, labelWidth }) =>
-    inline ? `${labelWidth} 1fr` : undefined};
+  grid-template-columns: ${({ inline }) => (inline ? `150px 1fr` : undefined)};
```
jhardy, line 160: I'll make a TODO for myself to investigate what the inline label width should be. I think 150 is fine for now, but would like to make sure we at least do a little due diligence from the design side to verify that.

lukelooker: Yeah - this is kind of a funny number :p

PR #1065, opened 2020-06-12
[LENS-926] Condense Slider and RangeSlider layouts

dbchristopher:
### :sparkles: Changes

- Smaller font size and tighter spacing to make the two slider components fit within a 36px vertical space

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x PR is ideally < ~400LOC

### :camera: Screenshots
Screen Shot 2020-06-11 at 5 45 52 PM
Screen Shot 2020-06-11 at 5 45 43 PM

### DIFF
--- a/packages/components/src/Form/Inputs/Slider/Slider.tsx
+++ b/packages/components/src/Form/Inputs/Slider/Slider.tsx
@@ -27,7 +27,13 @@
 import React, { forwardRef, Ref, SyntheticEvent, useState } from 'react'
 import isFunction from 'lodash/isFunction'
 import styled, { css } from 'styled-components'
-import { reset, space, SpaceProps } from '@looker/design-tokens'
+import {
+  reset,
+  space,
+  SpaceProps,
+  typography,
+  TypographyProps,
+} from '@looker/design-tokens'
 import { WidthProps, width } from 'styled-system'
 
 import { InputProps } from '../InputProps'
@@ -35,7 +41,8 @@ import { InputProps } from '../InputProps'
 export interface SliderProps
   extends SpaceProps,
     WidthProps,
-    Omit<InputProps, 'type'> {
+    Omit<InputProps, 'type'>,
+    TypographyProps {
   'aria-labelledby'?: string
   max?: number
   min?: number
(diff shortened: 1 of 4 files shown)


jhardy:
Overall looking really good and they fit well when aligned with the text fields, so should solve for our filter bar work. 

I only have a few minor nits 
The value text is a bit off center for `Slider`
image

The values on `RangeSlider` sit a little too close to the handles as well as have a different highlight style than `Slider`
image

I prefer the styling on the range slider and the distance on the single slider. I also think we should  adjust the value font size down to `xsmall`, here is my quick little browser hack that I think feels right 
image

dbchristopher:
@jhardy I just pushed a fix for `The value text is a bit off center for Slider`. TBH that's been bugging me for *a while*. Should be properly centered now.

packages/components/src/Form/Inputs/Slider/Slider.tsx
```diff
@@ -248,13 +262,12 @@ interface SliderValueProps extends SliderInputProps {
...
     disabled ? colors.neutral : colors.key};
-  line-height: 1;
   user-select: none;
-  transform: translateX(-50%) translateY(-1.3rem);
+  transform: translateX(-50%) translateY(-0.9rem);
   left: ${({ offsetPercent }) => offsetPercent}%;
   position: absolute;
   text-align: center;
-  padding: 0.2rem 0.5rem;
+  padding: 0 0.5rem;
```
jhardy, line 270: Everything is looking pretty good! Only question I have is should values here and above like `8px`, 16px`, `0.5rem`, which are available on the theme spacing,  if the hardcoded values are needed then I will 👍

dbchristopher: @jhardy in this case I'm actually not sure if the spacing values should be tied to the theme object. On this line, the padding controls the width of the background surrounding the numeric value. And every instance of 8/16px refers to either the radius or diameter of the slider knob:

Screen Shot 2020-06-15 at 11 35 05 AM

I feel like it would be considered a bug if a customer edits the theme spacing and causes those aspects of the slider change, right?

jhardy: @dbchristopher I think you are right, seems like something we want a bit more control over and not defer to the theme.

PR #1072, opened 2020-06-16
Fix SelectMulti create option & filtering keyboard nav

mdodgelooker:
### :sparkles: Changes

- `SelectMulti`: After the changes for `ComboboxOptionIndicator`, the create option had an awkwardly large space to the left – should've just updated it to `indicator={false}` during that change.
- `Select` & `SelectMulti`: Keyboard navigation was broken when filtering due to a typo-esque error during the performance improvements. Added a new test to catch any future breakage.

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

#### Both issues:
issues

### DIFF
--- a/packages/components/src/Form/Inputs/Combobox/ComboboxMultiOption.tsx
+++ b/packages/components/src/Form/Inputs/Combobox/ComboboxMultiOption.tsx
@@ -55,9 +55,8 @@ const ComboboxMultiOptionInternal = forwardRef(
       indicator,
       highlightText = true,
       scrollIntoView,
-      hideCheckMark,
       ...props
-    }: ComboboxOptionProps & { hideCheckMark?: boolean },
+    }: ComboboxOptionProps,
     forwardedRef: Ref<HTMLLIElement>
   ) => {
     const { label, value } = props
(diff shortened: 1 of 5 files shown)


packages/components/src/Form/Inputs/Combobox/ComboboxMultiOption.tsx
```diff
@@ -55,9 +55,8 @@ const ComboboxMultiOptionInternal = forwardRef(
       indicator,
       highlightText = true,
       scrollIntoView,
-      hideCheckMark,
       ...props
-    }: ComboboxOptionProps & { hideCheckMark?: boolean },
```
mdodgelooker, line 60: `hideCheckMark` is no longer necessary now that the same thing can be accomplished with `indicator={false}`.

packages/components/src/Form/Inputs/Select/Select.test.tsx
```diff
@@ -25,13 +25,13 @@
  */
 
 import { renderWithTheme } from '@looker/components-test-utils'
-import { cleanup, fireEvent } from '@testing-library/react'
-import React from 'react'
+import { cleanup, fireEvent, screen } from '@testing-library/react'
```
mdodgelooker, line 28: Updated the whole test to use `screen` while I was in there.

PR #1073, opened 2020-06-16
Consolidate tooling configuration into components monorepo

lukelooker:
### :sparkles: Changes

Improve ESLint config
  - Split header into seperate extendable configuration
  - Split ts-workspaces into seperate extendable configuration
  - Remove all overrides and integrate directly into eslint-config package

Move ESLint config from package.json to `.eslintrc.js` to prepare for future monorepository membership

Split Stylelint into its own package

- [x] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### DIFF
--- a/packages/eslint-config/tsconfig.build.json
+++ b/packages/eslint-config/tsconfig.build.json
@@ -1,0 +1,8 @@
+{
+  "extends": "../../tsconfig.build.json",
+  "compilerOptions": {
+    "rootDir": "./",
+    "outDir": "./lib"
+  },
+  "include": ["./"]
+}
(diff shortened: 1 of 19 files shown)


packages/eslint-config/tsconfig.build.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig.build.json",
+  "compilerOptions": {
+    "rootDir": "./",
+    "outDir": "./lib"
+  },
+  "include": ["./"]
+}
```
lukelooker, line 8: This is just to let build succeed. It doesn't actually do anything as the *-config projects don't need to be transformed.

PR #1094, opened 2020-06-17
Simplify properties accepted by Input* & TextArea components

lukelooker:
### :sparkles: Changes

- Major simplification to InputText interface
  - No longer supports typography or pseudo props
  - Switch from using defaultProps to `css` block to share common styles with other components
- Major simplification to TextArea interface
  - No longer supports border or typography props
- Deprecate `pseudo` support from design-tokens (prefer using `styled` when pseudo selectors are required)
  - Includes changes to `Box` & `IconButton`

- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


packages/components/src/Form/Inputs/InputColor/InputColor.tsx
```diff
@@ -193,21 +187,34 @@ export const InputColorComponent = forwardRef(
...
+    border-right: ${({ hideInput }) => (hideInput ? undefined : 'none')};
+    border-top-left-radius: ${({ theme: { radii } }) => radii.medium};
+  }
+
+  ${InputText} {
+    border-radius: ${({ theme: { radii } }) => radii.none};
+    border-bottom-right-radius: ${({ theme: { radii } }) => radii.medium};
+    border-top-right-radius: ${({ theme: { radii } }) => radii.medium};
+  }
+`
```
lukelooker, line 220: Use `styled` instead of component overrides since those are no longer available.

packages/components/src/Form/Inputs/InputColor/Swatch/__snapshots__/Swatch.test.tsx.snap
```diff
@@ -2,11 +2,13 @@
 
 exports[`Default Swatch 1`] = `
 .c0 {
+  background: #FFFFFF;
+  border: 1px solid #DEE1E5;
+  border-radius: 0.25rem;
+  color: #343C42;
```
lukelooker, line 8: Background is absorbing the `field` default background color (doesn't really cause any trouble) and the default `color` as well (again, doesn't do anything good orbad)

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
```diff
@@ -436,29 +428,26 @@ const InputDateRangeWrapper = styled.div`
...
-
+  ${inputCSS}
   align-items: center;
   display: inline-grid;
   grid-gap: ${({ theme }) => theme.space.xsmall};
   grid-template-columns: auto auto auto 1fr;
-  margin: ${({ theme: { space } }) => space.xxsmall} 0;
-  padding: 0 ${({ theme: { space } }) => space.small};
+  margin: ${({ theme: { space } }) => space.xxsmall + ' 0'};
+  padding: ${({ theme: { space } }) => '0 ' + space.small};
```
lukelooker, line 450: Oops... fixing this change

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
```diff
@@ -479,7 +468,8 @@ const IconWrapper = styled.div`
 `
 
 const InputTextWrapper = styled.div<{ inputLength: number }>`
-  padding: ${({ theme: { space } }) => space.xxsmall} 0;
+  padding: ${({ theme: { space } }) => space.xxsmall + ' 0'};
```
lukelooker, line 471: And this one...

packages/components/src/Form/Inputs/InputColor/InputColor.tsx
```diff
@@ -52,8 +51,7 @@ import { isValidColor } from './utils/color_utils'
 
 const colorWheelSize = 164
 
-export interface InputColorProps
-  extends Omit<InputTextProps, 'height' | 'size'> {
+export interface InputColorProps extends Omit<InputTextProps, 'height'> {
```
lukelooker, line 54: `size` no longer needs to be excluded as it already is done so by `InputTextProps`

packages/components/src/Form/Inputs/InputSearch/InputSearchBase.tsx
```diff
@@ -196,7 +200,7 @@ export const InputSearchBase = styled(InputSearchBaseComponent)`
     border: none;
     box-shadow: none;
     flex: 1;
-    height: ${(props) => getHeight(props.py)};
+    height: ${(props) => getHeight(props.py || 2)};
```
lukelooker, line 203: Not in love with this fix. 

However this code is a little funky (I think it would fall down if the component was specified with a t-shirt size like `xxxsmall`) so I don't think this is making it any worse than it was before. In the interest of keeping this PR focused I'm going to leave it as-is.

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -242,33 +236,16 @@ export const InputText = styled(InputComponent).attrs(
...
-  borderColor: 'ui2',
-  borderRadius: 'medium',
-  fontSize: 'small' as FontSizes,
-  height: '36px',
-  width: '100%',
-}
-
 InputText.defaultProps = {
-  ...inputTextDefaults,
+  height: inputHeight,
```
lukelooker, line 246: Eventually I'd like to remove support for height altogether but we need to implement a broad `size` property before I think it would make sense to do so.

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
```diff
@@ -476,117 +469,79 @@ const InputTimeInternal = forwardRef(
...
-                color="critical"
-                grid-area="warning"
-                size={20}
-              />
-            )}
-          </InputTimeLayout>
-        </InputTimeWrapper>
+        <InputText
+          id={id}
+          maxLength={2}
```
lukelooker, line 474: Restructured this to provide a much simpler interface here (single styled component wrapping all of the JSX)

The only user-facing change is that `hasInputValues` is no longer used to change the color of the separator `:` when the InputTime is in-active. Design seemed to agree that the simplification in code was worth the small end-user facing trade-off.

packages/components/src/Form/Inputs/TextArea/__snapshots__/TextArea.test.tsx.snap
```diff
@@ -17,10 +17,10 @@ exports[`TextArea default 1`] = `
 
 .c0 textarea {
   min-height: 6.25rem;
-  width: 100%;
```
lukelooker, line 20: Width was being double-outputted so this loss isn't actually a problem.

packages/components/src/MessageBar/MessageBar.tsx
```diff
@@ -72,7 +72,6 @@ const MessageBarLayout = forwardRef(
           id={id ? `${id}-iconButton` : undefined}
           ml="auto"
           onClick={onDismiss}
-          hoverStyle={{ background: 'none', border: 'none' }}
```
lukelooker, line 75: I'm going to address this via a separate PR as this workaround is also common in CORE product and points at a flaw in `IconButton`

packages/components/src/Form/Inputs/InputSearch/InputSearchBase.tsx
```diff
@@ -206,11 +210,9 @@ export const InputSearchBase = styled(InputSearchBaseComponent)`
...
+      outline: none;
+    }
   }
 `
-
-InputSearchBase.defaultProps = {
-  ...omit(inputTextDefaults, ['height', 'px']),
-  pr: 'xxsmall',
-  py: 2,
-}
```
mdodgelooker, line 216: `InputSearchBase` still needs `defaultProps = { height: '100%' }` to get it on the container `div`.
image

lukelooker: I'm guessing you meant "width" here instead of `{ height: '100%' }` ? I fixed that and things look better :)

PR #1095, opened 2020-06-17
IconButton improved hover/active states and no background on hover

lukelooker:
### :sparkles: Changes

- IconButton improved hover/active states and no background on hover
- Refined IconButton to remove the need for common overrides used to suppress default :hover background as well as provide more distinct :hover & :active psuedo states.
- Also addressed a weird bug where a border would be momentarily rendered when the cursor removed focus from a IconButton with no border specified

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots

image

### DIFF
--- a/packages/components/src/Modal/Layout/__snapshots__/ModalHeader.test.tsx.snap
+++ b/packages/components/src/Modal/Layout/__snapshots__/ModalHeader.test.tsx.snap
@@ -59,39 +59,9 @@ exports[`ModalHeader 1`] = `
   opacity: 0.25;
 }
 
-.c3 {
-  background: transparent;
-  border: 1px solid transparent;
-  color: #939BA5;
-  padding: 0 0.5rem;
-}
-
 .c3 .c7 {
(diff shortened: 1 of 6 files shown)


packages/components/src/Modal/Layout/__snapshots__/ModalHeader.test.tsx.snap
```diff
@@ -124,9 +94,23 @@ exports[`ModalHeader 1`] = `
...
 
+.c4:hover,
+.c4:focus,
+.c4.hover {
+  color: #707781;
+}
+
+.c4:active,
+.c4.active {
+  color: #4C535B;
```
jhardy, line 111: Key color is being applied here on hover. 

image

lukelooker: Oops. Fixed.

PR #1096, opened 2020-06-18
Badge now uses generated colors rather than palette references

lukelooker:
### :sparkles: Changes

- Badge now uses generated colors rather than palette references
- Colors are subtly different so I'd like design sign-off on the change. Color math uses functions provided by `design-tokens` library so improvements there should benefit `Badge` along the way.

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

#### Before

image

#### After

image


packages/components/src/Badge/Badge.tsx
```diff
@@ -26,20 +26,30 @@
...
 import { variant } from 'styled-system'
 
 export type BadgeSizes = SizeSmall | SizeMedium | SizeLarge
+type BadgeIntent =
+  | 'warn'
+  | 'positive'
+  | 'critical'
+  | 'inform'
+  | 'neutral'
+  | 'plain'
```
jhardy, line 51: I kind of dislike `plain` as an intent. A couple ideas I had were `quiet` or `simple` as a possible replacement?

packages/design-tokens/src/system/color/specifiable.ts
```diff
@@ -45,6 +45,14 @@ export interface SurfaceColors {
   inverseOn: string
 }
 
+export type IntentNames =
+  | 'link'
```
jhardy, line 49: `link` seems like an outlier here? Any specific reason for including it.

packages/components/src/Badge/Badge.tsx
lukelooker: It's not used at all today. I like "simple" but we could just "retire" it for now if you'd like since it's not used anywhere.

jhardy: I'd vote to retire `plain`.

packages/design-tokens/src/system/color/specifiable.ts
lukelooker: Agreed, I was just on auto-pilot copying over the "IntentColors" keys. 

`link` is a weird beast. Perhaps we should hang it off of `inform` eventually and remove it? A question for later perhaps...

packages/components/src/Badge/Badge.tsx
lukelooker: Done!

PR #1100, opened 2020-06-18
TreeItem: Hover and Selected states

ElliotTheHuman:
### 🛠️  To-Do

- Update test suite (follow-up PR)
- Update documentation (follow-up PR)

### :sparkles: Changes

`TreeItem` 
- Supports hover and selected visual states
- `detailAccessory` prop: allocates space for detail elements right of the label container
- `detailHoverDisclosure` prop: when true, the detail is shown / hidden on `TreeItem` hover
- Moved `TreeItem` CSS (non-indentation padding, size, etc.) from `Tree` to within `TreeItem`
  - Note: `Tree` now overrides its internal `TreeItem`'s styling so hover and border CSS
- Layout reworked to have a "label" and "detail" container
  - primary: always holds label and optional icon, also contains detail elements if `detailAccessory` === false
  - secondary: contains detail elements if `detailAccessory` === true

`Tree`
- No longer handles the bulk of `TreeItem` styling (still handles indentation / depth padding)
- Accepts `detailAccessory` and `detailHoverDisclosure` props
  - Applies these to `TreeContext` if received at the `Tree` level (as opposed to on a `TreeItem`)

`ButtonBase`
- No longer wraps `onKeyDown` within `handleKeyUp`

`design-tokens`
- Added Luke's transparency color math

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots
Screen Shot 2020-06-18 at 3 22 34 PM


packages/components/src/Tree/Tree.tsx
```diff
@@ -35,8 +35,8 @@ import {
   AccordionIndicatorProps,
 } from '../Accordion'
 import { IconNames } from '../Icon'
-import { TreeItem } from './TreeItem'
-import { TreeGroup, TreeGroupLabel } from './TreeGroup'
+import { TreeItem, TreeItemPrimary } from './TreeItem'
```
lukelooker, line 38: I don't love that `TreeItemPrimary` needs to be exposed outside of the `TreeItem` itself. Still pondering how to avoid that however.

packages/components/src/Tree/Tree.tsx
```diff
@@ -77,100 +82,28 @@ const indicatorProps: AccordionIndicatorProps = {
...
   ...restProps
 }) => {
-  const { border: contextBorder, depth } = useContext(TreeContext)
-  const isBorderEnabled = border || contextBorder
-  const nextDepth = depth + 1
+  const context = useContext(TreeContext)
+  const hasBorder = context.border || border
+  const hasDetailHoverDisclosure =
+    context.detailHoverDisclosure || detailHoverDisclosure
+  const hasDetailAccessory = context.detailAccessory || detailAccessory
```
lukelooker, line 100: Used `has*` here to make these easier to read

packages/components/src/Tree/Tree.tsx
```diff
@@ -194,10 +127,81 @@ const TreeLayout: FC<TreeProps> = ({
...
+}
+
+interface TreeBorderProps {
+  border?: boolean
+  depth: number
+}
+
+const TreeBorder = styled.div<TreeBorderProps>`
+  ${({ border, depth, theme }) => border && generateTreeBorder(depth, theme)}
+`
```
lukelooker, line 175: This should get incorporated into `TreeStyle` to reduce the number of wrapping `div`s

packages/components/src/Tree/Tree.tsx
```diff
@@ -194,10 +127,81 @@ const TreeLayout: FC<TreeProps> = ({
...
+  }
+
+  ${TreeGroupLabel} {
+    padding-left: ${({ depth, theme }) =>
+      `calc(${theme.space.xxsmall} + (${theme.space.xxsmall} + ${
+        theme.space.small
+      }) * ${depth + 1})`};
+  }
+
+  ${TreeItemPrimary} {
```
lukelooker, line 201: I wonder if `TreeItemLabel` would be a better name for this to be consistent with `TreeGroupLabel` and remove the idea of precedence.

packages/components/src/Tree/Tree.tsx
```diff
@@ -194,10 +127,81 @@ const TreeLayout: FC<TreeProps> = ({
...
+
+  ${AccordionDisclosure} ${TreeItem}:focus-within {
+    border-color: transparent;
+  }
+
+  ${TreeGroupLabel} {
+    padding-left: ${({ depth, theme }) =>
+      `calc(${theme.space.xxsmall} + (${theme.space.xxsmall} + ${
+        theme.space.small
+      }) * ${depth + 1})`};
```
lukelooker, line 198:
```suggestion
    padding-left: ${({
      depth,
      theme: {
        space: { small, xxsmall },
      },
    }) => `calc(${xxsmall} + (${xxsmall} + ${small}) * ${depth})`};
```

While this makes the de-structure a bit more complex  this change makes it easier to read-through the logic in play here.

packages/components/src/Tree/Tree.tsx
```diff
@@ -194,10 +127,81 @@ const TreeLayout: FC<TreeProps> = ({
...
+const TreeBorder = styled.div<TreeBorderProps>`
+  ${({ border, depth, theme }) => border && generateTreeBorder(depth, theme)}
+`
+
+const TreeStyle = styled.div<{ depth: number }>`
+  ${AccordionDisclosure} {
+    height: 25px;
+    padding: ${({ theme }) => theme.space.xxsmall};
+    padding-left: ${({ depth, theme }) =>
+      `calc(${theme.space.xxsmall} + (${theme.space.xxsmall} + ${theme.space.small}) * ${depth})`};
```
lukelooker, line 182: This chunk of code repeats 3 times. Make it a function :)

packages/components/src/Tree/Tree.tsx
ElliotTheHuman: I'm definitely for this. Marking comment as resolved.

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -55,20 +70,76 @@ export interface TreeItemProps {
...
+
+  const handleDetailKeyDown = (event: KeyboardEvent<HTMLElement>) => {
+    // Automatically prevents detail click from opening Accordion
+    const isDetailAcccessoryEnabled =
+      props.detailAccessory !== undefined
+        ? props.detailAccessory
+        : detailAccessory
+
+    isDetailAcccessoryEnabled && event.stopPropagation()
   }
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 8f56ac3.

Note: This commit also includes a fix for a bug where if `detailAccessory === false`, pressing enter on a focused detail element trigger's the `TreeItem onClick` twice. Need further investigation as to why this is happening without the "fix".

packages/components/src/Tree/Tree.tsx
ElliotTheHuman: Resolved in 291bc1f3f

packages/components/src/Tree/Tree.tsx
ElliotTheHuman: Resolved in 6c34a77e4

packages/components/src/Tree/Tree.tsx
ElliotTheHuman: Resolved in 6692ebcce

packages/components/src/Tree/Tree.tsx
ElliotTheHuman: Resolved in 6692ebcce

packages/components/src/Tree/TreeItem.tsx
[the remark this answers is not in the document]
ElliotTheHuman: Could be totally wrong here, but the "bug" where hitting enter on a focused button within the `detail` spot (when `detailAccessory === false`) is due to the fact that hitting enter on a focused button will trigger both `onClick` and `onKeyDown` handlers. 

As a result, two events bubble up (a click event and a key down event).
enter-key-bug

In this case, what's the best way for us to prevent both these events from firing off an "Enter" key press?

PR #1103, opened 2020-06-19
Popover now supports the preferred `cloneElement` style

lukelooker:
### :sparkles: Changes

- Popover now supports the preferred `cloneElement` style usage in addition to the existing render prop style
- Additionally the render prop style now exposes `aria-haspopup` for use

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


packages/components/src/Popover/Popover.tsx
```diff
@@ -527,16 +538,44 @@ export function Popover({
...
+        ? `${children.props.className} hover`
+        : children.props.className,
+      onClick: popoverProps.open,
+      ref: popoverProps.ref,
+    })
+  } else if (isRenderProp(children)) {
+    target = children(
+      popoverProps.open,
+      popoverProps.ref,
+      popoverProps.isOpen ? 'hover' : '',
```
mdodgelooker, line 562:
```suggestion
      popoverProps.isOpen ? 'active' : '',
```

packages/components/src/Popover/Popover.tsx
```diff
@@ -134,20 +136,29 @@ export interface UsePopoverProps {
...
 
+type PopoverRenderProp = (
+  onClick: (event: SyntheticEvent) => void,
+  /**
+   * Used by popper.js to position the OverlaySurface relative to the trigger
+   */
+  ref: Ref<any>,
+  className?: string,
+  ariaHaspopup?: boolean
+) => ReactNode
```
mdodgelooker, line 147: Do we want to follow `TooltipRenderProp` and just have a single object argument? https://github.com/looker-open-source/components/blob/master/packages/components/src/Tooltip/Tooltip.tsx#L108

Also, what do you think about going one step further with accessibility and adding `aria-expanded` (following the `isOpen` state): https://www.w3.org/TR/wai-aria-practices-1.1/#disclosure

packages/components/src/Popover/Popover.tsx
```diff
@@ -527,16 +538,44 @@ export function Popover({
...
+
+  // const popoverPropsLabeled = {
+  //   ...omit(popoverProps, ['popover', 'isOpen']),
+  // }
+
+  if (isValidElement(children)) {
+    target = cloneElement(children, {
+      'aria-haspopup': true,
+      className: popoverProps.isOpen
+        ? `${children.props.className} hover`
```
lukelooker, line 553:
```suggestion
        ? `${children.props.className} active`
```

packages/components/src/Popover/Popover.tsx
lukelooker: Great point! :) 

I'd originally not made this change to preserve the existing usage points but I think it's worth the change in this case because it makes it a no-op in the future to add more auto-spreading props. I went ahead and made the change as well as adding support for `aria-expanded` :)

PR #1104, opened 2020-06-19
Expose `wrap` option on Fieldset for use by OptionsGroup(s)

lukelooker:
### :sparkles: Changes

- Expose `wrap` option on Fieldset and use OptionsGroup(s)
- `CheckboxGroup` & `RadioGroup` options now properly wrap when the exceed the container width

image

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


lukelooker:
> Actually I think either the `FieldsetProps` also needs to be updated to be updated to remove `SpaceHelperProps` or those need to be added in its style block (or some kind of combo of the two). Right now it doesn't do anything with `width` or `padding` props, etc.

Improved!

PR #1105, opened 2020-06-19
Deprecate Drawer, rename Modal* to Dialog*

lukelooker:
### :sparkles: Changes

- `Drawer` is removed
- `Modal*` has moved to `Dialog*`
  - `ModalContext` is now `DialogContext`
  - `ModalContent` is now `DialogContent`
  - `ModalFooter` is now `DialogFooter`
  - `ModalHeader` is now `DialogHeader`
  - Aliases for the old names are in place to ease the upgrade but these aliases will be deprecated shortly

- [x] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


packages/components/src/Dialog/DialogContext.ts
```diff
@@ -27,7 +27,7 @@
 import { createContext, MutableRefObject } from 'react'
 import { FocusTrap } from 'focus-trap'
 
-export interface ModalContextProps {
+export interface DialogContextProps {
   closeModal?: () => void
```
mdodgelooker, line 31: Should this change to `closeDialog`?

I guess I'm not 100% clear on the motivation for renaming. Is it to get rid of `Modal` completely b/c removing `Drawer` means there's no need for generic naming? (If so, there's still a bunch of  components in the root of the `/Dialog` folder that need to be renamed.)

If not, `DialogContext` is used in `Popover` – could that be reason to keep it named `ModalContext`?

packages/components/src/Dialog/Dialog/Dialog.test.tsx
```diff
@@ -30,10 +30,10 @@ import {
   mountWithTheme,
```
mdodgelooker, line 30: Just making sure the double `/Dialog/Dialog` folder structure was intentional.

packages/components/src/Dialog/Layout/DialogContent.tsx
```diff
@@ -135,3 +135,8 @@ const Inner = styled.div<SpaceProps>`
 `
 
 Inner.defaultProps = { px: 'xlarge', py: 'large' }
+
+/**
+ * Legacy fallback until all existing call sites are updated
+ */
+export const ModalContent = DialogContent
```
mdodgelooker, line 142: Should there be a fallback for `ModalContext` call sites as well?

www/src/documentation/components/dialogs/confirm.mdx
```diff
@@ -1,10 +1,10 @@
 ---
```
mdodgelooker, line 1: `navigation.ts` needs to be updated with the new `/dialogs` path.

packages/components/src/Dialog/DialogContext.ts
lukelooker: Sorry, I should've detailed my plans more. Over a series of PRs I plan to deprecate `Modal` across the code base. 

This initial pass is just tearing out Drawer and moving the Layout/Modal* to Layout/Dialog*

In follow-up PRs I'll be wiping out the others. 

`ModalContext` is a strange one... Trying to untangle Dialog/Modal & Popover but haven't quite figured that out yet. Maybe it should get tied to the `Portal` component (coming in upcoming PR)

packages/components/src/Dialog/Dialog/Dialog.test.tsx
lukelooker: Yes, unwound in a later PR

packages/components/src/Dialog/Layout/DialogContent.tsx
lukelooker: Yes! Will fix in a moment

packages/components/src/Dialog/DialogContext.ts
mdodgelooker: Ok got it. ✅ `ModalContext` is a strange one indeed. For `Dialog` and `Popover` it enables management of the scroll locks and focus traps but for `Tooltip` (forgot it was in there too) it just provides `closeModal` for the hotkeys used in `OverlaySurface` (shared with `Popover`). Perhaps we don't need it for `Tooltip` anymore? I don't think the `ESC` key even works to close it...

www/src/documentation/components/dialogs/index.mdx
```diff
@@ -40,70 +40,70 @@ This provides the backdrop behind Modal containers. It can be customized via the
 </ModalSurface>
```
mdodgelooker, line 40: Should it be `DialogSurface`?
image

PR #1106, opened 2020-06-19
Rename ModalPortal to just Portal

lukelooker:
### :sparkles: Changes

- Rename ModalPortal to just Portal

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


packages/components/src/Portal/Portal.tsx
```diff
@@ -27,14 +27,27 @@
 import React, { forwardRef, Ref, useEffect, useRef, ReactNode } from 'react'
 import { createPortal } from 'react-dom'
 import styled from 'styled-components'
-import { getModalRoot } from './modalRoot'
 
-export const ModalPortal = forwardRef(
+const getPortalRoot = () => {
```
mdodgelooker, line 31: Is it ok if we still export this helper? It's being used in a workaround for some `react-select` issues in the core product.

lukelooker: Yes! Also added an alias for `getModalRoot` until we can remove the last of react-select uses :p

PR #1108, opened 2020-06-19
Dialog clean-up Phase III

lukelooker:
### TODO

- [ ] Remove DialogManager (later, separate PR)
- [ ] Solve DialogContext (used by Popover & Tooltip, sparingly)
  - [x] Remove Tooltip usage
  - [ ] Rename to PortalContext perhaps? (later, separate PR)
- [x] Should we be exporting Backdrop & Surface?
- [x] Rationalize closeModal (naming & should it be optional in context)

### :sparkles: Changes

- Deprecated all use of `Modal` except `closeModal`

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


packages/components/src/Dialog/Dialog.test.tsx
```diff
@@ -0,0 +1,139 @@
...
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+import 'jest-styled-components'
+import React from 'react'
+import {
+  assertSnapshotShallow,
+  mountWithTheme,
```
lukelooker, line 31: Really only mimimal changes here but Git doesn't want to see this as a `git mv`

packages/components/src/Dialog/DialogContext.ts
```diff
@@ -26,9 +26,10 @@
 
 import { createContext, MutableRefObject } from 'react'
 import { FocusTrap } from 'focus-trap'
+import noop from 'lodash/noop'
 
 export interface DialogContextProps {
-  closeModal?: () => void
+  closeModal: () => void
```
lukelooker, line 32: No longer optional! :)

storybook/src/Dialog/SaveChanges.stories.tsx
```diff
@@ -0,0 +1,165 @@
...
+        </DialogContent>
+        <DialogFooter>
+          <Button onClick={handleSave}>Save</Button>
+          <ButtonTransparent onClick={handleCancel}>Cancel</ButtonTransparent>
+        </DialogFooter>
+      </Dialog>
+      {confirm}
+    </>
+  )
+}
```
lukelooker, line 165: Moved this from the documentation to here because it felt overly complex in-context.

www/src/MDX/Pre/allComponents.ts
```diff
@@ -160,6 +159,7 @@ import {
   useToggle,
   VisuallyHidden,
 } from '@looker/components'
+import { Surface } from '@looker/components/src/Dialog/Surface'
```
mdodgelooker, line 162: This is because you don't want folks to directly use `Surface` but still want to demo the styling options via `surfaceStyles` in docs?

lukelooker: Yep

lukelooker: Now I'm wondering if this is going to _actually_ work in practice. Hmm...

packages/components/src/Dialog/__snapshots__/Dialog.test.tsx.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Dialog Active 1`] = `ShallowWrapper {}`;
+
+exports[`Dialog Inactive 1`] = `ShallowWrapper {}`;
+
+exports[`DialogManager - click events inactive 1`] = `ShallowWrapper {}`;
```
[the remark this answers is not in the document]
lukelooker: Agreed. Done :)

packages/components/src/Dialog/DialogManager.tsx
```diff
@@ -0,0 +1,123 @@
...
+}
+
+export const DialogManager: FC<DialogManagerProps> = ({
+  canClose,
+  content,
+  children,
+  onClose,
+  ...props
+}) => {
+  const triggerRef = useRef()
```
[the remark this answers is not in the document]
lukelooker: So there was a time when we re-assigned focus back to the trigger element when the Dialog was closed. It appears we don't do so anymore but I wanted to check-in with Jared if that's a "correct" behavior or not and restore it (in a separate PR) if so. I'll file a ticket to resolve whether that's true and if not remove this later-on.

www/src/MDX/Pre/allComponents.ts
mdodgelooker: I think it works. I think they won't need `Surface` because they can't do something like this b/c of the portal:
```
const StyledDialog = styled(Dialog)`
  ${Surface} {
    ...
  }
`
So the only option is via `surfaceStyles` right?

File: packages/components/src/Dialog/DialogManager.tsx (hunk 6 above)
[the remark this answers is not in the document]
Comment (mdodgelooker) [reply packages/components/src/Dialog/DialogManager.tsx:75]: Is it not working anymore? Oops. I thought `focus-trap` was doing that...

File: packages/components/src/Portal/Portal.tsx [hunk 7]
```diff
@@ -28,8 +28,10 @@ import React, { forwardRef, Ref, useEffect, useRef, ReactNode } from 'react'
 import { createPortal } from 'react-dom'
 import styled from 'styled-components'
 
+const rootId = 'modal-root'
```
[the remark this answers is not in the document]
lukelooker: It is referred to once in the CORE product so I just wanted to leave it alone for now. I actually think this is something we should make configurable at some point as it seems like something a developer might want to control in certain cases.

packages/components/src/Dialog/Dialog.tsx
```diff
@@ -66,33 +68,15 @@ export interface ManagedModalProps {
...
-  /**
-   * To implement Modal the Surface is supplied as a function so it can consume the animationState of the Modal.
-   * animationState will be null, 'exited', 'entering' or 'exiting' and can be used to set CSS class on Surface
-   * element to provide CSS transitions. (See DialogSurface for implementation examples)
-   */
-  render: (animationState: string) => JSX.Element
-}
-
-export function Modal({
+export const Dialog: FC<DialogProps> = ({
```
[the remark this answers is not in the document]
lukelooker: Perhaps? I'm kinda horrified with what people might do with that :|

I don't think there's a use case right now so I think I'd prefer to leave it off for now in the interest of cutting a fresh release with the pile of stuff already in play :)

PR #1110, opened 2020-06-23
autoResize on Select & InputText

mdodgelooker:
### Context
Several components in Looker filters combine one or more each of `Select` and `InputText` and require the `Select` to grow and shrink depending on the current value. (Allowing the `InputText` to also adjust is a nice-to-have.)

### :sparkles: Changes

- I extracted a base component `InlineInputTextBase` out of `InlineInputText` that has the resizing functionality but not the styling
- Building on the work done by others on `InputSearchBase` and `iconBefore`, etc, in `InputText` I was able to absorb the functionality of the former into the latter and then use `InlineInputTextBase` to provide the resizing feature

- [x] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### :camera: Screenshots
input-text-autoresize
select-autoresize


packages/components/src/Form/Fields/FieldSelect/FieldSelect.test.tsx
```diff
@@ -37,10 +37,6 @@ test('A FieldSelect', () => {
   assertSnapshot(<FieldSelect label="👍" name="thumbsUp" id="thumbs-up" />)
 })
 
-test('FieldSelect supports labelWeight', () => {
-  assertSnapshot(<FieldSelect label="👍" name="thumbsUp" id="thumbs-up" />)
-})
```
mdodgelooker, line 42: No more `labelWeight`

packages/components/src/Form/Fields/FieldSelect/FieldSelect.test.tsx
```diff
@@ -75,14 +71,6 @@ test('Should trigger onChange handler', () => {
   expect(handleChange).toHaveBeenCalledTimes(1)
 })
 
-test('A required FieldSelect', () => {
-  const component = createWithTheme(
-    <FieldSelect label="👍" name="thumbsUp" id="thumbs-up" required />
-  )
-  const tree = component.toJSON()
-  expect(tree).toMatchSnapshot()
-})
```
mdodgelooker, line 84: This component doesn't do anything specific with `required` so this snapshot isn't useful.

packages/components/src/Form/Inputs/InlineInputText/InlineInputText.tsx
```diff
@@ -28,26 +28,30 @@ import pick from 'lodash/pick'
...
+  TypographyProps,
+} from '@looker/design-tokens'
+import { inputPropKeys, InputProps, InputTextTypeProps } from '../InputProps'
+import { innerInputStyle } from '../innerInputStyle'
+
+export interface InlineInputTextProps
   extends TypographyProps,
-    Omit<InputProps, 'type'> {
+    Omit<InputProps, 'type'>,
+    InputTextTypeProps {
```
mdodgelooker, line 42: Since this is now being used inside `InputText` we want it to support the same `type` prop, not just `'text'`.

packages/components/src/Form/Inputs/InlineInputText/InlineInputText.tsx
```diff
@@ -28,26 +28,30 @@ import pick from 'lodash/pick'
...
-export const InlineInputTextInternal = forwardRef(
+const InlineInputTextLayout = forwardRef(
   (
     {
       className,
       onChange,
-      underlineOnlyOnHover,
       value: valueProp,
       placeholder,
-      simple = false,
```
mdodgelooker, line 50: `simple` wasn't conferring any styles to the `input` itself, so it's not necessary to pass down.

packages/components/src/Form/Inputs/InlineInputText/InlineInputText.tsx
```diff
@@ -62,70 +66,64 @@ export const InlineInputTextInternal = forwardRef(
 
     const handleChange = isFunction(onChange) ? onChange : handleValueChange
 
-    /**
-     * &#8203; is a zero-width space – ensures that visibleText gets accurate line-height applied
-     */
-
     return (
-      <div className={className} data-testid="inlineInputText">
```
mdodgelooker, line 70: Noticed a React error on the docs page for wrapping a `div` inside a `p`. Also, the `data-testid` wasn't being used.

packages/components/src/Form/Inputs/InlineInputText/InlineInputText.tsx
```diff
@@ -62,70 +66,64 @@ export const InlineInputTextInternal = forwardRef(
...
+          type={type}
           ref={ref}
-          {...pick(props, inputPropKeys)}
+          {...omitStyledProps(pick(props, inputPropKeys))}
         />
-        <VisibleText displayValue={displayValue}>
-          {displayValue || placeholder || ' '}
-        </VisibleText>
-      </div>
+        <StyledText>{displayValue || placeholder || ' '}</StyledText>
```
mdodgelooker, line 79: Simpler to keep the text transparent and let the input have its placeholder.

packages/components/src/Form/Inputs/InlineInputText/InlineInputText.tsx
```diff
@@ -62,70 +66,64 @@ export const InlineInputTextInternal = forwardRef(
...
   white-space: pre;
 `
 
-export const InlineInputText = styled(InlineInputTextInternal)`
+export const InlineInputTextBase = styled(InlineInputTextLayout)`
+  display: inline-flex;
+  justify-content: center;
+  min-width: 2rem;
+  position: relative;
+`
```
mdodgelooker, line 117: Separating out a "base" component without the bottom border, etc, for use in `InputText`.

packages/components/src/Form/Inputs/InlineInputText/InlineInputText.tsx
```diff
@@ -62,70 +66,64 @@ export const InlineInputTextInternal = forwardRef(
...
   text-transform: inherit;
   top: 0;
-  width: 100%;
+  &::-webkit-outer-spin-button,
+  &::-webkit-inner-spin-button {
+    appearance: none;
+  }
+  &[type='number'] {
+    appearance: textfield;
+  }
```
mdodgelooker, line 102: Now that `type="number"` is supported, the browser arrows look pretty bad.

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
```diff
@@ -60,8 +63,10 @@ export interface InputChipsControlProps {
 }
 
 export interface InputChipsCommonProps
-  extends Omit<InputSearchBaseProps, 'value' | 'defaultValue' | 'onChange'>,
+  extends Omit<InputTextBaseProps, 'defaultValue' | 'onChange'>,
     MaxHeightProps {
+  summary?: string
+  hideControls?: boolean
```
mdodgelooker, line 69: These 2 were in the docs but not explicitly on the interface.

packages/components/src/Chip/Chip.tsx
```diff
@@ -72,9 +72,6 @@ const ChipStyle = styled.span<ChipProps>`
   color: ${({ theme }) => theme.colors.keyInteractive};
   display: flex;
   justify-items: center;
-  & + & {
-    margin-left: ${({ theme }) => theme.space.xxsmall};
-  }
```
mdodgelooker, line 77: To make `InputText` work as the underlying component of `InputChipsBase` I was having to override this built-in spacing. But then I remembered what @lukelooker said about using `Space` to apply spacing to elements instead of elements hard-coding their own spacing. :)

packages/components/src/Form/Fields/FieldText/__snapshots__/FieldText.test.tsx.snap
```diff
@@ -317,8 +345,8 @@ exports[`A FieldText with label inline 1`] = `
     className="c3 "
   >
     <div
-      className="c4 c5 "
-      onClick={[Function]}
+      className="c4 "
+      onMouseDown={[Function]}
```
mdodgelooker, line 349: See changes to `InputText` below, but long story short, when merging in the functionality from `InputSearchBase` I realized that `onMouseDown` provides a more natural input focus experience.

packages/components/src/Form/Inputs/InlineInputText/InlineInputText.tsx
```diff
@@ -62,70 +66,64 @@ export const InlineInputTextInternal = forwardRef(
...
-     */
-
     return (
-      <div className={className} data-testid="inlineInputText">
-        <Input
+      <span className={className}>
+        <StyledInput
           onChange={handleChange}
-          simple={simple}
-          underlineOnlyOnHover={underlineOnlyOnHover}
```
mdodgelooker, line 74: These 2 don't confer any styling on the `input` itself. Also, they're doing the same thing on the wrapper: adding a dotted underline on hover only. Should we deprecate one of them?

packages/components/src/Form/Inputs/InlineInputText/InlineInputText.tsx
```diff
@@ -62,70 +66,64 @@ export const InlineInputTextInternal = forwardRef(
...
+  }
 `
 
-interface VisibleTextProps {
-  displayValue?: string
-}
-
-const VisibleText = styled.div<VisibleTextProps>`
-  color: ${({ displayValue, theme }) =>
-    displayValue ? 'transparent' : theme.colors.text5};
```
mdodgelooker, line 112: `innerInputStyle` already provides `text5` color on the placeholder.

packages/components/src/Form/Inputs/InlineInputText/__snapshots__/InlineInputText.test.tsx.snap
```diff
@@ -6,42 +6,81 @@ exports[`InlineInputText renders an input with a placeholder 1`] = `
...
+  -webkit-appearance: none;
+  -moz-appearance: none;
+  appearance: none;
+}
+
+.c1[type='number'] {
+  -webkit-appearance: textfield;
+  -moz-appearance: textfield;
+  appearance: textfield;
+}
```
mdodgelooker, line 57: A bunch more styles coming from `innerInputStyle` that are helpful now that this component supports some other values for `type`.

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
```diff
@@ -130,45 +136,62 @@ export const InputChipsBaseInternal = forwardRef(
...
   }
 )
 
 InputChipsBaseInternal.displayName = 'InputChipsBaseInternal'
 
+const inputHeightStyle = css`
+  height: calc(
+    ${inputHeight} - ((${({ theme: { space } }) => space.xxxsmall} + 1px) * 2)
+  );
+`
```
mdodgelooker, line 173: This replaces the awkward `getHeight` function that I had before (most recently living in `InputSearchBase`). That was trying to support custom vertical padding applied to the component, and shrinking down the internal `input` to make sure the total height was `36px` (at least until the chips start wrapping). But I don't think we want to support custom vertical padding on these components, right?

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
```diff
@@ -130,45 +136,58 @@ export const InputChipsBaseInternal = forwardRef(
...
+
 export const InputChipsBase = styled(InputChipsBaseInternal)`
   align-items: stretch;
   position: relative;
 
-  ${Flex} {
+  .inner {
+    align-content: flex-start;
+    display: flex;
+    flex-wrap: wrap;
```
mdodgelooker, line 180: Pulling in styles that were previously applied in `InputSearchBase` but are specifically for chips.

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
```diff
@@ -469,10 +469,7 @@ const IconWrapper = styled.div`
 
 const InputTextWrapper = styled.div<{ inputLength: number }>`
   padding: ${({ theme: { space } }) => space.xxsmall} 0;
-  ${InlineInputText} {
-    background: transparent;
-    border: none;
-    border-radius: 0;
```
mdodgelooker, line 475: With the new `InlineInputTextBase`, these overrides aren't needed.

packages/components/src/Form/Inputs/InputProps.ts
```diff
@@ -69,6 +92,7 @@ export const inputPropKeys = [
   'step',
   'value',
   'aria-autocomplete',
+  'aria-invalid',
```
mdodgelooker, line 95: Realized this was missing before.

packages/components/src/Form/Inputs/InputSearch/InputSearch.tsx
```diff
@@ -68,7 +61,7 @@ export const InputSearch = forwardRef(
     }: InputSearchProps,
     ref: Ref<HTMLInputElement>
   ) => {
-    const [value, setValue] = useState<string | undefined>()
+    const [value, setValue] = useState('')
```
mdodgelooker, line 64: I was getting a the "uncontrolled to controlled" React warning in the `defaultValue` test, due to having an undefined `value` on the initial render and then quickly updating to a string `value` after the `useEffect` runs.

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -26,121 +26,200 @@
...
+   */
+  autoResize?: boolean
+
+  onClick?: (e: MouseEvent<HTMLDivElement>) => void
+  onMouseDown?: (e: MouseEvent<HTMLDivElement>) => void
+  onMouseEnter?: (e: MouseEvent<HTMLDivElement>) => void
+  onMouseLeave?: (e: MouseEvent<HTMLDivElement>) => void
+  onMouseOver?: (e: MouseEvent<HTMLDivElement>) => void
+  onMouseOut?: (e: MouseEvent<HTMLDivElement>) => void
+  onMouseUp?: (e: MouseEvent<HTMLDivElement>) => void
```
mdodgelooker, line 61: This is pulled in from the old `InputSearchBase`. Since the wrapping element is a `div`, with small amounts of mouse-able area outside of the inner `input`, the mouse events should be applied to the `div` instead of the `input`.

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -26,121 +26,191 @@
...
-import { useForkedRef } from '../../../utils'
+import { useForkedRef, useWrapEvent } from '../../../utils'
+import { InlineInputTextBase } from '../InlineInputText'
 
-export interface InputTextProps
+export interface InputTextBaseProps
   extends Omit<SimpleLayoutProps, 'size'>,
-    Omit<InputProps, 'type'> {
+    Omit<InputProps, 'type'>,
+    InputTextTypeProps {
```
mdodgelooker, line 47: Extracted the `type` prop to its own interface in `InputProps.ts` for use in `InlineInputTextBase`.

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -221,31 +263,46 @@ export const inputCSS = css`
...
-      py: props.py || props.p || 'none',
-    }
-    if (props.prefix || props.iconBefore) {
-      padding.pl = 'xsmall'
-    }
-    if (props.suffix || props.iconAfter) {
-      padding.pr = 'xsmall'
-    }
-    return padding
-  }
```
mdodgelooker, line 237: This logic is a convoluted way of making sure we don't clobber padding props, but I don't think we want to encourage the padding props on this anymore, right?

And with the styling changes I've made, padding adjustments for `prefix`, etc, are no longer necessary.

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -221,31 +263,46 @@ export const inputCSS = css`
...
-)<InputTextProps>`
+export const InputText = styled(InputTextLayout)<InputTextProps>`
+  align-items: center;
+  cursor: text;
+  display: inline-flex;
+  justify-content: space-evenly;
+  padding: ${({ theme: { space } }) => `${space.xxxsmall} ${space.xxsmall}`};
+  width: ${({ autoResize }) => (autoResize ? 'auto' : '100%')};
```
mdodgelooker, line 272: Moving `width` up here from `defaultProps` so it can be toggled to `'auto'` for `autoResize` behavior. It can still be overridden via a `width` prop.

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
```diff
@@ -469,7 +469,8 @@ const InputTimeInternal = forwardRef(
         onBlur={onBlur}
         aria-invalid={validationType === 'error' ? 'true' : undefined}
       >
-        <InputText
+        <input
+          type="text"
```
mdodgelooker, line 473: `InputText` is now too heavy for this usage.

packages/components/src/Form/Inputs/InlineInputText/InlineInputText.tsx
```diff
@@ -28,26 +28,30 @@ import pick from 'lodash/pick'
...
-import { inputPropKeys, InputProps } from '../InputProps'
-
-interface InlineInputTextProps
+import {
+  omitStyledProps,
+  typography,
+  TypographyProps,
+} from '@looker/design-tokens'
+import { inputPropKeys, InputProps, InputTextTypeProps } from '../InputProps'
+import { innerInputStyle } from '../innerInputStyle'
```
mdodgelooker, line 37: `innerInputStyle` is the common styles for various components that render a text input inside of another element that's styled as the faux input.

packages/components/src/Form/Inputs/Select/Select.tsx
```diff
@@ -167,6 +176,7 @@ const SelectComponent = forwardRef(
             persistSelection
             windowedOptions={windowedOptions}
             indicator={indicator}
+            width={autoResize ? 'auto' : undefined}
```
mdodgelooker, line 179: If we don't set `width` to `'auto'` the effect is jarring. If the list width matches the input width, it jumps around and often causes the longer options to wrap.

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -26,121 +26,191 @@
...
+      'aria-invalid': validationType === 'error' ? true : undefined,
+      type,
+    }
+    const input = <input {...inputProps} ref={ref} />
+
+    const inner = children ? (
+      <div className="inner">
+        {children}
+        {input}
+      </div>
```
mdodgelooker, line 191: Support for `InputChips` and `SelectMulti`.

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -26,121 +26,191 @@
...
+    ) : autoResize ? (
+      <InlineInputTextBase {...inputProps} ref={ref} />
```
mdodgelooker, line 193: The heart of this PR. :)

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -26,121 +26,192 @@
...
+  onMouseLeave?: (e: MouseEvent<HTMLDivElement>) => void
+  onMouseOver?: (e: MouseEvent<HTMLDivElement>) => void
+  onMouseOut?: (e: MouseEvent<HTMLDivElement>) => void
+  onMouseUp?: (e: MouseEvent<HTMLDivElement>) => void
+}
+
+export interface InputTextProps extends InputTextBaseProps {
+  /**
+   * Experimental, will likely be removed in a future release
+   * @private
```
mdodgelooker, line 67: Marking as @private per convo with @lukelooker

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
dbchristopher: Yeah I can't see any reason we need to go out of our way to support that!

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
```diff
@@ -542,22 +545,15 @@ export const InputTime = styled(InputTimeInternal)`
   grid-template-columns: auto auto auto auto 1fr;
   padding: 0 ${({ theme }) => theme.space.xsmall};
 
-  ${InputText} {
-    background: transparent;
-    border: none;
-    border-radius: 0;
-    box-shadow: none;
-    color: inherit;
+  input {
```
dbchristopher, line 548: I think it would be good to break this out into a new styled Input component rather than using the css child selectors. It can still be defined in this file though

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `eslint-config`, `prettier-config` and `stylelint-config` packages moved from separate repository and updated to integrate any overrides needed previously.
 - `zIndexFloor` added to `theme` (used as base value for Overlay & Modal z-index value:q)
 - `MultiSelect` icon added
+- `autoResize` added to `InputText` and `Select`, allowing component width to adjust to the current value
+- `before` and `after` added to `InputText`
```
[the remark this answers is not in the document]
mdodgelooker: Good call, will do.

packages/components/src/Form/Fields/FieldText/__snapshots__/FieldText.test.tsx.snap
```diff
@@ -6,45 +6,73 @@ exports[`A FieldText with default label 1`] = `
...
+}
+
+.c4 .c6 {
+  height: 100%;
+  max-width: 100%;
+  width: 100%;
+}
+
+.c4 .c6 span {
+  padding: 0 0.5rem;
```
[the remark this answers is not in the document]
mdodgelooker: Oh shoot, thanks for catching that. I don't think it should be too bad to style – might just need to toggle on `display: inline-flex` on `Field` if `autoFocus` is found. I'll give it a go.

www/src/documentation/components/forms/input-text.mdx
```diff
@@ -34,6 +34,14 @@ If both `IconBefore` and `Prefix`, or `IcanAfter` and `Suffix` is passed an warn
...
 ```
 
+## autoResize
+
+Use `autoResize` to allow the width of the component to adjust to the current value or placeholder.
+
+```jsx
+<InputText autoResize placeholder="Start typing" />
+```
+
```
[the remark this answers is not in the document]
mdodgelooker: Will do!

packages/components/src/Form/Inputs/InlineInputText/InlineInputText.tsx
```diff
@@ -62,70 +66,64 @@ export const InlineInputTextInternal = forwardRef(
...
-interface VisibleTextProps {
-  displayValue?: string
-}
-
-const VisibleText = styled.div<VisibleTextProps>`
-  color: ${({ displayValue, theme }) =>
-    displayValue ? 'transparent' : theme.colors.text5};
+const StyledText = styled.span`
+  color: transparent;
+  line-height: inherit;
```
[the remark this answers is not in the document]
mdodgelooker: Oh good question – the `font: inherit;` on `StyledInput` catches line-height, but now I'm thinking `line-height: inherit;` isn't actually necessary here on `StyledText`. I think it may have been from before I converted this to a `span` but let me check it out...

packages/components/src/Form/Inputs/InputProps.ts
```diff
@@ -31,6 +31,26 @@ export interface InputProps extends CompatibleHTMLProps<HTMLInputElement> {
...
 }
 
+export interface InputTextTypeProps {
+  /**
+   *
+   * @default 'text'
+   */
+  type?:
+    | 'date'
+    | 'datetime-local'
```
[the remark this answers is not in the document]
mdodgelooker: Good call, will do.

packages/components/src/Form/Inputs/InputText/InputText.test.tsx
```diff
@@ -47,6 +47,10 @@ test('InputText default', () => {
   assertSnapshot(<InputText />)
 })
 
+test('InputText autoResize', () => {
+  assertSnapshot(<InputText autoResize />)
```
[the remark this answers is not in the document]
mdodgelooker: Sadly the behavior can't be tested b/c jsdom doesn't do layout. Do you think it would be better to just check for a css attribute or two?

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -26,121 +26,192 @@
...
+
+    const beforeToUse = before || iconBeforeOrPrefix
+
+    const iconAfterOrSuffix =
+      iconAfter || suffix ? (
+        <InputTextContent pl="xsmall" pr="xxsmall">
+          {iconAfter ? (
+            <Icon name={iconAfter} size={20} />
+          ) : (
+            <Text fontSize="small">{suffix}</Text>
```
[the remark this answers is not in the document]
mdodgelooker: I'd be down... @lukelooker , thoughts?

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -26,121 +26,192 @@
...
-        {after && after}
-        {validationType && (
-          <InputIconStyle paddingLeft="xsmall">
+    function handleMouseDown() {
+      // set focus to input on mousedown of container to mimic natural input behavior
+      // need requestAnimationFrame here due to browser updating focus _after_ mousedown is called
+      window.requestAnimationFrame(() => {
+        internalRef.current && internalRef.current.focus()
+      })
+    }
```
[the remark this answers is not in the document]
mdodgelooker: This is what I get if I delete it:
focus

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -26,121 +26,192 @@
...
-    if (iconBefore && prefix) {
+    if ((before && (iconBefore || prefix)) || (iconBefore && prefix)) {
       // eslint-disable-next-line no-console
-      console.warn(`Only use IconBefore or prefix not both at the same time. `)
+      console.warn(`Use only one of before, iconBefore, or prefix.`)
       return null
     }
 
-    if (iconAfter && suffix) {
+    if ((after && (iconAfter || suffix)) || (iconAfter && suffix)) {
```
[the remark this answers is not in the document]
mdodgelooker: Nice!

packages/components/src/Form/Inputs/InputText/InputText.tsx
mdodgelooker: I think `before` & `after` need to duke it out with `suffix` & `prefix`, it's confusing/overkill to have both available.  If we decide to keep the latter, we could bury these in an internal-only base component.

packages/components/src/Form/Inputs/InputText/InputText.tsx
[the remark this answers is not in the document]
dbchristopher: ah yes ok I get it! yeah worth keeping for sure 👍

packages/components/src/Form/Inputs/InputText/InputText.test.tsx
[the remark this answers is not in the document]
dbchristopher: yeah I think that narrowing down to a few essential styles would be more informative. the test would communicate what aspects we actually care about for this

mdodgelooker:
@dbchristopher thanks for the great feedback! I think I've addressed everything except the question of `prefix`/`suffix` vs `before`/`after`. I'll chat with @lukelooker about it in the morning and then hit you up for another review.

mdodgelooker:
@dbchristopher `prefix` / `suffix` are now replaced with `before` / `after` and ready for you re-review! :)

CHANGELOG.md
```diff
@@ -81,7 +91,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `MenuList`, `MenuGroup` & `MenuItem` no longer supports `customizationProps`
 - `pseudo` support from design-tokens (prefer using `styled` when pseudo selectors are required)
   - Includes changes to `Box` & `IconButton`
-- Icons: `FieldDistance` removed
+    <<<<<<< HEAD
```
dbchristopher, line 94: looks like a merge conflict snuck in here?

mdodgelooker: Oopsie! Fixed.

packages/components/src/Form/Fields/Field.tsx
```diff
@@ -168,18 +174,26 @@ const fieldLabelCSS = (inline?: boolean) =>
...
   grid-template-areas: ${({ inline }) =>
     inline
       ? '"label input detail" ". messages messages"'
       : '"label detail" "input input" "messages messages"'};
   grid-template-columns: ${({ inline }) => (inline ? '150px 1fr' : undefined)};
   height: fit-content;
   justify-content: space-between;
-  width: ${({ width }) => width || 'fit-content'};
+  width: ${({ autoResize, width }) =>
+    width || autoResize ? 'fit-content' : '100%'};
```
dbchristopher, line 186: nice solution!

packages/components/src/Form/Inputs/InputProps.ts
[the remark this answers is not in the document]
dbchristopher: still want to delete these?

packages/components/src/Form/Inputs/InputText/InputText.test.tsx
```diff
@@ -84,6 +98,46 @@ test('InputText with an error validation', () => {
...
+    Array [
+      Array [
+        "Use before or iconBefore, but not both at the same time.",
+      ],
+      Array [
+        "Use after or iconAfter, but not both at the same time.",
+      ],
+    ]
+  `)
+})
```
dbchristopher, line 139: nice 👍

packages/components/src/Form/Inputs/InputProps.ts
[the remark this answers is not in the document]
mdodgelooker: Oops, yes. Just deleted those 2 and `time`.

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -26,121 +26,195 @@
...
+      </InputTextContent>
+    )
+
+    const afterToUse = iconAfterOrSuffix ? (
+      <>
+        {iconAfterOrSuffix}
+        {validationIcon}
+      </>
+    ) : (
+      after || validationIcon
```
dbchristopher, line 180: am I reading this right that it will never display the validation icon if they pass a react element into the `after` prop? is that the correct behavior?

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
```diff
@@ -532,6 +526,24 @@ const InputTimeInternal = forwardRef(
...
+  line-height: ${({ theme }) => theme.lineHeights.medium};
+  padding: 0;
+  text-align: center;
+  width: 1.25rem;
+
+  &:focus {
+    background: ${({ theme }) => theme.colors.keyAccent};
+  }
+`
+
```
dbchristopher, line 546: 👍  thanks for setting that up

packages/components/src/Form/Inputs/InputText/InputText.tsx
dbchristopher: ah yep -- here's the error state when `after="%"`:

Screen Shot 2020-06-25 at 12 06 51 PM

and here it is when `after={<span>%</span>}`
Screen Shot 2020-06-25 at 12 06 33 PM

I feel like we should assume there would be room for both, or that the developers wouldn't be passing in a custom error icon into `after`. What do you think?

mdodgelooker: The main reason I did this is because `AdvancedInputControls` is now using `after` and it has its own error icon (to the left of the other icons) – which kind of demonstrates that we probably don't want to force a placement of the error icon, and that someone might even want to just replace it with something more custom (perhaps with a tooltip version). 

Another route we could go is to also accept a function on `after` that receives the error icon JSX so you can plop it where you want it? Or to keep it simple, just always put the error icon before the `after` (with Jared's blessing) and remove it from `AdvancedInputControls`. Thoughts?

dbchristopher: ah ok got it. It's probably reasonable to force them to take full responsibility for that if they want to start passing custom components there

PR #1111, opened 2020-06-23
Theme color adjustments

jhardy:
### :sparkles: Changes

Draft PR of some color tweaks to the theme. @lukelooker want to get your eyes on this. Most a few changes to how color is applied at darker background luminosities, found a few places where the theme was not applied quite right as well. I also changed our interactive colors to lighten instead of darken, because it performs a bit more consistently and that is where we are headed with material influences. Code can probably be cleaned up a bit more, but wanted to get your initial take. 

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


.node-version
```diff
@@ -1 +1 @@
-v12
+12.0.0
```
jhardy, line 1: I had to change this to get it to work on my remote instance, not sure what implications this has :D

packages/design-tokens/src/utils/color/blend.ts
```diff
@@ -49,6 +50,30 @@ export const mixColors = (
...
+    adjustment = lumnisanceScale.lower
+  } else if (colorLuminance < 0.08) {
+    adjustment = lumnisanceScale.lowest
+  }
+
+  const mixAdjustment =
+    colorLuminance > 0.3 ? mixAmount : scaleMixAmount(mixAmount, adjustment)
+  return mix(mixAdjustment / 100, foreground, background)
+}
+
```
jhardy, line 76: this function scales a color base luminance of the background color

.node-version
lukelooker: Weird. Yeah, we should change this because if someone has a newer version they'll get broken (I use 12.13.0 for instance)

jhardy:
@lukelooker I pushed around the color blending a bit and tried to document, what is going on a bit. Take a look and let me know what you think. I feel overall it works a lot better across various theming situations.

lukelooker:
Needs snapshot updates as well :)

packages/design-tokens/src/tokens/color/fallbacks.ts
```diff
@@ -73,24 +74,24 @@ export const fallbackStateful: StatefulColors = {
...
   criticalText: white,
   criticalBorder: red400,
 
   neutralSubtle: charcoal000,
   neutralAccent: charcoal100,
   neutralFocus: charcoal300,
-  neutralInteractive: charcoal500,
-  neutralPressed: charcoal600,
+  neutralInteractive: generateInteractive(charcoal500),
+  neutralPressed: generatePressed(charcoal500),
```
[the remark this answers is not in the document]
lukelooker: Change here is intentional. Ignore me :)

PR #1136, opened 2020-06-26
MenuItem fixes

lukelooker:
### :sparkles: Changes

- `MenuItem`
  - properly renders when `[aria-current='false']`
  - is now exported as Styled Component
  - renders properly in Safari / older Chrome implementations

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/Menu/MenuItemLayout.tsx
+++ b/packages/components/src/Menu/MenuItemLayout.tsx
@@ -40,7 +40,7 @@ export interface MenuListItemProps extends CompatibleHTMLProps<HTMLLIElement> {
  * All of this drama is to deal with SC's behavior of auto-spreading the Element interface
  * used when styled extends a base type. E.g. (styled.li has `color` prop)
  */
-const MenuItemLayoutInternal = forwardRef(
+const MenuItemWrapper = forwardRef(
   (props: MenuListItemProps, ref: Ref<HTMLLIElement>) => {
     return (
       <li {...omit(props, 'compact', 'focusVisible', 'hasIcon')} ref={ref} />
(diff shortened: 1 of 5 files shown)


packages/components/src/Menu/MenuItemLayout.tsx
```diff
@@ -111,7 +120,7 @@ export const MenuItemLayout = styled(MenuItemLayoutInternal)`
   }
 
   :hover,
-  &[aria-current] {
+  &[aria-current='true'] {
```
mdodgelooker, line 123:
```suggestion
  &[aria-current='page'] {
```

packages/components/src/Menu/MenuItemLayout.tsx
```diff
@@ -120,7 +129,7 @@ export const MenuItemLayout = styled(MenuItemLayoutInternal)`
     }
   }
 
-  &[aria-current] {
+  &[aria-current='true'] {
```
mdodgelooker, line 132:
```suggestion
  &[aria-current='page'] {
```

packages/components/src/Menu/MenuItemLayout.tsx
lukelooker: Doh! Great catch, reading on `aria-current` (https://www.digitala11y.com/aria-current-state/) I actually think `page` is the wrong key to use since `Menu` is often not used strictly for navigation. Switch to `true` as that seemed more apt.

Thoughts?

mdodgelooker: Great!

PR #1150, opened 2020-06-29
Allow for key-colored Link

lukelooker:
@jhardy While working on this I realized that we don't provide `:visited` and `:active` states. Any thoughts on that?

### :sparkles: Changes

- A few places in the product have links the leverage the `key` color and have to manually override styles

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/Link/Link.tsx
+++ b/packages/components/src/Link/Link.tsx
@@ -79,19 +79,13 @@ export const Link = styled(LinkLayout)`
     keyColor ? colors.key : colors.link};
   text-decoration: ${({ underline }) => (underline ? 'underline' : 'none')};
 
-  &:visited {
-    /* @TODO - We should probably support this */
-  }
-
   &:focus,
-  &:hover {
-    color: ${({ keyColor, theme: { colors } }) =>
-      keyColor ? colors.key : colors.link};
-    text-decoration: underline;
-  }
-
+  &:hover,
   &:active,
-  &.active {
-    /* @TODO - We should probably support this */
+  &.active,
+  &:visited {
+    color: ${({ keyColor, theme: { colors } }) =>
+      keyColor ? colors.keyInteractive : colors.linkInteractive};
+    text-decoration: underline;
   }
 `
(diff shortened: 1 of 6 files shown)


packages/components/src/Link/Link.tsx
```diff
@@ -26,35 +26,72 @@
...
+      keyColor ? colors.key : colors.link};
     text-decoration: underline;
   }
-`
 
-Link.defaultProps = {
-  color: 'link',
-  textDecoration: 'none',
-}
+  &:active,
```
jhardy, line 93: I think for now we can bundle active with the focus and hover

packages/components/src/Link/Link.tsx
```diff
@@ -26,35 +26,72 @@
...
   ${reset}
-  ${color}
   ${typography}
-  ${textDecoration}
+
+  color: ${({ keyColor, theme: { colors } }) =>
+    keyColor ? colors.key : colors.link};
+  text-decoration: ${({ underline }) => (underline ? 'underline' : 'none')};
+
+  &:visited {
```
jhardy, line 82: I think for now we can skip the visited state and let us look at it on the design side how the plays out across the system

packages/components/src/Link/Link.tsx
jhardy: @lukelooker, the only other thought I had is we can adjust the colors a bit here, would need to see how that plays with the theme. Do you think it makes sense to generate link colors with the theme or should we just pull in some of the color functions here and adjust in the file?

lukelooker: We should probably generate on the theme since I could see a world where someone has a collection of links on a page and having each do color math seems awfully expensive.

lukelooker: Ok, I went with the "cheap" route for now and just assigned `blue700` to `linkInteractive` in the theme. Once your PR with other color math improvements land we can generate that value intelligently.

packages/components/src/Link/Link.tsx
lukelooker: I went ahead and grouped `visited` with the other states so a pre-existing stylesheet (or browser default) doesn't cause trouble until we make an explicit choice.

packages/components/src/Link/Link.tsx
jhardy: 👍

PR #1155, opened 2020-06-30
Tooltip: add Portal

mdodgelooker:
### :sparkles: Changes

- Tooltip being rendered next to its trigger keeps causing issues: unwanted margin in `Space`/`SpaceVertical` and being blocked by `Menu`/`Popover`
- The only reason I can remember to _not_ use `Portal` for `Tooltip` was that it might be awkward when scrolling, but PopperJS seems to handle this well

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

#### Issue:
tooltip-portal-issue
#### Fixed:
tooltip-portal
#### It's not weird when you scroll:
scrolling

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED]
 
+### Changed
+
+- `Tooltip` now renders in a `Portal`
+
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


mdodgelooker:
Oh I guess one major difference in behavior would be if the `Tooltip` trigger's parent moves or transforms – with `Portal`, it won't automatically follow the trigger. I'll come back to this and create some stories to test after the next cut.

PR #1157, opened 2020-07-01
Focus Trap: fix return focus & Popover > Dialog > Select

mdodgelooker:
### :sparkles: Changes

- This started off with the goal of getting the "return focus to trigger on Dialog close" behavior working again
- In the meantime, another bug presented itself where you can't open pick from (or sometimes even open) a `Select` rendered in a `Dialog` that was opened from a (still open) `Popover` – instead that original `Popover` inappropriately regains focus.

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

#### Select bug:
overflow-edit-bug

#### Fixed:
overflow-edit-bug-fixed

### DIFF
--- a/packages/components/src/Popover/Popover.tsx
+++ b/packages/components/src/Popover/Popover.tsx
@@ -429,8 +429,13 @@ export function usePopover({
         // need to manually disable any parent focus trap
         parentFocusTrapRef &&
           parentFocusTrapRef.current &&
-          parentFocusTrapRef.current.deactivate({ returnFocus: false })
+          parentFocusTrapRef.current.pause()
       }
+    } else if (!focusTrap) {
+      // need to manually re-enable any parent focus trap
+      parentFocusTrapRef &&
+        parentFocusTrapRef.current &&
+        parentFocusTrapRef.current.unpause()
     }
   }, [focusTrap, parentFocusTrapRef, isOpen, enableFocusTrap])
 
--- a/packages/components/src/utils/useFocusTrap.ts
+++ b/packages/components/src/utils/useFocusTrap.ts
@@ -52,29 +52,46 @@ export function useFocusTrap(
   const { value, setOn, setOff } = useToggle(enabled)
 
   useEffect(() => {
+    function removeTrap() {
+      trapRef.current &&
+        trapRef.current.deactivate({ returnFocus: checkFocusLost() })
+      trapRef.current = undefined
+      enableFocusTrap && enableFocusTrap()
+    }
     if (element && value) {
       const autoFocusElement = element.querySelector(
         '[data-autofocus="true"]'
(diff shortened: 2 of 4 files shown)


packages/components/src/Popover/Popover.tsx
```diff
@@ -429,8 +429,13 @@ export function usePopover({
         // need to manually disable any parent focus trap
         parentFocusTrapRef &&
           parentFocusTrapRef.current &&
-          parentFocusTrapRef.current.deactivate({ returnFocus: false })
+          parentFocusTrapRef.current.pause()
```
mdodgelooker, line 432: `Select` uses a `focusTrap: false` version of `usePopover` so this logic is needed to disable any parent traps so that the options an be interacted with. Calling `deactivate` was triggering `useFocusTrap` to re-enable the grandparent trap. Calling `pause` instead avoids the latter.

packages/components/src/Popover/Popover.tsx
```diff
@@ -429,8 +429,13 @@ export function usePopover({
...
       }
+    } else if (!focusTrap) {
+      // need to manually re-enable any parent focus trap
+      parentFocusTrapRef &&
+        parentFocusTrapRef.current &&
+        parentFocusTrapRef.current.unpause()
```
mdodgelooker, line 438: What is paused must be unpaused. :)

packages/components/src/utils/useFocusTrap.ts
```diff
@@ -52,29 +52,46 @@ export function useFocusTrap(
...
       ) as HTMLElement
-      trapRef.current = createFocusTrap(element, {
-        clickOutsideDeactivates: true,
-        escapeDeactivates: false,
-        fallbackFocus: element,
-        onDeactivate: () => setOff(),
-        ...(autoFocusElement ? { initialFocus: autoFocusElement } : {}),
-      })
+      if (trapRef.current) {
+        trapRef.current.unpause()
```
mdodgelooker, line 66: Unpausing if the trap has already been paused by a superseding `Dialog` or `Popover` (see comment on line 81 below).

packages/components/src/utils/useFocusTrap.ts
```diff
@@ -52,29 +52,46 @@ export function useFocusTrap(
...
+        trapRef.current && trapRef.current.pause()
+      } else {
+        removeTrap()
+      }
     }
 
     return () => {
-      trapRef.current &&
-        trapRef.current.deactivate({ returnFocus: checkFocusLost() })
+      if (!element || document.body.compareDocumentPosition(element) !== 20) {
```
mdodgelooker, line 92: If the element's parent is unmounted, it can still be defined but will not be contained in `body`.

PR #1185, opened 2020-07-07
[LENS-912] MessageBar: Change canDismiss to default true

dbchristopher:
### :sparkles: Changes

- Sets canDismiss to true by default, and adds internal state tracking to allow clearing the messages out-of-the-box
- Adds a new `visible` prop to support controlled component pattern
- Improves testing and removes snapshots!
- Updated documentation to reflect these changes

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots
message-bar


packages/components/src/MessageBar/MessageBar.test.tsx
```diff
@@ -26,45 +26,144 @@
...
-    <MessageBar intent="warn" id="test-message-bar">
-      Warn
-    </MessageBar>
-  )
-})
+describe('clear button functionality', () => {
+  test('clearable by default', () => {
+    const { getByText, queryByText } = renderWithTheme(
+      <MessageBar>Message text</MessageBar>
+    )
```
lukelooker, line 40: Semi-unrelated ask but can we wrap this entire suite in a `describe('MessageBar', () => ` block so that the test output is easier to read:

e.g.:

```
MessageBar
  - clearable by default
  - controlled component
  - canDismiss ...
  - intents
```

Current:

image

packages/components/src/MessageBar/MessageBar.tsx
```diff
@@ -45,41 +48,75 @@ export interface MessageBarProps
...
+              size="small"
+              label={`Dismiss ${getIntentLabel(intent)}`}
+              aria-hidden
+            />
+          )}
+        </div>
+      )
+    } else {
+      return null
+    }
```
ElliotTheHuman, line 118: This suggestion could definitely be unnecessary, but what are you thoughts on making the return statement a bit tighter. Something like this maybe...

```
const messageBar = (
      <div
        aria-live="polite"
        ref={ref}
        role="status"
        {...omitStyledProps(props)}
      >
        <Status intent={intent} />
        <MessageBarContent canDismiss={canDismiss}>
          {children}
        </MessageBarContent>
        {canDismiss && (
          <IconButton
            ml="auto"
            onClick={handleDismiss}
            icon="Close"
            size="small"
            label={`Dismiss ${getIntentLabel(intent)}`}
            aria-hidden
          />
        )}
      </div>
)

return visible ? messageBar : null;
```

www/src/documentation/components/content/message-bar.mdx
```diff
@@ -4,38 +4,43 @@ propsOf: MessageBar
 github: MessageBar
 ---
 
-The `<MessageBar />` component is used to alert the user with `warn`, `critical`, `info`, or `positive` messages, settable via the `intent` property.
+The `<MessageBar />` component is used to alert the user with `warn`, `critical`, `info`, or `positive` messages, settable via the `intent` property. If no intent is set, it will fall back to `inform` by default.
```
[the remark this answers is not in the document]
dbchristopher: oh good catch!

packages/components/src/MessageBar/MessageBar.test.tsx
dbchristopher: good call. wrapped now!

Screen Shot 2020-07-08 at 9 47 00 AM

packages/components/src/MessageBar/MessageBar.tsx
```diff
@@ -45,41 +48,75 @@ export interface MessageBarProps
...
    */
   intent?: MessageBarIntent
+  /**
+   * Render the `X` which allows the MessageBar to be dismissed
+   * @default: true
+   */
   canDismiss?: boolean
   onDismiss?: () => void
   className?: string
+  visible?: boolean
```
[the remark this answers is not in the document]
dbchristopher: good call!

packages/components/src/MessageBar/MessageBar.tsx
dbchristopher: nice, that works for me 👍

dbchristopher:
Ok thanks for the feedback @lukelooker and @ElliotTheHuman! Everything should be updated now and ready for another look

PR #1188, opened 2020-07-08
[LENS-911] MessageBar custom actions

dbchristopher:
### :sparkles: Changes

- Updates MessageBar to render optional action buttons in place of the standard dismiss button.
- Updates Button documentation to clarify color and style variants.

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots

Screen Shot 2020-07-08 at 4 12 59 PM


packages/components/src/MessageBar/MessageBar.tsx
```diff
@@ -64,6 +71,8 @@ export interface MessageBarProps
    */
   visible?: boolean
   className?: string
+  primaryButton?: ReactElement<ButtonProps>
+  secondaryButton?: ReactElement<ButtonProps>
```
dbchristopher, line 75: Props match the naming in [`ConfirmLayout`](https://github.com/looker-open-source/components/blob/master/packages/components/src/Dialog/Confirm/ConfirmLayout.tsx#L41-L48)

packages/components/src/MessageBar/MessageBar.tsx
```diff
@@ -63,6 +70,14 @@ export interface MessageBarProps
...
    */
   visible?: boolean
+  /**
+   * Replaces the dismiss button with a custom react element
+   */
+  primaryButton?: ReactElement<ButtonProps>
+  /**
+   * Optional secondary action to be rendered alongside `primaryButton`
+   */
+  secondaryButton?: ReactElement<ButtonProps>
```
dbchristopher, line 80: Props match similar API pattern in in [`ConfirmLayout`](https://github.com/looker-open-source/components/blob/master/packages/components/src/Dialog/Confirm/ConfirmLayout.tsx#L41-L48)

dbchristopher: I wanted to keep these props flexible as I'm sure our fellow developers will come up with all kinds of variants and edge cases for things they want to render in these positions. This allows for more powerful component composition patterns (and fewer necessary props) than if I was styling the buttons inside the component here

packages/components/src/MessageBar/MessageBar.tsx
```diff
@@ -96,6 +113,27 @@ const MessageBarLayout = forwardRef(
...
+        icon="Close"
+        size="small"
+        label={`Dismiss ${getIntentLabel(intent)}`}
+        aria-hidden
+      />
+    )
+
+    const actions = primaryButton ? (
+      <>
+        {secondaryButton && secondaryButton}
```
dbchristopher, line 130: it will only render the `secondaryButton` if there is also a `primaryButton` passed in

dbchristopher:
@ElliotTheHuman @lukelooker here are more updates to the MessageBar component to check out! @jhardy I'm also tagging you here in case you want to double check the layout. I know in the Figma file the action buttons did not have bold text, but I figured it would be good to keep them styled as standard `TransparentButtons`

www/src/documentation/components/actions/button.mdx
```diff
@@ -55,42 +34,43 @@ Use a transparent button as a tertiary action on a screen, they are often used a
 <ButtonTransparent>Transparent</ButtonTransparent>
 ```
 
-#### Transparent states
+## Colors
+
+Buttons come in three color variants: `key`, `critical`, and `neutral`.
```
dbchristopher, line 39: I noticed we were only documenting the `Critical` variant under the Colors section for some reason! I had to dig into the code to remind myself what the value was to render secondary buttons. So I took the opportunity to update and reorganize the Button documentation a bit.

www/src/documentation/components/content/message-bar.mdx
```diff
@@ -20,7 +20,31 @@ The `<MessageBar />` component is used to alert the user with `warn`, `critical`
...
+    <ButtonTransparent
+      onClick={() => alert('Secondary Action')}
+      color="neutral"
+    >
+      Secondary Action
+    </ButtonTransparent>
+  }
+>
+  Render some custom actions!
+</MessageBar>
```
dbchristopher, line 47: Is this API pattern leaving too much for consumers to have to style themselves for those action buttons? It's always a tough call to balance convenience and flexibility, and I think I feel good about leaning towards flexibility for this kind of stuff

jhardy: I'd personally vouch to restrict the API here. I think this is a case of "It should just work" where the developer does not have to worry about button type or color. Leaving it open like this, I feel we will end up with a lot of inconsistencies  with `MessageBar` across the system.   Some cases I can think of are: 

- Primary key color buttons on top of critical message bar, or any type of `MessageBar` in general. 
- Inconsistent icon use in the buttons in the `MessageBar`, someone `iconBefore` for primary and `iconAfter` for secondary. 
- Different size buttons in different `MessageBar`'s, or even in the same `MessageBar`

dbchristopher: Yeah I totally get that! And ultimately might be the right way to go. But it's actually this case that concerns me about restricting the API:

> Inconsistent icon use in the buttons in the MessageBar, someone iconBefore for primary and iconAfter for secondary.

I just totally imagine that after we deploy this they **will** request the ability to include icons, and then eventually some design edge case will pop up where they'll want to flip the icon position lol. Then we'll end up with 8 props to accomplish what we could have done with 2:

* primaryActionText
* onPrimaryActionClick
* primaryActionIcon
* primaryActionIconPosition
* secondaryActionText
* onSecondaryActionClick
* secondaryActionIcon
* secondaryActionIconPosition

So that would be kind of a mess to maintain, and it seems like in general we have a philosophy of allowing component composition. Wouldn't there be some design QA to make sure they're using the right button style before any new features deploy?

dbchristopher: And then maybe they'll want to do an IconButton with a tooltip instead or something, then that's even more to document and configure in the props!

Not that I'm totally against restricting the API for this. I'm just feeling conflicted about the right approach to take for this kind of stuff and I want to make sure we're happy with it in the long run

dbchristopher: Yeah that feels ok I think. The need for additional event listener props gives me pause though... Would we have to be concerned about them doing something like this?

```jsx
<MessageBar 
  primaryButton={<Button>Dismiss</Button>} 
  onPrimaryButtonClick={() => alert('action taken')}>
  ...
</MessageBar>
```

packages/components/src/MessageBar/MessageBar.tsx
```diff
@@ -96,6 +113,27 @@ const MessageBarLayout = forwardRef(
...
+        {primaryButton}
+      </>
+    ) : (
```
ElliotTheHuman, line 133: I think the JSX could just look like this, right?

```suggestion
    const actions = primaryButton ? (
      <>
        {secondaryButton}
        {primaryButton}
      </>
    ) : (
```

I could be tripping up, but I think the logic already prevents `secondaryButton` from ever existing without a `primaryButton` since no `primaryButton` means the `dismissButton` is rendered instead.

www/src/documentation/components/content/message-bar.mdx
lukelooker: Hm... that's a reasonable point. Thinking through the "happy path" - I think folks would need to do:

```
<MessageBar 
  primary="Dismiss"
  secondary="Unsend"
  secondaryOnClick={doSomeCallbackHere}>
  Your thing was sent somewhere...
</MessageBar>
```

I think we should use event bubbling to capture any click events on Buttons within the "action" area (for lack of better descriptor) and close the MessageBar - this will act as a default `onClick` behavior for the primary and secondary buttons. 

If the developer specifies a `primaryOnClick` and/or `secondaryOnClick` then those should be explicitly assigned to the buttons and the event bubbling will take care of collapsing the message bar. (I suppose if if the developer wants to prevent the `MessageBar` from collapsing in that case they could specify `stopPropagation` which seems fine if there's a need)

In the case where the developer specifies their own Button the `MessageBar` can still use event bubbling to collapse itself and the developer can specify onClick directly on their button. I suppose in that case the `secondaryOnClick` / `primaryOnClick` is duplicative - I suppose we could get clever and detect if they've specified a non-string + a `onClick` handler and throw a warning? Or we could get clever and inject a wrapper to leverage bubbling yet again?

packages/components/src/MessageBar/MessageBar.tsx
dbchristopher: > I think the logic already prevents secondaryButton from ever existing without a primaryButton since no primaryButton means the dismissButton is rendered instead.

Yep that's correct! I did the `secondaryButton && secondaryButton` just because I assumed Typescript would yell at me if I tried to render `secondaryButton` without doing a truthy check for it. But it seems totally cool with that so definitely worth committing your suggestion. Thanks!

www/src/documentation/components/content/message-bar.mdx
dbchristopher: Clever! OK cool I'll give that a shot 👍

jhardy: I am not 100% opposed to having a more flexible API here, but from the design systems side we are going to explicitly disallow any button customization here or additional pieces to appear where the actions go, so the props would not expand because we do not plan on supporting that level of customization or flexibility

I get concerned if we allow it arbitrary JSX we are going to end up with unsupported patterns that the component was not designed to handle that we have to go unwind or update our spec to handle instead of restricting the API to its actual usage right now.

dbchristopher:
@jhardy @lukelooker @ElliotTheHuman I tightened up the API and refactored things a bit, and this is ready for another look!

www/src/documentation/components/content/message-bar.mdx
```diff
@@ -15,12 +15,78 @@ The `<MessageBar />` component is used to alert the user with `warn`, `critical`
 </SpaceVertical>
 ```
 
-## canDismiss
+## Customizing the action buttons
```
dbchristopher, line 18: I think it makes sense to start by reading through the docs for examples of intended use. I like where it's at right now -- seems to strike a good balance between flexibility and ease-of-use. I ended up consolidating a couple of the props as part of tightening up the API. 

Instead of having a separate `canDismiss` prop to hide the dismiss button, you can just set `primaryAction` to false. Doing that removed the need to have a separate `onDismiss` callback, and instead users should tie into the `onPrimaryClick` for that event hook.

storybook/src/MessageBar.stories.tsx
```diff
@@ -71,6 +74,63 @@ export const Controlled: FC = () => {
...
+      >
+        Custom Primary Action from a string label
+      </MessageBar>
+      <MessageBar
+        intent="positive"
+        secondaryAction="Secondary Action"
+        onSecondaryClick={handlePrimaryClick}
+      >
+        Custom Secondary Action from a string label
+      </MessageBar>
```
dbchristopher, line 101: After playing around with this component a bit more, I realized it's not the end of the world if they define a `secondaryAction` without also defining a `primaryAction`. It will just render the secondary action next to the default Dismiss button which seems like a totally reasonable UI to me:

Screen Shot 2020-07-13 at 12 24 17 PM

packages/components/src/MessageBar/MessageBar.tsx
```diff
@@ -50,44 +59,138 @@ export interface MessageBarProps
...
+              id={id ? `${id}-iconButton` : undefined}
+              onClick={onClick}
+              icon="Close"
+              size="small"
+              label={`Dismiss ${getIntentLabel(intent)}`}
+            />
+          )
+        : NoopComponent
+  }
+}
```
dbchristopher, line 133: These getter functions handle all the complexity of allowing the polymorphic action props to accept such disparate types

packages/components/src/MessageBar/MessageBar.tsx
```diff
@@ -50,44 +59,138 @@ export interface MessageBarProps
...
+      return ({ onClick }: DefaultDismissButtonProps) => (
+        <ButtonTransparent onClick={onClick} color="neutral">
+          {secondaryAction}
+        </ButtonTransparent>
+      )
+    case 'object':
+      // custom react component
+      return () => secondaryAction
+    default:
+      return NoopComponent
```
dbchristopher, line 159: I fall back to `NoopComponent` rather than `null` in order to remove one more logic flow check in the rendering block. Keeps the markup nice and simple

dbchristopher:
lol

PR #1194, opened 2020-07-10
Remove default title generation from Icon SVG

jhardy:
### :sparkles: Changes

Removes the default title generation for the `Icon` SVG to prevent presentational icons from displaying their title attribute on hover. Icons that need a title can now use the `title` prop. 

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


jhardy:
@lukelooker and @dbchristopher wanted to get your take on this idea here. 

Since our icons are presentational it make sense to avoid putting a title attribute on them unless explicitly set using the `title` prop. 

I know this blows up the world a bit, as I think a lot of tests use the icon title attribute for hooks But my question is should they? This feels like the right way to suppress the title attribute, but am I being to much of a purest here?  

If this is unreasonable we can suppress the title attribute by adding the `pointer-events: none` css rule to the `Icon`'s svg, which might be more realistic for now.

dbchristopher:
Hmm... I guess I need to clarify the intent for this change! If the issue is that these titles add unhelpful noise to screen reader output, then yeah we should probably remove them and deal with the test fallout. But if it's purely just to prevent the alt-text showing up on hover, then I'd prefer going with the `pointer-events: none` solution

jhardy:
Since our icons are used presentationally, for example, if they are used in a menu they have a label with them or if they are in an icon button they have a label supplied. The label should be used to describe the action or intent of the icon. The actual title of the artwork is not helpful to screen readers as right now it's just the name of the SVG file and not representative of the actions its taking, so we end up with a lot of issues like this everywhere icons are used. 
image

lukelooker:
I think it's worth removing them on the basis that the behavior is confusing at best. This will have some downstream test impacts but those are often examples of "bad behavior" anyway (hooking to text that's not really intended to visible so it seems worthwhile to catch and fix those)

dbchristopher:
Yeah that all makes sense to me! Let's kill it.

lukelooker:
Fixed up the tests on this. Removing myself as reviewer since I had a lot of hand's on here :)

packages/components/src/Button/IconButton.test.tsx
```diff
@@ -96,17 +96,18 @@ test('IconButton accepts events', () => {
   )
 })
 
-test('IconButton is a square by default', () => {
-  const { getByTitle } = renderWithTheme(
-    <IconButton id="test-iconButton" label="Gear" icon="Gear" size="large" />
+xtest('IconButton is a square by default', () => {
```
jhardy, line 99: is `xtest` a thing? Seems like a typo but just want to verify.

lukelooker: It is a thing - it disables the test without actually deleting it - I couldn't nail down why I couldn't get this behave but I also kinda hate this test (really should be done via visual testing anyway) but I wanted to keep a record of the test around for the time-being.

packages/components/src/Status/Status.tsx
```diff
@@ -87,6 +87,7 @@ export const Status = forwardRef(
         color={intent}
         name={getIntentIcon(intent)}
         size={size}
+        title={getIntentLabel(intent)}
```
jhardy, line 90: I am torn here. I think in some cases this will appear next to a label, like a Message bar, where you do not want the title. Though this component is presentational it does convey intent, I wonder if the intent label will always match up with what is being conveyed. 

Are titles like 'Inform' or 'Warn'  or 'Success' actually helpful? Maybe we require a label to be passed here for accessibility purposes?  `accessibilityTitle: string | false`  ?

packages/icons/bin/icon_template.js
```diff
@@ -24,17 +24,13 @@
...
 function template(
   { template },
   opts,
   { imports, componentName, props, jsx, exports }
 ) {
   const titleOverride = opts.titleProp
-    ? `const title = props.title || '${startCase(
-        opts.state.componentName.replace('Svg', '')
-      )}'
+    ? `const title = props.title
```
jhardy, line 33: One more thing I think would be valuable is that if we are hiding the title, we should probably add `aria-hidden="true"` and removing from the accessibility tree. Any thoughts here @lukelooker ?

packages/components/src/Status/Status.tsx
lukelooker: Hm... so I think there's actually a reasonable argument that these are actually useful as is for offering context. 

Thinking about the MessageBar use case if I have a "Success" or "Error" Status is can add a lot of context for a screen reader. E.g.

"Success: The file was saved"

"Error: The file was saved" 

I think it might make sense to allow for an override but I still think it's better to say "something" in cases like this rather than silently remove the contextual information.

We could require a label but in some components, such as MessageBar again, it would be yet-another-req'd-prop that I suspect would get mis-set often (I'm seeing more and more instances of `<IconButton label='' />` for instance so again I'd go for the optional prop and go with a reasonable default.

packages/icons/bin/icon_template.js
lukelooker: Yep - I think this makes sense in a case where the title has been completely omitted. Not like we can offer any context anyway. I'll dig into a fix.

packages/components/src/Status/Status.tsx
jhardy: That makes sense and sounds like a reasonable default

lukelooker:
> This all looks good to me! Seems like it would be good to add a note about the `title` prop to the Icon documentation for improved accessibility when possible

Great point! Added accessibility info to `Icon` :)

www/src/documentation/components/content/icon.mdx
```diff
@@ -91,6 +91,10 @@ import { IconList } from './demos'
 </Space>
 ```
 
+### Accessibility
+
+When using an `Icon` we strongly encourage specification of a `title` attribute. If an `Icon` is used without a epxlicitly specified `title` attribute `aria-hidden="true"` will be added to hide the icon from screen readers and the assumption will be made that other efforts have been made to give the user context.
```
dbchristopher, line 96: spelling: `a epxlicitly`

www/src/documentation/components/content/icon.mdx
```diff
@@ -91,6 +91,10 @@ import { IconList } from './demos'
 </Space>
 ```
 
+### Accessibility
+
+When using an `Icon` we strongly encourage specification of a `title` attribute. If an `Icon` is used without an explicitely specified `title` attribute `aria-hidden="true"` will be added to hide the icon from screen readers. The assumption will be made that other efforts have been made to give the user context.
```
lukelooker, line 96:
```suggestion
When using an `Icon` we strongly encourage specification of a `title` attribute. If an `Icon` is used without an explicitly specified `title` attribute `aria-hidden="true"` will be added to hide the icon from screen readers. The assumption will be made that other efforts have been made to give the user context.
```

dbchristopher: lol got me too. that word is actually impossible to spell

PR #1197, opened 2020-07-11
MenuGroup: Fix iconPlaceholder bug for MenuGroups with only 1 MenuItem

ElliotTheHuman:
### :sparkles: Changes

- `MenuGroup` now checks parent `MenuItemContext`s for `renderIconPlaceholder`
  - Doing this means that if a `MenuGroup` is the child of a `MenuList` or another `MenuGroup`, it will enable `renderIconPlaceholder` for its children assuming its parent context is set to true

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots
Screen Shot 2020-07-10 at 5 48 49 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `iconSizes` style function includes `xxsmall` case
 
 ### Changed
+
 - `MenuGroup` labels improvements when using `compact`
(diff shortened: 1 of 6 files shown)


ElliotTheHuman:
> Would be great to add a test to verify that this behavior is working / doesn't regress.

Added some new test cases to `MenuGroup` and `MenuList`!

Note: I added `data-testid` to the iconPlaceholder `div`. My additional tests boil down to "Is that iconPlaceholder `div` present or not?"

PR #1212, opened 2020-07-13
Enable bundle size statistics capture

lukelooker:
### :sparkles: Changes

This PR adds an "analyze" script to the playground. It allows for production of a graph showing the dependency sizes of our core package (at least as it's imported into playground)

You can try it out via `yarn workspace playground analyze` (for unknown reasons it hangs when complete but a quick `ctrl-c` kills off the hanging task. 🤷 

You can then view the result produced in `packages/playground/dist/analyze.html`

image

This could use additional tooling and documentation in the future (ideally it'd be part of our CI run to produce base line numbers for future comparison) but this initial pass is just being used to establish a baseline number and begin the process of isolating some of our more expensive dependencies.

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### DIFF
--- a/playground/webpack.config.js
+++ b/playground/webpack.config.js
@@ -24,23 +24,23 @@
 
  */
 
-const path = require('path')
-const HtmlWebpackPlugin = require('html-webpack-plugin')
-const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin')
+/* eslint-disable @typescript-eslint/no-var-requires */
 
-const PATHS = {
-  app: path.join(__dirname, 'src/index.tsx'),
-}
+const path = require('path')
+const HtmlWebPackPlugin = require('html-webpack-plugin')
+const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin')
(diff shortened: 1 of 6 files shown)


playground/webpack.config.js
```diff
@@ -24,23 +24,23 @@
...
 module.exports = {
   devServer: {
-    index: 'index.html',
-    proxy: {
-      '/api': 'http://localhost:3001',
-    },
+    port: 3000,
+    // proxy: {
+    //   '/api': 'http://localhost:3001',
+    // },
```
ElliotTheHuman, line 40: Question, is helpful to keep this comment around? Otherwise we can probably just delete it, right?

dbchristopher:
Awesome, this is super handy! I had no idea react-hotkeys was on par with react-day-picker in bundle size. It's just used to add a few keyboard event listeners. Seems like it might be worth trying to refactor that dependency out altogether

Screen Shot 2020-07-13 at 3 32 28 PM

dbchristopher:
> Didn't notice my terminal crash after running yarn workspace playground analyze fwiw

Same here btw. No terminal crashing on my end

PR #1214, opened 2020-07-14
Package size reduction via dropping exports

lukelooker:
### :sparkles: Changes

- `Calendar`, `FieldDate`, `FieldDateRange`, `InputDate` & `InputDateRange` now require "deep" imports
  - E.g.: `Calendar` imports from `@looker/components/lib/Calendar`)
- Broke up `DateTimeFormat` into three separate top-level directories (`DateFormat` & `TimeFormat`)

Savings: 

image

#### Before

Parsed: 756kb
Compressed: 145kb

before removal

#### After

Parsed: 669kb
Compressed: 112kb

after removal

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


www/src/documentation/components/content/calendar.mdx
```diff
@@ -18,6 +18,13 @@ import { CalendarPropTable } from './demos'
 <Calendar />
 ```
 
+<MessageBar my="large" canDismiss={false}>
+  Note: Ca;emdar must be imported from its specific location rather than the
+  package root.
+</MessageBar>
+
+Import path: `@looker/components/lib/Calendar`
```
dbchristopher, line 26: I think we should at least move this to the top of the file for now. Similar to how with the Material UI documentation [always starts with an Import example](https://material-ui.com/api/accordion/):

Screen Shot 2020-07-14 at 11 00 07 AM

This is probably out of scope for this PR, but it would be nice if there was a button to click to automatically copy that to clipboard. And it would definitely be great to automate that somehow. Do you think we should add that to the documentation roadmap?

www/src/documentation/components/content/date-format.mdx
```diff
@@ -14,6 +15,13 @@ This is the sibling component of [`TimeFormat`](time-format), which does the sam
 <DateFormat />
 ```
 
+<MessageBar my="large" canDismiss={false}>
+  Note: DateFormat must be imported from its specific location rather than the
+  package root.
+</MessageBar>
+
+Import path: `@looker/components/lib/DateFormat`
```
dbchristopher, line 23: Similar feedback with `calendar` -- lets make sure this is the first thing they see

www/src/documentation/components/content/date-format.mdx
```diff
@@ -14,6 +15,13 @@ This is the sibling component of [`TimeFormat`](time-format), which does the sam
 <DateFormat />
 ```
 
+<MessageBar my="large" canDismiss={false}>
+  Note: DateFormat must be imported from its specific location rather than the
+  package root.
+</MessageBar>
```
dbchristopher, line 21: I wonder if we actually don't need these MessageBar notifications? Especially since we're moving to a flatter world where you're expected to deep import everything. If we just start putting import examples at the top of every doc then we won't have to call out which ones are exceptions

www/src/documentation/components/content/time-format.mdx
```diff
@@ -14,6 +15,13 @@ This is the sibling component of [`DateFormat`](date-format), which does the sam
 <TimeFormat />
 ```
 
+<MessageBar my="large" canDismiss={false}>
+  Note: TimeFormat must be imported from its specific location rather than the
+  package root.
+</MessageBar>
+
+Import path: `@looker/components/lib/TimeFormat`
```
dbchristopher, line 23: move to top

www/src/documentation/components/forms/fields.mdx
```diff
@@ -232,6 +233,13 @@ This component is composed of an InputColor and a Label. It also has a `ColorWhe
 
 This component is composed of an `InputDate` and a `Label`.
 
+<MessageBar my="large" canDismiss={false}>
+  Note: FieldDate must be imported from its specific location rather than the
+  package root.
+</MessageBar>
+
+Import path: `@looker/components/lib/FieldDate`
```
dbchristopher, line 241: move to top

www/src/documentation/components/forms/fields.mdx
```diff
@@ -252,10 +260,17 @@ This component is composed of an `InputDate` and a `Label`.
...
+<Props of="FieldDateRange" />
 
 This component is composed of an `InputDateRange` and a `Label`.
 
+<MessageBar my="large" canDismiss={false}>
+  Note: FieldDateRange must be imported from its specific location rather than
+  the package root.
+</MessageBar>
+
+Import path: `@looker/components/lib/FieldDateRange`
```
dbchristopher, line 272: move to top

www/src/documentation/components/forms/input-date-range.mdx
```diff
@@ -4,8 +4,17 @@ propsOf: InputDateRange
...
+import { MessageBar } from '@looker/components'
+
 `InputDateRange` provides an interface to select to a date range, either through the calendar interface or text input boxes.
 
+<MessageBar my="large" canDismiss={false}>
+  Note: InputDatRange must be imported from its specific location rather than
+  the package root.
+</MessageBar>
+
+Import path: `@looker/components/lib/InputDateRange`
```
dbchristopher, line 16: move to top

www/src/documentation/components/forms/input-date.mdx
```diff
@@ -6,6 +6,15 @@ github: Form/Inputs/InputDate/InputDate.tsx
...
 `InputDate` provides a text input and calendar to select a single date, which can be used as a form input or filter.
 
+import { MessageBar } from '@looker/components'
+
+<MessageBar my="large" canDismiss={false}>
+  Note: InputDate must be imported from its specific location rather than the
+  package root.
+</MessageBar>
+
+Import path: `@looker/components/lib/InputDate`
```
dbchristopher, line 16: move to top

storybook/src/Forms/Date.stories.tsx
```diff
@@ -27,20 +27,19 @@ import React, { FC, useState, SyntheticEvent } from 'react'
...
-  DateFormat,
   Popover,
-  Locales,
   FieldSelect,
   InputText,
   FieldSlider,
   Paragraph,
   PopoverContent,
 } from '@looker/components'
+import { Locales, LocaleCodes } from '@looker/components/src/utils/i18n'
```
dbchristopher, line 39: I just realized that we don't have an i18n page on our site. Do you think people are going to need to import these utilities, and do we need to document the import paths and what's available anywhere?

www/src/documentation/components/content/date-format.mdx
lukelooker: We'll actually be able to move "back" later on once _everything_ is flattened - then we'll recommend importing from the package and using tooling to "unwind" those so these are going to temporarily be outliers while we're in the process of moving.

storybook/src/Forms/Date.stories.tsx
lukelooker: No - I've never seen anyone else import them thus-far.

www/src/documentation/components/content/calendar.mdx
lukelooker: Makes sense to move it up.

lukelooker:
Given that the behavior of these components is somewhat "aberrant" for now (requiring an explicit import path) I'd like to keep the notices in place until we get back to a consistent pattern.

Moved all of the notices to the top of the file per your suggestion.

PR #1216, opened 2020-07-15
theme.colors.text1-5 restructured

lukelooker:
### :sparkles: Changes

  - `text1-5` now go from lightest to darkest to match `ui1-5`
  - Reduced number of steps:
    - `text0` is now `text5` (consolidated the former `text0` & `text1`)
    - `text6` is now `text1` (consolidated the former `text6` & `text5`)

image

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/design-tokens/src/system/color/blends.ts
+++ b/packages/design-tokens/src/system/color/blends.ts
@@ -61,8 +61,7 @@ export interface UIColors {
 
 export interface TextColors {
   /**
-   * TODO - Should be defined
-   * Used for: TBD
+   * Lowest contrast text in UI, disabled states, placeholder text, for non-essential text only
    * @default charcoal400
    */
   text1: string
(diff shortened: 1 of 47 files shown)


packages/design-tokens/src/system/color/blends.ts
```diff
@@ -61,21 +61,15 @@ export interface UIColors {
...
-   * @default charcoal900
-   */
-  text0: string
-  /**
-   * Primary Text
-   * Used for: Headers, Labels, Dashboard tile titles
-   * @default charcoal800
+   * TODO - Should be defined
+   * Used for: TBD
+   * @default charcoal400
```
lukelooker, line 66: @jhardy Can you define what the intended use for `text1` is in this updated mapping?

jhardy: Lowest contrast text in UI, disabled states, placeholder text, for non-essential text only

PR #1237, opened 2020-07-21
Remove theme.colors.palette

lukelooker:
### :sparkles: Changes

- Remove theme.palette.colors by default
- Export `palette` from `@looker/design-tokens` for legacy consumers
- No longer exports `prismTheme` (not intended for re-use)

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


mdodgelooker (approved):
LGTM!

This line isn't needed anymore, but also not harmful at all: https://github.com/looker-open-source/components/blob/master/storybook/src/Theme/Swatches.stories.tsx#L35

lukelooker:
> This line isn't needed anymore, but also not harmful at all: https://github.com/looker-open-source/components/blob/master/storybook/src/Theme/Swatches.stories.tsx#L35

Good catch! I'm going to do some work on color documentation so I'll address that within that work so we don't have to do another review cycle on this PR :)

PR #1238, opened 2020-07-22
update Tab - scroll left and right

laraharrow:
### :sparkles: Changes

- scroll left and right

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


packages/components/src/Tabs/TabList.tsx
```diff
@@ -55,25 +55,32 @@ const TabListLayout: FC<TabListProps> = ({
...
-      ${(props) => props.theme.space.xsmall};
+    padding: ${({ theme: { space } }) =>
+      `${space.none} ${space.medium} ${space.xsmall}`};
   }
 `
 
 export const TabList = styled(TabListLayout)`
   border-bottom: 1px solid ${(props) => props.theme.colors.ui2};
-  ${({ distribute }) => distribute && distributeTabsCSS}
+  overflow-x: scroll;
```
jhardy, line 82:
```suggestion
  overflow-x: auto;
```

Setting the overflow to scroll will always show scroll bars even when you can't scroll
image

packages/components/src/Tabs/TabList.tsx
```diff
@@ -55,25 +55,32 @@ const TabListLayout: FC<TabListProps> = ({
...
+
+const distributeCSS = css`
   display: grid;
-  grid-gap: ${(props) => props.theme.space.none};
-  grid-template-columns: repeat(
-    auto-fit,
-    minmax(${(props) => props.theme.space.xlarge}, auto)
-  );
+  grid-auto-flow: column;
+  grid-template-columns: repeat(auto-fit, max-content);
```
jhardy, line 71: This rule is invalid and seems like we can remove it. Might be worth a few more stories testing a few tab amounts to see how this grid works. Would be good to see a tab set with 2 tabs, 4, tabs, 6, tabs, 20 tabs 
image

laraharrow: that's interesting, it was removed and year everything seems to be working fine. As for different options if you go to storybook > Tabs > Basic and in there look under knobs you will see an example that you can play with labels and number of tabs.

laraharrow:
@lukelooker @jhardy could ether of you do a final review here?
Thanks!

PR #1239, opened 2020-07-22
[LENS-1073] Replace react-hotkeys with lighter weight hook

dbchristopher:
### :sparkles: Changes

- Updates MenuList, Surface, and OverlaySurface to use react-hotkeys-hook to wire up keyboard listeners. 

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots

This change removes a full megabyte of uncompressed javascript from our bundle

#### react-hotkeys:
Screen Shot 2020-07-22 at 2 53 13 PM

#### react-hotkeys-hook:
Screen Shot 2020-07-22 at 2 53 05 PM


packages/components/src/Menu/MenuList.tsx
```diff
@@ -97,27 +97,42 @@ export const MenuListInternal = forwardRef(
   ) => {
     const { id, isOpen, setOpen, triggerElement } = useContext(MenuContext)
 
+    // track inner focus state to prevent components from clobbering each
+    // other's keyboard listeners when there are multiple Menus on the page
+    const [menuHasFocus, setMenuHasFocus] = useState(false)
+
```
dbchristopher, line 103: React-hotkeys-hook is much more bare-bones than our previous dependency, and as a result all listeners are just attached to the global scope. We'll have to do a little extra gymnastics in cases where we want to contain the keyboard event to the focused component

packages/components/src/Overlay/OverlaySurface.tsx
```diff
@@ -77,6 +77,8 @@ export const OverlaySurface = forwardRef(
     } = props
     const { closeModal } = useContext(DialogContext)
 
+    useHotkeys('esc', closeModal)
```
dbchristopher, line 80: The nice side effect of these keyboard events being global is that it automatically addresses the issue where you had to focus within a popover before the esc key listener activates

dbchristopher: I just hope we never have a situation where there is a Popover inside a Dialog or something crazy like that. In that case, pressing `esc` would close everything

dbchristopher:
@mdodgelooker this is ready for a look whenever you have time!

packages/components/src/Menu/MenuList.tsx
```diff
@@ -97,27 +97,52 @@ export const MenuListInternal = forwardRef(
   ) => {
     const { id, isOpen, setOpen, triggerElement } = useContext(MenuContext)
 
+    /*
+     * track inner focus state to prevent components from clobbering each
+     * other's keyboard listeners when there are multiple Menus on the page
+     */
+    const [menuHasFocus, setMenuHasFocus] = useState(false)
```
dbchristopher, line 104: React-hotkeys-hook is much more bare-bones than our previous dependency, and as a result all listeners are just attached to the global scope. We'll have to do a little extra gymnastics in cases where we want to contain the keyboard event to the focused component

packages/components/src/Overlay/OverlaySurface.tsx
lukelooker: So we actually have a lot of these because of `Select`. 

I tested against the Select + Dialog example in Storybook (`/?path=/story/forms-select--select-demo`) and the `esc` is definitely a little over-aggressive.

packages/components/src/Menu/MenuList.tsx
```diff
@@ -97,27 +97,52 @@ export const MenuListInternal = forwardRef(
...
+    }
+
+    const handleBlur = () => {
+      /*
+       * delay focus change to prevent state from updating multiple times as
+       * the event bubbles up the dom
+       */
+      clearTimeout(focusId.current)
+      focusId.current = setTimeout(() => setMenuHasFocus(false), 0)
+    }
```
mdodgelooker, line 141: What you have totally works, but you could get away with something simpler too:
```suggestion
    function handleArrow(e: KeyboardEvent, direction: number, initial: number) {
      if (
        wrapperRef.current &&
        wrapperRef.current.contains(document.activeElement)
      ) {
        moveFocus(direction, initial, wrapperRef)
      }
    }

    useHotkeys('down', () => handleArrow(e, 1, 0))
    useHotkeys('up', () => handleArrow(e, -1, -1))
```

packages/components/src/Menu/MenuList.tsx
```diff
@@ -97,27 +97,52 @@ export const MenuListInternal = forwardRef(
...
-        <HotKeys
-          innerRef={innerRef}
-          keyMap={{ MOVE_DOWN: 'down', MOVE_UP: 'up' }}
-          handlers={{
-            MOVE_DOWN: () => moveFocus(1, 0, innerRef),
-            MOVE_UP: () => moveFocus(-1, -1, innerRef),
-          }}
-          style={{ borderRadius: 'inherit' }}
-        >
+        <div ref={wrapperRef} onFocus={handleFocus} onBlur={handleBlur}>
```
mdodgelooker, line 145: Again, not urgent, but you could get rid of this div by merging the refs, something like:
```
const ref = useForkedRef(forwardedRef, wrapperRef)
```

packages/components/src/Overlay/OverlaySurface.tsx
mdodgelooker: Something like this might work (as well as something similar on the Dialog side):
```suggestion    
    const wrapperRef = useRef<HTMLElement>(null)
    const forkedRef = useForkedRef(ref, wrapperRef)

    function handleEsc(e: KeyboardEvent) {
      if (
        wrapperRef.current &&
        wrapperRef.current.contains(document.activeElement)
      ) {
        closeModal()
        e.stopPropagation()
      }
    }
    useHotkeys('esc', handleEsc)
```

dbchristopher:
ok cool, that totally makes sense about needing to preserve focus context for those keyboard shortcuts. Thanks for the suggestions @mdodgelooker I'll play around with that a bit!

packages/components/src/Menu/MenuList.tsx
dbchristopher: wow that's such a better solution 😍

dbchristopher:
@mdodgelooker `e.stopPropagation()` doesn't seem to be making a difference for those nested event listeners. My theory is that it's because react uses event delegation, so all events are just attached to the window rather than individual dom elements. I don't think there actually is any bubbling for it to cancel 🤔  

I'll have to think on this a bit, and I hope an idea presents itself in the morning

mdodgelooker:
Oh shoot @dbchristopher that's right. Good luck! :)

dbchristopher:
@lukelooker @mdodgelooker I need to do some cleanup/comments/refactoring, but I think I've finally landed on a final strategy and I wanted to update you on that. 

Basically, it's really hard to handle this on a global level in a way that handles the specific edge cases well. I think 99% of the time, especially when the keyboard event doesn't include modifier keys, we should just use the component level onKeyDown/onKeyUp props. That way react will automatically deal with the event bubbling and focus issues. I updated the Menu components to reflect that, and it works really well. 

For handling multiple layered focus traps, I set up a `useGlobalHotkey` helper that will manage an "event stack" of sorts, where it will keep track of the multiple listeners for the same key command and only fire the callback at the top of the stack. You can see that running where I can press `esc` once to close the select dropdown and then press `esc` again to close the modal:

stacked-events

This is actually a big improvement over react-hotkeys, where I can never close the modal once a child element has been focused:

stacked-events-master

The UX actually seems really broken in master at the moment considering how users could get stranded in a single context like that, so I'm glad to have had a chance to dig into this!

packages/components/src/Menu/MenuList.tsx
```diff
@@ -93,33 +99,39 @@ export interface MenuListProps
...
+      handleArrowDown: (e: KeyboardEvent<HTMLLIElement>) => {
+        e.preventDefault()
+        handleArrowKey(1, 0)
+        return false
+      },
+      handleArrowUp: (e: KeyboardEvent<HTMLLIElement>) => {
+        e.preventDefault()
+        handleArrowKey(-1, -1)
+        return false
+      },
```
dbchristopher, line 125: the best solution for the menu was to elevate these callbacks to the context and then fire them on individual MenuItem components

packages/components/src/utils/useGlobalHotkeys.ts
```diff
@@ -0,0 +1,115 @@
...
+    const rect1 = ev1.target.getBoundingClientRect()
+    const rect2 = ev2.target.getBoundingClientRect()
+
+    if (!doRectsIntersect(rect1, rect2)) {
+      // no intersection
+      return 0
+    } else {
+      // elements intersect! sort by dom stacking order
+      const { x, y } = calculateIntersectionPoint(rect1, rect2)
+      const stackedElements = document.elementsFromPoint(x, y)
```
dbchristopher, line 78: this was a really cool method for reliably sorting events by visual stacking order. I can query the viewport at any pixel and get an array of how they are layered.

packages/components/src/Form/Inputs/Combobox/utils/useKeyDown.ts
```diff
@@ -172,7 +172,7 @@ export function useKeyDown() {
       }
       case 'Escape': {
         if (state !== ComboboxState.IDLE) {
-          transition && transition(ComboboxActionType.ESCAPE)
+          // transition && transition(ComboboxActionType.ESCAPE)
```
dbchristopher, line 175: @mdodgelooker do you know if anything happens to the combobox on `ESCAPE` other than just closing it? This was superseding the escape listener on OverlaySurface, so I had to disable this event in order to get the modal + select to work together. So I'm not sure if it's cool to delete this altogether, or if it will need to be restored

packages/components/src/Form/Inputs/Combobox/utils/useKeyDown.ts
```diff
@@ -170,12 +170,6 @@ export function useKeyDown() {
         }
         break
       }
-      case 'Escape': {
-        if (state !== ComboboxState.IDLE) {
-          transition && transition(ComboboxActionType.ESCAPE)
-        }
-        break
-      }
```
dbchristopher, line 178: Removed to prevent it from colliding with the OverlaySurface event listener

mdodgelooker: I think it's probably ok to remove this as long as your changes make it so that the list always closes on `Escape` (I think this wasn't working before because focus was on the `input` rather than inside the `Popover`). This line does trigger some other behavior, like reverting to the existing option value if the user has changed the input value but not actually selected a new option. But this should also be triggered by the `Popover` – I'll test and make sure once your ready for the next round. :)

dbchristopher: Cool! Yeah I think I found and fixed those issues today

dbchristopher:
@mdodgelooker @lukelooker this is ready for another look! 

I spent some time trying to figure out if there was a good way to test the useGlobalHotkeys hook, but couldn't come up with much. So I think I'm going to leave it as is, considering we haven't really bothered to test other hooks in isolation.

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Reduced number of steps:
     - `text0` is now `text5` (consolidated the former `text0` & `text1`)
     - `text6` is now `text1` (consolidated the former `text6` & `text5`)
+- Refactor out react-hotkeys and improve keyboard ux for multiple stacked focus traps (effects `MenuList`, `Surface`, and `OverlaySurface`)
```
mdodgelooker, line 46: Got a new [UNRELEASED] section now :)

dbchristopher:
@mdodgelooker 😱 yep I'm seeing that too! investigating...

packages/components/src/utils/useGlobalHotkeys.ts
```diff
@@ -155,7 +155,12 @@ export const useGlobalHotkeys = (
     executeFirstKeyCommand(e, orderedEventListeners)
   }
 
-  // `enableOnTags` enables listener on form inputs
-  // for use in searchable Select and MultiSelect variants
-  useHotkeys(keyCommand, wrappedCb, { enableOnTags: ['INPUT'] })
```
dbchristopher: @mdodgelooker this `enableOnTags: ['INPUT']` setting is what broke the listener when the button was focused. I added this in because it wasn't dispatching the `esc` event in a searchable combobox. The docs were a little misleading and I didn't realize that made it **only** allow events from an input element. It should be resolved now!

mdodgelooker:
A belated big ol' 👍 to this!

PR #1240, opened 2020-07-23
Fix Select/SelectMulti windowed options issues

mdodgelooker:
### :sparkles: Changes

- **Issue 1:** 
  1. Go to Storybook > Form > SelectMulti > 1k (windowed) options
  2. Type in "ab", then keyboard navigate down to "Anabelle" then delete the "b"
  3. Keyboard nav now appears broken. "Anabelle" is still technically selected but since it is now outside the newly re-windowed options, the user can't tell, and they can't see anything happening with keyboard navigation.
  **Fix:** refine the logic for jumping to the currently selected option in `useWindowedOptions`.
- **Issue 2:** 
  1. Go to Storybook > Form > SelectMulti > 1k (windowed) options
  2. Type in "ab", then delete the "b"
  3. Re-type the "b", then keyboard navigate through the entire list. The selection moves around in unexpected ways. Because of `key={index}` and going from un-windowed to windowed to un-windowed again, some of the options are umnounted/re-mounted while others are simply re-rendered, which messes up the order used for keyboard nav. 
  **Fix:** make the key unique every time the options list changes.
- **Issue 3:** 
  1. Go to Storybook > Form > SelectMulti > 1k (windowed) options
  1. Focus to open the menu
  1. Resize the window vertically
  1. Empty white space is added to the bottom of the list even though there are more options available
  **Fix:** update the state the keeps track of the list height on window resize

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

#### Issue 1
issue-1

#### Issue 2
issue-2

#### Issue 3
issue-3


packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
```diff
@@ -221,9 +221,18 @@ const ComboboxListInternal = forwardRef(
         scrollListener()
       }
 
+      const resizeListener = throttle(() => {
+        if (contentContainer && setListClientRect) {
+          setListClientRect(contentContainer.getBoundingClientRect())
+        }
+      })
```
dbchristopher, line 228: you probably want to specify a `wait` value, yeah? otherwise throttle will be set to 0ms by default

dbchristopher:
I found another edge case when there are a bunch of selected options in a windowed instance. I don't think this needs to be a blocker for this pr, but I just wanted to give you a heads up.

I'm not sure why, but it seems that the more options you select, the slower the windowed scrolling becomes. It slows to the point that if I hold down the up or down arrow it will quickly lose the focused index altogether. I have to then mouse over the option list to reestablish the highlighted index and then I can scroll again. Maybe it would benefit from just a throttle on the keyboard listener?

selected-windowed-behavior

dbchristopher:
It is weird that it seems to be correlated to the amount of options selected. It seems to work perfectly when there is nothing selected. I'm not sure why the existence of selected options would slow down render performance that badly 🤔

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
```diff
@@ -122,23 +123,35 @@ SelectOptionGroupTitle.defaultProps = {
   variant: 'subdued',
 }
 
+function useOptionKeyPrefix(options?: SelectOptionProps[]) {
+  // Create a new unique prefix every time the list of options changes
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  return useMemo(() => uuid(), [options])
```
dbchristopher, line 129: really good call on useMemo here!

packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
mdodgelooker: Whoopsie, thank you for catching that!

mdodgelooker:
@dbchristopher great catch on the scroll performance issue! I'll see if it's a straightforward fix – if so, I'd love to get it in now.

mdodgelooker:
Shoot @dbchristopher actually I'm having trouble reproducing the scroll performance issue. I definitely want to keep trying, but save it for a future PR. Fixed the missing `wait` in `resizeListener`. :)

PR #1241, opened 2020-07-23
Fix SelectMulti freeInput + honor click outside

mdodgelooker:
### :sparkles: Changes

- Convention seems to say that a [_select_](https://material-ui.com/components/selects/) (no typing allowed) should stay focused after the click that dismisses the list, while a [_combobox_ / _autocomplete_](https://material-ui.com/components/autocomplete/) (typing allowed) should lose focus in this scenario.
- **Issue 1:** Because we don't currently follow this convention, there's a bug with `SelectMulti` > `freeInput` where you lose the input value on dismiss click, but the expected behavior is that the value should be tokenized (similar to `InputChips`).
- Updating to follow convention and allow the `Select`/`SelectMulti` input to lose focus (and whatever was clicked to receive that event / focus) on the list-dismissal click when typing is allowed (`isFilterable` or `freeInput`) fixes this bug.
- Added `cancelClickOutside` (default `true`) prop to `Popover`/`usePopover` to support this.
- **Issue 2** with `freeInput`, when you type into the input _then_ select from the list, but the value you typed and the option you clicked are added.

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

#### Issue 1:
issue-1

#### Issue 2:
issue-2

### DIFF
--- a/packages/components/src/Form/Inputs/Combobox/ComboboxContext.ts
+++ b/packages/components/src/Form/Inputs/Combobox/ComboboxContext.ts
@@ -77,11 +77,14 @@ export interface ComboboxContextProps<
   setListClientRect?: Dispatch<SetStateAction<DOMRect | undefined>>
 }
 
-export type ComboboxMultiContextProps = ComboboxContextProps<
-  ComboboxMultiData,
-  ComboboxMultiCallback,
-  ComboboxTransition<ComboboxMultiActionPayload>
->
+export interface ComboboxMultiContextProps
+  extends ComboboxContextProps<
+    ComboboxMultiData,
+    ComboboxMultiCallback,
+    ComboboxTransition<ComboboxMultiActionPayload>
+  > {
+  freeInputPropRef?: MutableRefObject<boolean>
+}
 
 export const defaultData: ComboboxData = {
   // the value the user has typed, we derived this also when the developer is
--- a/packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
+++ b/packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
@@ -53,7 +53,6 @@ import { ComboboxOptionIndicatorProps } from './ComboboxOptionIndicator'
 import { ComboboxContext, ComboboxMultiContext } from './ComboboxContext'
 import { useBlur } from './utils/useBlur'
 import { useKeyDown } from './utils/useKeyDown'
-import { ComboboxActionType } from './utils/state'
 
 export interface ComboboxListProps
   extends Pick<ComboboxOptionIndicatorProps, 'indicator'>,
@@ -128,7 +127,6 @@ const ComboboxListInternal = forwardRef(
       closeOnSelectPropRef,
       windowedOptionsPropRef,
       indicatorPropRef,
-      transition,
       wrapperElement,
       isVisible,
       optionsRef,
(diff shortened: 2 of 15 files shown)


dbchristopher:
As far as I can tell this is all working perfectly! So much nuance here  👍

packages/components/src/Form/Inputs/Combobox/ComboboxContext.ts
```diff
@@ -77,11 +77,14 @@ export interface ComboboxContextProps<
...
-  ComboboxMultiCallback,
-  ComboboxTransition<ComboboxMultiActionPayload>
->
+export interface ComboboxMultiContextProps
+  extends ComboboxContextProps<
+    ComboboxMultiData,
+    ComboboxMultiCallback,
+    ComboboxTransition<ComboboxMultiActionPayload>
+  > {
+  freeInputPropRef?: MutableRefObject<boolean>
```
mdodgelooker, line 86: `freeInput` needs to get stored in the context so it can be accessed not just from `ComboboxMultiInput` but also `ComboboxList` (multi). When either the input is blurred or the list is closed, we need to either clear the current input value (default behavior) or leave it to allow the underlying `InputChips` blur handler to tokenize it.

packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
```diff
@@ -177,12 +183,14 @@ const ComboboxListInternal = forwardRef(
 
     const setOpen = (isOpen: boolean) => {
       if (!isOpen) {
-        transition && transition(ComboboxActionType.BLUR)
+        // Without passing an event, this just handles state change required when closing the list
+        handleBlur()
```
mdodgelooker, line 187: `handleBlur` now contains the logic for how to handle `freeInput` vs default.

packages/components/src/Form/Inputs/Combobox/ComboboxMulti.tsx
```diff
@@ -118,7 +118,7 @@ export const ComboboxMultiInternal = forwardRef(
 
     const isVisible = useComboboxToggle(state, options, onOpen, onClose)
 
-    const { ref, ...commonRefs } = useComboboxRefs(forwardedRef)
+    const { ref, ...commonRefs } = useComboboxMultiRefs(forwardedRef)
```
mdodgelooker, line 121: New `useComboboxMultiRefs` contains `freeInputPropRef`, which is only applicable to `ComboboxMulti`, not `Combobox`.

packages/components/src/Form/Inputs/Combobox/ComboboxMultiInput.tsx
```diff
@@ -92,7 +92,7 @@ export const ComboboxMultiInputInternal = forwardRef(
       isVisible,
     } = useContext(ComboboxMultiContext)
 
-    useInputPropRefs(props, ComboboxMultiContext)
+    useInputMultiPropRefs(props)
```
mdodgelooker, line 95: Similar to my last comment – contains multi-specific `freeInputPropRef`.

mdodgelooker:
@dbchristopher I fixed a few things I had missed (had some funky logic causing a lot of tests to fail) – sorry to jump the gun but it should be all good to go now.

PR #1249, opened 2020-07-24
Avatar Improvements

lukelooker:
### :sparkles: Changes

- `AvatarCombo`, `AvatarIcon` & `AvatarUser` now supports `role="button"`
  - Added support for common DOM properties and event handlers (e.g.: `onClick`)
  - Improved a11y for all `Avatar*` components
  - Added Storybook with knobs

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


lukelooker:
> Looks great, the only thing I would suggest is to add this to the documentation?

Great point! Added :)

packages/components/src/Avatar/AvatarUser.tsx
```diff
@@ -38,24 +39,33 @@ export interface AvatarUserProps extends AvatarProps {
...
+    <BaseElement {...omitStyledProps(props)} aria-label={name}>
+      <AvatarInitials color={color} aria-hidden>
         {size === 'xxsmall'
           ? `${firstInitial}`
           : `${firstInitial}${lastInitial}`}
       </AvatarInitials>
       {user && user.avatar_url && (
-        <AvatarPhoto color={color} type="image/png" data={user.avatar_url} />
+        <AvatarPhoto
+          aria-hidden
```
jhardy, line 62: This should be `aria-hidden="true"` does react do that for us?

jhardy: Same on line 55 for `<AvatarInitials color={color} aria-hidden>`

lukelooker: Yep. (I had to double-check myself) 

image

PR #1275, opened 2020-07-29
Update color application to ButtonOutline and ButtonTransparent

jhardy:
### :sparkles: Changes

This PR adjusts how we apply color to our `ButtonOutline` and `ButtonTransparent` Colors. The biggest changes it to `ButtonOutline` so that it matches closer to the design spec and color is applied in a similar fashion to `ButtonTransparent`. This PR also adjusts the default neutral color so that our neutral buttons are now accessible. Before they were not passing contrast checks.

#### Before
image

#### After

image

- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


packages/components/src/Button/IconButton.tsx
```diff
@@ -237,7 +237,8 @@ export const IconButton = styled(IconButtonComponent)<IconButtonProps>`
 
   background: none;
   border: none;
-  color: ${({ theme, color = iconButtonDefaultColor }) => theme.colors[color]};
+  color: ${({ theme, color = iconButtonDefaultColor }) =>
+    color === 'neutral' ? theme.colors.ui4 : theme.colors[color]};
```
jhardy, line 241: @lukelooker I had to end up adjusting this slightly due to the switch of the neutral color. I could not find a better solution, open to any other ideas here.

lukelooker: I was playing with a way to use neutral and derive the color: 

```suggestion
  color: ${({ theme, color = iconButtonDefaultColor }) =>
    lighten(0.14, theme.colors[color])};
``` 

Gets pretty close but pushes around the `critical` / `key` versions a smidge.

image

I'm mixed on which approach is more ideal but I tend to prefer this. Today we don't expose `neutral` as an intent color but I could foresee an argument for doing so eventually perhaps.

Thoughts?

jhardy: Only downside I see here is if we lighten here, in the case of a dark background the icons can get a bit brighter. I can go either way in this case for now as long as we mark it as something to go back and investigate.

PR #1277, opened 2020-07-29
update MenuItem detail to be part of the clickable area

laraharrow:
### :sparkles: Changes

- detail is now part of MenuItem's clickable area 

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/Menu/MenuItem.tsx
+++ b/packages/components/src/Menu/MenuItem.tsx
@@ -133,13 +133,18 @@ const MenuItemInternal: FC<MenuItemProps> = (props) => {
       onKeyUp={handleOnKeyUp}
       className={className}
     >
-      <Component href={href} role="menuitem" target={target}>
+      <Component
+        href={href}
+        role="menuitem"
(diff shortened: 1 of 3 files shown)


packages/components/src/Menu/MenuItem.tsx
```diff
@@ -133,13 +133,18 @@ const MenuItemInternal: FC<MenuItemProps> = (props) => {
       onKeyUp={handleOnKeyUp}
       className={className}
     >
-      <Component href={href} role="menuitem" target={target}>
+      <Component
+        href={href}
+        role="menuitem"
+        style={{ display: 'flex' }}
```
dbchristopher, line 139: would it work for this style rule to be added to MenuItem instead?

laraharrow: the way I understand is that it needs to be added to the <a> tag otherwise it wont be clickable.

PR #1279, opened 2020-07-30
Action List: Control Bar

ElliotTheHuman:
### 🛠️ TODO
- Docs (possibly separate PR)

### :sparkles: Changes

- `useActionListSelectManager` hook added
  - Generates selected items state piece, onSelect, and onSelectAll for `ActionList` usage
- `ActionListControlBar` component added
- `ActionList` 
  - conditionally renders `ActionListControlBar` based on `bulk` prop
  - added `bulk` prop that accepts an object for bulk actions configuration

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots
730e46df-64c3-4f82-a121-7fd81e66ff8b


packages/components/src/ActionList/ActionListControlBar.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  // TODO: Figure out if there's better logic for the displayed items message
+  let message
+  if (pageCount) {
+    if (itemsSelected.length < pageCount) {
+      message = `${itemsSelected.length} of ${pageCount} displayed items selected`
+    } else if (itemsSelected.length === pageCount) {
+      message = `All ${pageCount} displayed items selected`
+    } else if (totalCount && itemsSelected.length === totalCount) {
+      message = `All ${totalCount} items selected`
```
lukelooker, line 69: Yeah - we should look at using `i18next` for this as it should also support proper pluralization. Not something we need to do as part of this PR but definitely something we could polish up a bit going that route.

packages/components/src/ActionList/utils/useActionListSelectManager.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import { useState } from 'react'
+
+export const useActionListSelectManager = (selectableItems: string[]) => {
+  const [selections, setSelections] = useState<string[]>([])
+
+  const onSelect = (selectionId: string) => {
+    // Note: In the event that selections includes the item being selected, we call filter only on selectableItems.
+    // This is to avoid the situation where you have non-displayed items selected but only some displayed items.
+    // Doing the above will mean you have selected items that cannot be unselected (i.e. there's no way to interact with non-displayed items).
```
lukelooker, line 35: Pure nit-pick... for multi-line comments use 

```
/* 
...
*/

OR: 

/**
  * Comment here....
  * style comments (prettier will auto-format these)
  **/

```

packages/components/src/ActionList/ActionListControlBar.tsx
```diff
@@ -0,0 +1,140 @@
...
+  onTotalClearAll,
+  onTotalSelectAll,
+  pageCount,
+  totalCount,
+}) => {
+  const { itemsSelected } = useContext(ActionListContext)
+  const bulkActionsButton = (
+    <Button iconAfter="ArrowDown" minWidth={bulkActionsButtonWidth}>
+      <Text fontSize="xsmall">Bulk Actions</Text>
+    </Button>
```
lukelooker, line 58: Why is the `minWidth` needed here?

Rather than `<Text>` inside the Button can you use `<Button size="xsmall">` (or whatever size correct)

packages/components/src/ActionList/ActionListControlBar.tsx
```diff
@@ -0,0 +1,140 @@
...
+        {selectedItemsText}
+        {selectTotalResultsButton}
+      </Space>
+    </div>
+  )
+}
+
+export const ActionListControlBar = styled(ActionListControlBarLayout)`
+  align-items: center;
+  background-color: ${({ theme }) => theme.colors.neutralSubtle};
```
lukelooker, line 125: `neutralSubtle` should _probably_ be `ui1` instead?

packages/components/src/ActionList/ActionListControlBar.tsx
```diff
@@ -0,0 +1,140 @@
...
+    // TODO: Export this width from Checkbox.tsx
+    const checkboxInputWidth = '1rem'
+    const actionListCheckboxWidth = '2.75rem'
+    return `calc((${actionListCheckboxWidth} - ${checkboxInputWidth}) / 2)`
+  }};
+  padding-right: ${() => {
+    const checkboxInputWidth = '1rem'
+    const actionListCheckboxWidth = '2.75rem'
+    return `calc((${actionListCheckboxWidth} - ${checkboxInputWidth}) / 2 + ${bulkActionsButtonWidth})`
+  }};
```
lukelooker, line 139: I'm not really understanding why this is needed. Seems like a complexity to figure out that the padding-x values are `1rem` 👍 

I think it'd be worth pursing a way to simplify this.

packages/components/src/ActionList/ActionListControlBar.tsx
```diff
@@ -0,0 +1,140 @@
...
+    <Text color="text2" fontSize="xsmall">
+      {message}
+    </Text>
+  )
+
+  !selectedItemsText &&
+    // eslint-disable-next-line no-console
+    console.warn(
+      'An ActionList received a bulk prop object without a pageCount property. As a result, the "displayed items selected" text within the ActionList control bar is disabled.'
+    )
```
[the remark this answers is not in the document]
ElliotTheHuman: I'm a big fan for making these required! I thought enforcing fewer requirements offered flexibility for the user that outweighed the checks.

Will go ahead and refactor those properties to be required.

packages/components/src/ActionList/utils/useActionListSelectManager.tsx
ElliotTheHuman: Noted!

packages/components/src/ActionList/ActionListControlBar.tsx
ElliotTheHuman: Resolved in 6262aaf!

packages/components/src/ActionList/utils/useActionListSelectManager.tsx
ElliotTheHuman: Resolved in 6262aaf!

packages/components/src/ActionList/ActionListControlBar.tsx
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+
+  const selectedItemsText = pageCount !== undefined && (
+    <Text color="text2" fontSize="xsmall">
+      {message}
+    </Text>
+  )
+
+  !selectedItemsText &&
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 6262aaf!

packages/components/src/ActionList/ActionListControlBar.tsx
```diff
@@ -0,0 +1,140 @@
...
+      message = `All ${pageCount} displayed items selected`
+    } else if (totalCount && itemsSelected.length === totalCount) {
+      message = `All ${totalCount} items selected`
+    }
+  }
+
+  const selectedItemsText = pageCount !== undefined && (
+    <Text color="text2" fontSize="xsmall">
+      {message}
+    </Text>
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 6262aaf!

packages/components/src/ActionList/ActionListControlBar.tsx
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 59381f9!

packages/components/src/ActionList/ActionListRow.tsx
```diff
@@ -43,6 +43,11 @@ export interface ActionListItemLayoutProps
    * @default true
    */
   supportsRaised?: boolean
+  /**
+   * If true, this row's background color will be the theme object's keySubtle color
+   * @default false
+   */
+  selected?: boolean
```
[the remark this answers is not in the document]
ElliotTheHuman: You've got it. The only difference is the visual treatment (i.e. `checked` affects whether the checkbox is checked or not, while `selected` affects whether the row's background is `color: focus` or not).

Based on the specs, the header should NOT take a colored background when checked. I wanted to tie the `color: focus` to the `checked` prop, but because both the header and the rows use `ActionListRow` elements, the header ends up adopting the colored background on checked with this method.

I could do something like this, which works! I'm just a afraid of `styled-components` here since the styling from `ActionListRow`'s default className and the className introduced by `ActionListHeader` don't seem to have a clear winner:

```
const ActionListHeader = styled(ActionListHeaderInternal)`
   ${ActionListRow} {
     background: transparent;
   }
`
```

Screen Shot 2020-08-03 at 9 45 19 AM

packages/components/src/ActionList/ActionListControlBar.tsx
ElliotTheHuman: I used `minWidth` here originally to guarantee a specific width for the button. I'm using the `bulkActionsButtonWidth` to help set the right padding of the actual control bar (which jankily "centers" the control bar text and select all non-hidden and hidden rows button).

~~In retrospect, I'm wondering if there's a cleaner way to achieve that "one element left aligned, one element centered, empty on the right"~~

As I was writing the above I realized using grid would probably be perfect here 🤦  Going to try that out.

packages/components/src/ActionList/ActionListControlBar.tsx
ElliotTheHuman: Yeah, in retrospect, I really should've used grid to achieve the layout I did rather than relying on super fragile padding math. This should be cleared up after my latest updates. Stay tuned!

packages/components/src/ActionList/ActionListRow.tsx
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 90606ae

packages/components/src/ActionList/ActionListControlBar.tsx
ElliotTheHuman: In pursuing the grid route, I _think_ I would need to use an explicit width for the button as well.

In talking to Meredith in #team-components, I think using `flexbox` to center the text + select all button and using `absolute` positioning to keep the "Bulk Actions" button from offsetting the aforementioned centering is the best solution here (could be wrong though, in which case, open to alternatives!)

packages/components/src/ActionList/ActionListControlBar.tsx
ElliotTheHuman: See separate comment:
```
In pursuing the grid route, I _think_ I would need to use an explicit width for the button as well.

In talking to Meredith in #team-components, I think using `flexbox` to center the text + select all button and using `absolute` positioning to keep the "Bulk Actions" button from offsetting the aforementioned centering is the best solution here (could be wrong though, in which case, open to alternatives!)
```

ElliotTheHuman: Resolved (maybe) in 95eac50

packages/components/src/ActionList/ActionListControlBar.tsx
ElliotTheHuman: Resolved (maybe) in 95eac50

packages/components/src/ActionList/ActionListControlBar.tsx
```diff
@@ -0,0 +1,117 @@
...
+  border-bottom: solid 1px ${(props) => props.theme.colors.ui2};
+  display: flex;
+  justify-content: center;
+  padding: ${({ theme }) => theme.space.small};
+  position: relative;
+
+  & > div:first-child {
+    left: ${({ theme }) => theme.space.small};
+    position: absolute;
+  }
```
lukelooker, line 116:
```suggestion
  ${BulkActionsButton} {
    left: ${({ theme }) => theme.space.small};
    position: absolute;
  }
```

packages/components/src/ActionList/ActionListControlBar.tsx
```diff
@@ -0,0 +1,117 @@
...
+  onTotalSelectAll,
+  pageCount,
+  totalCount,
+}) => {
+  const { itemsSelected } = useContext(ActionListContext)
+  const bulkActionsButton = (
+    <Button iconAfter="ArrowDown" size="xsmall">
+      Bulk Actions
+    </Button>
+  )
```
lukelooker, line 57: Extract this as `BulkActionsButton` with a SC wrapper so style can be attached to it (see below)

packages/components/src/ActionList/ActionListControlBar.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  return (
+    <div className={className}>
+      <div>
+        <Menu>
+          <MenuDisclosure>{bulkActionsButton}</MenuDisclosure>
+          <MenuList>{actions}</MenuList>
+        </Menu>
+      </div>
```
lukelooker, line 95: Let's remove the superfluous `<div>` here (see other comments on this file)

```suggestion
        <Menu>
          <MenuDisclosure>{bulkActionsButton}</MenuDisclosure>
          <MenuList>{actions}</MenuList>
        </Menu>
```

ElliotTheHuman: Resolved in 0f451b4!

packages/components/src/ActionList/ActionListControlBar.tsx
ElliotTheHuman: Resolved in 0f451b4!

packages/components/src/ActionList/ActionListControlBar.tsx
ElliotTheHuman: Resolved in 0f451b4!

PR #1282, opened 2020-07-31
IE11 Fixes

dbchristopher:
Fix specific issues IE11 issues:

-  Menu layout breaks when icons and details are used or text overflow container
- DialogContent not displayed


packages/components/src/Dialog/Layout/DialogContent.tsx
```diff
@@ -119,7 +119,7 @@ const Outer = styled.div<{ borderBottom?: boolean }>`
   ${({ borderBottom, theme: { colors } }) =>
     borderBottom && `border-bottom: 1px solid ${colors.ui2};`}
 
-  flex: 8;
+  flex: 1 1 auto;
```
dbchristopher, line 122: This was the simple fix for [LENS-1027 Dialogs don't have contents on IE11](https://looker.atlassian.net/browse/LENS-1027)

First off I think the original `flex: 8` was an error? I think it can only be 1 or 0.

Anyways, by setting` flex: 1` IE11 interpreted the intent that flex-basis should be 0, which caused the content to be hidden. Explicitly setting `flex-basis` to `auto` fixed the issue

playground/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "private": true,
   "scripts": {
-    "start": "webpack-dev-server --host=0.0.0.0 0 --disable-host-check",
+    "start": "webpack-dev-server --host=0.0.0.0 --disable-host-check",
```
dbchristopher, line 13: fixed the playground startup script

playground/src/template.html
```diff
@@ -7,6 +7,16 @@
...
+    <script>
+      const modernBrowser = 'fetch' in window && 'assign' in Object
+
+      if (!modernBrowser) {
+        const scriptElement = document.createElement('script')
+        scriptElement.async = false
+        scriptElement.src = '/polyfills.bundle.js'
+        document.head.appendChild(scriptElement)
+      }
+    </script>
```
dbchristopher, line 19: I'm not sure if this is the best way to load the polyfill packages in the playground app before the rest of the bundle loads. I copied this pattern from the [webpack shimming documentation](https://webpack.js.org/guides/shimming/#loading-polyfills)

dbchristopher:
@lukelooker I found one fix. I was in the middle of investigating [LENS-1109](https://looker.atlassian.net/browse/LENS-1109) but for some reason the browserstack IE11 stopped loading and I haven't been able to figure out why 😖.

I did find one potential lead on ticket that ticket though. The menu items have some grid-based layout, and the styled-components autoprefixer doesn't convert grid styles to the early `-ms-grid` equivalents. The reasoning is that there isn't an exact 1:1 relationship between the grid rules that Microsoft initially implemented and what made it into the final spec. 

The creators of Stylis say that they'd be more concerned about introducing unexpected side effects if they tried to automate that process. If IE11 grid layouts actually matter you should manually write it out using the old rules. 

Unfortunately I was never able to confirm if that's _actually_ causing the layout issues, but it seems like a good place to start.

storybook/package.json
```diff
@@ -22,6 +22,7 @@
     "@storybook/addon-a11y": "^5.3.19",
     "@storybook/addon-backgrounds": "^5.3.19",
     "@storybook/addon-knobs": "^5.3.19",
-    "@storybook/addon-storysource": "^5.3.19"
+    "@storybook/addon-storysource": "^5.3.19",
+    "babel-loader": "^8.1.0"
```
lukelooker, line 26: Not sure why I needed to add this to make Storybook happy again but 🤷

PR #1293, opened 2020-08-03
Tabs keyboard shortcut

laraharrow:
### :sparkles: Changes

- Tabs keyboard shortcut

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots

Screen Shot 2020-08-03 at 4 36 10 PM


lukelooker:
@jhardy - I think we can handle it in a follow-up but I wonder if we should look at an improved focus design (something like what you did for `Menu` as the blurry-edges that happen here aren't the most ideal :)

PR #1301, opened 2020-08-07
Fix InputChips/InputText layout issues

mdodgelooker:
### :sparkles: Changes

- `SelectMulti` & `InputChips`:
  - Enable vertical scrolling if a max-height is set.
  - The `white-space: nowrap` on `Chip` causes it to break out of the flex layout – fixed with `min-width: 0` (see screenshot 1) https://stackoverflow.com/questions/38223879/white-space-nowrap-breaks-flexbox-layout
- `InputSearch`, `Select`, `InputChips` & `SelectMulti`
  - with `autoResize` and a max-width that has been reached, the x `IconButton` becomes unclickable (blocked by the hidden text) – see screenshot 2

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

#### Screenshot 1
image-2020-07-31-15-12-03-402

### DIFF
--- a/packages/components/src/Form/Fields/Field.tsx
+++ b/packages/components/src/Form/Fields/Field.tsx
@@ -25,6 +25,7 @@
  */
 
 import React, { FunctionComponent, ReactNode, useContext } from 'react'
+import { width } from '@looker/design-tokens'
 import styled, { css } from 'styled-components'
 import { ResponsiveValue, TLengthStyledSystem } from 'styled-system'
 import omit from 'lodash/omit'
@@ -160,7 +161,10 @@ FieldDetail.defaultProps = {
   fontSize: 'xsmall',
 }
 
-const InputArea = styled.div``
+const InputArea = styled.div`
+  /* Workaround for Chip's truncate styling breaking flexbox layout in FieldChips */
+  min-width: 0;
+`
 const MessageArea = styled.div``
 
 const fieldLabelCSS = (inline?: boolean) =>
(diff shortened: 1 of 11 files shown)


packages/components/src/Form/Fields/Field.tsx
```diff
@@ -188,8 +192,8 @@ export const Field = styled(FieldLayout)<FieldPropsInternal>`
   grid-template-columns: ${({ inline }) => (inline ? '150px 1fr' : undefined)};
   height: fit-content;
   justify-content: space-between;
-  width: ${({ autoResize, width }) =>
-    width || autoResize ? 'fit-content' : '100%'};
+  width: ${({ autoResize }) => (autoResize ? 'fit-content' : '100%')};
+  ${width}
```
mdodgelooker, line 196: I can't think of a better way to make sure the width prop wins out but what was there before wasn't working (my bad).

packages/components/src/Form/Fields/Field.tsx
```diff
@@ -160,7 +161,10 @@ FieldDetail.defaultProps = {
   fontSize: 'xsmall',
 }
 
-const InputArea = styled.div``
+const InputArea = styled.div`
+  /* Workaround for Chip's truncate styling breaking flexbox layout in FieldChips */
+  min-width: 0;
```
mdodgelooker, line 166: Oddly, without this here as well as the one in `InputText` this issue shows up again in `FieldChips` and `FieldSelectMulti`.

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
```diff
@@ -178,6 +178,9 @@ export const InputChipsBase = styled(InputChipsBaseInternal)`
     align-content: flex-start;
     display: flex;
     flex-wrap: wrap;
+    /* Workaround for Chip's truncate styling breaking flexbox layout */
+    min-width: 0;
+    overflow-y: auto;
```
mdodgelooker, line 183: The comment only applies to `min-width`. The `overflow-y` is to allow scrolling if a max-height is set.

PR #1302, opened 2020-08-07
Fix Select popover-related issues

mdodgelooker:
#### `Select` scrollbar issue
When focus leaves the input, the handler in `useBlur` tries to determine if focus landed somewhere in the list. To do that it was using a ref on the `ul` but when you click on the scrollbar to drag it, focus lands on `OverlaySurface.Outer`. Since `ComboboxList` already has access to `popperInstanceRef`, I updated `listRef` to take the `elements.popper` value from there.

#### `SelectMulti` list height issue
`Popover` has an internal `useVerticalHeight` hook that keeps long popovers from going off the page (which would be bad, because of the scroll lock). It updates on window resize, but wasn't updating with the popover's position changed (via `popoverInstanceRef.update()`). I'm piggy-backing on the effects of that update – a changed value for `style.transform` from `usePopper` – to trigger an update in `useVerticalHeight`.

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `Select`/`SelectMulti` list closing when trying to scroll by dragging the scrollbar
+- `SelectMulti` list height now adjusts as needed when chips are added
 - `Tree` default color is text4 (was previously a different color due to browser button defaults)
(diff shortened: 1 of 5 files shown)


lukelooker:
To clarify my comment on this - I _think_ I saw good behavior on SelectMulti working with list height change but I'm not 100% sure I repro'd :)

PR #1305, opened 2020-08-07
InputChipsBase: fix click/mousedown/focus issues

mdodgelooker:
### :sparkles: Changes

- `InputChipsBase` (used in `InputChips` and `SelectMulti`) renders an `InputText`, which moves focus to the inner input just after mousedown
- This behavior disrupts clicking on other elements in the component, like a `Chip` (the whole thing, if a user wants to select a specific chip to delete, or copy its text, as well as the x icon), and the clear field x icon button.
- Some of the disruption can only be seen if the inner input is below the fold (see screen capture below) whereas trying to select a specific chip or copy its text just never works.
- The fix is to add `stopPropagation` on the `Chip` and clear field button, but then move focus to the inner input after the action is complete (where appropriate)

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots

d63df948-ffd3-4ba9-b4c1-d7c96dc946e1 (1)

### DIFF
--- a/packages/components/src/Form/Inputs/AdvancedInputControls.tsx
+++ b/packages/components/src/Form/Inputs/AdvancedInputControls.tsx
@@ -29,16 +29,17 @@ import flatMap from 'lodash/flatMap'
 import tail from 'lodash/tail'
 import compact from 'lodash/compact'
 import { Icon } from '../../Icon'
-import { InputSearchControls } from './InputSearch/InputSearchControls'
+import {
+  InputSearchControls,
+  InputSearchControlsProps,
+} from './InputSearch/InputSearchControls'
 
-interface AdvancedInputControlsProps {
+export interface AdvancedInputControlsProps
+  extends Omit<InputSearchControlsProps, 'height' | 'showClear'> {
   validationType?: 'error'
   renderSearchControls?: boolean
   isVisibleOptions?: boolean
   hasOptions?: boolean
-  disabled?: boolean
-  onClear: () => void
-  summary?: string
 }
 
 // inserts a divider line between each control element (item1 | item2 | item3)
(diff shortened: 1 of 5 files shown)


packages/components/src/Form/Inputs/AdvancedInputControls.tsx
```diff
@@ -29,16 +29,17 @@ import flatMap from 'lodash/flatMap'
...
 import { Icon } from '../../Icon'
-import { InputSearchControls } from './InputSearch/InputSearchControls'
+import {
+  InputSearchControls,
+  InputSearchControlsProps,
+} from './InputSearch/InputSearchControls'
 
-interface AdvancedInputControlsProps {
+export interface AdvancedInputControlsProps
+  extends Omit<InputSearchControlsProps, 'height' | 'showClear'> {
```
mdodgelooker, line 38: Needs to pass-thru `onMouseDown` to `InputSearchControls` but streamlining the props interface a bit while I'm at it.

packages/components/src/Form/Inputs/InputChips/InputChips.test.tsx
```diff
@@ -148,6 +148,18 @@ tag2`
...
+    const onChangeMock = jest.fn()
+    renderWithTheme(
+      <InputChips onChange={onChangeMock} values={['tag1', 'tag2']} />
+    )
+    const clear = screen.getByText('Clear Field')
+
+    fireEvent.click(clear)
+    expect(onChangeMock).toHaveBeenCalledTimes(1)
+    expect(onChangeMock).toHaveBeenCalledWith([])
+  })
```
mdodgelooker, line 161: Not related to the bug, but I realized this was an oversight in the test suite.

PR #1304, opened 2020-08-07
Select: make input value reflect change to option label

mdodgelooker:
### :sparkles: Changes

- Fixes this bug https://codesandbox.io/s/nifty-smoke-k232p?file=/src/index.tsx
- See storybook example at /?path=/story/forms-select--update-options

### :white_check_mark: Requirements

- [x] Includes test coverage for all changes
- [x] Build and tests are passing
- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] Checked for i18n impacts
- [ ] Checked for a11y impacts
- [x] PR is ideally < ~400LOC

PR #1328, opened 2020-08-12
Select: support empty string value

mdodgelooker:
### :sparkles: Changes

- Currently `Select` only respects an empty string as initial `value` or `defaultValue` if a `placeholder` is defined (and not also an empty string 😛) or if `isClearable` is true
  - There has been at least one use case in the core product that needed an empty initial value but neither `isClearable` nor `placeholder` were desirable. (`placeholder=" "` was the workaround used)
- Even with `isClearable` or `placeholder` select was not respecting a `value` change to `''` (see the new example in `Select.stories.tsx`)
  - The same use case in the core product that revealed this bug, also showed that the `name` prop wasn't getting passed down to the input.
- Also added support for empty string `value`/`defaultValue` and an option with _its_ value also empty (for example picking "Any option" from a dropdown filter)
- I found and fixed a related bug caused by https://github.com/looker-open-source/components/pull/1304 where in a `Select` with filtering behavior, if you pick an option and then start filtering, the input value reverts to the current value once that option has been "filtered out" (see screenshot below).

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots
560307b4-bafe-4bcb-9e5a-940a135a83c5

### DIFF
--- a/packages/components/src/Form/Inputs/Combobox/utils/getComboboxText.ts
+++ b/packages/components/src/Form/Inputs/Combobox/utils/getComboboxText.ts
@@ -30,7 +30,7 @@ export function getComboboxText(
   value?: string | ComboboxOptionObject,
   options?: ComboboxOptionObject[]
 ): string {
-  if (!value) return ''
+  if (value === undefined) return ''
   if (typeof value === 'string') {
     if (options && options.length > 0) {
       const currentOption = options.find((option) => option.value === value)
--- a/packages/components/src/Form/Inputs/Select/Select.test.tsx
+++ b/packages/components/src/Form/Inputs/Select/Select.test.tsx
@@ -29,7 +29,7 @@ import {
   renderWithTheme,
 } from '@looker/components-test-utils'
 import { cleanup, fireEvent, screen } from '@testing-library/react'
-import React, { useState } from 'react'
+import React, { useState, useMemo } from 'react'
 
 import { Button } from '../../../Button'
 import { ComboboxOptionIndicatorFunction } from '../Combobox'
@@ -540,11 +540,12 @@ describe('Select / SelectMulti', () => {
 })
 
 describe('Select', () => {
+  const options = [
+    { label: 'Foo', value: 'FOO' },
+    { label: 'Bar', value: 'BAR' },
+  ]
+
   test('value', () => {
-    const options = [
-      { label: 'Foo', value: 'FOO' },
-      { label: 'Bar', value: 'BAR' },
-    ]
     renderWithTheme(
       <Select
         options={options}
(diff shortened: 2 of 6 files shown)


packages/components/src/Form/Inputs/Combobox/utils/getComboboxText.ts
```diff
@@ -30,7 +30,7 @@ export function getComboboxText(
   value?: string | ComboboxOptionObject,
   options?: ComboboxOptionObject[]
 ): string {
-  if (!value) return ''
+  if (value === undefined) return ''
```
mdodgelooker, line 33: This change adds support for an option with an empty string as its value. (Before, if the value was an empty string, the method would return on this line and not even look for a matching option.)

packages/components/src/Form/Inputs/Select/Select.test.tsx
```diff
@@ -540,11 +540,12 @@ describe('Select / SelectMulti', () => {
 })
 
 describe('Select', () => {
+  const options = [
+    { label: 'Foo', value: 'FOO' },
+    { label: 'Bar', value: 'BAR' },
+  ]
```
mdodgelooker, line 546: Same sample options are used in most the tests, might as well just define them once.

packages/components/src/Form/Inputs/Select/Select.test.tsx
```diff
@@ -621,59 +632,118 @@ describe('Select', () => {
...
-          <Button onClick={() => setLabel('Updated label')}>
-            Update label
-          </Button>
-          <Select value="value_stays_the_same" options={options} />
-        </>
-      )
-    }
-    renderWithTheme(<Component />)
+  describe('updating options', () => {
+    test('displayed value changes when option label changes', () => {
```
mdodgelooker, line 676: I added this test in https://github.com/looker-open-source/components/pull/1304 so I'm adding the test for the bug caused by that change right after.

packages/components/src/Form/Inputs/Select/Select.test.tsx
```diff
@@ -676,4 +715,32 @@ describe('Select', () => {
     // Close popover to silence act() warning
     fireEvent.click(document)
   })
+
+  test('filtering out the selected option does not trigger a change', () => {
```
mdodgelooker, line 719: Testing the fix for the bug caused by https://github.com/looker-open-source/components/pull/1304

packages/components/src/Form/Inputs/Select/utils/options.ts
```diff
@@ -46,9 +46,11 @@ export function flattenOptions(options: SelectOptionProps[]) {
...
 export function getOption(value?: string, options?: SelectOptionProps[]) {
   const flattenedOptions = options && flattenOptions(options)
-  return value
-    ? { label: getComboboxText(value, flattenedOptions), value }
-    : undefined
+  const label = getComboboxText(value, flattenedOptions)
+  // If this is a filterable Select and the current option has been filtered out
+  // leave label out, so that the matching against the option saved in ComboboxContext won't fail
+  const labelProps = label ? { label } : {}
+  return value !== undefined ? { ...labelProps, value } : undefined
```
mdodgelooker, line 53: Fixes both:
- the issue of not respecting when the `value` prop is changed _to_ an empty string (by explicitly checking if it's undefined vs falsy), and...
- the bug caused by https://github.com/looker-open-source/components/pull/1304 by not adding an empty string as the label (caused by the current value no longer being present in the options b/c it was filtered out), which was making the [match check](https://github.com/looker-open-source/components/pull/1304/files#diff-ab2109d76bfdcc03b5a7f67c1837474eR126) fail

storybook/src/Forms/Select.stories.tsx
```diff
@@ -62,6 +63,7 @@ export const All = () => (
     <InlineError />
     <SelectDemo />
     <UpdateOptions />
+    <ClearValue />
```
dbchristopher, line 66: Maybe this should be called `EmptyValue`? `Clear` sounds more like an action than a description to me

mdodgelooker:
Good call @dbchristopher !

PR #1331, opened 2020-08-13
Select: option icons

mdodgelooker:
### :sparkles: Changes

- `Select` options can now receive an `icon` (mapping to `Icon`'s `name` or `artwork` properties)
- The icon displays to the left of the option label in both the list and the input, when selected

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots
image
image


packages/components/src/Form/Inputs/Combobox/ComboboxInput.tsx
```diff
@@ -66,7 +66,7 @@ export interface ComboboxInputCommonProps {
 }
 
 export interface ComboboxInputProps
-  extends Omit<InputTextBaseProps, 'autoComplete' | 'value' | 'defaultValue'>,
+  extends Omit<InputTextProps, 'autoComplete' | 'value' | 'defaultValue'>,
```
mdodgelooker, line 69: Need to use the `before` prop to display the selected option's icon.

packages/components/src/Form/Inputs/Select/Select.tsx
```diff
@@ -160,6 +161,7 @@ const SelectComponent = forwardRef(
       >
         <ComboboxInput
           {...ariaProps}
+          before={<SelectInputIcon options={options} />}
```
mdodgelooker, line 164: If option icons aren't used, this will render nothing.

packages/components/src/Form/Inputs/Select/SelectInputIcon.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export interface SelectInputIconProps {
+  options?: SelectOptionProps[]
+}
+
+export function SelectInputIcon({ options }: SelectInputIconProps) {
+  const {
+    data: { option, inputValue },
+  } = useContext(ComboboxContext)
+  if (!options || !option) return null
```
mdodgelooker, line 54: Getting the current option from `ComboboxContext` rather than a prop b/c if the `Select` is uncontrolled, the current value isn't available there.

packages/components/src/Form/Inputs/Select/SelectInputIcon.tsx
```diff
@@ -0,0 +1,74 @@
...
+    // Don't show the icon if the user is filtering
+    return null
+  }
+
+  const icon = option && getOptionIcon(option.value, options)
+  return (
+    <>
+      {icon && (
+        <Icon
+          {...getSelectOptionIconProps(icon)}
```
mdodgelooker, line 66: Detects whether to use `name` or `artwork`.

storybook/src/Forms/Select.stories.tsx
```diff
@@ -208,7 +210,7 @@ export function SelectContent() {
     return options1k.reduce(optionReducer(searchTerm), [])
   }, [searchTerm])
   return (
-    <Box p="large">
+    <>
```
mdodgelooker, line 213: Random cleanup (don't need that margin).

storybook/src/Forms/Select.stories.tsx
```diff
@@ -326,16 +328,14 @@ export const SelectDemo = () => {
   const handleClick = () => setOpen(true)
   const handleClose = () => setOpen(false)
   return (
-    <>
+    <SpaceVertical align="start">
```
mdodgelooker, line 331: Random cleanup (better spacing).

storybook/src/Forms/options1k.ts
```diff
@@ -385,8 +385,6 @@ export const options1k = [
   { label: 'Bonfire', value: 'Bonfire' },
   { label: 'Bonus', value: 'Bonus' },
   { label: 'BonWorth', value: 'BonWorth' },
-  { label: 'Boob', value: 'Boob' },
-  { label: 'BOOB-EEZ', value: 'BOOB-EEZ' },
```
mdodgelooker, line 389: Oops, gotta keep it classing. (Good catch @ElliotTheHuman )

jhardy:
The checkbox icon and the custom artwork are different sizes so they are not quite aligned. 
image

jhardy:
One thing I noticed, but having a hard time capturing, and might be worth fixing in another PR,  is that the select always has a cursor of type `text`, I think that makes sense on selects that you can type into, but in other cases ideally we should use the cursor type `pointer `

jhardy:
@mdodgelooker would it be possible for when using stock components icons to apply color like we do on `Menu`. [ Looks like it is using](https://github.com/looker-open-source/components/blob/master/packages/components/src/Menu/MenuItem.tsx#L143) `text1`

mdodgelooker:
Thanks @jhardy ! I made some updates (including fixing the cursor issue, thanks for catching that), but I just wanted to confirm I went the right route on these 2:
- To fix the alignment, I made the icons and option text all vertically center aligned inside respective containers of `theme.lineHeights.small` (= 1.25rem = 20px), but should I have instead just made the two icons the same size? (The selected check icon is `xsmall`, while the icon from the prop is `small`)
- The icon color (when not using a custom svg) is now `text1` – is it ok that it stays this color, whether the option is highlighted or not? The text and check icon turn to `theme.colors.key` when the option is highlighted. And actually now that I think of it, the text color for options with descriptions isn't changing when highlighted. Should it? image
image

storybook/src/Forms/options1k.ts
dbchristopher: stay classy

jhardy:
@mdodgelooker I think the icons should be the same size, the `small` should be good for both the custom icon and the check mark

I also think we should remove the key color from the text on the selected/hover state, which looks like it could be achieved by [taking this off here](https://github.com/looker-open-source/components/blob/master/packages/components/src/Form/Inputs/Combobox/ComboboxOption.tsx#L213)

I know this may affect other areas that use combobox but I can fall on that sword with the design team

mdodgelooker:
Sounds good @jhardy , those 2 changes are pushed!

PR #1336, opened 2020-08-14
Assorted Firefox layout bugs

dbchristopher:
https://looker.atlassian.net/browse/LENS-1110
https://looker.atlassian.net/browse/LENS-1047

### :sparkles: Changes

- Update Space gap for consistent rendering across browsers
- Fix issue where Select inputs would lose padding at small screen sizes
- Resolve Slider style inconsistencies by rendering a div rather than the vanilla slider input 

- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### :camera: Screenshots

Screen Shot 2020-08-14 at 10 11 34 AM

Screen Shot 2020-08-14 at 10 31 58 AM

Screen Shot 2020-08-14 at 10 33 01 AM

### DIFF
--- a/packages/components/src/Form/Inputs/InputText/InputText.tsx
+++ b/packages/components/src/Form/Inputs/InputText/InputText.tsx
@@ -198,18 +198,15 @@ const InputTextLayout = forwardRef(
       'aria-invalid': validationType === 'error' ? true : undefined,
       type,
     }
-    const input = <input {...inputProps} ref={ref} />
 
     const inner = children ? (
       // Support for rendering chips in InputChips and SelectMulti
       <div className="inner">
         {children}
-        {input}
+        {<InlineInputTextBase {...inputProps} ref={ref} />}
       </div>
-    ) : autoResize ? (
-      <InlineInputTextBase {...inputProps} ref={ref} />
     ) : (
-      input
+      <InlineInputTextBase {...inputProps} ref={ref} autoResize={autoResize} />
     )
 
     return (
--- a/packages/components/src/Form/Inputs/InputText/InputText.tsx
+++ b/packages/components/src/Form/Inputs/InputText/InputText.tsx
@@ -203,10 +203,12 @@ const InputTextLayout = forwardRef(
       // Support for rendering chips in InputChips and SelectMulti
       <div className="inner">
         {children}
-        {<InlineInputTextBase {...inputProps} ref={ref} />}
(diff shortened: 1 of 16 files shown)


packages/components/src/Layout/Space/Space.tsx
```diff
@@ -110,7 +110,7 @@ const fauxGap = ({ gap = defaultGap, reverse }: SpaceHelperProps) => css`
 
 const flexGap = ({ gap = defaultGap, reverse }: SpaceHelperProps) => css`
   @supports (-moz-appearance: none) {
-    gap: ${({ theme: { space } }) => space[gap]};
+    gap: 0 ${({ theme: { space } }) => space[gap]};
```
dbchristopher, line 113: I hope this doesn't cause other layout side effects. Previously Firefox was inserting extra vertical space when the horizontal Space component children would wrap to multiple lines

0365b0cd-2d87-45cb-ba1b-dc11e3ef7690

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -228,6 +227,15 @@ const InputTextLayout = forwardRef(
 
 InputTextLayout.displayName = 'InputComponent'
 
+const StyledInput = styled.input`
+  ${innerInputStyle}
+  flex: 1;
+  font-size: ${(props) => props.theme.fontSizes.small};
+  max-width: 100%;
+  min-width: 2rem;
```
dbchristopher, line 235: @mdodgelooker since you pointed out the issue isn't happening on autoResized select inputs, I was able to narrow down to this style difference. I'm actually not sure why setting a small min-width on the input would be related to that layout issue, but I'll take it 🤷 

min-width

dbchristopher:
@mdodgelooker this is ready if you have any time to look monday! Most of the line diffs are snapshot changes, so it should be a quick review hopefully 🤞

packages/components/src/Layout/Space/Space.tsx
mdodgelooker: Oh good catch! I think this is perfect b/c it more accurately matches the `fauxGap` style used in other browsers. My only thought would be "should this also be updated in `SpaceVertical`?" but it seems less applicable there – I can't really picture someone wanting to use a wrapping `SpaceVertical`.

PR #1337, opened 2020-08-15
Adding nix configuration support

jkaster:
### :sparkles: Changes

This PR adds support for `nix` configuration rather than using `nodenv`. This is currently a very lightweight `shell.nix` that only resolves the `node` version requirement.

The `.envrc` file 

Almost all of these are not applicable

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


jkaster:
Sure, I can add the documentation to the README. However, I'm not trying to make it the _preferred_ route, just a supported route. I think we may end up recommending Nix for configuration purposes in the future, but for right now I'm just trying to support those of us who may choose to use Nix.

lukelooker:
i'd _prefer_ not to add infrastructure that isn't either exercised by CI or used by everyday developers / users as breakage on these happens and doesn't get caught until far later. Is it possible to get this to utilize configuration files used by the build or more widely used developer workflows?

jkaster:
I'll close this PR and just use the nix configuration for my own needs

PR #1350, opened 2020-08-19
Build ES module version of each package

dntrkv:
I believe there was some work done in the past to export the packages as ES modules, but I don't think the config was exactly correct. I believe the babel config option that was used was `targets.esmodules: true`, but the correct one is `modules: false`. With this option, Babel will leave the output as ES modules. Currently, the files we publish are using the `exports` object, which is still the CommonsJS format.

This PR adds an additional build script specifically for ES modules. I have also updated the package.json in each package with the `module: "esm/file.js"` field which allows us to export both build types. On the consuming end, Webpack will prioritize the module field.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -57,6 +57,7 @@ module.exports = (api) => {
       [
         '@babel/preset-env',
         {
+          modules: process.env.BABEL_ENV === 'build' ? false : 'cjs',
           targets: {
             esmodules: true,
           },


lukelooker:
@dntrkv Is this PR still relevant? Happy to pull it in - just lost track of it / wasn't sure if it was still useful or not.

dntrkv:
> @dntrkv Is this PR still relevant? Happy to pull it in - just lost track of it / wasn't sure if it was still useful or not.

It is still relevant, but I need to make some changes to it. It's kind of a side project thing I'm doing and haven't had a chance to get back to it. I should be able to next week.

lukelooker:
> > @dntrkv Is this PR still relevant? Happy to pull it in - just lost track of it / wasn't sure if it was still useful or not.
> 
> It is still relevant, but I need to make some changes to it. It's kind of a side project thing I'm doing and haven't had a chance to get back to it. I should be able to next week.

Sounds good. Let me know if I can do anything to help out!

PR #1369, opened 2020-08-24
InputChips/SelectMulti select/copy/paste chips

mdodgelooker:
### :sparkles: Changes

- Added `selectedValues` state and keydown/click behavior to support selecting & copying chip values
- Inspired by the hidden input used in Gmail's email picker field
- For `SelectMulti` to accept pasted chips, it needs `freeInput` (copying does not require it)
  - This wasn't what I was planning at first, but trying to constrain pasted values to `options` wouldn't work because of how some filter components work in the core product: options are limited on the back-end to 999, so even if the pasted values aren't present in the options, they could still be valid. 
  - This is what inspired me to support `validate` and the other validation-related callbacks on `SelectMulti`. If we really want to constrain pasted values to valid options, we can use the `validate` callback to do so.

- [x] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### :camera: Screenshots
86de2e58-444e-4a0c-ad8e-ea0904952e19


packages/components/src/Form/Inputs/InputChips/InputChips.tsx
```diff
@@ -47,10 +47,7 @@ import {
  * but also allows (validated) user inputs to be stored as 'chips' (see the Chip element)
  */
 
-export interface InputChipsProps
-  extends Omit<InputChipsCommonProps, 'onValidationFail'>,
-    InputChipsControlProps,
-    Partial<InputChipsInputControlProps> {
+export interface InputChipsValidationProps {
```
mdodgelooker, line 50: Extracted the validation callbacks so they could be easily included in `ComboboxMultiProps`. Pasting chips from one `SelectMulti` into another can only work gracefully if `freeInput` is true, which made me realize how useful these callbacks could be there.

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
```diff
@@ -104,49 +120,197 @@ export const InputChipsBaseInternal = forwardRef(
     forwardedRef: Ref<HTMLInputElement>
   ) => {
     const internalRef = useRef<HTMLInputElement>(null)
+    const hiddenInputRef = useRef<HTMLInputElement>(null)
```
mdodgelooker, line 123: For the hidden input, inspired by gmail's email picker field – it simplifies keydown handling and enables copying to clipboard.

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
```diff
@@ -104,49 +120,197 @@ export const InputChipsBaseInternal = forwardRef(
...
+            break
+          case 'x':
+          case 'c':
+            hiddenInputRef.current && hiddenInputRef.current.select()
+            document.execCommand('copy')
+            break
+        }
+        if (e.key === 'x') {
+          deleteSelected()
+        }
```
mdodgelooker, line 263: Is this awkward? I first had it in the switch above right under `case 'x':` but apparently that's considered an "evil" fall through. Any suggestions? Maybe the switch itself is awkward?

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
```diff
@@ -104,49 +120,197 @@ export const InputChipsBaseInternal = forwardRef(
...
     const chips = values.map((value) => {
-      function onChipDelete() {
-        handleDeleteChip(value)
+      function onChipDelete(e?: SyntheticEvent) {
+        handleDeleteChip(value, e)
       }
+      const isSelected = selectedValues.includes(value)
       return (
         <Chip
+          disabled={disabled}
```
mdodgelooker, line 305: This should've always been there, just noticed the oversight.

packages/components/src/Chip/Chip.tsx
```diff
@@ -112,18 +120,18 @@ const ChipJSX = forwardRef((props: ChipProps, ref: Ref<HTMLSpanElement>) => {
...
     }
     onKeyDown && onKeyDown(event)
     setFocusVisible(false)
   }
 
-  const handleDelete = () => {
+  const handleDelete = (e: SyntheticEvent) => {
     if (!disabled) {
-      onDelete && onDelete()
+      onDelete && onDelete(e)
```
mdodgelooker, line 134: Need to pass the event so `stopPropagation` can be called on it over in `InputChipsBase.tsx`.

storybook/src/Forms/SelectMulti.stories.tsx
```diff
@@ -44,21 +45,21 @@ import {
...
+  { value: 'Apples2' },
+  { value: 'Bananas2' },
+  { value: 'Oranges2' },
+  { value: 'Pineapples2' },
+  { value: 'Kiwis2' },
+  { value: 'Apples3' },
+  { value: 'Bananas3' },
+  { value: 'Oranges3' },
+  { value: 'Pineapples3' },
+  { value: 'Kiwis3' },
```
mdodgelooker, line 62: `freeInput` and copy-pasting don't work well when the label is distinct from the value. You would past the labels, but the corresponding options in the list would not show as selected.

mdodgelooker:
@jhardy can you take a look at the UX of selecting & copy-pasting chips? @lukelooker this is here if you need a distraction. 😁

packages/components/src/Form/Inputs/Combobox/ComboboxMultiInput.tsx
```diff
@@ -52,12 +53,14 @@ import { useInputMultiPropRefs } from './utils/useInputPropRefs'
...
 export interface ComboboxMultiInputProps
   extends Omit<InputChipsCommonProps, 'autoComplete'>,
+    InputChipsValidationProps,
     ComboboxInputCommonProps,
     Partial<InputChipsInputControlProps> {
   onClear?: () => void
   /**
    * Allows inputting of values outside of options via typing or pasting
    * Not recommended for use when options have labels that are different from their values
+   * Use validate, onValidationFail, and onDuplicate to control behavior around free inputting
```
dbchristopher, line 63: I'm not sure how often this would come up, but it does allow you to paste in values that don't exist in the options array. Rather than accepting a custom validation function, I wonder if a better path forward would just be to offer a built-in that options existence check?

Screen_Shot_2020-08-25_at_10_22_01_AM

mdodgelooker: @dbchristopher I did initially try what you're suggesting – to not require `freeInput` for the chip-pasting behavior to work, and to match pasted values against the provided `options`. But then I realized this won't work for the `TagList` and `MultiStringInput` filters in the core product, which are the driving motivation for this feature. Due to truncation at 999 on the back-end, the provided `options` can't be assumed to be exhaustive. 

Currently in dashboards legacy, where pasting chips is supported (it's not yet in dashboards-next), values are not validated. But I'm going to look into whether back-end validation is something we want to add, when this feature goes into dashboards-next.
a9b3cd58-26c0-4e38-9d9c-f8079d019ef0

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
```diff
@@ -104,49 +120,197 @@ export const InputChipsBaseInternal = forwardRef(
...
+          inputValue === '' && handleDeleteChip(values[values.length - 1])
+        } else if (isCtrlCmdPressed(e) && e.key === 'a') {
+          focusInput(hiddenInputRef)
+          selectAll()
+        } else if (e.key === 'ArrowLeft') {
+          focusInput(hiddenInputRef)
+          selectPrevious(e)
+        }
+      }
+    }
```
dbchristopher, line 246: It might be nice to break this up using react-hotkeys-hook.  This sort of pattern reads a little easier to me:

```
    const isFocused = () =>
      internalRef.current &&
      internalRef.current.contains(document.activeElement)

    useHotkeys('cmd+a, ctrl+a', () => {
      isFocused() && selectAll()
    })

    useHotkeys('ArrowLeft', (e) => {
      isFocused() && selectPrevious(e)
    })

    useHotkeys('backspace', () => {
      isFocused() &&
        inputValue === '' &&
        handleDeleteChip(values[values.length - 1])
    })
```

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
```diff
@@ -104,49 +120,197 @@ export const InputChipsBaseInternal = forwardRef(
...
+        setSelectedValues([values[values.length - 1]])
+      } else {
+        const curIndex = values.indexOf(selectedValues[0])
+        if (curIndex > 0) {
+          const newSelectedValue = values[curIndex - 1]
+          if (e.shiftKey) {
+            setSelectedValues([newSelectedValue, ...selectedValues])
+          } else {
+            setSelectedValues([newSelectedValue])
+          }
```
dbchristopher, line 145: it's really cool how well this works! the whole element feels very native

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
```diff
@@ -104,49 +120,197 @@ export const InputChipsBaseInternal = forwardRef(
...
+      if (isCtrlCmdPressed(e)) {
+        // Select all, copy, cut
+        switch (e.key) {
+          case 'a':
+            selectAll()
+            break
+          case 'x':
+          case 'c':
+            hiddenInputRef.current && hiddenInputRef.current.select()
+            document.execCommand('copy')
```
dbchristopher, line 258: 🤯

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
dbchristopher: yeah that's pretty awkward 😬. I think I'd extract that copy-to-clipboard functionality and then stick to canonical Switch syntax

```js
   const copyToClipboard = () => {
      hiddenInputRef.current && hiddenInputRef.current.select()
      document.execCommand('copy')
    }
   
    switch (e.key) {
      case 'a':
        selectAll()
        break
      case 'x':
        copyToClipboard()
        deleteSelected()
        break
      case 'c':
        copyToClipboard()
        break
    }
```

storybook/src/Forms/SelectMulti.stories.tsx
dbchristopher: I wonder if this is going to be a dealbreaker internally. Will people be ok ensuring that labels and values are identical in order to use this functionality? And if so, should we just stop supporting the `label` attribute altogether? It feels like this limitation is going to cause a lot of confusion in the future 😞

dbchristopher (dismissed):
This works so well! Suggested a couple changes, but it's ok if you don't want to think about reworking key commands at this point. So the main thing is just cleaning up that cut/copy switch fall-through, and then figuring out whether the label/value mismatch is a blocker

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
mdodgelooker: Looks good to me, will do! Thank you 🙏

storybook/src/Forms/SelectMulti.stories.tsx
mdodgelooker: Good point – I've been so focused on the filter components use cases, where label and value are the same, but I should do some research and see if there's a way to make it work for when label != value. I wonder if I should again take a page from Gmail and allow those to be copied like `I'm the label<I'm the value>`...?

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
mdodgelooker: Nice! Let me give it a try.

storybook/src/Forms/SelectMulti.stories.tsx
jhardy: +1 to seeing if we can make it work with `label!=value`

jhardy:
@mdodgelooker this works really well! I did notice a few accessibility issues with screen readers, but I need to do a a bit more testing to narrow down the issue and I can file a separate ticket. The copy and pasting features are 💯

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
```diff
@@ -76,19 +83,32 @@ export interface InputChipsCommonProps
    * @default true
    */
   removeOnBackspace?: boolean
+  /**
+   * Format the value for display in the chip
+   */
+  formatChip?: (value: string) => string
```
mdodgelooker, line 89: This new prop allows `ComboboxMultiInput` to support copying options in string form as `'Option Label<optionvalue>'` while displaying a friendly "Option Label" in the chip.

mdodgelooker:
@dbchristopher and @jhardy thanks for the encouragement to make this work for `SelectMulti` where label!=value. 😁 I think I have a workable solution but don't re-review yet! I just chatted with @lukelooker who had a good idea that I want to put in. Stay tuned.

mdodgelooker:
@dbchristopher @jhardy and @lukelooker question for you guys. I updated this PR with support for copy/paste when label != value by using the "email" format `'Label<value>'` in the pasted string. But then I got a good suggestion from Luke to try using a JSON string instead, as a more universal standard. It ended up requiring a good amount more code than the email format, mostly due to the status of comma as delimiter when entering chips, and so I wanted to run it past you all before moving forward. 

Here's a branch with JSON format, compared against this one: https://github.com/looker-open-source/components/pull/1372

Do we like JSON enough to make the extra code worth it?

lukelooker:
My $.02 is the flexibility of JSON copy is worthwhile as it does a really nice of job of leaving the interface "open to extension" (https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle#:~:text=In%20object%2Doriented%20programming%2C%20the,without%20modifying%20its%20source%20code) and will set a really nice path for supporting things like an embedded filter expression.

dbchristopher:
Yeah that is a fair bit of process to support JSON. But I think @lukelooker's justification make sense -- probably best to stick to well-understood standards when possible

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
```diff
@@ -99,56 +127,210 @@ export const InputChipsBaseInternal = forwardRef(
...
+        switch (e.key) {
+          case 'a':
+            selectAll()
+            break
+          case 'x':
+            copyToClipboard()
+            deleteSelected()
+            break
+          case 'c':
+            copyToClipboard()
```
mdodgelooker, line 277: Less awkward `switch`, thanks to @dbchristopher !

mdodgelooker: (I tried `useHotkeys` but it got a bit messy because the behaviors for select-all and backspace are slightly different depending on whether the main input or the hidden one is focused, so I stuck with the keydown handlers).

mdodgelooker:
@dbchristopher it's ready for re-review! 🙏 I pulled in the commits from my json format branch, and updated `splitInputValue` per your suggestion.

lukelooker (approved):
Read through in detail and tested as well. Works great for me.

PR #1386, opened 2020-08-31
[Draft] Activated Icon Button

jhardy:
### :sparkles: Changes

This is a first pass of activated icon button pattern we have seen. @lukelooker interested in your take here. I changed how color is applied to the icon buttons to achieve this. This matches closer to where we want to go but I don't remember why we were lightening it before. 

The other idea I was toying around with is instead of all the tertiary applications of color here, we could create a new `IconToggleButton` component.

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/Button/IconButton.tsx
+++ b/packages/components/src/Button/IconButton.tsx
@@ -81,6 +81,10 @@ export interface IconButtonProps
     SpaceProps {
   type?: 'button' | 'submit' | 'reset'
   /*
+   * If the button has a 'toggled' appearance
+   */
+  activated?: boolean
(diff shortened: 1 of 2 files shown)


lukelooker:
This alters the resting, hover and active states a bit as well:

Before:
image

After:
image

I'm worried that designers might not be happy with the filled background state as previously we've seen folks make modifications to `IconButton` when they had built-in backgrounds in some states.

Thoughts? Perhaps we make "activated" more stand-alone behaviorally?

PR #1387, opened 2020-09-01
Select: support showCreate

mdodgelooker:
### :sparkles: Changes

- Previously only `SelectMulti` supported `showCreate` and `formatCreateLabel` (when `isFilterable` is true)
- While `SelectMulti` may have more frequent use cases for these props, `Select` also needs them, since it otherwise has no way for the user to enter a custom value
- Noticed and fixed a couple related bugs:
  - `SelectMulti` failing to appropriately show "No options" when `showCreate` is used
  - `Select` overwriting search value with the current option value if the option's value and label are different

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots
59dfa0c4-d522-4730-b93b-7147c9592835

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED]
 
+### Added
+
+- `Select` now supports `showCreate` and `formatCreateLabel` that were previously only supported in `SelectMulti`
+
 ### Changed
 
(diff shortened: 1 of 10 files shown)


mdodgelooker:
@lukelooker good call – I did add a couple lines in a test for the "No options" bug:
https://github.com/looker-open-source/components/pull/1387/files#diff-8311f703e9a25dbd25875e432e1dac91R580-R582
I'll see if I can capture the overwriting bug.

mdodgelooker:
Captured it! @lukelooker ready for re-review 👍

PR #1388, opened 2020-09-02
Chip onClick keyboard a11y

mdodgelooker:
### :sparkles: Changes

- New `useClickable` hook that provides role & tabIndex attributes, focusVisible prop for styling, and enter/space keyup handler for keyboard accessibility on clickable & focus-able components
- `Chip` now uses `useClickable`

- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


mdodgelooker:
@jhardy & @lukelooker my goal was to make the `useClickable` as re-usable as possible which is why I didn't name it `useButton` – in theory we could use this for elements that are `role="tab"` or `role="option", etc. Speaking of which, I was thinking of trying this hook over in `Tab` as well – would that be good to see in this PR as proof of re-usability, or better in a later PR?

packages/components/src/Chip/Chip.tsx
```diff
@@ -24,30 +24,34 @@
...
-const ChipStyle = styled.span<ChipProps>`
+const ChipStyle = styled.span<FocusVisibleProps>`
   ${reset}
 
   align-items: center;
   background: ${({ theme }) => theme.colors.keySubtle};
   border-radius: 4px;
   color: ${({ theme }) => theme.colors.keyInteractive};
-  display: flex;
+  display: inline-flex;
```
mdodgelooker, line 54: Opportunistic. I think `inline-flex` is what we want here – it'll fix this kind of thing:
image

lukelooker (approved):
Really like this pattern. 

Especially like how this wraps up the `focusVisible` pattern too!  I wonder if we could (in a follow-up) snag the other implementations of that to consolidate that throughout the components. :)

I'd like to hear @jhardy weigh in on a11y on this just in case I missed something or he has more thoughts but it definitely solves the specific use case really nicely :)

packages/components/src/Chip/Chip.tsx
```diff
@@ -96,72 +101,58 @@ const ChipLabel = styled(Text)<TextProps & TruncateProps>`
...
+        onKeyDown={useWrapEvent(handleKeyDown, onKeyDown)}
+        ref={ref}
+        {...props}
+      >
+        <ChipLabel truncate={truncate}>{children}</ChipLabel>
+        {onDelete && !disabled && (
+          <IconButton
+            disabled={disabled}
+            icon="Close"
+            label="Delete"
```
jhardy, line 146: The removable chip button has a strange accessible label 
image

The button gets announced as "Delete me delete button", what is happening is the `ChipButton`'s label and the `IconButton` label is being combined. 

I think since we have a button inside a `role=button` here the labels are being combined into one. 

The only way I can think how to avoid this is to have each button be siblings but styled to look like there are composed.

Something like

```
<ChipButtonWrapper>
  <ChipButton />
  <RemoveButton />
</ChipButtonWrapper>
```

jhardy:
I found a small issue with the label on removable chips. 

I think a larger conversation we should have is how we navigate chips via keyboard, right now we rely on tab to navigate between pieces. Ideally we use arrows to navigate between groups of chips. 

It would look something like this, where tab (1) gets you to the list of chips, then arrow keys are the next movements to navigate between individual chips

image

packages/components/src/utils/useClickable.ts
```diff
@@ -0,0 +1,96 @@
...
+    Pick<GenericClickProps<E>, 'onClick'> {}
+
+export interface UseClickableResult<E extends HTMLElement>
+  extends Pick<CompatibleHTMLProps<E>, Attributes | 'onClick' | 'tabIndex'>,
+    FocusVisibleProps {}
+/**
+ * This hook provides keyboard accessibility for any component that renders a non-button element
+ * that is both focus-able and clickable. The component should handle styling for focusVisible.
+ */
+export function useClickable<E extends HTMLElement>({
```
jhardy, line 52: this is cool! 🎉

mdodgelooker:
@jhardy thanks for catching that. How do you feel about the accessibility of Material UI Chip? (Check out the console when interacting with the chips) https://material-ui.com/components/chips/#chip-2
Their delete icon is not a button and not focusable (thought it is clickable). To delete the chip via keyboard, you need to use Backspace or Delete.

The sibling buttons idea is cool too but I worry that it could interfere with other behavior and would require more complex code than the Material UI solution.

lukelooker (approved):
LGTM

PR #1390, opened 2020-09-03
theme.fonts Improvements

lukelooker:
### theme.fonts updates

- Adds `theme.fonts.body` and uses that as the default font-family
- Removes font specification from the GlobalStyle to prevent font "leakage"
- Adds font specification to reset to ensure that it's broadly applied
- Apply `fonts.brand` to `Button*`, 'Heading` and `Tabs`
- Allow font specification on `Heading` (sometimes folks might want a heading with `fonts.body`
- Removed `fontWeights.extraBold` (not used anywhere in components or product)
- Removed `fontWeights.light` (remove usage, product usage will be replaced with `normal`)

### GoogleFontsLoader

- Introduced new GoogleFontsLoader to allow for easy loading of Google
fonts in standalone applications
- Leveraged GoogleFontsLoader in ComponentsProvider

### Storybook fixes

- Fixed a variety of small Storybook configuration errors to correct console warnings

- [x] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


packages/components/src/Avatar/__snapshots__/Avatar.test.tsx.snap
```diff
@@ -2,6 +2,7 @@
 
 exports[`Avatar AvatarCombo renders Avatar and its secondary avatar 1`] = `
 .c7 {
+  font-family: 'Roboto','Noto Sans JP','Noto Sans CJK KR','Noto Sans Arabic UI','Noto Sans Devanagari UI','Noto Sans Hebrew','Noto Sans Thai UI','Helvetica','Arial','sans-serif','Noto Sans';
```
jhardy, line 5: A couple things here. 'sans-serif' in quotes won't work, it will default to a serif font. 

'Noto Sans' is at the end here, should be before `Helvetica...sans-serif`

jhardy: I think you might have explained this to me before, but do we need the `Noto` fonts here? 

Would be nice to just keep our base font stack to system fonts `Helvetica, Arial, sans-serif` and then let users choose their font stack at implementation level.

packages/components/src/Text/__snapshots__/Code.test.tsx.snap
```diff
@@ -2,8 +2,9 @@
 
 exports[`A Code component aligned 1`] = `
 .c0 {
+  font-family: 'Roboto','Noto Sans JP','Noto Sans CJK KR','Noto Sans Arabic UI','Noto Sans Devanagari UI','Noto Sans Hebrew','Noto Sans Thai UI','Helvetica','Arial','sans-serif','Noto Sans';
   text-align: right;
-  font-family: 'Monaco','Menlo','Ubuntu Mono','Consolas','source-code-pro',monospace;
+  font-family: 'Roboto Mono','Monaco','Menlo','Ubuntu Mono','Consolas','source-code-pro','monospace';
```
jhardy, line 7: font-family declared twice here. Should just be the mono family

packages/components/src/Text/__snapshots__/Heading.test.tsx.snap
```diff
@@ -19,14 +22,17 @@ exports[`A <h1> Heading 1`] = `
 
 exports[`A <h1> Heading sized to <h2> 1`] = `
 .c0 {
+  font-family: 'Roboto','Noto Sans JP','Noto Sans CJK KR','Noto Sans Arabic UI','Noto Sans Devanagari UI','Noto Sans Hebrew','Noto Sans Thai UI','Helvetica','Arial','sans-serif','Noto Sans';
   font-size: 1.375rem;
+  font-family: 'Red Hat Display','Noto Sans JP','Noto Sans CJK KR','Noto Sans Arabic UI','Noto Sans Devanagari UI','Noto Sans Hebrew','Noto Sans Thai UI','Helvetica','Arial','sans-serif','Noto Sans';
```
jhardy, line 27: headings have two font family declarations

packages/design-tokens/src/tokens/typography/font_families.ts
```diff
@@ -26,7 +26,66 @@
...
+
+  faces = [...faces, ...fallbacks]
+
+  return faces.map((face) => `'${face}'`).join(', ')
+}
+
+export const defaultFonts: FontFamilyChoices = {
+  body: 'Roboto',
+  brand: 'Red Hat Display',
+  code: 'Roboto Mono',
```
jhardy, line 71: I wonder if we should default to system fonts? I don't have a strong opinion here just more of an open question.

jhardy (dismissed):

packages/components/src/Text/__snapshots__/Heading.test.tsx.snap
lukelooker: Fixed. `reset` now handles font-family globally so we get only one instance.

packages/components/src/Text/__snapshots__/Code.test.tsx.snap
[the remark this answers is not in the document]
lukelooker: Same as `Heading`, `reset` handles this now

packages/components/src/Avatar/__snapshots__/Avatar.test.tsx.snap
lukelooker: Doh - I had strings properly double-quoted at one point to make this work and then un-did it (thinking I had made a mistake). Fixed now to _actually_ work right.

`Noto Sans` is at the end of the font-stack in the CORE product and I originally just left that because I hadn't dug into _why_. On further reflection I think you're right that it doesn't make any sense to be there. Our previous stock font "Open Sans" had a lot of missing characters so Noto Sans was used to fill in those gaps, now that we're defaulting to better quality font without missing characters (Roboto) I think that fallback is no longer needed.

The entries for the rest of the Noto fonts are to address broad i18n support for characters that aren't included in most fonts (Korean, Japanese, etc.).  We leave it to the implementor to actually load those fonts but I'd like to leave them in place because applications running within the CORE application such as Extensions will have access to them. In general I think having them in the font-stack shouldn't have any impact itself – it's really whether or not they're loaded which is a separate concern.

packages/design-tokens/src/theme.ts
```diff
@@ -100,3 +100,8 @@ export const theme: Theme = {
   transitions,
   zIndexFloor: 1,
 }
+
+declare module 'styled-components' {
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  export interface DefaultTheme extends Theme {}
+}
```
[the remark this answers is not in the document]
mdodgelooker: Rad!

mdodgelooker (changes requested):
This is so neat! Just a couple of minor suggestions.

mdodgelooker:
Oh and quick question about `Heading` – is it always supposed to be Red Hat Display or do we want to sometimes style it as `font-family: inherit`, like in `Select` with descriptions: 
image

packages/components/src/Button/Button.test.tsx
```diff
@@ -36,6 +36,13 @@ test('Button is rendered ', () => {
   assertSnapshotShallow(<Button>click here</Button>)
 })
 
+const renderTest = (component: JSX.Element) => {
+  const { container } = renderWithTheme(component)
+  expect(container).toBeVisible()
+}
+
+test('Render', () => renderTest(<Button />))
```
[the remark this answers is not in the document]
lukelooker: Whoops... this was entirely unrelated (I was verifying a text fix in another repo and accidentally committed this)

lukelooker:
> Oh and quick question about `Heading` – is it always supposed to be Red Hat Display or do we want to sometimes style it as `font-family: inherit`, like in `Select` with descriptions:
> image

Great point. Went through and audited all uses of `Heading` internally - specified `body` explicitly on `CalendarNav, MenuGroupLabel, SelectOptions, SidebarGroup` components :)

packages/design-tokens/src/GoogleFontsLoader/GoogleFontsLoader.tsx
```diff
@@ -0,0 +1,85 @@
...
+  let uri = `${family.replace(/ /g, '+')}:`
+
+  let weights = font.weights.map((weight) => `0,${weight}`)
+  if (italic) {
+    uri += 'ital,'
+    const italicizedWeights = font.weights.map((weight) => `1,${weight}`)
+    weights = [...weights, ...italicizedWeights]
+  }
+
+  uri += `wght@${weights.join(';')}`
```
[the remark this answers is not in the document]
lukelooker: Doh! You're totally right :)

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
```diff
@@ -140,7 +140,12 @@ export function SelectOptionWithDescription({
 }: SelectOptionObject) {
   return (
     <Box>
-      <Heading fontSize="small" fontWeight="semiBold" pb="xxsmall">
+      <Heading
+        fontFamily="body"
+        fontSize="small"
+        fontWeight="semiBold"
+        pb="xxsmall"
+      >
```
mdodgelooker, line 148: Should `SelectOptionGroupTitle` below also have font="body"?

mdodgelooker (commented):
Everything looks good! Just had one question about the font on `SelectOptionGroupTitle`.

packages/components/src/Avatar/__snapshots__/Avatar.test.tsx.snap
jhardy: Sounds good to me. No harm in declaring the font families.

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
lukelooker: Yep! (missed it, doh)

mdodgelooker (approved):
LGTM!

PR #1391, opened 2020-09-03
Fix Popover click outside issue with onMouseUp

mdodgelooker:
### :sparkles: Changes

- Something changed in 16.9 (looked but I couldn't find exactly what it was) such that if a page has both an `onMouseUp` (anywhere) and `click` and `mouseup` event listeners added directly to the `document` with `useCapture=true` _after_ a `mousedown`, the click handler will not be called.
- Removing `useCapture` on the `mouseup` listener fixes this.
- Here's a sandbox showing the issue issue https://codesandbox.io/s/boring-platform-tbbf4
- See it fixed at [your local storybook]/?path=/story/overlays-popover--mouse-up
- To see it without `@looker/components`, check out this sandbox: https://codesandbox.io/s/magical-bush-70ngs
- I tried capturing this with a unit test but apparently jsdom is not subject to this behavior.

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/Popover/Popover.tsx
+++ b/packages/components/src/Popover/Popover.tsx
@@ -369,13 +369,13 @@ function usePopoverToggle(
       // popover was closed via mousedown, but still need to cancel next click
       document.addEventListener('click', handleClickOutside, true)
       // and then cleanup mouseDownTarget
-      document.addEventListener('mouseup', handleMouseUp, true)
+      document.addEventListener('mouseup', handleMouseUp)
     }
 
     return () => {
       document.removeEventListener('mousedown', handleMouseDown, true)
       document.removeEventListener('click', handleClickOutside, true)
-      document.removeEventListener('mouseup', handleMouseUp, true)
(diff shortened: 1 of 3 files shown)


packages/components/src/Popover/Popover.tsx
```diff
@@ -369,13 +369,13 @@ function usePopoverToggle(
       // popover was closed via mousedown, but still need to cancel next click
       document.addEventListener('click', handleClickOutside, true)
       // and then cleanup mouseDownTarget
-      document.addEventListener('mouseup', handleMouseUp, true)
+      document.addEventListener('mouseup', handleMouseUp)
```
mdodgelooker, line 372: Luckily, `useCapture` is actually not needed for `handleMouseUp` since all it's doing is clearing the `mouseDownTarget` value – no event cancelling like with `mousedown` and `click`.

lukelooker (approved):
LGTM. (nice catch!)

PR #1392, opened 2020-09-03
ActionListItemAction: Renamed "role" prop to "itemRole" prop

ElliotTheHuman:
### :sparkles: Changes

- `ActionListItemAction` Renamed "role" prop to "itemRole" prop
  - The purpose of the prop is to pass either "link" or "button" down to the underlying `MenuItem`'s `itemRole` prop. However, the name of that prop on `ActionListItemAction` was `role` rather than `itemRole`. Because props from `ActionListItemAction` are destructured  onto the underlying `MenuItem`, the underlying `MenuItem` was not able to receive a value for its `itemRole` via the `ActionListItemAction`.

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots
`itemRole` prop receiving "link", "button", and not being used:
Screen Shot 2020-09-03 at 2 00 38 PM

`ActionListItemAction` with `itemRole="link"`:
Screen Shot 2020-09-03 at 2 01 22 PM

`ActionListItemAction` with `itemRole="button"`:
Screen Shot 2020-09-03 at 2 01 30 PM

`ActionListItemAction` with no explicitly passed in `itemRole` prop:
Screen Shot 2020-09-03 at 2 01 44 PM

### DIFF
--- a/packages/components/src/ActionList/ActionListItemAction.tsx
+++ b/packages/components/src/ActionList/ActionListItemAction.tsx
@@ -34,7 +34,12 @@ export interface ActionListItemActionProps
   children?: ReactNode
   detail?: ReactNode
   icon?: IconNames
-  role?: 'button' | 'link'
+  /**
+   * Determines if the ActionListItemAction is an <a/> or <button/> element
+   * Note: The value passed into this prop is passed into the underlying MenuItem's itemRole prop
+   * @default 'button'
+   */
+  itemRole?: 'link' | 'button'
 }
 
 /**


mdodgelooker (approved):
LGTM – I think if you end up making more updates to this interface, it might be worthwhile to do something like this, but probably overkill for right now.
```
export type ActionListItemActionProps = Omit<MenuItemProps, 'compact'|'current'>
```

PR #1408, opened 2020-09-11
Truncate component and Tree layout updates

dbchristopher:
https://looker.atlassian.net/browse/LENS-952
https://looker.atlassian.net/browse/LENS-1212
https://looker.atlassian.net/browse/LENS-1211
https://looker.atlassian.net/browse/LENS-1041

### :sparkles: Changes

- Adds a new Truncate component
- Updates tree layout to allow for items and labels to either wrap or truncate based on desired behavior. 
- When truncating, text will be cut off after one line, but a tooltip will be provided to view the rest of the text

- [x] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### :camera: Screenshots

tree-labels


lukelooker:
Neat! (screenshot looks awesome) 💯

dbchristopher:
@jhardy I implemented your feedback, and this is ready for another look. I also added additional logic so that the truncated text component only renders the tooltip when the text actually overflows:

text-overflow-behavior

jhardy:
@dbchristopher only small nit I see now is when tree labels wrap the icons are misaligned, other than this is so cool  
image

packages/components/src/Accordion/AccordionDisclosureGrid.tsx
```diff
@@ -36,13 +36,20 @@ interface AccordionDisclosureGridProps
 }
 
 const Indicator = styled.div`
-  align-items: center;
+  align-items: align-items;
```
jhardy, line 39: Meta :D

packages/components/src/Text/useIsTextTruncated.ts
```diff
@@ -0,0 +1,65 @@
...
+        return
+      }
+
+      resizeObserver.disconnect()
+      window.removeEventListener('resize', handleResize)
+    }
+  }, [handleResize, element])
+
+  return isTruncated
+}
```
jhardy, line 65: This is really neat! I would love to see it documented a bit more, just because I want to learn to write more things like this. I think it would be great to have a bit more verbose descriptions of what everything is doing. 

I don't know best practices are here, so feel free to ignore this request as well :)

dbchristopher: oh yeah for sure! I'm definitely planning to clean up and document this stuff better

packages/components/src/ActionList/ActionListItemColumn.tsx
```diff
@@ -34,7 +34,9 @@ interface ActionListItemColumnProps {
   className?: string
 }
 
-const ActionListItemColumnInnerLayout = styled.div``
+const ActionListItemColumnInnerLayout = styled.div`
+  width: 100%;
```
dbchristopher, line 38: had to add a width style in order for the Truncate component to work in the actionlist context

www/src/documentation/components/content/action-list.mdx
```diff
@@ -587,3 +587,73 @@ The `bulk` prop is used to configure an `ActionList`'s control bar. The object p
   )
 }
 ```
+
+## Truncation
```
dbchristopher, line 591: Added a truncation example to the Action list, as I believe that could be a common use case for it.

packages/components/src/Overlay/OverlaySurface.tsx
```diff
@@ -79,6 +87,7 @@ export const OverlaySurface = forwardRef(
         {...eventHandlers}
         tabIndex={-1}
         data-placement={placement}
+        maxWidth={maxWidth}
```
dbchristopher, line 90: added maxWidth support here for a little more style control over the long-text tooltips

packages/components/src/Text/Truncate.tsx
```diff
@@ -0,0 +1,110 @@
...
+      <HiddenText {...restProps} ref={textRef}>
+        {children}
+      </HiddenText>
+    </TextWrapper>
+  )
+}
+
+const TextStyle = styled.div`
+  display: block;
+  font: inherit;
```
dbchristopher, line 90: I'm just doing a blunt inherited font style rather than allowing people to pass fontSize and other props directly to this component. My thinking is that Truncate should just be concerned with the truncation behavior, and not any other styling

packages/components/src/Overlay/OverlaySurface.tsx
dbchristopher: I wonder if we might just want to hard-code a sane default max-width rather than accepting this as a prop? @jhardy what do you think?

packages/components/src/Layout/FlexItem/FlexItem.tsx
```diff
@@ -50,4 +50,9 @@ export const FlexItem = styled.div<FlexItemProps>`
   ${flexBasis}
   ${flex}
   ${order}
+  /*
+   * A min-width must be set here to resolve a firefox bug where any children
+   * with style of text-overflow: ellipsis; will otherwise not truncate the
+   * text appropriately. */
+  min-width: 0; /* IMPORANT!! Do not delete! */
```
dbchristopher, line 57: This was... 🤯  to debug

packages/components/src/Text/Truncate.tsx
```diff
@@ -0,0 +1,104 @@
...
+        }}
+      >
+        <TextStyle>{children}</TextStyle>
+      </TooltipComponent>
+      {/*
+       * Tooltip clobbers refs on any children.
+       * We must use a hidden text element to reliably detect whether content
+       * is truncated.
+       */}
+      <HiddenText ref={textRef}>{children}</HiddenText>
```
mdodgelooker, line 81: This is a good use case for `useTooltip`
```suggestion
  const { tooltip, ...triggerProps } = useTooltip({
    content: children,
    disabled: !isTruncated,
    maxWidth: '700px',
    placement: 'top-start',
    surfaceStyles: {
      backgroundColor: 'background',
      border: '1px solid',
      borderColor: 'ui3',
      color: 'text3',
    },
    textAlign: 'left',
    triggerElement: domNode,
    width: 'auto',
  })

  return (
    <TextWrapper>
      {tooltip}
      <TextStyle {...triggerProps} ref={textRef}>
        {children}
      </TextStyle>
```

mdodgelooker (changes requested):
This is so great, can't wait to use `Truncate` in `InputChips` and `SelectMulti`.

I think `useTooltip` will streamline `Truncate` a bit, and I noticed a small height issue when `detail` is used:
image
Before:
image

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -195,7 +208,11 @@ const TreeItemLayout: FC<TreeItemProps> = ({
           {props.icon && (
             <PrimaryIcon name={props.icon} size={defaultIconSize} />
           )}
-          <FlexItem flex="1">{children}</FlexItem>
+          <FlexItem flex="1" width="100%" pr={truncate ? 'large' : 'none'}>
```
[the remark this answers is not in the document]
dbchristopher: Initially I kept running into a bug where the text would continue outside the container a bit before getting truncated, and this was may hacky one-off fix to resolve that. But I just double checked, and it doesn't seem to be an issue any more. Maybe that FlexItem `min-width` change helped fix things here too 🤷 I'll delete!

packages/components/src/Text/Truncate.tsx
dbchristopher: Oh awesome! Thanks that's much better

dbchristopher:
ok @mdodgelooker I updated it so that it could align `detail` prop better for both truncated and wrapped text. Should be ready for final review!
 
Screen Shot 2020-09-24 at 3 10 54 PM

mdodgelooker (approved):
Looks great! 🚢

packages/components/src/Overlay/OverlaySurface.tsx
lukelooker: It seems like setting a maxWidth here via `defaultProps` could make sense in that it at least addresses the awkwardness of having a "magic value" (`700px`) deep within `Truncate`. 

If it was set here `Truncate` could simply absorb that and we could, if later requested, add an optional `maxWidth` to `Truncate` (or other `OverlaySurface` consumers) to override it.

I'd argue for `30rem` or `48rem` since either could be derived from our pre-existing breakpoints.

packages/components/src/Text/Truncate.tsx
```diff
@@ -0,0 +1,84 @@
...
+   */
+  const { tooltip, ...triggerProps } = useTooltip({
+    content: children,
+    disabled: !isTruncated,
+    maxWidth: '700px',
+    placement: 'top-start',
+    surfaceStyles: {
+      backgroundColor: 'background',
+      border: '1px solid',
+      borderColor: 'ui3',
```
lukelooker, line 60: `ui2` is more common border color.

lukelooker: Verified in spec that`ui2` is correct value here:

image

packages/components/src/Text/Truncate.tsx
```diff
@@ -0,0 +1,84 @@
...
+      color: 'text3',
```
lukelooker, line 61:
```suggestion
      color: 'text5',
```

packages/components/src/Text/Truncate.tsx
lukelooker, line 60:
```suggestion
      borderColor: 'ui2',
```

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -177,6 +188,8 @@ const TreeItemLayout: FC<TreeItemProps> = ({
     </HoverDisclosure>
   )
 
+  const TextWrapper = truncate ? Truncate : Identity
```
lukelooker, line 191: Named `[React].Fragment` can be used here instead.

```suggestion
  const TextWrapper = truncate ? Truncate : Fragment
```

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -91,13 +92,23 @@ export interface TreeItemProps
...
+   * Prevent text wrapping on long labels and instead render truncated text
+   **/
+  truncate?: boolean
 }
 
+/*
+ * Identity component: transparently pass through children
+ */
+const Identity: FC<any> = ({ children }) => <>{children}</>
+
```
lukelooker, line 105: Not needed (see below)

```suggestion
```

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -195,7 +208,11 @@ const TreeItemLayout: FC<TreeItemProps> = ({
           {props.icon && (
             <PrimaryIcon name={props.icon} size={defaultIconSize} />
           )}
-          <FlexItem flex="1">{children}</FlexItem>
+          <FlexItem flex="1">
+            <Paragraph fontSize="xsmall">
+              <TextWrapper>{children as string}</TextWrapper>
```
lukelooker, line 213: `Fragment` doesn't require `any` workaround.

```suggestion
              <TextWrapper>{children}</TextWrapper>
```

lukelooker (changes requested):
Found a few small clean-ups and color tweaks that are off-spec.

Noticed in testing that the "field picker" tree example in Storybook has some fairly large hover-state issues (in master it appears to work fine):

12aecc92-dbc5-465e-acf8-c1f99948e4d3

www/src/documentation/components/content/action-list.mdx
lukelooker: This super cool. I'm working on updating `ActionList` to use a `table` underneath and I'm going to take a look at doing "baked in" support similar to `Tree` as I think you're totally correct about how useful this will be. :)

packages/components/src/Tree/TreeItem.tsx
dbchristopher: oh that's awesome!

packages/components/src/Overlay/OverlaySurface.tsx
jhardy: My vote would be for something that is a comfortable line length, `48rem` is on the larger end of that comfortable reading line length, but I think its a good starting point and a better choice than `30rem`

www/src/documentation/components/content/tree.mdx
```diff
@@ -70,6 +70,42 @@ Use the `border` prop if you would like to have vertical lines expand from your
...
+
+Occasionally tree item labels will wrap to a second line when the text content is very long, or the available space is very narrow. The text content will wrap by default, but if that is undesirable you can use the `truncate` prop to limit text content to a single line.
+
+When text content overflows, it will render a tooltip to view the entire content, which is consistent with the [Truncate](/components/text/truncate) component.
+
+```jsx
+<Heading>Default Text Wrapping Layout</Heading>
+<Tree
+  border
+  label="A very label that wraps to a second line. Both the label and the tree item will wrap to two or more lines because there is just too much text."
```
jhardy, line 83:
```suggestion
  label="A very long label that wraps to a second line. Both the label and the tree item will wrap to two or more lines because there is just too much text."
```

jhardy:
This is looking really great, two things I think we may want to consider and wanted to your's and @lukelooker  feedback here:

1.  I think we should have a longer delay before showing the truncation tooltip. This would be something we need to test a bit, but my first guess would be somewhere between 200-400ms, possibly longer, but would need to test it out

2. If possible, I think we might want to try hiding the truncation tooltip  when the user leaves the trigger area, even if their mouse is over the tooltip. 

I attached a gif here to try to demonstrate why I am suggesting these changes:

-  In the scenario where you have a long list of items that could be truncated, when you are moving through that list quickly you really don't need to see the truncation tooltip until you rest on an item. 
- When moving up through a lists of truncated items, the tooltip blocks you from hovering over the previous item, or items if the tooltip has lots of content,  in the list.  I think it would be a better experience if we just close the tooltip right away so you can get back to the previous item, with out having to move your mouse left or right outside of the truncation target area

packages/components/src/Overlay/OverlaySurface.tsx
lukelooker: It's worth noting that `OverlaySurface` is used underneath `Tooltip` & `Popover` so this would be a fairly global value. We may want to set a larger threshold here and then narrow it for `Truncate` specifically perhaps?

lukelooker:
> This is looking really great, two things I think we may want to consider and wanted to your's and @lukelooker feedback here:
> 
> 1. I think we should have a longer delay before showing the truncation tooltip. This would be something we need to test a bit, but my first guess would be somewhere between 200-400ms, possibly longer.

I like the idea of a delay but I wonder if it might be better implemented as global functionality to `Tooltip` itself – we've heard comments about tooltip being "too aggressive" and giving it a subtle delay would probably be helpful with that feedback as well.

> 2. If possible, I think we might want to try hiding the truncation tooltip  when the user leaves the trigger area, even if their mouse is over the tooltip.

Having using this pattern in other products (cloud console is my reference) I would actually find this frustrating because it would preclude me from being able to select the text for copy & paste – something that's common with long non-breaking strings in technical products (e.g.: server address, GUID, etc)

jhardy:
> Having using this pattern in other products (cloud console is my reference) I would actually find this frustrating because it would preclude me from being able to select the text for copy & paste – something that's common with long non-breaking strings in technical products (e.g.: server address, GUID, etc)

I think you are right here and this was the use case that almost made me not suggest that change but figured it was worth having a discussion about 😄 

I can see cases in like an long list of tree items (exploring) where it may be annoying, but that seems to be the tradeoff between wrapping and truncation

lukelooker:
> I think you are right here and this was the use case that almost made me not suggest that change but figured it was worth having a discussion about 😄

Yeah - I was actually running this a bit yesterday when trying to paste something that was truncated :)

Thoughts on a general `Tooltip` delay?

jhardy:
I like the idea of the global tool tip delay.

lukelooker:
> I like the idea of the global tool tip delay. I could see a way to opt out of delay being helpful, things like an info icon where its holding more info, but for our icons button and things like truncation a delay makes a lot of sense

Sounds awesome. I've added a ticket to track that change separately from this PR.

dbchristopher:
> Sounds awesome. I've added a ticket to track that change separately from this PR.

Thanks @lukelooker! Punting the general tooltip behavior changes for another sprint makes sense to me. 

> I would actually find this frustrating because it would preclude me from being able to select the text for copy & paste – something that's common with long non-breaking strings in technical products (e.g.: server address, GUID, etc)

Yeah I had that exact internal battle. It's annoying that it blocks the content when you move up the list, but it seems important to allow them to copy the text. One potential solution: if they click the truncated text it could automatically copy the full content to your clipboard. I'm not sure how best to indicate that behavior in the UI though. Any thoughts @jhardy or @lukelooker? 

Might be worth leaving that as a future optimization either way though

packages/components/src/Overlay/OverlaySurface.tsx
dbchristopher: I added the default value to Tooltip, so it should be reasonably global but shouldn't affect any Popover content

packages/components/src/Tree/Tree.tsx
```diff
@@ -237,6 +244,13 @@ export const TreeStyle = styled.div<TreeStyleProps>`
   & > ${Accordion} > ${AccordionContent} > ${TreeItem} > ${TreeItemLabel} {
     ${({ depth, theme }) => generateIndent(depth + 1, theme)}
   }
+
+  ${TreeItemLabel} {
+    ${IconButton} {
+      max-height: ${({ theme }) => theme.lineHeights.xsmall};
+      overflow: visible;
+    }
+  }
```
dbchristopher, line 253: I hate having to do style overrides like this 😕 This pattern makes the child components inherently impure and harder to debug, as you can't just look at IconButton to reason about why it's rendering the way it is. But it's consistent with how most a lot of tree content is styled, and seemed to be the only way I could get the FieldPicker example to behave

storybook/src/Tree/FieldPicker.stories.tsx
```diff
@@ -128,6 +128,7 @@ const PickerItem = () => {
 const addButton = (
   <ButtonTransparent
     size="xxsmall"
+    mt="xxxsmall"
```
dbchristopher, line 131: had to do a little one-off fix to fix alignment on the FieldPicker story itself. We'll probably have to double check any [INTERNAL REPO] instances of tree in case they're using this pattern and the flex-start alignment throws anything off.

I had to go with flex-start alignment to support how we want the detail content to line up when text wraps 

Screen Shot 2020-09-28 at 5 19 00 PM

dbchristopher:
ok @lukelooker I implemented feedback and resolved those tricky Field Picker bugs! Should be ready for another review

lukelooker (changes requested):
My apologies for the delay on reviewing this – I want to have image-snapshots in place because there's a lot of nuance within this component.

I went ahead and built a quick branch (`dc/tree-truncate-rebased`) with this work on top of the new Tree image snapshots (#1498) - in so doing I found two issues:

1. TreeItem height changed from `25px` to `26px` - after discussing with Jared the correct value should actually be `24px` so regardless of what happens snapshot updates will be needed.
2. When using the `Truncate` component within the prototype field picker story the `detail` items fall outside of the `Tree` space itself:
image

dbchristopher:
@lukelooker it appears that these screenshots aren't always consistent with what a real browser renders:
Screen Shot 2020-10-02 at 10 35 49 AM

Probably most useful to alert you when something unintentionally changed, rather than to be used as a primary vehicle for QA. I'll condense the vertical space a little bit though!

dbchristopher:
sorry @lukelooker I just fully processed what you were saying. It's friday and I'm tired lol. I'll look into the truncation/fieldpicker combo now

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -228,13 +239,19 @@ interface TreeItemLabelProps {
...
   height: 100%;
+  max-width: 100%;
   outline: none;
-  padding: ${({ theme: { space } }) => space.xxsmall};
+  /*
+   * Subtract 1px from each padding value to compensate for
+   * a transparent border on TreeItemSpace
+   */
+  padding-bottom: calc(${({ theme }) => theme.space.xxsmall} - 1px);
+  padding-top: calc(${({ theme }) => theme.space.xxsmall} - 1px);
```
dbchristopher, line 254: had to do a manual tweak here to get it to add up pixel perfectly

Screen_Shot_2020-10-02_at_11_34_55_AM

packages/components/src/utils/HoverDisclosure/HoverDisclosure.tsx
```diff
@@ -36,5 +37,11 @@ export const HoverDisclosure: FC<HoverDisclosureProps> = ({
...
   const context = useContext(HoverDisclosureContext)
-  return visible || context.visible ? <>{children}</> : null
+  return (
+    <Disclosure visible={visible || context.visible}>{children}</Disclosure>
+  )
 }
+
+const Disclosure = styled.span<HoverDisclosureProps>`
+  visibility: ${({ visible }) => (visible ? 'visible' : 'hidden')};
+`
```
dbchristopher, line 47: I'm not sure what unintended consequences this could cause. But it seems like we need to make sure the content actually exists in the layout to prevent situations like this when it toggles into view:

disclosure-hover

dbchristopher:
Ok this component might be the death of me but I **think** I've gotten it all buttoned up for now 😂 😭 

field-picker

Ready for another look @lukelooker!

packages/components/src/Tree/Tree.tsx
```diff
@@ -237,6 +246,24 @@ export const TreeStyle = styled.div<TreeStyleProps>`
...
+      max-height: calc(${({ theme }) => theme.lineHeights.xsmall} - 2px);
+      transform: translateY(-1px);
+    }
+    ${ButtonTransparent} {
+      /*
+       * Caution: Magic Numbers! Necessary FieldPicker adjustment
+       */
+      transform: translateY(-1px);
+    }
+  }
```
dbchristopher, line 266: eye twitch

packages/components/src/Tree/TreeItem.tsx
dbchristopher: I do have to say that I really hope the new image snapshots don't push us towards a ton of magic numbers and finicky tweaks sprinkled out through the codebase. I feel that if our normal T-shirt sizing doesn't add up to the exact pixels shown in Figma, then the answer is to update Figma.

Ultimately there isn't a material difference between 24px rows and 26px rows. I'm fine with this as an exception, but I just really hope this is not a glimpse into the future of our codebase.

lukelooker (commented):
> I do have to say that I really hope the new image snapshots don't push us towards a ton of magic numbers and finicky tweaks sprinkled out through the codebase. I feel that if our normal T-shirt sizing doesn't add up to the exact pixels shown in Figma, then the answer is to update Figma.

I totally agree with your point here - I don't want to be chasing arbitrary number – that said, `24` is a codified `size` in our theme and that got me thinking about simpler ways to chase this problem – I added a commit that backs out the complexity through two major changes:

```min-height: ${({ theme }) => theme.sizes.medium}```

+ changing all use of `border: 1px solid transparent` (and its focusVisible alternate) to use `outline` instead. This effectively makes the the math a moot-point as "outline" doesn't mess with the box-sizing.

Additionally I flipped all of the flex arrangement back to positioning everything vertically centered so that as the `TreeItem` (and `AccordionDisclosure`) height changes everything just tracks along with the change in height.

Try pulling my changes and see what you think – I wasn't able to find any glaring issues

File: packages/components/src/Accordion/AccordionDisclosureLayout.tsx [hunk 21]
```diff
@@ -68,8 +61,23 @@ const Layout: FC<AccordionDisclosureLayoutProps> = ({
   )
 }
 
+const Indicator = styled.div`
+  align-items: flex-start;
+  display: flex;
+  justify-content: center;
+  padding-top: ${({ theme }) => theme.space.xxsmall};
```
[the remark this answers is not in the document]
dbchristopher: @lukelooker yeah exactly. Jared requested that all the icons align with the first line of text when it starts wrapping. So we can't rely on flex API to handle our vertical alignment for us :( 

Screen Shot 2020-10-05 at 10 59 31 AM

package.json
```diff
@@ -44,7 +44,7 @@
     "test": "yarn jest",
     "storybook": "yarn workspace storybook start",
     "storybooks-build": "yarn lerna run storybook-build --scope '@looker/*' --stream --parallel",
-    "preimage-snapshots": "yarn storybooks-build",
+    "-preimage-snapshots": "yarn storybooks-build",
```
dbchristopher, line 47: @lukelooker is `-preimage-snapshots` an intentional change here?

packages/components/src/Accordion/AccordionDisclosure.tsx
```diff
@@ -130,13 +130,12 @@ export const AccordionDisclosureStyle = styled.div<
...
-  border: 1px solid transparent;
-  border-color: ${({ focusVisible, theme }) =>
-    focusVisible && theme.colors.keyFocus};
   ${({ color }) => (color ? colorStyleFn : 'color: currentColor;')}
   cursor: pointer;
   display: flex;
-  outline: none;
+  outline: 1px solid transparent;
+  outline-color: ${({ focusVisible, theme }) =>
+    focusVisible && theme.colors.keyFocus};
```
dbchristopher, line 138: `outline` is a great call 👍 👍

dbchristopher:
@lukelooker thanks for helping simplify things and talking us down from the more finicky vertical alignment. I just gave it another once-over and I think things are ready to merge whenever you want to give the ✅ !

lukelooker (dismissed):

lukelooker (approved):

PR #1446, opened 2020-09-16
Hack Extract: Gatsby/Google Analytics

mdodgelooker:
### :sparkles: Changes

- added `gatsby-plugin-google-analytics`
- added some event tracking in the Search component

### DIFF
--- a/www/src/components/Search.tsx
+++ b/www/src/components/Search.tsx
@@ -24,8 +24,9 @@
 
  */
 
+import { trackCustomEvent } from 'gatsby-plugin-google-analytics'
 import startCase from 'lodash/startCase'
-import React, { useState } from 'react'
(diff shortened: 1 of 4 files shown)


www/src/components/Search.tsx
```diff
@@ -72,10 +73,37 @@ const SearchField = ({ index, store }: FauxSearchProps) => {
...
+    const t = window.setTimeout(() => {
+      if (query) {
+        const noResults = results.length === 0 ? '-no-results' : ''
+        trackCustomEvent({
+          action: `input${noResults}`,
+          category: 'Search',
+          label: query,
+        })
+      }
+    }, 500)
```
mdodgelooker, line 86: A fairly sizable debounce to avoid capturing search values mid-typing.

PR #1448, opened 2020-09-18
Layout (semantic) improvements

lukelooker:
### :sparkles: Changes

- Remove `Main` (use `as="main"` on `Section` instead)
- Various flex configuration improvemeents

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


jhardy:
Layout not quite working as expected

image

storybook/src/Layout/ApiExplorer.stories.tsx
```diff
@@ -58,7 +57,7 @@ export const ApiExplorer = () => (
         <Aside p="large" width="200px">
           Aside
         </Aside>
-        <Main p="xxlarge">
+        <Section p="xxlarge" as="main">
```
jhardy, line 60: since this is as `main` its not picking up the section styling in `Layout.tsx`, following lines do not style Section when cast as main

```
 & > ${Section} {
    width: 0;
  } 
```

lukelooker: > Layout not quite working as expected
> 
> image

Ack! So in this PR I removed the auto-magic behaviors around `hasAside` – basically the developer _for now_ has to explicitly specify if the `Layout` `hasAside` within it. I'm going to re-introduce that behavior later but for the initial implementation was finicky so I decided to just focus on core layout challenges now and then deal with component developer experience a little later along in the process.

Examples should be working smoothly now.

jhardy (approved):

PR #1475, opened 2020-09-24
feat(TreeItem): export the internal TreeItemDetail

pablolmiranda:
### :sparkles: Changes

Export the TreeItemDetail component from TreeItem to allow changes on the styling from upstream components.
Today if we want to remove the `padding-right` from the `TreeItemDetail` we can't today because the component is hidden inside the `TreeItem` component. In this case, it is not possible to do:
```
const TreeItemWithoutDetailPadding = styled(TreeItem)`
  ${TreeItemDetail} {
    padding-right: 0
  }
`
```

Exporting the `TreeItemDetail` component allow us to start styling it.

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [X] Checked for i18n impacts
- [X] Checked for a11y impacts
- [X] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/Tree/TreeItem.tsx
+++ b/packages/components/src/Tree/TreeItem.tsx
@@ -237,7 +237,7 @@ export const TreeItemLabel = styled(Space)<TreeItemLabelProps>`
   padding: ${({ theme: { space } }) => space.xxsmall};
 `
 
-const TreeItemDetail = styled.div<{ detailAccessory: boolean }>`
+export const TreeItemDetail = styled.div<{ detailAccessory: boolean }>`
   align-items: center;
   display: flex;
   height: 100%;
--- a/packages/components/src/Tree/TreeItem.tsx
+++ b/packages/components/src/Tree/TreeItem.tsx
@@ -237,12 +237,10 @@ export const TreeItemLabel = styled(Space)<TreeItemLabelProps>`
   padding: ${({ theme: { space } }) => space.xxsmall};
 `
 
-export const TreeItemDetail = styled.div<{ detailAccessory: boolean }>`
+const TreeItemDetail = styled.div<{ detailAccessory: boolean }>`
   align-items: center;
   display: flex;
   height: 100%;
-  padding-right: ${({ detailAccessory, theme }) =>
-    detailAccessory && theme.space.xxsmall};
 `
 
 export const TreeItem = styled(TreeItemLayout)`


packages/components/src/Tree/TreeItem.tsx
```diff
@@ -237,7 +237,7 @@ export const TreeItemLabel = styled(Space)<TreeItemLabelProps>`
   padding: ${({ theme: { space } }) => space.xxsmall};
 `
 
-const TreeItemDetail = styled.div<{ detailAccessory: boolean }>`
```
lukelooker, line 240: @ElliotTheHuman - TreeItemDetail currently get's an `xxsmall` padding when `detailAccessory` is enabled but my memory on _why_ is fuzzy. Can you help me remember why?

lukelooker (commented):
I'd prefer not to export `TreeItemDetail` as it's meant to be an internal implementation detail and exporting it means we need to support customization like this (something that often causes integration wrinkles).

Let's explore if we can make `padding-right: 0` the default instead perhaps.

packages/components/src/Tree/TreeItem.tsx
lukelooker: Reference: https://github.com/looker-open-source/components/blob/master/packages/components/src/Tree/TreeItem.tsx#L244-L245

ElliotTheHuman: Ah I remember this. @lukelooker Kind of long story but I'll do my best to sum it up briefly. When `detailAccessory` is enabled, we get this hover effect that only applies to the `TreeItem`'s label container and NOT the detail elements. We do this by placing the detail elements within the label container (or their own detail container) depending on `detailAccessory`.

The label container has `xxsmall` padding by default, and so if `detailAccessory` is enabled, the detail elements don't end up having `xxsmall` padding right of them. When `detailAccessory` is disabled, there IS `xxsmall` padding right of the detail elements do the label container's `xxsmall` padding all around.

So basically, if we don't have `padding-right: xxsmall` when `detailAccessory` is enabled, misalignment can occur between two `TreeItem`s in the rare scenario where `detailAccessory` is enabled for one item and not the other:
Screen Shot 2020-09-24 at 11 59 55 AM

ElliotTheHuman: I'm trying to think of reasons you'd want `detailAccessory` to be enabled for one `TreeItem` and disabled for another sibling, but can't really think of any 🤷‍♀️ .

I know that there are cases where you want `detailAccessory` to be disabled for a parent `Tree` but NOT for a child `TreeItem`. But based on current mockups and use cases, the details in parent `Tree`s and the details in their child `TreeItem`s tend to be different so the misalignment isn't as visually noticeable.

@pablolmiranda Could you elaborate a bit more on why you'd like to be able to overwrite that default `xxsmall` padding? Wondering if we should start a larger conversation on how padding is applied to these `Tree`-related elements.

lukelooker: @jhardy and I discussed this a bit. This component currently breaks our "no internal margin" rule (using padding but to roughly the same effect). 

The action item is to go ahead and consistently remove the internal padding and if the end-user needs it they can add it to the content the place inside `detail`.

Does that make sense to you @ElliotTheHuman? 

@pablolmiranda, would that solve your issue?

ElliotTheHuman: From what I can tell that should be fine! The set height of 25px for `TreeItem` should do a good enough job of creating "spacing" between items. Alternatively, I imagine you could actually use a `Space` element within a `Tree` to get more customized gap sizing.

pablolmiranda: @lukelooker, removing the padding solve the issue on our side.

pablolmiranda: @lukelooker I removed the `padding-right` from `TreeItemDetail` and kept the component inside `TreeItem`.

PR #1478, opened 2020-09-25
Elliot/tree padding removal

ElliotTheHuman:
### :sparkles: Changes
- Removed all padding from `Tree`
- Removed all padding from `TreeItem`
- Removed px from `TreeGroup`

**Note:** `Tree` elements still receive left padding using our auto-indent pattern.

It's worth noting that (afaik) these changes do not visually change any of our `Tree` components. Using `display: flex` and `align-items: center` and `height: 25px` essentially handles the "spacing" of `Tree`-related elements.

In addition, `TreeGroup` still has default `py` in order to keep with the Explore Team's desired specs. It might be worth giving `TreeGroup` a set height a similar fashion to `Tree` and `TreeItem`, but then we wouldn't be able to do what we do now which is have different amounts of spacing above and below a `TreeGroup`.

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### :camera: Screenshots
Screen Shot 2020-09-25 at 10 38 20 AM

### DIFF
--- a/packages/components/src/Tree/TreeItem.tsx
+++ b/packages/components/src/Tree/TreeItem.tsx
@@ -251,6 +251,7 @@ export const TreeItemLabel = styled(Space)<TreeItemLabelProps>`
   min-height: ${({ theme }) => theme.sizes.medium};
   min-width: 0;
   outline: none;
+  padding: ${({ theme: { space } }) => space.xxsmall};
 `
 
 const TreeItemDetail = styled.div<{ detailAccessory: boolean }>`
--- a/packages/components/src/Tree/TreeItem.tsx
+++ b/packages/components/src/Tree/TreeItem.tsx
@@ -251,6 +251,7 @@ export const TreeItemLabel = styled(Space)<TreeItemLabelProps>`
   min-height: ${({ theme }) => theme.sizes.medium};
   min-width: 0;
   outline: none;
+  /* TODO: Make sure to remove this (or just remove px) after Truncation PR is merged in */
   padding: ${({ theme: { space } }) => space.xxsmall};
 `
 
--- a/packages/components/src/Tree/TreeItem.tsx
+++ b/packages/components/src/Tree/TreeItem.tsx
@@ -257,8 +257,6 @@ const TreeItemDetail = styled.div<{ detailAccessory: boolean }>`
   align-items: center;
   display: flex;
   height: 100%;
-  padding-right: ${({ detailAccessory, theme }) =>
-    detailAccessory && theme.space.xxsmall};
 `
 
 export const TreeItem = styled(TreeItemLayout)`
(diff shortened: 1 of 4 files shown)


packages/components/src/Tree/TreeItem.tsx
```diff
@@ -234,15 +234,12 @@ export const TreeItemLabel = styled(Space)<TreeItemLabelProps>`
   font-size: ${({ theme: { fontSizes } }) => fontSizes.xsmall};
   height: 100%;
   outline: none;
-  padding: ${({ theme: { space } }) => space.xxsmall};
```
dbchristopher, line 237: Just a heads up this line will conflict with my [PR #1408](https://github.com/looker-open-source/components/pull/1408/files#diff-a161a32b0dd7dffaf90fe387ebceb408L237)

I removed the horizontal padding, but it seems to need vertical-padding for the layout to work in my branch. How tight are they wanting these rows?

ElliotTheHuman: Hey Daniel. Afaik, the main priority is removing the horizontal padding from `TreeItem` at the request of the Explore Team, so I'm cool with keeping the vertical padding in if that's necessary!

I removed all padding outright since, from what I could tell, the thing that was ultimately "spacing" `TreeItem`s from each other was the set `height: 25px` that have on them. I could definitely be missing something here, but on this existing branch, I _think_ we end up with the same amount of "spacing" even without the vertical padding within `TreeItem`.

Screen Shot 2020-09-28 at 4 08 20 PM

dbchristopher: Ok that works for me! Maybe actually would you mind just reverting this change here so that it doesn't conflict with the other PR?

ElliotTheHuman: I can do that!

ElliotTheHuman: Actually, quick question for you @dbchristopher. Would it be alright if I just took the `padding-bottom` and `padding-top` additions from your PR and added them to my PR?
```
  padding-bottom: ${({ theme }) => theme.space.xxsmall};
  padding-top: ${({ theme }) => theme.space.xxsmall};
```

Asking because it would be nice to get the `TreeItem` px removal into master for the Explore team. At the same time, if your PR is likely to go in before the next release cut or if doing this on my end would still risk conflict, I could revert back here for simplicity.

lukelooker: Let's just hold this change until we land visual snapshots & truncation - should be before end-of-week anyway and then we can verify that nothing moves unexpectedly :)

ElliotTheHuman: Ah gotcha! Didn't realize taht visual snapshots and truncation would be landing so soon. 

I'll hold off and revert on my end in that case. Thanks team.

lukelooker:
Replaced with #1523

PR #1497, opened 2020-10-01
Scroll lock refactor

mdodgelooker:
### :sparkles: Changes

A couple use cases made it obvious that the current scroll lock implementation was not ideal (see https://codesandbox.io/s/modest-feynman-siyjk for a demo):

1. **A `Popover` that opens a controlled `Dialog` that is not its child, and the `Popover` stays open**
    Both scroll locks will stay enabled because there is no shared `DialogContext` to manage them, and the `Popover`'s lock will prevent scrolling in the `Dialog`
2. **A `Popover` that opens a `Dialog` that IS its child, the `Popover` stays open, and 2nd `Popover` is opened inside of the `Dialog`**
    The 2nd `Popover`'s scroll lock will disable the `Dialog`'s scroll lock, which will in turn re-enable the 1st `Popover`'s scroll lock, which will prevent scrolling inside the 2nd `Popover`.*

*This issue could probably have been solved by adding a pause function that wouldn't re-enable its parent scroll lock. I did try this, but 1) I couldn't quite get it to work, 2) It made the code even more convoluted than it already was, and 3) It wouldn't have solved the 1st issue above.

- [x] Includes test coverage for all changes*
- [x] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

*I added test for the body style change but the scroll event listener and scroll lock layering behavior are not well suited to jsdom. The storybook example at `:3333/?path=/story/overlays-popover--overlay-open-dialog` is a good overall test though.


packages/components/src/utils/useScrollLock.ts
```diff
@@ -24,86 +24,37 @@
...
-  const { value, setOn, setOff } = useToggle(enabled)
-
-  // save the existing body overflow value
-  const bodyOverflowRef = useRef(
-    typeof document !== 'undefined' ? document.body.style.overflow : ''
-  )
+import { ScrollLockContext } from '@looker/components-providers'
+import { useContext, useEffect } from 'react'
+import { useCallbackRef, useID } from './'
+
```
mdodgelooker, line 30: Using `ScrollLockContext` as the "global" single source of truth for scroll locks allows this hook to be vastly simplified. No more enable/disable confusion – just detecting when the element is added & removed from the DOM, adding & removing a lock in the context accordingly, and letting the context decide which lock should be enabled.

storybook/package.json
```diff
@@ -10,6 +10,7 @@
   },
   "dependencies": {
     "@looker/components": "*",
+    "@looker/components-providers": "*",
```
mdodgelooker, line 13: Added this to have access to `ScrollLockContext` for en example of how to disable the current scroll lock.

package.json
```diff
@@ -68,6 +69,7 @@
     "jest-styled-components": "^6.3.4",
     "lerna": "^3.22.1",
     "lint-staged": "^10.4.0",
+    "lodash": "^4.17.20",
```
mdodgelooker, line 72: I can't remember how this is supposed to work... do shared packages go in the main `package.json`'s `devDependencies` _and_ in the specific package's `dependencies`?

mdodgelooker:
@lukelooker and @dbchristopher this is ready for review with the caveat that I still need to give it a test run in the core product where some of the thorniest use cases occur.

The best example to test with is :3333/?path=/story/overlays-popover--overlay-open-dialog but also any `Popover` (including `Menu` and `Select`) or `Dialog`.

mdodgelooker:
Setup a sandbox to show the bugs with the current scroll lock implementation: https://codesandbox.io/s/modest-feynman-siyjk

mdodgelooker:
I tested these changes with the tricky use cases in the core product, and they look good. 👍

lukelooker (commented):
Works really well and code looks good to me. 

Minor note on root `package.json` clean-up.

For some reason React dev-tools is refusing to load for me and I'd love to verify that the profiler looks happy (rule of hooks and such) before final sign-off but functionally and code read-through looks great to me.

packages/components-providers/src/ScrollLock/utils.ts
```diff
@@ -0,0 +1,133 @@
...
+  if (previousStyle) {
+    document.body.style.paddingRight = previousStyle.paddingRight
+    document.body.style.overflow = previousStyle.overflow
+  }
+}
+
+export function getActiveScrollLock(lockMap: ScrollLockMap) {
+  // Sort the elements according to dom position and return the last
+  // which we assume to be stacked on top since all components using Portal
+  // share a single zIndexFloor and use dom order to determine stacking
```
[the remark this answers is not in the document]
mdodgelooker: Good call, gonna test this over in the CORE product now.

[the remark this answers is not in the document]
mdodgelooker: Testing looked good so far where `zIndexFloor` gets lifted.

lukelooker (approved):
Figured out react-devtools and verified that all appears well. 🚢

PR #1499, opened 2020-10-01
feat(TreeItem/TreeItemLabel): change the TreeItemLabel hover behavior

pablolmiranda:
### :sparkles: Changes

The current behavior turn off the hover when the tree item is selected.
On this PR we change the logic so TreeItemLabel always have a hover state, even when selected.

- [X] Build and tests are passing
- [ ] Update documentation
- [ ] Updated CHANGELOG
- [X] PR is ideally < ~400LOC


jhardy (approved):
This change works for me for now since it supports our current primary use case for Tree, but I think we will need to revisit how we apply color on interactive states because I can see others wanting the opposite behavior and I have seen some recent prototypes that use color differently than this.

PR #1502, opened 2020-10-02
Fix lodash import to "single method"

mdodgelooker:
Would it be worth adding `eslint-plugin-lodash` to the `eslint-config` package if just for the import-scope rule?

https://github.com/wix/eslint-plugin-lodash/blob/master/docs/rules/import-scope.md


lukelooker (approved):

lukelooker (approved):
Also, that lint check would be a great addition :) 

Follow-up PR :)

PR #1495, opened 2020-09-30
[draft] Use consistent icon sizes across components

jhardy:
### :sparkles: Changes

This PR adjust icons in various components to ensure consistent sizing 

### :white_check_mark: Requirements

- [ ] Includes test coverage for all changes
- [ ] Build and tests are passing
- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] Checked for i18n impacts
- [ ] Checked for a11y impacts
- [ ] PR is ideally < ~400LOC

PR #1503, opened 2020-10-02
Add visual snapshots for MenuItem and MenuGroup

jhardy:
### :sparkles: Changes
Adds visual snap shots for Menu Items and Menu Groups
These components would be affected by #1495, so adding visual snap shots to account for change

- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


packages/components/src/Menu/MenuGroup.story.tsx
```diff
@@ -0,0 +1,80 @@
...
+)
+
+export const MenuGroupExample = ListTemplate.bind({})
+MenuGroupExample.args = {
+  children: MenuListExample,
+}
+
+export const CompactMenuGroupExample = ListTemplate.bind({})
+CompactMenuGroupExample.args = {
+  children: MenuListExample,
```
lukelooker, line 73:
```suggestion
  ...Basic.args,
```

packages/components/src/Menu/MenuGroup.story.tsx
```diff
@@ -0,0 +1,80 @@
...
+  </>
+)
+
+export const MenuGroupExample = ListTemplate.bind({})
+MenuGroupExample.args = {
+  children: MenuListExample,
+}
+
+export const CompactMenuGroupExample = ListTemplate.bind({})
+CompactMenuGroupExample.args = {
```
lukelooker, line 72:
```suggestion
export const Compact = ListTemplate.bind({})
Compact.args = {
```

packages/components/src/Menu/MenuGroup.story.tsx
```diff
@@ -0,0 +1,80 @@
...
+    </MenuGroup>
+
+    <MenuGroup>
+      <MenuItem icon="TrashOutline">Move to Trash</MenuItem>
+    </MenuGroup>
+  </>
+)
+
+export const MenuGroupExample = ListTemplate.bind({})
+MenuGroupExample.args = {
```
lukelooker, line 67: There's a convention (admittedly loose at this point) to name the initial example `Basic` (see `Button*` stories for this)

```suggestion
export const Basic = ListTemplate.bind({})
Basic.args = {
```

packages/components/src/Menu/MenuItem.story.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+ */
+
+import React from 'react'
+import { Story } from '@storybook/react/types-6-0'
+
+import { MenuItem } from './MenuItem'
+import { MenuList, MenuListProps } from './MenuList'
+
+const ListTemplate: Story<MenuListProps> = (args) => <MenuList {...args} />
```
lukelooker, line 33: I _think_ this should be `MenuItem` here instead:

```suggestion
const Template: Story<MenuItemProps> = (args) => <MenuItem {...args} />
```

packages/components/src/Menu/MenuItem.story.tsx
```diff
@@ -0,0 +1,114 @@
...
+    Menu Item
+  </MenuItem>
+)
+
+const CurrentMenuItem = <MenuItem current>Menu Item</MenuItem>
+
+export const SingleMenuItem = ListTemplate.bind({})
+SingleMenuItem.args = {
+  children: <MenuItem>Menu Item</MenuItem>,
+}
```
lukelooker, line 67:
```suggestion
export const Basic = Template.bind({})
SingleMenuItem.args = {
  children: "Menu Item"
}
```

packages/components/src/Menu/MenuItem.story.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+export const SingleMenuItemWithIcon = ListTemplate.bind({})
+SingleMenuItemWithIcon.args = {
+  children: MenuItemWithIcon,
+}
```
lukelooker, line 72:
```suggestion
export const Icon = Template.bind({})
Template.args = {
  ...Basic.args,
  icon: 'User'
}
```

packages/components/src/Menu/MenuItem.story.tsx
```diff
@@ -0,0 +1,114 @@
...
+)
+SingleMenuItemWithDescriptionAndDetailAndIcon.args = {
+  children: MenuItemWithDescriptionAndDetailAndIcon,
+}
+
+export const SingleMenuItemCurrent = ListTemplate.bind({})
+SingleMenuItemCurrent.args = {
+  children: CurrentMenuItem,
+}
+
```
lukelooker, line 110: These rest of these should repeat pattern above. You can keep extending args to create more complex examples and this will allow for storybook users to see the args in use via the "Controls" add-on UI.

lukelooker (changes requested):
The coverage looks great – I'd just like to heavily leverage the Storybook "args" and Controls pattern more heavily here.

packages/components/src/Menu/MenuGroup.story.tsx
jhardy: extending basic here kept failing so just passed in children again

packages/components/src/Menu/MenuItem.story.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { MenuItem, MenuItemProps } from './MenuItem'
+import { MenuList } from './MenuList'
+
+const MenuItemTemplate: Story<MenuItemProps> = (args) => (
+  <MenuList>
+    <MenuItem {...args}>Menu Item</MenuItem>
+  </MenuList>
+)
+
```
jhardy, line 38: Rending outside of a `<MenuList />` kept inserting a `•` bullet in the visual snapshots

packages/components/src/Menu/MenuItem.story.tsx
```diff
@@ -0,0 +1,97 @@
...
+  ...Basic,
+  detail: 'A Detail',
+}
+
+export const IconAndDetail = MenuItemTemplate.bind({})
+IconAndDetail.args = {
+  ...Basic,
+  detail: 'A Detail',
+  icon: 'User',
+}
```
jhardy, line 58: I tried extending like
```
IconAndDetail.args = {
...Icon,
detail: 'A Detail'
}
```

but it would not render all the pieces of the component, so had to keep extending `Basic` for all of these
not sure if I was doing something wrong.

packages/components/src/Menu/MenuItem.story.tsx
lukelooker: Fixed by applying `list-style-type: none;` within `MenuItem`

packages/components/src/Menu/MenuGroup.story.tsx
```diff
@@ -0,0 +1,102 @@
...
+    detail: 'A longer detail',
+    icon: 'Undo' as IconNames,
+  },
+]
+
+interface CustomStoryProps extends MenuGroupProps {
+  icons?: boolean
+  detail?: boolean
+  description?: boolean
+}
```
lukelooker, line 54: Lets Storybook expose these as toggles to allow for easy action.

packages/components/src/Menu/MenuGroup.story.tsx
```diff
@@ -0,0 +1,102 @@
...
+NoDescriptions.args = {
+  ...Basic.args,
+  description: false,
+}
+
+export const NoIcons = Template.bind({})
+NoIcons.args = {
+  ...Basic.args,
+  icons: false,
+}
```
lukelooker, line 97: Because the MenuContext doesn't expect for `Icons` to disappear from MenuItems this the `icons: false` only collapses the icon gutter space when the template is loaded with this prop from the get-go.

lukelooker (approved):
I just pushed updates to these tests in this PR. 

Generally we want to keep tests narrowly focused on the component that we're looking at so that changes from outside don't cause unexpected issues.

PR #1504, opened 2020-10-02
First visual snapshots for text fields

jhardy:
### :sparkles: Changes

This is the last pieces that would be affected by #1495 
I only decided to snapshot text fields for right now to account for any icon changes in the PR referenced above. 

I am not sure if we want to create snapshots at the component level, or stick to just fields. I can adjust depending on feedback 

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


packages/components/src/Form/Inputs/InputText/InputText.story.tsx
```diff
@@ -0,0 +1,71 @@
...
+}
+
+const Template: Story<InputTextProps> = (args) => <InputText {...args} />
+
+export const Basic = Template.bind({})
+
+export const WithPlaceholder = Template.bind({})
+WithPlaceholder.args = {
+  placeholder: 'Placeholder',
+}
```
lukelooker, line 42: I think I'd skip the `With` portion of these story names as it'll look cleaner in the Storybook UI to see:

- Placeholder
- Value
- Etc.

lukelooker (changes requested):
Minor story-naming change and then I think this will be GTG :)

lukelooker (approved):
LGTM!

Note (mostly) to self... I wish the Storybook doc-generator could figure out nested typings. For some reason it just gives up trying to find the extended types so in this case it thinks `InputText` has no props:

image

More things to putter on I suppose...

PR #1522, opened 2020-10-06
Icon sizing consistency 

jhardy:
### :sparkles: Changes

This PR address a few issues with icon sizes across various components.
It make sure that icons in `Button`'s and `IconButton`'s are the same across sizes 
It uses named sizes for icons in `MenuItem`'s
It updates a few component that had an icon size specified in a number to using one of the icon sizes

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


dbchristopher (approved):
looks good!

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Tooltip` now has a default `maxWidth` of `30rem` (this can be overridden)
 - `Span` replaced all library-internal usage of `Text` with `Span`
 - `Code`, `CodeBlock` & `Paragraph` now explicitly use `theme.colors.text` as default color
+- Adjusted icons sizes in `IconButtons`, `Button`, `MenuItems`, `TextField` to ensure consitent sizing across components
```
lukelooker, line 29:
```suggestion
- Adjusted icons sizes in `IconButtons`, `Button`, `MenuItems`, `InputText`, `InputSearch` to ensure consistent sizing across components
```

lukelooker (commented):
`IconButton` `small` and `large` icon sizes changed significantly - I'm guessing that's intentional but just wanted to verify that it was intended. 

I also re-ran `image-snapshots` after rebasing and found a couple of `InputSearch` snapshots that needed updating (no appreciable change however)

packages/components/src/Button/ButtonBase.tsx
```diff
@@ -135,6 +140,8 @@ const ButtonJSX = forwardRef(
       onBlur && onBlur(event)
     }
 
+    const iconSize = buttonIconSizeMap[props.size || 'small']
+
```
jhardy, line 144: look up icon size for the button

packages/components/src/Button/ButtonBase.tsx
```diff
@@ -143,9 +150,9 @@ const ButtonJSX = forwardRef(
         onBlur={handleOnBlur}
         ref={ref}
       >
-        {iconBefore && <ButtonIcon name={iconBefore} />}
+        {iconBefore && <ButtonIcon name={iconBefore} size={iconSize} />}
         {children}
-        {iconAfter && <ButtonIcon name={iconAfter} />}
+        {iconAfter && <ButtonIcon name={iconAfter} size={iconSize} />}
       </ButtonOuter>
```
jhardy, line 156: here we apply the icon size directly to the button instead of a css function like before

packages/components/src/Button/size.ts
```diff
@@ -64,6 +64,26 @@ export const buttonSizeMap = {
...
+/* The size of an icon relative to the Button size inside a Button, ButtonOutline and ButtonTransparent */
+/* eslint-disable sort-keys-fix/sort-keys-fix */
+export const buttonIconSizeMap = {
+  xxsmall: 'xxxsmall',
+  xsmall: 'xxxsmall',
+  small: 'xxsmall',
+  medium: 'xsmall',
+  large: 'small',
+}
+
```
jhardy, line 86: Icon buttons and buttons have slightly different icon size requirements, so we have a scale for each component. 

Open to better suggestions for names for these.

packages/components/src/Form/Inputs/TextArea/TextArea.tsx
```diff
@@ -68,7 +68,7 @@ const TextAreaLayout: FC<TextAreaProps> = ({
         aria-invalid={validationType === 'error' ? 'true' : undefined}
         {...textareaProps}
       />
-      {validationType && <Icon name="Error" color="critical" size={20} />}
+      {validationType && <Icon name="Error" color="critical" size="small" />}
```
jhardy, line 71: use an icon size from the scale

packages/components/src/Form/Inputs/InputTime/InputTime.tsx
```diff
@@ -516,7 +516,7 @@ const InputTimeInternal = forwardRef(
             name="Error"
             color="critical"
             grid-area="warning"
-            size={20}
+            size="small"
```
jhardy, line 519: use an icon size from the scale

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -179,7 +179,7 @@ const InputTextLayout = forwardRef(
           color="critical"
           name="Error"
           title="Validation Error"
-          size={20}
+          size="small"
```
jhardy, line 182: use an icon size from the scale

packages/design-tokens/src/tokens/size.ts
```diff
@@ -29,9 +29,10 @@ import { SizeRamp } from '../system'
...
-  xxsmall: rem(12),
-  xsmall: rem(16),
-  small: rem(18),
+  xxxsmall: rem(12),
+  xxsmall: rem(16),
+  xsmall: rem(18),
+  small: rem(20),
   medium: rem(24),
   large: rem(32),
 }
```
jhardy, line 38: This is the change that I am not 100% sure f the downstream effects. This gives us some flexibility with icon sizes across the components.

packages/components/src/Dialog/Layout/__snapshots__/DialogHeader.test.tsx.snap
```diff
@@ -3,8 +3,8 @@
 exports[`DialogHeader Snapshot 1`] = `
 .c7 {
   font-family: 'Roboto','Noto Sans JP','Noto Sans CJK KR','Noto Sans Arabic UI','Noto Sans Devanagari UI','Noto Sans Hebrew','Noto Sans Thai UI','Helvetica','Arial',sans-serif;
-  width: 22px;
-  height: 22px;
+  width: 1rem;
+  height: 1rem;
```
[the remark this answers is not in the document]
jhardy: I updated this to use a value from the scale.

lukelooker (commented):
Overall this looks good to me. 

I'd _really_ like to figure out a way to make the snapshots pick-up the overall size of the component so we can see if anything changed "external" size (aka: icon is smaller but the button still takes the same amount of space).

With some tweaks I made today for `Dialog` snapshots I think I understand a route to get better-trimmed snapshots so let me take a pass at that tomorrow morning and see if we can use that as one more signal.

packages/design-tokens/src/tokens/size.ts
jhardy: would it worth being more explicit here and call this iconSizes?

packages/components/src/Button/size.ts
jhardy: After sleeping on it I kind of like `iconSizeForIconButton` and `iconSizeForButton` instead. How important is it to have the word `Map` in the name?  

Could also prepend with `get` so `getIconSizeForIconButton` and `getIconSizeForButton`.  🤷

packages/design-tokens/src/tokens/size.ts
jhardy: added visual snapshot tests in #1527

lukelooker: It _theoretically_ applies to any place that the Styled System `size` helper is used so for now I'd keep it as is. 

We'd need to do some more work to change the name of this within the theme, which might very well be worthwhile, but I think we could do it later / separately.

packages/components/src/Button/size.ts
lukelooker: Sold! I totally dig `iconSizeForIconButton` :)

`get` is usually a name for a "getter" method since this is just an object I'd say we'd forego that piece.

lukelooker (dismissed):

lukelooker (approved):
Rebased this on top of the other changes that landed and I think this is good to go.

🚢  - it

PR #1527, opened 2020-10-08
Add visual snapshots for Status

jhardy:
### :sparkles: Changes

Add visual snapshots for `Status` component.

I went ahead and snapshotted the different sizes, not a realistic use case for this component, but I figured we should capture the component as is. 

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


lukelooker (approved):
LGTM. Do you think we should remove size support eventually for these?

lukelooker:
Merging this so it goes in with `0.9.17`

PR #1529, opened 2020-10-09
Popover improvements

lukelooker:
### :sparkles: Changes

- Introduce domProps pattern from useDialog
- Break apart Popover into separate files
- Replace `function` keywords with arrow functions

- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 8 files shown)


mdodgelooker (dismissed):
LGTM! But do you think we should call out that the `active` class will no longer be added? I think there may be 1 or 2 places in the CORE product where `.active` is being assigned custom styles.

lukelooker:
> LGTM! But do you think we should call out that the `active` class will no longer be added? I think there may be 1 or 2 places in the CORE product where `.active` is being assigned custom styles.

Great catch! I did find a few spots that will need to be tweaked. Updated the CHANGELOG to warn

mdodgelooker (approved):

PR #1551, opened 2020-10-12
Message Bar uses flex instead of grid (IE fix)

lukelooker:
### :sparkles: Changes

- MessageBar story and preliminary snapshots (**initial snapshots still exactly match after flexbox refactor**)
- Reactor MessageBar grid to use flexbox layout
- Added `flex-shrink: 0` to `Space` to prevent it from taking less room than required for its content
- Wrapped `Status` in `styled` to allow for style-extension (used in `MessageBar` refactor

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


packages/components/src/MessageBar/MessageBar.tsx
```diff
@@ -234,21 +234,27 @@ const MessageBarLayout = forwardRef(
...
-  margin-left: ${({ theme: { space } }) => space.large};
+  ${Space},
+  ${Status} {
+    flex-shrink: 0;
+  }
+
+  ${MessageBarContent} {
+    flex-grow: 1;
+    padding: 0 ${({ theme: { space } }) => space.large};
+  }
```
dbchristopher, line 257: any reason to style this here rather than where `MessageBarContent` is declared on line 237?

packages/components/src/MessageBar/MessageBar.tsx
```diff
@@ -234,21 +234,27 @@ const MessageBarLayout = forwardRef(
...
-  display: grid;
+  display: flex;
   font-size: ${({ theme: { fontSizes } }) => fontSizes.small};
-  grid-template-columns: auto 1fr auto;
-`
 
-const MessageBarContent = styled.div`
-  flex: 1;
-  margin-left: ${({ theme: { space } }) => space.large};
+  ${Space},
```
dbchristopher, line 249: Looks like this same rule was [added to Space directly](https://github.com/looker-open-source/components/pull/1551/files?short_path=e532496#diff-99f58127be25b99bc85728c9e9456080R145), so probably no reason for the override here? Could it be added to Status directly too?

packages/components/src/MessageBar/MessageBar.story.tsx
```diff
@@ -0,0 +1,99 @@
...
+ SOFTWARE.
+
+ */
+
+import React from 'react'
+import { Story } from '@storybook/react/types-6-0'
+import { Button } from '../Button'
+import { MessageBar, MessageBarProps } from './MessageBar'
+
+const Template: Story<MessageBarProps> = (args) => <MessageBar {...args} />
```
dbchristopher, line 32: Remind me, how does this storybook file alert the system that it's set up for snapshots? Does it literally just look for any story with a `Template` object like this?

dbchristopher (changes requested):
Layout looks good! I just suggested a couple changes to MessageBar.tsx to try and minimize use of css side effects for styling

packages/components/src/MessageBar/MessageBar.story.tsx
lukelooker: Storyshots picks up any stories declared within the `packages/components/` as defined in the path glob defined here:

https://github.com/looker-open-source/components/blob/master/packages/components/src/stories.shots.ts

They can be explicitly excluded via the parameters option on the story:

```
MyStory.parameters = {
}
```

The `Template` pattern is just an easy way to enable the "args" (https://storybook.js.org/docs/react/writing-stories/args)  & "controls" (https://storybook.js.org/docs/react/essentials/controls) features of Storybook.

packages/components/src/MessageBar/MessageBar.tsx
lukelooker: You're totally right - adding to `Status` and removing from here altogether :)

packages/components/src/MessageBar/MessageBar.tsx
lukelooker: Nope - I think I was just copying around things as they were in grid-land but it's admittedly odd.

packages/components/src/MessageBar/MessageBar.story.tsx
dbchristopher: ah, cool 👍

dbchristopher (approved):
Looks good!

PR #1556, opened 2020-10-12
TreeItem: Support Cmd + Enter callback prop

ElliotTheHuman:
### :sparkles: Changes

- Added "onMetaEnterKeyDown" prop to `TreeItem`

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/Tree/TreeItem.tsx
+++ b/packages/components/src/Tree/TreeItem.tsx
@@ -93,7 +93,7 @@ export interface TreeItemProps
    * Callback that is triggered on CMD + Enter on Mac
    * or Windows key + Enter on PC
    */
-  onMetaEnter?: () => void
+  onMetaEnterKeyDown?: () => void
   /**
    * Determines if this TreeItem is in a selected state or not
    */
@@ -108,7 +108,7 @@ const TreeItemLayout: FC<TreeItemProps> = ({
   children,
   className,
   gapSize = 'xsmall',
-  onMetaEnter,
+  onMetaEnterKeyDown,
   selected,
   truncate,
   ...props
@@ -177,7 +177,7 @@ const TreeItemLayout: FC<TreeItemProps> = ({
     }
 
     if (event.keyCode === 13 && event.metaKey) {
-      onMetaEnter && onMetaEnter()
+      onMetaEnterKeyDown && onMetaEnterKeyDown()
     }
 
     onKeyDown && onKeyDown(event)
--- a/packages/components/src/Tree/TreeItem.tsx
+++ b/packages/components/src/Tree/TreeItem.tsx
@@ -44,6 +44,7 @@ import {
   uiTransparencyBlend,
 } from '@looker/design-tokens'
 import Omit from 'lodash/omit'
+import noop from 'lodash/noop'
 import { Space, FlexItem } from '../Layout'
 import { Icon, IconNames } from '../Icon'
 import { useHovered } from '../utils/useHovered'
@@ -108,7 +109,7 @@ const TreeItemLayout: FC<TreeItemProps> = ({
   children,
   className,
   gapSize = 'xsmall',
-  onMetaEnter,
+  onMetaEnter = noop,
   selected,
   truncate,
   ...props
@@ -119,7 +120,13 @@ const TreeItemLayout: FC<TreeItemProps> = ({
   const [isHovered] = useHovered(itemRef)
   const [isFocusVisible, setFocusVisible] = useState(false)
 
-  const { onBlur, onClick, onKeyDown, onKeyUp, ...restProps } = Omit(props, [
+  const {
+    onBlur,
+    onClick = noop,
+    onKeyDown,
+    onKeyUp,
+    ...restProps
+  } = Omit(props, [
     'color',
     'detail',
     'detailAccessory',
@@ -168,16 +175,8 @@ const TreeItemLayout: FC<TreeItemProps> = ({
       return
     }
 
-    if (
-      event.key === 'Enter' &&
-      !event.metaKey &&
-      event.target === event.currentTarget
-    ) {
-      onClick && onClick()
-    }
-
-    if (event.key === 'Enter' && event.metaKey) {
-      onMetaEnter && onMetaEnter()
+    if (event.key === 'Enter') {
+      event.metaKey ? onMetaEnter() : onClick()
     }
 
     onKeyDown && onKeyDown(event)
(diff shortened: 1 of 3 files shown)


packages/components/src/Tree/TreeItem.tsx
```diff
@@ -162,10 +168,18 @@ const TreeItemLayout: FC<TreeItemProps> = ({
...
+    if (
+      event.keyCode === 13 &&
+      !event.metaKey &&
+      event.target === event.currentTarget
+    ) {
       onClick && onClick()
     }
 
+    if (event.keyCode === 13 && event.metaKey) {
+      onMetaEnter ? onMetaEnter() : onClick && onClick()
```
dbchristopher, line 180: Just a quick sanity check -- is that fallback behavior actually what we want? That typing `cmd+enter` would fire the click handler if there's no meta+enter callback?

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -162,10 +168,18 @@ const TreeItemLayout: FC<TreeItemProps> = ({
       return
     }
 
-    if (event.keyCode === 13 && event.target === event.currentTarget) {
+    if (
+      event.keyCode === 13 &&
```
dbchristopher, line 172: I just remembered that `event.keyCode` is actually deprecated now. Maybe a good time to switch to `event.key === 'Enter'` while you're touching the code?

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -162,10 +168,18 @@ const TreeItemLayout: FC<TreeItemProps> = ({
...
+      !event.metaKey &&
+      event.target === event.currentTarget
```
dbchristopher, line 174: I think still want that `event.target === event.currentTarget` constraint on both `Enter` events to be safe... Or you actually don't need it on either lol. But either way, the logic flow for both branches is actually pretty similar, so I'd combine them into a single `if` statement.

Also, if you assign `noop` as the default value for onMetaEnter and onClick, you also skip the existence checks and condense the logic branching even more:

```
// ({
//  onMetaEnter = noop,
// onClick = noop,
//  ...
// }) => {

if(event.key === 'Enter' && event.target === event.currentTarget) {
  event.metaKey ? onMetaEnter() : onClick()
}
```

Let me know if that makes sense!

dbchristopher: For context, the `event.target === event.currentTarget` pattern was something we needed when using the react-hotkeys-hook dependency. It's a global key listener, so it would actually fire the event for _every_ element that was listening for that combination.

But since this component is actually just passing the click handler to the standard react `onKeyDown` listener, I think you could be safe deleting it altogether

www/src/documentation/components/content/tree.mdx
```diff
@@ -181,12 +181,17 @@ Use the `gapSize` prop to control the distance between elements in your `TreeIte
 
 Use the `selected` prop to display a light grey background on a given `TreeItem`.
 
+Use the `onClick` prop to trigger a callback on click or when pressing the "Enter" key on a focused `TreeItem`.
+
+Use the `onMetaEnter` prop to trigger a callback when pressing the "Enter" key while holding the meta key down (i.e. "Command" key or "Windows" key) on a focused `TreeItem`. If `onMetaEnter` receives a callback, then the `onClick` callback will not be triggered when pressing the "Enter" key plus the meta key.
```
dbchristopher, line 186: It's kind of awkward logic to explain in writing lol, so here's a couple of possible edits to make it read a little more smoothly:

```suggestion
Use the `onMetaEnter` prop to trigger a callback when pressing the "Enter" key in combination with the meta key (i.e. "Command" or "Windows" key) on a focused `TreeItem`. If no `onMetaEnter` handler is specified, then it will fall back to the `onClick` handler by default.
```

dbchristopher (changes requested):
This looks great @ElliotTheHuman! I just suggested a small refactor to simplify the logic branching a bit and move away from the deprecated `keyCode` property.

packages/components/src/Tree/TreeItem.tsx
ElliotTheHuman: I actually wasn't sure! I thought it could be confusing for a user if pressing `cmd+enter` didn't do anything (when pressing `enter` does), so figured it might be best to this fallback behavior.

That being said, I'm definitely not tied to this! What do you recommend regarding the "`TreeItem` has an onClick but no onMetaEnter"?

dbchristopher: I think I'd assume that typing `cmd+enter` just wouldn't do anything in that case!

dbchristopher: And now that I write that out, I think it does feel like the right choice. Each key or key chord combination should really only be tied to one possible side effect

ElliotTheHuman: Yeah, that does make more sense that you mention it 🤦‍♀️ 

Thanks for explaining!

ElliotTheHuman: Resolved in ec76144

packages/components/src/Tree/TreeItem.tsx
ElliotTheHuman: Didn't know that `keyCode` was deprecated. Neat! This seems way more readable that `keyCode`.

Resolved in ec76144

packages/components/src/Tree/TreeItem.tsx
ElliotTheHuman: Agreed on all fronts. Also verified that removing the `event.target === event.currentTarget` pattern doesn't affect behavior of the `TreeItem`. 

Also, spoke over Slack, but really like using "noop" as a default value to remove undefined checks for passed-in handlers. Going to apply that to all passed-in handlers.

Resolved in f59a36f

www/src/documentation/components/content/tree.mdx
ElliotTheHuman: Yeah definitely not my finest sentence ever written haha. Thanks for the suggestion, much cleaner than the original.

Also removed the fallback behavior note since that's been removed (i.e. hitting `cmd+enter` will only ever trigger `onMetaEnter` or a noop).

Resolved in 352b24e

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -197,7 +211,7 @@ const TreeItemLayout: FC<TreeItemProps> = ({
         onKeyDown={handleKeyDown}
         onKeyUp={handleKeyUp}
         ref={itemRef}
-        tabIndex={onClick ? 0 : -1}
+        tabIndex={isTreeItemTabbable}
```
dbchristopher, line 214: good catch!

dbchristopher (approved):
looks good!

PR #1566, opened 2020-10-14
IconButton toggle / remove color support

jhardy:
### :sparkles: Changes

- This pr add's a `toggle` state to `IconButton` that lets it represent a toggled on and toggled off state
- Added RTL tests where using snapshots were previous used poorly
- Added image-snapshots for toggle on/off IconButton
- Remove the color prop on IconButton (neutral only now)

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


packages/components/src/Button/IconButton.story.tsx
```diff
@@ -100,6 +100,12 @@ CriticalColorIconButton.args = {
   color: 'critical',
 }
 
+export const Toggle = Template.bind({})
+Toggle.args = {
+  ...Basic.args,
+  toggle: true,
+}
+
```
jhardy, line 108: added this but did not run yet, will run with snapshots are fixed

packages/components/src/Button/IconButton.tsx
```diff
@@ -185,6 +190,7 @@ const IconButtonComponent = forwardRef(
     return (
       <ButtonBase
         aria-describedby={ariaDescribedBy}
+        aria-pressed={toggle ? 'true' : 'false'}
```
jhardy, line 193: In this case aria-pressed would alway be false, which seems like the wrong approach, so wanted to chat a bit about this. Perhaps we should only set this to true if there if the toggle prop is true and do not include `aria-pressed` if its false?

packages/components/src/Button/IconButton.tsx
```diff
@@ -250,8 +256,14 @@ export const iconButtonColor = css<Pick<IconButtonProps, 'color'>>`
   &[aria-expanded='true'],
   &:active,
   &.active {
-    color: ${({ theme, color = iconButtonDefaultColor }) =>
-      theme.colors[`${color}Pressed`]};
+    color: ${({ theme, toggle, color = iconButtonDefaultColor }) =>
+      toggle !== undefined
+        ? theme.colors.key
+        : theme.colors[`${color}Pressed`]};
+  }
```
jhardy, line 263: I'd love to remove the color prop on `IconButton`. It would clean this up and be closer to the spec we want to head to.

jhardy:
@lukelooker wanted your take on this approach before I move out of draft

lukelooker:
@mdodgelooker - This PR became a Jared / Luke co-effort so in the interest of impartial review I added you as a reviewer :)

packages/components/src/Button/IconButton.test.tsx
```diff
@@ -26,8 +26,7 @@
 
 import 'jest-styled-components'
 import React from 'react'
-import noop from 'lodash/noop'
-import { assertSnapshot, renderWithTheme } from '@looker/components-test-utils'
+import { renderWithTheme } from '@looker/components-test-utils'
```
lukelooker, line 29: Down with snapshot tests!

lukelooker (approved):

mdodgelooker (commented):
@jhardy should `toggle` get `keyInteractive` and `keyPressed` for hover and expanded (e.g. with an open popover) states? And should the border color match the icon color or always be neutral?

packages/components/src/Button/IconButton.test.tsx
```diff
@@ -37,128 +36,47 @@ import { Tooltip } from '../Tooltip'
...
-      <IconButton
-        id="test-iconButton"
-        label="Test"
-        icon="Favorite"
-        aria-haspopup
-      />
-    )
+    renderWithTheme(<IconButton label="Test" icon="Favorite" />)
+    const button = screen.getByRole('button')
+    expect(button).not.toHaveAttribute('aria-haspopup')
```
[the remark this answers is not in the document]
lukelooker: Not sure what I was doing here. That _was_ my intent I swear. 🤦

jhardy:
> @jhardy should `toggle` get `keyInteractive` and `keyPressed` for hover and expanded (e.g. with an open popover) states? And should the border color match the icon color or always be neutral?

@mdodgelooker right now we are just going to apply `key` on `toggle`. When we resolve a few other inconsistencies with Icon Button we are going to be adding support for background color updates when hovered and toggled.

mdodgelooker (approved):
Got it @jhardy 👍 LGTM!

PR #1567, opened 2020-10-14
FieldTimeSelect Field Label aria fix

dbchristopher:
### :sparkles: Changes

- Properly passes through IDs and Aria Labels for better accessibility on FieldTimeSelect
- Sets up snapshot integration for FieldTimeSelect

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


.gitignore
```diff
@@ -12,6 +12,7 @@ node_modules
 /coverage
 .cache
 .eslintcache
+.jest-image-snapshot-touched-files
```
dbchristopher, line 15: this popped up as a new untracked file... I assume it's not supposed to be checked in?

dbchristopher:
@mdodgelooker here's a small one I'd love to get your eyes on, to make sure I'm handling accessibility correctly with combobox.

.gitignore
lukelooker: Let's not add this to `.gitignore` because it should get reviewed and explicitly deleted and I'm worried that if `.gitignore` we'll miss another breakage (it's signal that image-snapshots found "orphaned" snapshot images)

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Space` revert mistakenly applied `flex-shrink: 0`
 - Reverts: `HoverDisclosure` toggles visibility with css rather than inserting elements into the DOM
 - Fix `image-snapshots` issue
+- Aria accessibily fixes for `FieldTimeSelect` label
```
lukelooker, line 24: `0.9.18` cut so this needs to move up ☝️

lukelooker (commented):
Needs changelog update and gitignore removal. Then it should be GTG.

.gitignore
dbchristopher: ah ok cool, that makes sense!

dbchristopher:
ok @lukelooker, updated!

lukelooker (approved):
👍

mdodgelooker:
Belated LGTM. I spaced – I should've added the `ariaProps` util here to begin with. 

I was going to suggest that the `ariaProps` also be applied to `ComboboxList` so that screen readers will announce the currently highlighted option when keyboard navigating, but a quick check of `Select` with VoiceOver tells me that's not working like I thought it was. Here's a ticket to fix that behavior this https://looker.atlassian.net/browse/LENS-1311 and this time I will not forget about `InputTimeSelect`! 😉

dbchristopher:
Awesome, thanks @mdodgelooker!

PR #1575, opened 2020-10-16
InputTimeSelect UX Improvements

dbchristopher:
https://looker.atlassian.net/browse/LENS-1257

### :sparkles: Changes

- Allows you to enter a time that is not included in the select dropdown options

- [] Includes test coverage for all changes
- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


packages/components/src/Form/Inputs/InputTimeSelect/InputTimeSelect.tsx
```diff
@@ -301,6 +314,17 @@ const InputTimeSelectLayout = forwardRef(
...
+      throttle(
+        (v: MaybeComboboxOptionObject) => {
+          handleChange(v)
+        },
+        50,
+        { trailing: false }
+      ),
+      []
+    )
+
```
dbchristopher, line 327: I hope this isn't just band-aiding over a real problem. The issue I found was that when you have an option selected and then try to enter something custom (not one of the pre-defined options), it would accept the change and then combobox would quickly revert it to the last known valid option. You can see the console output here for illustration:

combobox-revert

So my solution was simply to throttle the change handler so that only the first change event would be accepted. @mdodgelooker I'm curious how you feel about this approach. I'm hesitant to start trying to change the combobox internals as it seems like that "bug" is actually the behavior we'd want most of the time.

packages/components/src/Form/Inputs/InputTimeSelect/InputTimeSelect.tsx
```diff
@@ -182,13 +184,24 @@ const createOptionFromStringValue = (format: TimeFormats, value: string) => {
...
+const convert12to24hr = (hour: number, period: 'am' | 'pm') => {
+  if (hour + period === '12am') {
+    return 0
+  } else if (period === 'pm' && hour < 12) {
+    return hour + 12
+  } else {
+    return hour
+  }
+}
+
```
dbchristopher, line 197: Expanded logic fixes a bug where it would change anything around 12:00 **am** to 12:00 **pm** upon hitting `enter`

midnight-bug

packages/components/src/Form/Inputs/InputTimeSelect/InputTimeSelect.tsx
```diff
@@ -309,13 +335,21 @@ const InputTimeSelectLayout = forwardRef(
...
+      const { key } = e
+      if (arrowKeys.includes(key)) {
+        setIsNavigating(true)
+      } else if (key === 'Enter' || key === 'Tab') {
         if (inputTextValue.length) {
           // allow entering shortcuts like `2pm` to select `02:00 pm` on enter
           const option = createOptionFromLabel(format, inputTextValue)
-          handleChange(option)
+          // don't fire change event here if user is selecting an option from the dropdown list
+          !isNavigating && throttledHandleChange(option)
```
dbchristopher, line 349: Fixes a bug (caused by throttling the change event), where you could no longer select an option from the ComboboxList after typing a custom value. Again because of competing change events between the text input and combobox internals

packages/components/src/Form/Inputs/InputTimeSelect/InputTimeSelect.tsx
dbchristopher: I also wonder if this input _should_ be more constrained? I wonder if I should just encourage them to use [InputTime](https://components.looker.com/components/forms/input-time/) instead if they really need down-to-the-minute precision. @jhardy do you have any thoughts about that?

jhardy: I'm on the fence about constraining the component. I think if were were to remove the ability to have down-to-the-minute precision we would need to remove the ability to filter down the list, which would make the component clunky and harder to use.  We could potentially look at other filtering patterns, but that would need a bunch more work to go and design and implement anything new.

Ideally we would support the ability to have the minute precision, but at the end of the day, I believe, consistency in functionality matters more, so if we need to remove that ability for technical reasons right now then we should bias to having it work consistently across keyboard and mouse interactions. If we do decide to constrain the component, we would need to do a little messaging with some teams before we make that change, just to make sure they are on board and do not have any major objections.

mdodgelooker (commented):
@dbchristopher I wish I had a clear answer for you about how to avoid the `Combobox` value clobbering the `ComboboxInput` value in this situation. I thought maybe the new `freeInput` prop on `ComboboxInput` would be helpful here – it enables `InputSearch` to keep the input's value as primary – but fiddling around with it for a while and just could not get it to work. I couldn't figure out how to how to let the input value be the actual value but only if it was a valid time. You might have better luck, if the `freeInput` prop seems promising to you.

The only other suggestion I have would be that you could do something like this inside `handleKeyDown` to stop the `Combobox` change from happening.
```
if (!matchStringLabelToOption(timeOptions, inputTextValue)) {
  e.preventDefault()
}
```
However, I played around with what you've got here and it seems to fix the issue just fine, so I have no issue with this solution if neither of the above seem viable.

packages/components/src/Form/Inputs/InputTimeSelect/InputTimeSelect.test.tsx
```diff
@@ -28,9 +28,10 @@ import React from 'react'
 import { fireEvent } from '@testing-library/react'
 import { renderWithTheme } from '@looker/components-test-utils'
 import map from 'lodash/map'
-
 import { InputTimeSelect, InputTimeSelectProps } from './InputTimeSelect'
 
+jest.mock('lodash/throttle', () => (fn: any) => fn)
```
dbchristopher, line 33: I need to mock `throttle` to get the tests to pass at all, which prevents me from writing a test that depends on the throttling behavior 🤦

dbchristopher:
@mdodgelooker I tried poking around at this a bit more, and I think the throttling is going to be the best option. Using `preventDefault` I did occasionally wind up in a state where I couldn't select the option using the up/down arrow keys, and I'm not sure why. 

So I think I'll count this as ready for final review if you have time to give it another look!

mdodgelooker (approved):
LGTM!

PR #1592, opened 2020-10-20
Tooltip Delay

mdodgelooker:
### :sparkles: Changes

- Leveraged `useAnimationState` to delay `Tooltip` show/hide
  - I didn't add =actual animation, just made it hidden while `entering` and shown while `exiting`
  - I had to make a slight modification to `useAnimation` to prevent undesirable behavior (see screenshot below)
- Added `TooltipSurface`, a slight variation of `OverlaySurface` (which does actually have a fade-in animation built-in, still works great)
- Added `disableDelay` prop

- [x] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### :camera: Screenshots
If the `mouseout` event occurs during `entering`, the state now changes to `exited` instead of `exiting`, to prevent this behavior (you can't see my mouse but I'm hovering over the button "accidentally" and not pausing). This applies to `Dialog` as well, but I don't think it really impacts the user. It did impact the unit tests.
e651b93d-8032-4326-9827-f93bb3d50157


packages/components/src/Button/IconButton.test.tsx
```diff
@@ -28,13 +28,26 @@ import 'jest-styled-components'
...
+})
+afterEach(() => {
+  jest.runOnlyPendingTimers()
+  jest.useRealTimers()
+})
+const runTimers = () =>
+  act(() => {
+    jest.runOnlyPendingTimers()
+  })
+
```
mdodgelooker, line 50: Annoyingly, the change I made to `useAnimationState` (see the PR description) caused a bunch of tests to fail because elements now unmount immediately when you open and close them without giving time fore the `entering` to complete. So I added a this.

packages/components/src/Button/IconButton.tsx
```diff
@@ -148,13 +148,15 @@ const IconButtonComponent = forwardRef(
...
-      'aria-describedby': ariaDescribedBy,
-      ref,
+      domProps: {
+        'aria-describedby': ariaDescribedBy,
+        ref,
+        onFocus,
+        onBlur,
+        onMouseOver,
+        onMouseOut,
+      },
```
mdodgelooker, line 158: Figured it was a good time to change the interface to match `usePopover`.

packages/components/src/Menu/MenuDisclosure.tsx
```diff
@@ -130,7 +133,9 @@ export const MenuDisclosure: FC<MenuDisclosureProps> = ({
       'aria-describedby': ariaDescribedBy,
       'aria-expanded': isOpen,
       'aria-haspopup': true,
-      className: `${childProps.className || ''}${isOpen ? ' active' : ''}`,
+      className: `${childProps.className || ''}${
+        isOpen ? ' active' : ''
+      } ${tooltipClassName}`,
```
mdodgelooker, line 138: Realized this was missing the "hover" className that keeps the hover styles on while the tooltip is open.

packages/components/src/Overlay/OverlaySurface.tsx
```diff
@@ -56,21 +56,22 @@ export interface SurfaceStyleProps extends BorderProps, BoxShadowProps {
 
 export interface OverlaySurfaceProps extends SurfaceStyleProps {
   children: ReactNode
-  eventHandlers?: DOMAttributes<{}>
+  className?: string
```
mdodgelooker, line 59: For `entering`, `entered`, ...etc.

www/src/MDX/Pre/CodeSandbox.tsx
```diff
@@ -234,9 +234,8 @@ const SandboxWrapper = styled.div`
 `
 
 const PreviewWrapper = styled.div`
-  ${({ theme: { space } }) => `
-    padding: ${space.small};
-  `}
+  color: ${({ theme }) => theme.colors.text};
```
mdodgelooker, line 237: Fixes a white text on white background situation in the `Truncate` page example.

packages/components/src/Tooltip/TooltipSurface.tsx
```diff
@@ -0,0 +1,39 @@
...
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+import styled from 'styled-components'
+import { OverlaySurface } from '../Overlay/OverlaySurface'
+
+export const TooltipSurface = styled(OverlaySurface)`
```
jhardy, line 30: The behavior we are going for adds a pause before the tooltip is rendered.


I was able to get it by adding a few lines here: 

1. Only transition the opacity (otherwise the delay applies to moving the tooltip when its first inserted into the dom)
2. Delay transition and animation for the same amount of time

```  
transition: opacity;
transition-delay: 500ms;
animation-delay: 500ms;
```

jhardy:
Another thing on my mind, is do we want to make tooltip delay customizable at all? My initial thinking is that somewhere between 300-500ms is a good pause for a tooltip delay, but in some cases perhaps for truncation we want that to be a bit longer. Any thoughts there?

jhardy:
Another piece of feedback we received was that tooltips in some cases can be transient. Where mousing over the tooltip content does not keep the tooltip open. 

Is something like that in scope for this change? I am not sure how realistic or hard adding that change in would be, but it would be nice option for things like Icon Buttons, lists of items with lots of tooltips, and in general most places tooltips are used. This could also clear up some distinctions between tooltip vs truncation component. 

If out of scope its something we can add a new ticket for and track later

packages/components/src/Tooltip/TooltipSurface.tsx
mdodgelooker: Ooh @jhardy I totally should've asked whether the delay on hide was needed. It'll make things so much simpler not to have that. :) I'm not totally following the 3 lines though – do they replace or add to the existing CSS? Either way, just need to make sure the "hover up a truncated `Tree`" behavior still works smoothly on http://meredithdodge.c.googlers.com:8000/components/content/tree#truncating-tree-items--labels

mdodgelooker:
> Another thing on my mind, is do we want to make tooltip delay customizable at all? My initial thinking is that somewhere between 300-500ms is a good pause for a tooltip delay, but in some cases perhaps for truncation we want that to be a bit longer. Any thoughts there?

Totally up to you – very doable.

mdodgelooker:
> > Another thing on my mind, is do we want to make tooltip delay customizable at all? My initial thinking is that somewhere between 300-500ms is a good pause for a tooltip delay, but in some cases perhaps for truncation we want that to be a bit longer. Any thoughts there?
> 
> Totally up to you – very doable.

This should already be in place. Wherever it's not working is a bug.

mdodgelooker:
> Another piece of feedback we received was that tooltips in some cases can be transient. Where mousing over the tooltip content does not keep the tooltip open.
> 
> Is something like that in scope for this change? I am not sure how realistic or hard adding that change in would be, but it would be nice option for things like Icon Buttons, lists of items with lots of tooltips, and in general most places tooltips are used. This could also clear up some distinctions between tooltip vs truncation component.
> 
> If out of scope its something we can add a new ticket for and track later

This should already be in place. Wherever it's not working is a bug.

mdodgelooker:
@jhardy I wasn't able to get a CSS-only solution (I think that's what your first comment was about, but I may have misinterpreted?) that covered:
1. Gets announced in VoiceOver (`visibility: hidden` doesn't)
2. Plays nicely with a truncated tree (need `pointer-events: none` if using `opacity: 0` – during the delay only)
3. Cross-browser support (Safari ignores `pointer-events: none` when it's in an animation)

But I was able to update the `useAnimationState` implementation and it works nicely (and no need for all the updates to `Dialog` tests!). Let me know what you think – and whether you'd like the delay to be customizable.

@lukelooker the `Dialog` image snapshots are broken, but they are in master too, due to `useAnimationState` – I think `CSSTransition` didn't animate at all for `defaultOpen`. Wanna chat about a solution?

mdodgelooker:
Whoops my changes broke the gatsby build. Fixing that and also, after chatting with @lukelooker, making the delay customizable FYI @jhardy :)

jhardy:
> > Another piece of feedback we received was that tooltips in some cases can be transient. Where mousing over the tooltip content does not keep the tooltip open.
> > Is something like that in scope for this change? I am not sure how realistic or hard adding that change in would be, but it would be nice option for things like Icon Buttons, lists of items with lots of tooltips, and in general most places tooltips are used. This could also clear up some distinctions between tooltip vs truncation component.
> > If out of scope its something we can add a new ticket for and track later
> 
> This should already be in place. Wherever it's not working is a bug.

@mdodgelooker is this a prop that needs to be placed on the tooltip? Everywhere I did a spot check with a components tooltip the tooltip stays open when moving to its surface. Here is an example from the docs

mdodgelooker:
@jhardy ah sorry I totally misread your earlier comment. Yes, currently the tooltip stays open if you hover into the surface. I can see how adding the "transient" option could would help with the tricky `Tree` behavior, but wouldn't the inability to copy-paste the full content of the truncated item also be annoying? @lukelooker thoughts?

Otherwise, ready for re-review. 🙂

jhardy:
I think in an ideal world what would happen is our tooltip is now would have a boolean prop to be transient as an option and we would probably set that prop to true on most of our tooltips. We do have a few places where people put links inside of tooltips but that is frowned upon by our parent design system, so could be a good forcing function to help us move away from those patterns.  And then in my mind `Truncate` would turn always be false so you could get to and copy the content if needed.

mdodgelooker:
@jhardy that makes sense to me! I know there’s some urgency for the truncated Tree so perhaps we can leave the transient feature for a future PR since it wouldn’t affect that use case?

packages/components/src/Button/IconButton.tsx
lukelooker: Not for this PR but I wonder if we could use something like `data-hover` so we don't clobber existing className in the future.

lukelooker (approved):
There was a minor issue in the story where it was still referring to `disableDelay` (instead of `delay="none"`. I pushed a fix for that and otherwise this looks great!

I agree that we should address the "point-events: none" enabling feature separately – might be a pretty easy lift but it seems tangental (although not unrelated) to this change.

lukelooker (approved):

PR #1597, opened 2020-10-21
Select: assorted fixes

mdodgelooker:
### :sparkles: Changes

1. Fixed: On a mobile device or with the Mac trackpad "tap to click" setting enabled, clicking on an option closes & immediately re-opens the list.
2. Fixed: In a controlled `Select`, if there's any delay (even 0ms) between a `onChange` being called and `value` updating, the `Select` loses focus.
    - Broken: https://codesandbox.io/s/brave-frost-eh9lf
    - Fixed: `:3333/?path=/story/fieldselect--delay-update`
3. Fixed: Screen reader not announcing options during keyboard navigation via up & down arrows. (It was a misplaced `aria-activedescendent`.)

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots
Trackpad setting:
image


CHANGELOG.md
```diff
@@ -25,10 +25,6 @@ NOTE: This is the CHANGELOG for the @looker/components MONOREPO. Each package ha
     - `develop` - disables "Docs" addon
     - `publish` - enables Typescript extraction and full `addon-essentials` support
 
-### Fixed
-
-- `MenuItems` within the same `MenuList` align consistently when icons are used
-
```
mdodgelooker, line 31: Noticed this was in the wrong spot during a conflict merge.

packages/components/src/Form/Fields/FieldSelect/FieldSelect.story.tsx
```diff
@@ -593,6 +599,34 @@ CreateOption.parameters = {
   storyshots: { disable: true },
 }
 
+export const DelayUpdate = () => {
```
mdodgelooker, line 602: Example showing fix for issue #2

packages/components/src/Form/Inputs/Combobox/ComboboxMultiInput.tsx
```diff
@@ -151,44 +159,37 @@ export const ComboboxMultiInputInternal = forwardRef(
...
-        // this is most likely the initial value so we want to
-        // update the value without transitioning to suggesting
-        transition &&
-          transition(ComboboxActionType.CHANGE_SILENT, {
-            inputValue: controlledInputValue,
-          })
       }
-    }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [controlledInputValue])
```
mdodgelooker, line 185: This is a bit of code , tangentially related to the changes above, for handling a controlled `inputValue` (as opposed to `values`, a.k.a. the `Chip`s) that I realized was 1) completely broken, but 2) not being used in `SelectMulti`, which is the only implementation of this component (that I know of).

I had some indecision over whether to fix it or just remove the ability to control `inputValue`. I decided to fix it in the end, since I know implementations of `Combobox` will only get more complex. I could see an email address input needing to control the `inputValue`, for example.
Broken: https://codesandbox.io/s/epic-bush-g3lvl
Fixed: `:3333/?path=/story/combobox--controlled-input-value`

packages/components/src/Form/Inputs/Combobox/utils/state.ts
```diff
@@ -305,11 +308,14 @@ export function getOptionsFromValues(
...
-const reducerMulti: Reducer<
+export const reducerMulti: Reducer<
   ComboboxMultiData,
   ComboboxMultiActionWithPayload
 > = (data, action) => {
-  const nextState = { ...data, lastActionType: action.type }
+  const nextState = { ...data }
+  if (action.type.indexOf('_SILENT') === -1) {
+    nextState.lastActionType = action.type
+  }
```
mdodgelooker, line 318: This was the fix for the 2nd issue, where a controlled `Select` with any delay b/t `onChange` and an updated `value` was losing focus when clicking an option. This was because the mismatch between the prop `value` and the value from context was triggering a `SELECT_SILENT` – to sync the latter to the former – and the `useFocusManagement` hook that returns focus to the input after a click on an option, is looking for the `lastActionType` to be `SELECT_WITH_CLICK`.

Excluding "SILENT" action types here feels like the right move – they're used only to sync context state to controlled props, not to trigger interactive behavior in the component.

packages/components/src/Form/Inputs/Combobox/utils/state.ts
```diff
@@ -386,44 +392,27 @@ const reducerMulti: Reducer<
...
+    stateRef.current = state
+  }, [state])
 
   function transition(
     action: ComboboxActionType,
-    payload: ComboboxMultiActionPayload = { inputValues: [], options: [] }
+    payload: TPayload = defaultPayload
   ) {
-    const currentState = stateChart.states[state]
+    const currentState = stateChart.states[stateRef.current]
```
mdodgelooker, line 415: Removing the `requestAnimationFrame` delay in `useBlur` to fix the "tap to click" bug caused a bug where the list would fail to close on blur for a `SelectMulti` with `freeInput`. This was caught by "creates value and closes list on blur" in `SelectMulti.test.tsx`! 🎉

packages/components/src/Form/Inputs/Combobox/utils/state.ts
```diff
@@ -386,53 +392,31 @@ const reducerMulti: Reducer<
...
-  const [data, dispatch] = useReducer(reducerMulti, initialData)
+export function useReducerMachine<
+  TData = ComboboxData,
+  TPayload = ComboboxActionPayload
+>(
+  reducerFn: Reducer<TData, ComboboxAction & TPayload>,
+  initialData: TData,
+  defaultPayload: TPayload
+): [ComboboxState, TData, ComboboxTransition<TPayload>] {
+  const stateRef = useRef(stateChart.initial)
```
mdodgelooker, line 403: Removing the `requestAnimationFrame` delay in `useBlur` to fix the "tap to click" bug caused a different bug where the list would fail to close on blur for a `SelectMulti` with `freeInput`. 

The issue was that `transition` is called with `BLUR` action, then immediately with `CHANGE_VALUES` action due to the tokenization of the input value on blur. Previously, the `requestAnimationFrame` in `useBlur` put the `BLUR` transition in the next cycle _after_ `CHANGE_VALUES`. 

Because `useState` [batches updates](https://stackoverflow.com/questions/53048495/does-react-batch-state-update-functions-when-using-hooks), the state update from `SUGGESTING` to `IDLE` from the `BLUR` action "loses" to the `CHANGE_VALUES` action, which keeps the state as `SUGGESTING`. Using a ref to capture the intermediate change to `IDLE` gets around this issue.

packages/components/src/Form/Inputs/InputChips/InputChips.tsx
```diff
@@ -154,12 +155,15 @@ export const InputChipsInternal = forwardRef(
       ? controlledInputValue || ''
       : uncontrolledValue
 
-    const setInputValue = (val: string) => {
+    const setInputValue = (
+      val: string,
+      event?: FormEvent<HTMLInputElement>
```
mdodgelooker, line 160: Adding the event here so `ComboboxMultiInput` can distinguish between changes directly from the user typing and "side-effect" changes, like the input value reverting to empty after a value is entered.

packages/components/src/Form/Inputs/InputProps.ts
```diff
@@ -85,6 +85,7 @@ export const inputPropKeys = [
   'pattern',
   'step',
   'value',
+  'aria-activedescendant',
```
mdodgelooker, line 88: Adding this to the list of attributes that get passed down to the inner input as opposed to the wrapper div fixes the screen reader issue.

packages/components/src/Form/Inputs/Select/SelectMulti.test.tsx
```diff
@@ -238,26 +233,30 @@ describe('closeOnSelect', () => {
...
+          />
+        </>
       )
 
       const input = screen.getByPlaceholderText('Search')
+      input.focus()
       fireEvent.change(input, { target: { value: 'baz' } })
       expect(screen.getByRole('listbox')).toBeVisible()
-      fireEvent.blur(input)
+      screen.getByText('A random button').focus()
```
mdodgelooker, line 251: This test needed some adjustments after removing the `requestAnimationFrame` in `useBlur`.

mdodgelooker:
@lukelooker & @dbchristopher this is finally ready for review. The bulk of the changes are for the 1st issue – the "tap to click" re-opening the list – and it's pretty convoluted. Let me know if you'd like chat about it for more clarity. Thank you! 🙏

packages/components/src/Form/Inputs/Combobox/utils/useBlur.ts
```diff
@@ -48,38 +48,36 @@ export function useBlur<
...
-          closeList()
-        }
-      })
+      if (focusInList && state !== ComboboxState.INTERACTING) {
+        // focus landed inside the select, keep it open
+        transition && transition(ComboboxActionType.INTERACT)
+      } else if (!focusInList && nextFocusTarget !== inputElement) {
+        // focus landed outside the select, close it
+        closeList(ComboboxActionType.BLUR)
+      }
```
mdodgelooker, line 80: Fixes the tap-to-click list re-opening bug because with tap-to-click there is no delay between blur & click.

With a traditional mouse click there's a tiny delay after blur & before click, so that even with the `requestAnimationFrame`, the `SELECT_WITH_CLICK` action (closes the list) was still coming after the `INTERACT` action (keeps the list open). With tap-to-click, the `requestAnimationFrame` causes `INTERACT` to come after `SELECT_WITH_CLICK`, thus closing then re-opening the list.

The `requestAnimationFrame` was originally in here b/c the `focusInList` check was looking at `document.activeElement`, which gets updated _after_ the blur event. This is no longer the case with `getNextFocusTarget`.

The other thing it was doing was in `SelectMulti` with `freeInput`, making sure the `BLUR` action came after the `CHANGE_VALUES` action that is triggered when the input value gets tokenized on blur. This was tricky to fix but you'll read more about it in my other comments. :)

packages/components/src/Form/Inputs/Combobox/ComboboxMultiInput.tsx
```diff
@@ -151,44 +159,37 @@ export const ComboboxMultiInputInternal = forwardRef(
...
     const isControlled = controlledInputValue !== undefined
     // [*]... and when controlled, we don't trigger handleValueChange as the user
-    // types, instead the developer controls it with the normal input onChange
-    // prop
+    // types, instead the developer controls it with the onInputChange prop
     const handleInputChange = useCallback(
-      (value: string) => {
-        isInputting.current = true
+      (value: string, event?: FormEvent<HTMLInputElement>) => {
+        isInputting.current = event !== undefined
```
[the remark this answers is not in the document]
dbchristopher: smart

packages/components/src/Form/Inputs/Combobox/stories/Combobox.story.tsx
```diff
@@ -150,6 +150,37 @@ ComboboxDemo.parameters = {
...
+          <ComboboxMultiOption value="Pineapples" />
+        </ComboboxMultiList>
+      </ComboboxMulti>
+    </SpaceVertical>
+  )
+}
+
+ControlledInputValue.parameters = {
+  storyshots: { disable: true },
+}
```
dbchristopher, line 182: I think there's a weird bug here I don't quite understand. It seems like once I clear the list, and then add an option again, the "Change Input Value" button no longer works

combobox-bug

packages/components/src/Form/Inputs/Combobox/utils/state.ts
dbchristopher: yeah that makes sense to me!

dbchristopher (changes requested):
Wow, this is some really epic sleuthing and debugging. This component is so dang complicated! I have no feedback on the code changes, but it seems like that controlled combobox bug should probably be looked into before merging. 

But of course, this generally seems _better_ than the current combobox in significant ways, so I'm happy to merge as-is and revisit that issue later if you want

packages/components/src/Form/Inputs/Combobox/stories/Combobox.story.tsx
mdodgelooker: Great catch! Got a fix in the works now...

mdodgelooker: @dbchristopher the issue was that the `inputValue` in `ComboboxContext` changes to `''` when a new option is selected, but `onInputChange` wasn't getting called, so the controlled input value was still on `'bananas'` even though the input showed an empty value. Then clicking the button did nothing b/c it wasn't actually changing the state value (it was already `'bananas'`).

`onInputChange` _should_ be called with the input value is reset to `''` due to an option being selected. I made some updates so that this does happen now, and actually I was able to remove a couple lines of code in `SelectMulti.tsx` as a result. 👍 Thanks again for catching this.

dbchristopher (approved):
Awesome, I'm glad you were able to get that dialed in!! It all looks great 👍 👍

PR #1602, opened 2020-10-22
Fix: IE11 click option in Select, etc

mdodgelooker:
### :sparkles: Changes

IE11 doesn't add `relatedTarget` to a blur event so the logic to keep the list open on input blur if the user is clicking into the list was failing, the list was closing prematurely, and the click on the option failed to register.

I added a utility `getNextFocusTarget` that uses `relatedTarget` if available and `document.activeElement` if not. The reason to not use `document.activeElement` across the board is that only IE11 updates that value _before_ the blur event. Other browsers don't update till _after_ the blur event.

`InputChips` was also using a blur event's `relatedTarget` for selecting chips behavior so I used `getNextFocusTarget` there as well.

Both components can be tested in the playground.

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `Select` / `SelectMulti` / `InputTimeSelect` click to select option in IE11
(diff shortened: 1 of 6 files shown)


lukelooker (approved):
Nice fix (silly IE).

Verified in Chrome and IE11.

PR #1604, opened 2020-10-22
Disable react docgen

mdodgelooker:
### :sparkles: Changes

React docgen is one of the best features of storybook, but it's making it almost unusably slow. ~~Unfortunately, removing it does not appear to impact the speed of `yarn image-snapshots`.~~ Actually, it does appear to speed up snapshots. It finished in under 10 min.

https://github.com/storybookjs/storybook/blob/next/addons/docs/react/README.md#typescript-props-with-react-docgen

https://github.com/storybookjs/storybook/issues/7998


lukelooker:
Supplanted with #1612

PR #1606, opened 2020-10-23
Add groupedMenusRef prop to menu to support better UX in a group of menus

hkbarton:
### :sparkles: Changes

- **Thank you for your contribution to Looker Components!**
- To ensure a successful review, please double check our checklist of requirements below.
- Also including a short description of changes and relevant screenshots will help us more quickly understand the ultimate goal of your work (🗑 replace this introduction with your summary of changes).

- [x] Update documentation
- [ ] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/Menu/Menu.tsx
+++ b/packages/components/src/Menu/Menu.tsx
@@ -50,11 +50,19 @@ export interface MenuProps {
    * Use for controlled menu
    */
   setOpen?: (isOpen: boolean) => void
+
+  /**
+   * This prop exposes the groupedPopoversRef prop of the wrapped Popover component rendered by MenuList.
+   * This allows transitions between menus in a group without clicking twice
+   * (first to dismiss the current menu, and again to open the target menu)
+   */
+  groupedMenusRef?: RefObject<HTMLElement>
 }
 
 export const Menu: FC<MenuProps> = ({
   children,
   disabled,
+  groupedMenusRef,
   hoverDisclosureRef,
   id: propsID,
   isOpen: controlledIsOpen = false,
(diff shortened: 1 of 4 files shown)


packages/components/src/Menu/Menu.tsx
```diff
@@ -50,11 +50,19 @@ export interface MenuProps {
    * Use for controlled menu
    */
   setOpen?: (isOpen: boolean) => void
+
+  /**
+   * This prop exposes the groupedPopoversRef prop of wrapped Popover component in MenuList.
+   * It solves the problem when clicking between different menu in a group of menus, you don't
+   * end up with clicking twice(first click dismiss current menu, second click opens target menu).
```
AprilArcus, line 57: Minor grammatical suggestions:

"of wrapped" -> "of the wrapped"
"component in MenuList" -> "component rendered by MenuList"
"This allows transitions between menus in a group without clicking twice (first to dismiss the current menu, and again to open the target menu"

Also, maybe include an example of how to use it?

AprilArcus (changes requested):

AprilArcus:
LGTM. Can we get a review from someone on the components team? @lukelooker @mdodgelooker ?

lukelooker (approved):
LGTM. 

Two caveats -

1. All changes need a CHANGELOG entry (but I'm cutting a release immediately after this lands so I'll go ahead and do so since branching a fork is a fun adventure).
2. `groupedPopoversRef` is slated for deprecation and removal next sprint (next week). After a lot of discussion we've decided to reverse the default of `cancelClickOutside` (it will be disabled by default). While there are arguments that this is slightly more "dangerous" in certain scenarios (an errant click to dismiss a `Popover` could also trigger a destructive action) the amount of confusion the feature has caused has outpaced our ability / energy to keep educating folks about. With that in mind - expect `groupedMenusRef` to be removed at the same time. On the plus side, when removed it will act the same since the change `cancelClickOutside` will effectively imbue this behavior everywhere by default.

PR #1608, opened 2020-10-23
Fix Select + Dialog focus issue

mdodgelooker:
### :sparkles: Changes

- Test at `:3333/?path=/story/fieldselect--select-demo`
- Clicking an option on a `Select` in a `Dialog` results in the `Select` losing focus
- Because the focus trap is being re-enabled when the `Select` list closes, and now surface itself is set to `initialFocus`
- (This was probably already happening if there was a `[data-autofocus="true"]` element in the `Dialog`)

- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


lukelooker (approved):
Ship it :)

PR #1614, opened 2020-10-26
Correct buttonCSS typing

lukelooker:
### :sparkles: Changes

- Fix TS issue that 3.9.x TS makes evident.

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/Button/ButtonBase.tsx
+++ b/packages/components/src/Button/ButtonBase.tsx
@@ -76,13 +76,13 @@ export interface ButtonProps extends ButtonBaseProps, ButtonIconProps {
   fullWidth?: boolean
 }
 
-export const buttonCSS = css<ButtonBaseProps>`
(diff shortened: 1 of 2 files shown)


packages/components/src/Button/ButtonBase.tsx
```diff
@@ -76,13 +76,13 @@ export interface ButtonProps extends ButtonBaseProps, ButtonIconProps {
   fullWidth?: boolean
 }
 
-export const buttonCSS = css<ButtonBaseProps>`
```
lukelooker, line 79: `ButtonBaseProps` includes `color` but `IconButton` no longer has `color` so this would fails in TS 3.9.7 where string literal templates type-checking is somewhat improved.

dbchristopher (approved):

PR #1615, opened 2020-10-26
InputColor now properly supports readOnly

lukelooker:
### :sparkles: Changes

- Added more explicit tests for disabled & readonly
- Add storybook screenshot tests

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


mdodgelooker (dismissed):
LGTM! 
The swatch isn't using the aria attributes from the new & improved `usePopover`, but then again, the color wheel isn't accessible, so it almost seems like it would be worse/misleading to use them.

mdodgelooker (dismissed):

lukelooker:
@mdodgelooker - PR needed a CHANGELOG fix-up so I fixed the `usePopover` issue as well. 

I don't disagree about the a11y of the ColorWheel but we can always take that on as a separate task in the future :)

mdodgelooker (approved):
Cool! LGTM.

PR #1618, opened 2020-10-26
Select: options-related fixes

mdodgelooker:
### :sparkles: Changes

1. Fixed `Select` alignment of options when some have icons and others do not.
2. `Select` inappropriately shows "No options" when the current value is empty, after filtering a long list down to a single option then deleting the filter text. 
  _Test with the 2nd Select on :3333/?path=/story/fieldselect--select-content_
3. Fixed `Select` remounting its options when the options prop changes due to this fix for a different issue: https://github.com/looker-open-source/components/pull/1240. This causes select via click to fail if the options prop "changes" (same values, new array instance) after a mousedown, e.g. if a parent component contains `usePopover`. 
  _Test with the Open Popover button on :3333/?path=/story/fieldselect--select-demo_

- [x] Includes test coverage for all changes (except the 3rd issue, which I wasn't able to simulate in RTL)
- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

#### Issue 1 Before
Screen Shot 2020-10-26 at 1 44 30 PM

#### Issue 1 After
Screen Shot 2020-10-26 at 1 47 08 PM

#### Issue 2 Before
2e00cb9a-09d5-4e76-9b62-4194ddfae0e0

#### Issue 2 After
772cfcc2-972e-4f12-9e53-a621a4e617bc

#### Issue 3 Before
d8ed5024-c49c-4897-80ae-6390d070c05d

#### Issue 3 After
fa86b795-3b88-4a08-8ef5-beb36bbe7683

### DIFF
--- a/packages/components/src/Form/Inputs/Select/SelectOptions.tsx
+++ b/packages/components/src/Form/Inputs/Select/SelectOptions.tsx
@@ -84,16 +84,17 @@ const StyledIcon = styled(Icon)`
   /* For proper alignment with option text and check icon */
   height: ${({ theme }) => theme.lineHeights.small};
 `
-interface SelectOption {
+
+interface OptionLayoutProps {
   option: SelectOptionObject
   scrollIntoView?: boolean
 }
 
-const SelectOption = ({ option, scrollIntoView }: SelectOption) => {
+const OptionLayout = ({ option, scrollIntoView }: OptionLayoutProps) => {
   const { description, detail, icon, ...rest } = option
   const { hasIcons } = useContext(SelectOptionsContext)
 
-  if (detail || icon || description) {
+  if (detail || hasIcons || description) {
     const iconToUse = icon ? (
       <StyledIcon
         size="small"
@@ -103,7 +104,7 @@ const SelectOption = ({ option, scrollIntoView }: SelectOption) => {
         data-testid="option-icon"
       />
     ) : hasIcons ? (
-      <IconPlaceholder size="small" />
+      <IconPlaceholder size="small" data-testid="option-icon-placeholder" />
     ) : null
 
     return (
@@ -121,17 +122,12 @@ const SelectOption = ({ option, scrollIntoView }: SelectOption) => {
   return <ComboboxOption {...rest} />
 }
 
-const renderMultiOption = (
-  option: SelectOptionObject,
-  key: string,
-  scrollIntoView?: boolean
-) => {
+const MultiOptionLayout = ({ option, scrollIntoView }: OptionLayoutProps) => {
   const { description, detail, ...rest } = option
   if (description || detail) {
     return (
       <ComboboxMultiOption
         {...rest}
-        key={key}
         py="xxsmall"
         scrollIntoView={scrollIntoView}
       >
@@ -144,7 +140,7 @@ const renderMultiOption = (
       </ComboboxMultiOption>
     )
   }
-  return <ComboboxMultiOption {...rest} key={key} />
+  return <ComboboxMultiOption {...rest} />
 }
 
 export function SelectOptionWithDescription({
@@ -198,9 +194,9 @@ export const SelectOptionGroup = ({
       {options.map((option, index) => {
         const key = `${keyPrefix}-${index}`
         return isMulti ? (
-          renderMultiOption(option, key)
+          <MultiOptionLayout option={option} key={key} />
         ) : (
-          <SelectOption option={option} key={key} />
+          <OptionLayout option={option} key={key} />
         )
       })}
     </SelectOptionGroupContainer>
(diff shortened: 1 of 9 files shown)


packages/components/src/Form/Inputs/Select/SelectOptions.tsx
```diff
@@ -25,14 +25,14 @@
...
 import { IconNames, iconNameList } from '@looker/icons'
-import React, { ReactNode, useContext, useMemo } from 'react'
+import React, { createContext, ReactNode, useContext, useMemo } from 'react'
 import styled from 'styled-components'
-import { v4 as uuid } from 'uuid'
 import { Icon, IconProps } from '../../../Icon'
 import { Spinner } from '../../../Spinner'
 import { Box } from '../../../Layout'
-import { ListItemDetail, ListItem } from '../../../List'
+import { IconPlaceholder, ListItemDetail, ListItem } from '../../../List'
```
mdodgelooker, line 33: Moved `IconPlaceholder` into `/List` from `MenuItem`. More commonality for list-based components! 🎉

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
```diff
@@ -83,31 +85,30 @@ const StyledIcon = styled(Icon)`
...
+    ) : null
 
     return (
-      <ComboboxOption
-        {...rest}
-        key={key}
-        py="xxsmall"
-        scrollIntoView={scrollIntoView}
-      >
+      <ComboboxOption {...rest} py="xxsmall" scrollIntoView={scrollIntoView}>
```
mdodgelooker, line 111: No longer need a `key` here since I made this a component instead just a function, so that `useContext` would work (the `key` still gets used in `SelectOptions`).

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
```diff
@@ -181,18 +177,12 @@ SelectOptionGroupTitle.defaultProps = {
   variant: 'subdued',
 }
 
-function useOptionKeyPrefix(options?: SelectOptionProps[]) {
-  // Create a new unique prefix every time the list of options changes
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  return useMemo(() => uuid(), [options])
-}
```
mdodgelooker, line 188: This was the culprit for the overly sensitive options unmount/remount. It was to fix a bug with windowed options + filtering, but the change below (with `useID` and the option length) fixes that issue just as well.

packages/components/src/Form/Inputs/Select/utils/useWindowedOptions.tsx
```diff
@@ -123,8 +123,10 @@ export function useWindowedOptions(
         'value',
         navigationOption.value,
       ])
-      start = selectedIndex
-      end = selectedIndex
+      if (selectedIndex > -1) {
+        start = selectedIndex
+        end = selectedIndex
+      }
```
mdodgelooker, line 129: This was the fix for the "No options" but. With an empty value, the `navigationOption` is defined as `{value: ''}` but doesn't match any of the actual options, so the `selectedIndex` is -1, and when starting and ending the options window with -1, you're not gonna get much. 😜

packages/components/src/List/IconPlaceholder.tsx
```diff
@@ -0,0 +1,35 @@
+/*
```
mdodgelooker, line 1: Moved here from `MenuItem.tsx` in order to share more between list-like components.

packages/storybook-config/src/main.js
```diff
@@ -38,6 +38,9 @@ const addonEssentials = {
 const config = {
   addons: [addonEssentials],
   stories: ['../**/*.story.tsx'],
+  typescript: {
+    reactDocgen: false,
+  },
```
lukelooker, line 43: This should get removed as it will break the `storybookMode` logic. We could def look at making a CLI flag to enable `fast` if needed.

lukelooker (changes requested):
Need to remove react-docgen override but otherwise this looks great. Tested in a bunch of different ways and I wasn't able to see anything amiss the identified problems appear to be gone. :)

packages/storybook-config/src/main.js
mdodgelooker: Whoops how did that get in there. Thanks!

lukelooker (approved):

lukelooker:
Had a build error - I've seen this once or twice before. TS can't infer the type so it requires explicitly specifying the return type to make everything happy.

lukelooker (approved):
LGTM

PR #1620, opened 2020-10-28
usePopper: render off screen until true position is calculated

mdodgelooker:
### :sparkles: Changes

- On very slow pages, Popover flashes in briefly in the upper left corner of the page (due the initial `styles` value) before moving to its proper place, with `PopperJS`-calculated positioning
- Hiding the surface with `opacity` or `visibility` causes all sorts of unit test havoc, but moving it off the screen with a negative `top` value should do the trick.
- `display: none` is definitely a no-go – it would break `PopperJS`'s ability to calculate the correct position based on the surface's dimensions
- I tried and failed to create an example page slow enough to reproduce the bug, but I'm 99% sure this will fix the issue – and it doesn't break anything
 
- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `Popover` flashing in the upper left corner of the screen on slow pages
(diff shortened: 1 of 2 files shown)


lukelooker (approved):
LGTM

PR #1621, opened 2020-10-29
Popover: honor 1st click outside

mdodgelooker:
### :sparkles: Changes

- Reversed the default of `Popover`'s `cancelClickOutside` prop from true to false
- Removed `groupedPopoversRef` and the very recent `groupedMenusRef` since they won't be needed anymore

## Questions
1. Do we want to keep the `cancelClickOutside` prop? If so, should we document it or keep it on the DL?
2. Currently a **non-filterable** `Select` has `cancelClickOutside={true}` to mimic a native browser select. Do we still want this?

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- `Popover` and `Menu` no longer cancel the first click outside by default (use `cancelClickOutside` to override this)
+- `Popover`, `Menu`, and `Select` no longer cancel the first click outside by default (use `cancelClickOutside` to override this)
 - `DialogContent` no longer has `py` unless it overflows the available space (acting as `overflow: scroll`)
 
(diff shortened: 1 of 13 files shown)


lukelooker (approved):
Code change looks good. 

I think we should maintain `cancelClickOutside` for at least a few releases in case we see an unanticipated use case – if it seems like after a few months nobody protests we can reevaluate totally removing it.

Looking at Material Web Components example it appears Material takes a pretty universal approach of simply allowing the next click to trigger's it's normal flow: https://material-components.github.io/material-components-web-catalog/#/component/select so I'd say we include `Select` in the `cancelClickOutside=false` pattern unless somebody has a strong objection.

At the very lease it's internally consistent. :)

mdodgelooker:
Removed `cancelClickOutside` from `Select`, etc, so now none of our components turn it "on", though it is still available for the time-being.

lukelooker (approved):
LGTM :)

PR #1628, opened 2020-11-02
Support autoFocus in Combobox-based components

mdodgelooker:
### :sparkles: Changes

- Apply `autoFocus` and `data-autofocus` props to the `ComboboxInput` in all `Combobox`-based components

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `Select`, `SelectMulti`, `InputTimeSelect`, and `InputSearch` now support `autoFocus` as well as `data-autofocus` (for use in `Dialog` and `Popover`)
+- The `autoFocus` prop will now work for inputs in `Popover` and `Dialog`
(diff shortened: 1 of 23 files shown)


lukelooker (commented):
Discussed during stand-up. 

Exploring if it's possible to hide internal need to use `data-autofocus` so that developers only need to think about / assign `autoFocus` :)

mdodgelooker:
@lukelooker I've added a couple utils, `pickInputProps` and `getAutoFocusProps` that apply `data-autofocus` wherever `autoFocus` is found in all our inputs except `InputHidden` and `InputFilters`. I tried applying `autoFocus` to the `Select` inside of `InputFilters`, but it pre-opens the list, which I think is probably not desired behavior?

lukelooker (approved):
Rebased this PR to fix-up merge conflict.

The code looks great otherwise - really nice clean-up to the autoFocus prop.

On the select autoFocus behavior I think your approach makes sense - look at MDN's docs I think it fits: https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/autofocus

PR #1643, opened 2020-11-09
DataTable supports onClick of Link/Anchor as text inside rows and does not select the row

laraharrow:
### :sparkles: Changes

- DataTable supports onClick of Link/Anchor as text inside rows and does not select the row

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/DataTable/Item/DataTableRow.tsx
+++ b/packages/components/src/DataTable/Item/DataTableRow.tsx
@@ -87,14 +87,17 @@ const DataTableRowLayout = forwardRef(
     ) : (
       <div />
     )
-
     return (
       <tr
         ref={ref}
         className={className}
         onKeyDown={onKeyDown}
         tabIndex={tabIndex}
-        onClick={onClick}
+        onClick={
+          event && (event.target as HTMLElement).tagName === 'A'
+            ? onClick
+            : undefined
+        }
       >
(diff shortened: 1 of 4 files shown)


packages/components/src/DataTable/Item/DataTableRow.tsx
```diff
@@ -82,19 +82,23 @@ const DataTableRowLayout = forwardRef(
       }
     })
 
+    const handleOnClick = (event: React.MouseEvent<HTMLElement>) =>
+      event.target instanceof HTMLAnchorElement
+        ? undefined
+        : onClick && onClick(event)
```
dbchristopher, line 88: Oh wow, this is way better. TS for the win!

lukelooker (changes requested):
Awesome! Can you update the PR title to describe the work here and look at adding a test that validates this (mostly so it doesn't break in the future) :)

packages/components/src/DataTable/Item/DataTableRow.tsx
laraharrow: 🤩 Thanks @lukelooker for the suggestion!

lukelooker (approved):
LGTM

PR #1644, opened 2020-11-09
Focus trap refactor

mdodgelooker:
### :sparkles: Changes

- Added `FocusTrapContext` and `FocusTrapProvider` in `@looker/components-providers`
- Created `TrapStackProvider` and `useTrapStack` as the base for both scroll lock and focus trap
- Pulled in the necessary utility functions from `focus-trap` and removed that dependency (it had way more than we needed, making it awkward for us to use... plus a troublesome bug)

- [x] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


packages/components/src/Dialog/SurfaceBase.tsx
```diff
@@ -24,24 +24,17 @@
...
 
   const ref = useRef<null | HTMLDivElement>(null)
 
-  useEffect(() => {
-    const t = setTimeout(() => {
-      enableFocusTrap && enableFocusTrap()
-    }, theme.transitions.moderate)
-    return () => clearTimeout(t)
-  }, [enableFocusTrap])
-
```
mdodgelooker, line 44: I think this was a relic of CSSTransition. Now it works fine to start the focus trap as soon as the designated element is rendered in the dom.

packages/components/src/Popover/usePopover.tsx
```diff
@@ -194,25 +184,6 @@ export const usePopover = ({
...
-          parentFocusTrapRef.current &&
-          parentFocusTrapRef.current.pause()
-      }
-    } else if (!focusTrap) {
-      // need to manually re-enable any parent focus trap
-      parentFocusTrapRef &&
-        parentFocusTrapRef.current &&
-        parentFocusTrapRef.current.unpause()
-    }
-  }, [focusTrap, parentFocusTrapRef, isOpen, enableFocusTrap])
```
mdodgelooker, line 214: We still need to disable any current focus trap directly if `focusTrap` is false (so that the contents of the non-focus-trapped `Popover` can receive focus) but I moved this logic to `useTrapStack` since it should apply to any use of focus trap or scroll lock.

packages/components/src/Tooltip/useTooltip.tsx
```diff
@@ -142,7 +148,11 @@ export function useTooltip({
   const element =
     typeof triggerElement === 'undefined' ? newTriggerElement : triggerElement
 
-  const handleOpen = () => setIsOpen(true)
+  const handleOpen = () => {
+    if (element && !element.dataset.notooltip) {
+      setIsOpen(true)
+    }
+  }
```
mdodgelooker, line 155: If the trigger element has `data-notooltip` (because it was focused due to focus trap's "return focus on deactivate" behavior), don't show the tooltip.

packages/components-providers/src/FocusTrap/utils.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    // If focus lands on the body, move it back to where it was before the trap
+    if (checkFocusLost() && nodeFocusedBeforeActivation) {
+      const elementToFocus = nodeFocusedBeforeActivation as HTMLElement
+      // When returning focus to the element focused prior to focus trap activation
+      // opening the tooltip at that moment is undesirable
+      elementToFocus.setAttribute('data-notooltip', 'true')
+      elementToFocus.focus()
+      // Clear the data attribute immediately so that tooltip will work the next time
+      elementToFocus.removeAttribute('data-notooltip')
```
mdodgelooker, line 156: I needed to make sure this refactor would support the no-tooltip-on-menu-close feature, so I decided to just include it in this PR.

packages/components/src/Form/Inputs/Combobox/utils/useInputEvents.ts
```diff
@@ -96,13 +96,8 @@ export function useInputEvents<
 
   const handleBlur = useBlur(context)
 
-  function handleFocus(e: FocusEvent<HTMLInputElement>) {
-    const input = e.currentTarget
-    if (readOnly && input) {
-      window.requestAnimationFrame(() => {
-        input.selectionEnd = input.selectionStart = 0
```
mdodgelooker, line 103: This was a workaround for https://github.com/focus-trap/focus-trap/issues/114, and no longer needed 🎉

packages/components/src/Dialog/Prompt/PromptDialog.tsx
```diff
@@ -159,11 +159,13 @@ export const PromptDialog: FC<PromptDialogProps> = ({
           <Label htmlFor="promptInput">{inputLabel}</Label>
         </VisuallyHidden>
         <InputText
+          autoFocus
           onKeyDown={onKeyDown}
           id="promptInput"
           placeholder={inputLabel}
           onChange={onChange}
           width="100%"
+          my="xxxsmall"
```
mdodgelooker, line 168: The overflow style on `DialogContent` is chopping off a few pixels of the focus ring shadow.

lukelooker (changes requested):
This is really cool - works great. Couple of minor nits to address in comments.

packages/components-providers/src/FocusTrap/utils.ts
```diff
@@ -0,0 +1,164 @@
...
+import { tabbable, isFocusable, FocusableElement } from 'tabbable'
+
+const isSelectableInput = (
+  node: FocusableElement
+): node is HTMLInputElement => {
+  const nodeAsInput = node as HTMLInputElement
+  return (
+    nodeAsInput.tagName !== undefined &&
+    nodeAsInput.tagName.toLowerCase() === 'input' &&
+    typeof nodeAsInput.select === 'function' &&
```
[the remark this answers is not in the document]
mdodgelooker: That's in the original `focus-trap` – it's checking that `select` is a function and then it calls `node.select()` down on line 105. As far as I can tell, as long as it's an `input`, `select` will be available, so I think I can remove this line. Think I should?
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement

packages/components/src/Dialog/useDialog.tsx
```diff
@@ -177,10 +171,6 @@ export const useDialog = ({
     <DialogContext.Provider
       value={{
         closeModal: handleClose,
```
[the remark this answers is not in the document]
mdodgelooker: I agree, it's obnoxiously named. The only drawback of `close` is that if you somehow try to access `DialogContext` incorrectly, you won't get an error for `close` being undefined, due to `window.close`. (Though calling it in this case would only actually close your window if it was opened via `window.open`).

playground/src/index.tsx
```diff
@@ -23,39 +23,328 @@
...
+  PopoverContent,
+  Dialog,
+  DialogContent,
+  Space,
+  Card,
+  CardContent,
+  Form,
+  Text,
+  SelectOptionProps,
+  SelectOptionGroupProps,
```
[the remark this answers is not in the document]
mdodgelooker: Omg that's brilliant.

mdodgelooker (commented):
Thanks @lukelooker ! Ready a re-review. :)

packages/components-providers/src/FocusTrap/utils.ts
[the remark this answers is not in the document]
lukelooker: Out of abundance of caution (basically I'm not sure why they introduced it originally) I think it's fine to keep it in and not possibly introduce a breakage.

packages/components/src/Dialog/useDialog.tsx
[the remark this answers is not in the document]
lukelooker: 🤦  - we've totally had this conversation below and you're still right. I'm just going to wander off now... :)

lukelooker (dismissed):
Looks great! 

Let's wait to merge this until we've shipped 0.9.24 and perhaps SC5 upgrade?

lukelooker:
Oh - just noticed this has a jest snapshot failure (and probably a matching image-snapshot failure) due to the padding change on `DialogContent`

lukelooker (approved):
LGTM

mdodgelooker:
@lukelooker just rebased and need another ✅, 🙏. There were a few more image snapshot updates – just from that `xxxsmall` vertical padding I added on the (non-scrolling) `DialogContent` for focus rings.

lukelooker (approved):
Sold! Ship it :)

PR #1645, opened 2020-11-10
0.9.24 Fixes

lukelooker:
### :sparkles: Changes

- `useDataTable` hook for simple `DataTable` rendering (used for testing)
- `Avatar` will consistently maintain it's 1:1 aspect ratio within flex layouts
- `CodeBlock` should now uses `overflow-y: auto` instead of `overflow-y: scroll`
- `DataTable` with number columns now properly aligns the cell content
- `DataTableCell` truncates `description` appropriately now
- `DataTableCell` with a link is presented properly when it gains focus via keyboard
- `DataTableColumn` `width` again supports `number` for percentage-based column widths
- `DataTableHeaderCell` properly aligns sort direction icon when column `size` is specified
- `DataTableItem` supports a single column
- `DialogContent` now will only enter "overflow" state exactly once per render lifecycle
- `DialogFooterLayout` is no longer exported

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


packages/components/src/Dialog/DialogSurface.tsx
```diff
@@ -57,11 +66,11 @@ export interface DialogSurfaceProps {
   width?: DialogWidth
 
   /**
-   * Explicitly specifying a minHeight will set the Surface to be the lesser of
-   * the specified height, the content height or the viewport height. Default will
-   * cause the Surface to auto-size to its content.
+   * Explicitly specifying a height will set the Surface to be the greater of
+   * the specified height, the content height but no greater than the viewport height.
```
mdodgelooker, line 70: I think you updated to remove "the content height" but maybe didn't push?

mdodgelooker (dismissed):
Looks good – but I think you didn't push your update to the Dialog height jsdoc comment?

packages/components/src/Dialog/DialogSurface.tsx
lukelooker: Doh. Pushed!

mdodgelooker (approved):

PR #1646, opened 2020-11-11
FadeIn component

mdodgelooker:
### :sparkles: Changes

- Adds a base component `Animate` and a `FadeIn` component that wraps it with opacity keyframes
- Currently the only use case in the core product is for `FadeIn`
- An "out" transition (e.g. `FadeOut`) could be done with `useAnimationState` but I think that can be left for a future PR / when there's a use case

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots
2818b5dd-f0ae-4b5a-94d0-0f9d2854f454


packages/components/src/Animate/Animate.tsx
```diff
@@ -0,0 +1,68 @@
...
+  animation-duration: ${({ duration = 'quick', theme }) =>
+    `${theme.transitions[duration]}ms`};
+`
+
+export const Animate = styled((props: AnimateProps) => (
+  <div {...omit(props, ['delay', 'duration'])} />
+))`
+  ${simpleLayoutCSS}
+  ${animateCSS}
+`
```
mdodgelooker, line 62: This component isn't particularly useful as is... should it even be exported?

packages/components/src/Animate/Animate.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const FadeIn = styled(Animate)`
+  animation-fill-mode: forwards;
+  animation-name: ${fadeIn};
+  opacity: 0;
+`
```
mdodgelooker, line 68: I couldn't decide on a testing strategy. An image snapshot won't work, but I hesitate to add more DOM snapshots, though that would capture the relevant CSS, beyond which there's nothing much to test. Thoughts?

lukelooker: I think a snapshot would at least point at the CSS being stable / the component rendering (not throwing an error) so I'd be good with that for now :)

packages/components/src/Animate/Animate.tsx
lukelooker: I'm not sure... I suppose it gives easy access to the theme.durations if someone wanted to build their own FadeX or other animation component.

I'm ambivalent honestly.

packages/components/src/Animate/Animate.tsx
lukelooker: It just occurred to me that you might be able to use the Jest timers to measure opacity at the beginning of the duration, then check that it's resolved at the end. That would "prove" that the animation ends in a good state I think?

lukelooker (changes requested):
I think figuring out some kind of test (even if just snapshot is viable) would be the only addition I'd like to see.

packages/components/src/Animate/Animate.tsx
mdodgelooker: I added a couple snapshots (default props & non-default), but sadly the jest timers don't seem to have any affect on the css values.

lukelooker (approved):

PR #1648, opened 2020-11-13
InputSearch now supports disabled and readOnly

laraharrow:
### :sparkles: Changes

- InputSearch now supports disabled and readOnly

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


packages/components/src/Form/Inputs/InputSearch/InputSearch.tsx
```diff
@@ -208,4 +211,11 @@ const InputSearchLayout = forwardRef(
...
 InputSearchLayout.displayName = 'InputSearch'
 
-export const InputSearch = styled(InputSearchLayout)``
+export const InputSearch = styled(InputSearchLayout)`
+  input {
+    color: ${({ disabled, theme }) =>
+      disabled ? theme.colors.text1 : 'inherit'};
+    cursor: ${({ readOnly, disabled }) =>
+      readOnly || disabled ? 'not-allowed' : undefined};
+  }
```
dbchristopher, line 220: Whenever possible we need to avoid indirect style overrides like this. It can make things a lot harder to debug, and we'll likely end up with conflicting style rules. 

It also looks like combobox input already has disabled styles set up (screenshot from master):

Screen Shot 2020-11-13 at 10 03 56 AM

If the text color needs adjusting, it would be better to update ComboboxInput directly so that all instances have the proper styling. 

Also, it looks like ComboboxInput is intentionally styled to have default curser when it is readOnly. We should update that, or at least hear the case for why it's not using the `not-allowed` cursor in the first place:

Screen Shot 2020-11-13 at 10 08 15 AM

@mdodgelooker is that intentional that `readOnly` combobox inputs should use the default cursor?

www/src/documentation/components/forms/input-search.mdx
```diff
@@ -67,6 +67,17 @@ Use the `autoResize` prop to allow the input's width to adjust to its current va
...
+## Attributes
+
+InputSearch supports `disabled` and `readOnly`
+
+```jsx
+<SpaceVertical align="start">
+  <InputSearch disabled value="Disabled" />
+  <InputSearch readOnly value="Read Only" />
+</SpaceVertical>
+```
```
dbchristopher, line 79: I'm not sure if there is any reason for these two attributes to be grouped under a single heading. For the sake of consistency I think they could be listed separately to match the `InputText` docs:

Screen Shot 2020-11-13 at 10 12 01 AM

dbchristopher (changes requested):
This works well! I think we need to re-think the style overrides here though. It's important to keep things maintainable and consistent, so the style changes should be made directly to `ComoboxInput`.

packages/components/src/Form/Inputs/InputSearch/InputSearch.tsx
mdodgelooker: It is intentional – when a `Select` is not filterable, it applies `readOnly` to the `ComboboxInput`. https://github.com/looker-open-source/components/blob/master/packages/components/src/Form/Inputs/Select/Select.tsx#L173
The `cursor: default` style here is meant to match a native browser `select`, which also gets the default cursor.

www/src/documentation/components/forms/input-search.mdx
laraharrow: updated!

packages/components/src/Form/Inputs/Combobox/ComboboxInput.tsx
```diff
@@ -219,20 +219,13 @@ export const ComboboxInputInternal = forwardRef(
...
-      : ''}
-`
-
 export const ComboboxInput = styled(ComboboxInputInternal)`
-  ${comboboxStyles}
+  input {
+    color: ${({ disabled, theme }) =>
+      disabled ? theme.colors.text1 : 'inherit'};
+    cursor: ${({ readOnly, disabled }) =>
+      readOnly || disabled ? 'not-allowed' : undefined};
```
mdodgelooker, line 227: Hi @laraharrow actually it does need to have `cursor: default`, for `Select`.

mdodgelooker (changes requested):

packages/components/src/Form/Inputs/Combobox/ComboboxInput.tsx
dbchristopher: It looks like native text inputs don't render the `not-allowed` cursors for read only fields either. 

https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_input_readonly

So if that's the logic, then I see no reason for doing a special cursor override for this case!

mdodgelooker: Oh good point @dbchristopher

lukelooker:
@laraharrow - Getting a jest-test failure.

lukelooker (changes requested):
@laraharrow Getting a jest-test failure.

laraharrow:
this pr is solved at: https://github.com/looker-open-source/components/pull/1650

PR #1674, opened 2020-11-21
MenuList windowing

mdodgelooker:
### :sparkles: Changes

- New `useWindow` hook
  - "fixed" mode using static pre-defined item height (similar to current `Select` windowing)
  - "variable" mode using a function that returns the height of each item
- `MenuList` uses `useWindow` in variable mode if the 1st item is a `MenuGroup`, otherwise fixed mode

- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


packages/components/src/Popover/usePopover.tsx
```diff
@@ -257,14 +257,16 @@ export const usePopover = ({
           boxShadow={3}
           color="text5"
         >
-          <Box
+          <Flex
+            flexDirection="column"
+            alignItems="flex-start"
```
mdodgelooker, line 262: Using `Flex` allows `MenuList` to fill the available space as soon as it mounts / before having to render all the children, so `useWindow` can get an accurate reading.

I'm thinking this should be low impact b/c most times the direct child will be `PopoverContent` which won't be hurt by filling available space.

packages/components/src/Tabs/TabList.tsx
```diff
@@ -73,7 +73,7 @@ const TabListLayout = forwardRef(
     )
 
     function handleArrowKey(direction: number, initial: number) {
-      moveFocus(direction, initial, wrapperRef)
+      moveFocus(direction, initial, wrapperRef.current)
```
mdodgelooker, line 76: Slightly changed `moveFocus` to accept an element rather than a ref.

packages/components/src/Form/Inputs/Select/utils/useWindowedOptions.tsx
```diff
@@ -102,11 +102,11 @@ export function useWindowedOptions(
   let { start, end } = useMemo(
     () =>
       getWindowedListBoundaries({
-        containerHeight,
-        containerScrollPosition: listScrollPosition,
         enabled: windowedOptions,
+        height: containerHeight,
```
mdodgelooker, line 106: Just name changes on these props – "container" felt superfluous.

mdodgelooker:
Ok fixed some lint errors and updated docs. @lukelooker this is ready for review :)

lukelooker (approved):
Works great and code looks awesome :)

PR #1675, opened 2020-11-23
updating FieldInline to support detail and description

laraharrow:
### :sparkles: Changes

- updating FieldInline to support detail and description
- components affected:
    * FieldCheckbox
    * FieldRadio
    * FieldToggleSwitch

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


packages/components/src/Form/Fields/FieldInline.tsx
```diff
@@ -45,71 +45,76 @@ interface FieldInlinePropsInternal extends FieldBaseProps {
...
-  grid-column: 1;
-  grid-row: 1;
-  padding-right: ${({ theme }) => theme.space.xsmall};
-  /* stylelint-disable  */
-  -ms-grid-column: 1;
-  -ms-grid-row: 1;
-  /* stylelint-enable */
+}) => {
+  return (
+    <div className={className}>
```
dbchristopher, line 56: Ok I don't actually know what the issue is, where the tests were indicating that expected strings were no longer in the document. But apparently they depend on this wrapper being a `label` rather than a `div`. Reverted this and things pass again 🤷‍♂️ :

Screen_Shot_2020-11-23_at_12_05_47_PM

laraharrow: thanks yeah I think that fix the problem. I changed that because my grid wasn't working and I thought maybe it needed a div but then I found the problem and forgot to revert that back. thanks

packages/components/src/Form/Fields/FieldInline.tsx
```diff
@@ -45,71 +45,76 @@ interface FieldInlinePropsInternal extends FieldBaseProps {
...
+const InputArea = styled.div``
+
+const MessageArea = styled.div``
+
 export const FieldInline = styled(FieldInlineLayout)`
-  align-items: center;
   display: grid;
-  grid-template-columns: auto 1fr;
+  grid-template-areas: 'input label detail .' '. messages messages messages';
+  grid-template-columns: auto auto auto auto;
```
dbchristopher, line 92: Unfortunately IE11 doesn't support grid areas. And because of inconsistencies of IE11 grid support, autoprefixers don't add in `-ms*` style rules at all. So you'll have to make sure to explicitely write out all IE equivalent grid styles, or find another way to make this layout work without grid 😞 

If you choose to write the IE grid styles, here's a guide that looks pretty decent:

https://medium.com/@elad/supporting-css-grid-in-internet-explorer-b38669e75d66

dbchristopher (changes requested):
I was inspecting the snapshot changes, and saw that a lot of elements had additional padding added to them. I compared in snapshot on master against this branch, and got the same result:

Screen_Shot_2020-11-23_at_3_00_30_PM

So I think you'll have to track down where that extra space is coming from before we can merge. Also left a note about internet explorer's unfortunate lack of Grid support 😞 

Otherwise, it all looks good to me!

packages/components/src/Form/Fields/FieldInline.tsx
```diff
@@ -45,71 +45,76 @@ interface FieldInlinePropsInternal extends FieldBaseProps {
...
+
+const FieldDescription = styled(Paragraph)`
+  color: ${({ theme }) => theme.colors.text2};
+  font-size: ${({ theme }) => theme.fontSizes.xsmall};
 `
 
+const InputArea = styled.div``
+
+const MessageArea = styled.div``
+
```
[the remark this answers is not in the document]
laraharrow: I used the `styled component` because latter on the style part I'm using it to define the area for the `grid`

```
${MessageArea} {
    grid-area: messages;
}
```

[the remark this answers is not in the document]
dbchristopher: Ahh ok. You should define styles directly on the styled component itself!

dbchristopher (changes requested):
Looks like it reverted to a previous state where a bunch of form elements got extra padding when they shouldn't have 😩 

onion-skin

packages/components/src/Form/Fields/FieldInline.tsx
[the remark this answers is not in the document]
dbchristopher: These elements should be styled directly!

dbchristopher (dismissed):
It all looks good to me! Do we need @mdodgelooker to verify whether the `-ms-grid` stuff works as expected on the IE cloudtop?

mdodgelooker:
I'll check it out on IE!

laraharrow:
Thanks!

mdodgelooker:
The detail is in the wrong spot on IE11 (hiding under the switch). I'll play around with it and try to get it working, but I'm pretty bad at `grid`. @laraharrow any ideas, based on the screenshot?
image

packages/components/src/Form/Fields/FieldInline.tsx
```diff
@@ -45,56 +45,81 @@ interface FieldInlinePropsInternal extends FieldBaseProps {
...
+        {description && <FieldDescription>{description}</FieldDescription>}
+        {validationMessage ? (
+          <ValidationMessage {...validationMessage} />
+        ) : null}
+      </MessageArea>
+    </label>
+  )
+}
+
+const FieldDetail = styled(Span)`
```
mdodgelooker, line 73: This would fix the IE11 issue. I think IE11 doesn't handle inline elements in a grid.
```suggestion
const FieldDetail = styled.div`
```

mdodgelooker (changes requested):

packages/components/src/Form/Fields/FieldInline.tsx
mdodgelooker: Ah actually `styled(Paragraph)` works better than `styled.div` b/c it includes our default font styles. Updated and pushed!

mdodgelooker (approved):

mdodgelooker (approved):

PR #1677, opened 2020-11-25
Range Slider Touch Support

dbchristopher:
### :sparkles: Changes

- Adds touch events to range slider and dependencies  

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots

touch-events


dbchristopher:
@lukelooker this PR is ready for a look whenever you have a chance! I was able to validate this on the iPhone simulator, and I think that should be sufficient to verify that the touch events work. 

@jhardy do you know if there is any good way to test on an actual touch device? I have a feeling that in a real setting, your finger is probably going to obscure the value which would be pretty annoying. Maybe we should make a ticket for future touch-friendly design updates?

dbchristopher:
@mdodgelooker here's my touch event PR if you have any time to take a look! Obviously no rush, especially while we're blocked by the travis stuff 😵

packages/components/src/utils/useMouseDragPosition.ts
```diff
@@ -42,43 +43,53 @@ export function useMouseDragPosition(
...
+    setMousePos({ x: pageX || clientX, y: pageY || clientY })
   }
 
-  const handleMouseUp = () => {
+  const handleStart = (e: globalThis.TouchEvent | globalThis.MouseEvent) => {
+    // update mouse down state AFTER mouse position state is updated
+    requestAnimationFrame(() => {
+      setIsMouseDown(true)
+    })
+    updateMousePos(e)
```
mdodgelooker, line 59: Just curious what makes the `requestAnimationFrame` necessary now as opposed to how it was before?

packages/components/src/utils/useMouseDragPosition.ts
```diff
@@ -42,43 +43,53 @@ export function useMouseDragPosition(
...
+        window.removeEventListener('touchend', handleEnd)
+        window.removeEventListener('mouseup', handleEnd)
+        window.removeEventListener('mousemove', handleMove)
+        window.removeEventListener('mouseleave', handleEnd)
       }
-    },
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [isMouseDown, targetRef]
-  )
+    }
```
mdodgelooker, line 91: The `exhaustive-deps` rule still flags a warning here, so might as well keep the disable comment. Alternatively, if you did want to include the 3 handler functions in the deps array, you'd need to protect them from changing every render with `useCallback`. (And I seem to remember it being surprisingly awkward to do this with `throttle`).

mdodgelooker (changes requested):
Looks good! I'd say just add back in the eslint disable comment in `useMouseDragPosition` and it's gtg.

packages/components/src/utils/useMouseDragPosition.ts
dbchristopher: ack ok! VSCode didn't give me the red squiggle or anything so I assumed that meant the rule was disabled altogether 🤦‍♂️  I'll add it back in!

packages/components/src/utils/useMouseDragPosition.ts
dbchristopher: I am honestly not sure why it wasn't necessary before. After implementing the other changes I started getting race conditions between the mouseMove and mouseDown handlers, which caused RangeSlider to act like the mouse was always down. As a result, that caused the RangeSlider to forever try and maintain focus with the first slider thumb you interact with. This gif illustrates the bug, where I'm unable to select the max endpoint:

bad-focus

Let me know if the explanation makes sense, it's kind of hard to explain in text lol. I can take another step through all the changes tomorrow morning to see if I can at least understand why this is happening now when it wasn't before

mdodgelooker: Ah, got it. 🙂

dbchristopher: Ok after playing around with this a bit more, it turns out it is indeed a race condition issue related to the order that the state updates. So I can make the RangeSlider work correctly with either of these two options:

```
    requestAnimationFrame(() => {
      setIsMouseDown(true)
    })
    updateMousePos(e)
```
or 

```
  updateMousePos(e)
  setIsMouseDown(true)
```
But with the second option I need the additional check of `(isMouseDown && prevMouseDown)` before firing the move event, to prevent mousedown and mousemove listeners from firing simultaneously.

So given that the order of events matter so much I kind of feel like it's going to be less error-prone if we just explicitly delay it with `requestAnimationFrame`. What do you think?

dbchristopher:
@mdodgelooker this should be ready for final ✅  whenever you get a chance!

packages/components/src/utils/useMouseDragPosition.ts
mdodgelooker: Tricky. I think it's fine as-is. 👍

mdodgelooker (approved):
Ship it! 👏

PR #1678, opened 2020-11-30
Deprecate class components

lukelooker:
### :sparkles: Changes

- Moves 3 demo components from class to functional 

(it's on our tech-debt list to be consistent - this leaves us with just ColorWheel which is a larger project to port)

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


www/src/documentation/components/layout/demos/BoxSpacingRecipeTable.tsx
```diff
@@ -122,51 +122,37 @@ const ListRender = (
...
+export const BoxSpacingRecipeTable: FC<BoxSpacingRecipeTableProps> = ({
+  types,
+  sides,
+  sizes,
+}) => (
+  <SpacingTable>
+    <div>
+      {SpaceListHeader('1. Type')}
+      <List pl="none">
+        {types.map((col, i) => ListRender(col.value, col.label, i))}
```
dbchristopher, line 140: Getting an error here
Screen Shot 2020-12-01 at 10 10 56 AM

www/src/documentation/components/layout/demos/BoxSpacingRecipeTable.tsx
```diff
@@ -122,51 +122,37 @@ const ListRender = (
...
-    sizes: ColumnExample[]
-  }
-> {
-  constructor(props: {}) {
-    super(props)
-    this.state = {
-      sides: spacingSides,
-      sizes: spacingSizes,
-      types: spacingTypes,
-    }
```
dbchristopher, line 139: looks like the class version stored these values in state, but the functional component is expecting to receive them as props. Since the variables are defined in this file, maybe don't do either and just treat them as hard-coded constants?

www/src/documentation/system/demos/BreakpointTable.tsx
```diff
@@ -79,66 +79,48 @@ export interface BreakpointExample {
...
+  labels: string[]
 }
+
+export const BreakpointTable: FC<BreakpointTableProps> = ({
+  breakpoints,
+  labels,
+}) => (
+  <DocTable>
+    <TableHead>
+      <TableRow>{labels.map((label, i) => TableLabel(label, i))}</TableRow>
```
dbchristopher, line 118: Same issue in this demo

Screen Shot 2020-12-01 at 10 15 25 AM

dbchristopher (changes requested):
I'm glad we'll be leaving class components behind us for good 😄 Just gotta update a couple of those variables! Really strange that those dummy values were stored in state to begin with 🤔

lukelooker:
@dbchristopher Apologies for my mistakes here - I thought I'd tried these out in the final state but obviously not (eek!) fixed up (and actually tested) all 3. Ready for re-review.

dbchristopher (changes requested):
Cool, form validation and box spacing looks good! 

I think there might have been a merge conflict or something with the recent gatsby upgrades though. I'm seeing two directories, `foundations` and `system`, which are basically duplicates of each other.

Screen_Shot_2020-12-10_at_10_15_32_AM

The top search result takes me to `system/breakpoints`, which is still broken

Screen_Shot_2020-12-10_at_10_14_28_AM

I'm guessing that directory just needs to be deleted?

lukelooker:
@dbchristopher - I went and checked, I think the `system` folder was an artifact on your instance - I looked and it appears to be removed on the branch:

https://github.com/looker-open-source/components/tree/luke/deprecate-class-components/www/src/documentation

dbchristopher (approved):
👍

PR #1679, opened 2020-12-03
InputText - safari color disabled update

laraharrow:
### :sparkles: Changes

- InputText and InlineInputText safari color disabled update

- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


jhardy (changes requested):
Disabled state for InputText looks good now.

But issue with the date range input now

image

packages/components/src/InputDateRange/InputDateRange.tsx
```diff
@@ -474,9 +474,9 @@ const InputTextGroupWrapper = styled.div<InputTextGroupWrapperProps>`
     ${inputTextFocus}
   }
 
-  ${({ disabled }) => disabled && inputTextDisabled}
-
   ${inputTextValidation}
+
+  ${({ disabled }) => disabled && inputTextDisabled}
```
dbchristopher: I think you can delete this altogether actually! InputDateRange relies on `InlineInputText` for the text inputs, which also needed a disabled text fix in safari:

Screen Shot 2020-12-03 at 2 55 03 PM

I think I found a small fix though, which I just pushed onto the branch if you want to double check!

packages/components/src/Form/Inputs/InlineInputText/InlineInputText.tsx
```diff
@@ -105,6 +105,7 @@ const StyledInput = styled.input`
 
   :disabled {
     color: ${(props) => props.theme.colors.text1};
+    -webkit-text-fill-color: ${(props) => props.theme.colors.text1};
```
dbchristopher, line 108: fix for both InlineInputText and InputDateRange

packages/components/src/InputDateRange/InputDateRange.tsx
```diff
@@ -456,6 +456,13 @@ interface InputTextGroupWrapperProps {
   validationType?: 'error'
 }
 
+const InputTextDisabled = css`
+  span {
```
jhardy, line 460: is there specific item this is trying to center? Can we reference the item directly instead of a `span` incase the underlying implementation changes?

laraharrow: Good call, this is actually inside `InlineInputText` and I added the style there. Should be good now.

PR #1680, opened 2020-12-03
fixing checkbox overflow issue

laraharrow:
### :sparkles: Changes

- fixing checkbox overflow issue

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


packages/components/src/Form/Inputs/Checkbox/Checkbox.tsx
```diff
@@ -102,17 +102,21 @@ CheckboxLayout.displayName = 'CheckboxLayout'
 export const Checkbox = styled(CheckboxLayout)`
   ${reset}
   ${space}
-
   height: 1rem;
+  margin: 0.19rem;
```
mdodgelooker, line 106: I think `3px` here would be better, since `0.19rem` could be confusing (and this might make the image snapshots unchanged), but defer to @jhardy if one of our tshirt sizes might be best.

mdodgelooker (changes requested):
This fixes the issue! 🎉 It does appear to slightly affect the visual spacing, as the image snapshots show. I'll defer to to @jhardy on this, but my first thought would be to use pixels instead of ems for the margin.

lukelooker (changes requested):
The position change of the Checkbox should get fixed - it was properly positioned before (but with weird margin of course) so don't want to introduce a regression along with the margin fix :)

lukelooker (approved):
LGTM

PR #1692, opened 2020-12-08
Select: fix click caret to open while focused

mdodgelooker:
#### Issue 1
- **Issue:** Clicking on the caret icon _while_ `Select` is focused fails to open the list.
9008e81d-603a-48cc-8fde-6a7039f9c9fd
- **Cause:** The caret click triggered a blur event on the input, which tells `Combobox` to close... _after_ the mousedown event tells it to open. "Blur event -> close" wins.
- **Fix:** `InputText` focuses the input after any mousedown on a non-button element inside the wrapper div – e.g. the caret icon. Adding a check inside the mousedown handler for if the input is already focused and then `e.preventDefault()` avoids the unnecessary blur & refocus, thus fixing the `Select` issue.

#### Issue 2
- **Issue:** The 1st click on the caret icon on a `SelectMulti` doesn't focus the field.
5c445a8e-f692-4832-b181-324d3ac6e0b5
- **Cause:** In `InputChipsBase.tsx` there was a `onMouseDown={stopPropagation}` that was put in place to prevent the default behavior from `InputText` of focus moving to the input, because this breaks the Clear button. 
- **Fix:** Adding a check for if the mousedown occurred inside a button to `InputText` and removing `onMouseDown={stopPropagation}` from `InputChipsBase` allows the caret to move focus but the Clear button still works.

#### Issue 3
- **Issue:** The 1st click on the caret icon on a `InputTimeSelect` with no current value fails to open the list.
e8cef040-2d50-4ba3-8f95-59ed83b70f9d
- **Cause:** The "scroll current time into view" behavior was triggering a `NAVIGATE` action, which in turn (unnecessarily) re-focused the input, causing a blur event on the caret. (Who knew `.focus()` would trigger a blur even when the element is already focused?)
- **Fix:** Remove behavior that triggers `.focus()` on the input after a `NAVIGATE` action. This was in the original code but I can't figure out why it would be there. `NAVIGATE` is only triggered via `mouseenter` and `keydown` (arrow keys only) events, neither of which move focus. I also removed `ESCAPE` (closes the list in various scenarios) from the same bit of code – it wasn't actually doing anything except breaking my new tests.

- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

#### Issue 1 fixed
dce8387d-94b1-4436-bc2b-893eb5e3dc0a

#### Issue 2 fixed
8c9b52d9-9fdc-479f-840f-0645d9cad831

#### Issue 3 fixed
3f922166-18ce-45b6-af7b-32d3035c3631

### DIFF
--- a/packages/components/src/Form/Inputs/Combobox/utils/useFocusManagement.ts
+++ b/packages/components/src/Form/Inputs/Combobox/utils/useFocusManagement.ts
@@ -38,8 +38,6 @@ export function useFocusManagement(lastActionType?: ComboboxActionType) {
   // of awkwardly at the beginning, unclear to my why ...
   useLayoutEffect(() => {
     if (
-      lastActionType === ComboboxActionType.NAVIGATE ||
-      lastActionType === ComboboxActionType.ESCAPE ||
       lastActionType === ComboboxActionType.SELECT_WITH_CLICK ||
       lastActionType === ComboboxActionType.INTERACT
     ) {
--- a/packages/components/src/Form/Inputs/Combobox/utils/useInputEvents.ts
+++ b/packages/components/src/Form/Inputs/Combobox/utils/useInputEvents.ts
@@ -31,7 +31,11 @@ import {
   useCallback,
   Context,
 } from 'react'
-import { useMouseDownClick, useWrapEvent } from '../../../../utils'
+import {
+  targetIsButton,
+  useMouseDownClick,
+  useWrapEvent,
+} from '../../../../utils'
 import { ComboboxInputProps } from '../ComboboxInput'
 import { ComboboxMultiInputProps } from '../ComboboxMultiInput'
 import {
@@ -41,20 +45,6 @@ import {
 import { ComboboxActionType, ComboboxState } from './state'
 import { useBlur } from './useBlur'
 import { useKeyDown } from './useKeyDown'
-
-// Walks up the dom tree from an element to a containingAncestor checking for a button
-// Used for determining whether a mousedown/click should toggle the option list
-function checkForButton(
-  element: Element,
-  containingAncestor: Element
-): boolean {
-  if (element === containingAncestor) return false
-  if (!element.parentElement) return false
-  if (element.tagName === 'BUTTON') {
-    return true
-  }
-  return checkForButton(element.parentElement, containingAncestor)
-}
 
 export function useInputEvents<
   TProps extends
@@ -129,7 +119,7 @@ export function useInputEvents<
       // Without this, when clicking on a "clear" or "remove value" icon button
       // the list will flash open & closed (if closed)
       // or unnecessarily close (if open)
-      if (checkForButton(e.target as Element, e.currentTarget as Element)) {
+      if (targetIsButton(e)) {
         return
       }
       if (state === ComboboxState.IDLE) {
--- a/packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
+++ b/packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
@@ -188,9 +188,10 @@ export const InputChipsBaseInternal = forwardRef(
       }
     }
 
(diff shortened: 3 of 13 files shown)


packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -122,12 +122,17 @@ const InputTextLayout = forwardRef(
...
       // set focus to input on mousedown of container to mimic natural input behavior
       // need requestAnimationFrame here due to browser updating focus _after_ mousedown is called
-      window.requestAnimationFrame(() => {
-        internalRef.current && internalRef.current.focus()
-      })
+      if (document.activeElement === internalRef.current) {
+        // Avoid triggering the blur event
+        e.preventDefault()
+      } else {
+        setTimeout(() => {
```
mdodgelooker, line 132: Effectively the same as raf but actually testable with jest. :)

mdodgelooker:
Don't review this just yet, I found some related issues that might be good to fix in the same PR.

packages/components/src/Form/Inputs/Combobox/utils/useFocusManagement.ts
```diff
@@ -38,8 +38,6 @@ export function useFocusManagement(lastActionType?: ComboboxActionType) {
   // of awkwardly at the beginning, unclear to my why ...
   useLayoutEffect(() => {
     if (
-      lastActionType === ComboboxActionType.NAVIGATE ||
-      lastActionType === ComboboxActionType.ESCAPE ||
```
mdodgelooker, line 42: Issue 3 fix.

packages/components/src/Form/Inputs/Combobox/utils/useInputEvents.ts
```diff
@@ -42,20 +46,6 @@ import { ComboboxActionType, ComboboxState } from './state'
...
-  containingAncestor: Element
-): boolean {
-  if (element === containingAncestor) return false
-  if (!element.parentElement) return false
-  if (element.tagName === 'BUTTON') {
-    return true
-  }
-  return checkForButton(element.parentElement, containingAncestor)
-}
-
```
mdodgelooker, line 58: Moved this to a separate util as part of Issue 2 fix.

packages/components/src/Form/Inputs/InputChips/InputChipsBase.tsx
```diff
@@ -363,7 +364,6 @@ export const InputChipsBaseInternal = forwardRef(
             disabled={disabled}
             summary={summary}
             showCaret={showCaret}
-            onMouseDown={stopPropagation}
```
mdodgelooker, line 366: Part of Issue 2 fix.

packages/components/src/Form/Inputs/Select/Select.test.tsx
```diff
@@ -719,7 +719,9 @@ describe('Select / SelectMulti', () => {
     renderWithTheme(<Component />)
     const input = screen.getByPlaceholderText('Search')
     fireEvent.mouseDown(input)
-    fireEvent.click(screen.getByText('FOO'))
+    const option = screen.getByText('FOO')
+    fireEvent.blur(input, { relatedTarget: option })
+    fireEvent.click(option)
```
mdodgelooker, line 724: This test was failing after the fix for Issue 3, but a more accurate sequence of events fixed it. I tried using `userEvent.click()` but it doesn't put the right `relatedTarget` on the blur event.

mdodgelooker (commented):
@lukelooker this is ready for review now :)

packages/components/src/Form/Inputs/Select/Select.test.tsx
lukelooker: Interesting - TIL :)

lukelooker (approved):
LGTM! :)

PR #1699, opened 2020-12-11
Mobile Timeframe Filter Components

dbchristopher:
### :sparkles: Changes

- Updates `InputDateRange` with responsive layout adjustments
- Adds a new `Breakpoint` component for toggling different layouts in react
- Updates `Popover` with new props `renderMobileDialog` and `dialogHeader` for rendering mobile friendly UX

- [x] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### :camera: Screenshots

mobile-demo

### Next Steps

- [ ] update INTERNAL REPO to use the `renderMobileDialog` prop for the timeframe filter once this is released


packages/components/src/Dialog/Layout/DialogHeader.tsx
```diff
@@ -113,9 +119,13 @@ const Detail = styled.div`
 export const DialogHeader = styled(DialogHeaderLayout)`
   ${reset}
   ${space}
+  ${border}
+  ${position}
   align-items: center;
+  background: ${({ theme }) => theme.colors.background};
   display: flex;
   flex-shrink: 0;
+  z-index: 1;
```
dbchristopher, line 128: Had to take the z-index escape hatch 🙈 

This allows use of `position: sticky` on DialogHeader, while ensuring that the header stacks higher than siblings with modified `position` and `display` values

z-index

packages/components/src/Popover/Popover.tsx
```diff
@@ -82,8 +118,17 @@ export const Popover = ({
...
+      <Breakpoint from={renderMobileDialog ? 'tablet' : 'mobile'}>
+        {popover}
+        {triggerShown && children}
+      </Breakpoint>
+      {/* if renderMobileDialog is true, render render Dialog on mobile instead */}
+      {renderMobileDialog && (
+        <Breakpoint from={'mobile'} to={'mobile'}>
+          <DialogRender {...dialogProps}>{children}</DialogRender>
+        </Breakpoint>
+      )}
```
dbchristopher, line 131: Happy to take suggestions on alternate APIs for the Breakpoint component. I went with the props `from` and `to`, as I think it makes sense to say `render this from mobile to tablet`, or `from mobile to xl` for example. 

I can also imagine an alternate version where you pass an array of devices you want it rendered or hidden on. E.G.

```
<Breakpoint hideOn=['mobile`]>...</Breakpoint>
<Breakpoint showOn=['tablet', 'laptop', 'desktop']>...</Breakpoint>
```

@lukelooker @mdodgelooker do you have any preferences or suggestions on what feels the most natural?

packages/components/src/Popover/Popover.tsx
```diff
@@ -55,14 +57,48 @@ export interface PopoverProps extends UsePopoverProps {
...
+          borderBottomColor="ui2"
+          position="sticky"
+          top="0"
+          p="medium"
+          px="medium"
+          pr="medium"
+        >
+          {dialogHeader}
+        </DialogHeader>
+        {content}
```
lukelooker, line 97: I'm pretty retiscent to open up lots of new behavior on `DialogHeader` and I think a simple recomposition will allow you to unwind all of the changes to `Dialog*`

```suggestion
  <DialogHeader>{dialogHeader}</DialogHeader>
  <DialogContent>{content}</DialogContent>
```

packages/components/src/Popover/Popover.tsx
```diff
@@ -55,14 +57,48 @@ export interface PopoverProps extends UsePopoverProps {
...
   hoverDisclosureRef?: HTMLElement | null | RefObject<HTMLElement>
+  /**
+   * Convert popover to full-screen dialog on mobile
+   * @default false
+   */
+  renderMobileDialog?: boolean
+  /**
+   * Populate dialog header content when renderMobileDialog is true
+   */
+  dialogHeader?: string
```
lukelooker, line 68: I think "label" might be more appropriate here.

```suggestion
  label?: string
```

packages/components/src/Dialog/Layout/DialogHeader.tsx
lukelooker: See comment below on Popover. I think these changes can be unwound by using `DialogContent` below

packages/components/src/Dialog/DialogSurface.tsx
```diff
@@ -126,6 +126,10 @@ export const DialogSurface = styled(SurfaceBase)<DialogSurfaceProps>`
     border-radius: ${({ theme }) => theme.radii.medium};
   }
 
+  @media screen and (max-width: ${breakpoints[1]}) {
+    overflow-y: auto;
+  }
+
```
lukelooker, line 132: I think this can be removed with introduction of `DialogContent`

packages/components/src/Breakpoint/Breakpoint.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { useResize } from '../utils'
+
+type NamedBreakpoints = 'mobile' | 'tablet' | 'laptop' | 'desktop' | 'xl'
+const NAMED_BREAKPOINTS: NamedBreakpoints[] = [
+  'mobile',
+  'tablet',
+  'laptop',
+  'desktop',
+  'xl',
+]
```
lukelooker, line 38: This is great. We should lift these changes to `@looker/design-tokens` package

packages/components/src/Breakpoint/Breakpoint.tsx
```diff
@@ -0,0 +1,74 @@
...
+export interface BreakpointProps {
+  children: ReactNode
+  from?: NamedBreakpoints
+  to?: NamedBreakpoints
```
lukelooker, line 42: I think this begs for a tuple.

```suggestion
  display: [ NamedBreakpoints, NamedBreakpoints ]
```

I also think it should be required since the component is kinda silly without this.

packages/components/src/Breakpoint/Breakpoint.tsx
```diff
@@ -0,0 +1,74 @@
...
+}
+
+function convertRemToPixels(rem: number) {
+  return rem * parseFloat(getComputedStyle(document.documentElement).fontSize)
```
lukelooker, line 46: Honestly I think it's okay to just use `16` - for better or worse it's a fairly safe assumption throughout our system.

This might be nice to lift into a util from `@looker/design-tokens` since that'll be the source of truth for that.

lukelooker: Also, we've semi-codified that going forward we prefer arrow functions over `function ...` syntax. 

@mdodgelooker I'm struggling to find where we put that document?

lukelooker (changes requested):

packages/components/src/Breakpoint/Breakpoint.tsx
mdodgelooker: @lukelooker we didn't put it anywhere yet, but I just shared it with the team. I can add it to our internal site if you think it's ready?

packages/components/src/Popover/Popover.tsx
lukelooker: `Breakpoint` is super cool - I read through / played with this PR a bit and added some thoughts around this and other pieces.

Now that I'm thinking deeper I wonder if there's a better case for:

```jsx
interface Breakpoint {
  show: [ NamedBreakpoints | undefined, NamedBreakpoints | undefined ]
}

<Breakpoint show={[ undefined, 'large' ]} />
<Breakpoint show={[ 'small', undefined ]} />
<Breakpoint show={[ 'medium', 'xlarge' ]} />

```

packages/components/src/Breakpoint/Breakpoint.tsx
lukelooker: Oh - I'm a dork. It's linked on our "Team Documents" page: https://sites.google.com/corp/google.com/looker-components/team-resources?authuser=0

mdodgelooker: Oh whoops, I forgot I snuck it up there 😂

packages/components/src/Popover/Popover.tsx
dbchristopher: I can't really use `DialogContent` here though because in this context the `content` is likely already wrapped in `PopoverContent`, so it would be doubling up the whitespace. 

Also, jared specified smaller margins in mobile, and it's REALLY helpful to have the header be sticky on iPhone SE:

unnamed

Maybe DialogHeader should just be responsive to begin with?

dbchristopher: Ah ok, I'm following now. So the content would scroll rather than the surface. That makes sense. Maybe I can just strip the padding from `DialogContent`, and then make the header more responsive

packages/components/src/Popover/Popover.tsx
dbchristopher: I'm just not sure if making that prop name more generic would cause confusion, considering 99.9% of popovers wouldn't even use it?

packages/components/src/Popover/Popover.tsx
dbchristopher: nice, yeah I like that better!

dbchristopher: @lukelooker should we support a single value tuple? I can imagine if they only passed something like `show={['tablet']}` that could be shorthand for `show={['tablet', 'tablet']}`. Do you think that makes sense, or should we require them to always pass two values?

lukelooker: Hm... I suppose that could make sense (to specify "only display exactly at this breakpoint"

A tuple always has exactly two elements so in that case we could just support giving us 1 breakpoint value as well:

```
interface Breakpoint {
  show: [ NamedBreakpoints | undefined, NamedBreakpoints | undefined ] | NamedBreakpoints
}
```

dbchristopher: oh yeah good call, passing a string value rather than a single-value array makes more sense for that case 👍 

> A tuple always has exactly two elements

That actually is not true! I had to refresh myself on the terminology to make sure the phrase "a single value tuple" even makes sense lol. I found this:

> In mathematics, a tuple is a finite ordered list of elements.

And of course type theory and mathematics don't always correlate in practice, so I double checked the typescript docs:

> Tuple types allow you to express an array with a fixed number of elements whose types are known, but need not be the same.  

So if I'm understanding correctly, a tuple is just an array with a predefined shape, but the lengths can vary

giphy

packages/components/src/Breakpoint/Breakpoint.story.tsx
```diff
@@ -0,0 +1,198 @@
...
+`
+
+const Link = styled.a<TextColorProps>`
+  ${textColor}
+`
+
+const NatureCard = () => (
+  <Card raised>
+    <CardMedia
+      image="https://placeimg.com/630/480/nature"
```
lukelooker, line 87: This will introduce variance into snapshots (different image can potentially load each time). Let's just use a solid color or something that'll stay stable / not require CI to load resources from the public internet.

packages/components/src/Breakpoint/Breakpoint.tsx
```diff
@@ -0,0 +1,67 @@
...
+  convertRemToPx,
+  NamedBreakpoints,
+  NAMED_BREAKPOINTS,
+} from '@looker/design-tokens'
+import isArray from 'lodash/isArray'
+import { ThemeContext } from 'styled-components'
+import { useResize } from '../utils'
+
+export interface BreakpointProps {
+  children: ReactNode
```
lukelooker, line 38: This is inferred by the `FC` wrapper below

packages/components/src/Breakpoint/Breakpoint.tsx
```diff
@@ -0,0 +1,67 @@
...
+  show: NamedBreakpoints | [NamedBreakpoints, NamedBreakpoints]
```
lukelooker, line 39: JSDoc for this prop

packages/components/src/Breakpoint/Breakpoint.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  useResize(document.body, handleResize)
+
+  const screenMin = from === 'mobile' ? 0 : breakpointPx[fromIndex - 1] // mobile screens start at 0px
+  const screenMax = to === 'xl' ? Infinity : breakpointPx[toIndex] // xl includes xl breakpoint and above
+
+  return (
+    <>{screenWidth > screenMin && screenWidth <= screenMax ? children : null}</>
+  )
```
lukelooker, line 66: Does this need to be wrapped in a fragment or would this work?

```suggestion
  return screenWidth > screenMin && screenWidth <= screenMax ? children : null
```

packages/components/src/Dialog/Layout/DialogContent.tsx
```diff
@@ -70,38 +75,46 @@ const DialogContentLayout: FC<DialogContentLayoutProps> = ({
     <div
       className={overflow ? `overflow ${className}` : className}
       ref={internalRef}
-      {...omit(props, ['renderedHeight'])}
+      {...omit(props, [...paddingProps, 'renderedHeight'])}
```
lukelooker, line 78: I _think_ `pickStyledProps` could do this (wouldn't to `renderedHeight` but that could be done on a separate line explicitly

packages/components/src/Dialog/Layout/DialogContent.tsx
```diff
@@ -70,38 +75,46 @@ const DialogContentLayout: FC<DialogContentLayoutProps> = ({
...
+          padding-top: ${theme.space.large};
+        `
+      : ''}
 `
 
+Inner.defaultProps = {
+  px: ['medium', 'xlarge'],
+  py: 'xxxsmall',
+}
+
```
lukelooker, line 106: Can the defaultProps be specified on the `DialogContent` itself instead? (Makes them more discoverable for codegen stuff that I've been working on). It's also looking like we're going to need to remove defaultProps altogether for tree-shaking work so moving stuff "up" should make that refactor easier as well

packages/design-tokens/src/GlobalStyle/GlobalStyle.tsx
```diff
@@ -26,10 +26,12 @@
 
 import { css, createGlobalStyle } from 'styled-components'
 
+export const BASE_FONT_SIZE = 16
```
lukelooker, line 29: Why is this needed?

packages/design-tokens/src/tokens/breakpoints.ts
```diff
@@ -32,4 +36,58 @@
...
+
 export const breakpoints = ['30rem', '48rem', '64rem', '75rem', '90rem']
+
+export const NAMED_BREAKPOINTS: [MOBILE, TABLET, LAPTOP, DESKTOP, XL] = [
+  'mobile',
+  'tablet',
+  'laptop',
+  'desktop',
+  'xl',
+]
```
lukelooker, line 61: It seems like `Breakpoints` or `BreakpointRamp` would more internally consistent with other named theme objects.

packages/design-tokens/src/tokens/breakpoints.ts
```diff
@@ -32,4 +36,58 @@
...
+type DESKTOP = 'desktop'
+type XL = 'xl'
+
+export type NamedBreakpoints =
+  | undefined
+  | MOBILE
+  | TABLET
+  | LAPTOP
+  | DESKTOP
+  | XL
```
lukelooker, line 51: I'd argue for excluding `undefined` from this because I could see a case where someone might want to specify that the develop _must_ specify a explicit breakpoint and they'd have to alter the type to do so rather than leaving them the choice to mark the type as optional. E.g.:

```ts
interface FauxThing {
  required: NamedBreakpoints
  optional?: NamedBreakpoints
}

interface AltFauxThing {
  required: NamedBreakpoints 
  optional: Exclude<NamedBreakpoints, undefined>
}
```

packages/design-tokens/src/tokens/breakpoints.ts
```diff
@@ -32,4 +36,58 @@
...
+      : `${sizePx * 0.55}px`
+  return {
+    ...map,
+    [name as string]: {
+      name: startCase(name),
+      styles: { height, width },
+      type: name,
+    },
+  }
+}, {})
```
lukelooker, line 93: Since this isn't really relevant to general use of `design-tokens` I'd prefer to see this lifted. Perhaps park it in the `storybook-config` package?

packages/design-tokens/src/utils/convertRemToPx.ts
```diff
@@ -0,0 +1,31 @@
...
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+import { BASE_FONT_SIZE } from '../GlobalStyle'
```
lukelooker, line 27: I _think_ it'd be ok to just declare `const BASE_FONT_SIZE = 16px`

I don't see a world where we'd expect / design for folks to change that value

packages/design-tokens/src/utils/index.ts
```diff
@@ -0,0 +1,40 @@
...
+export { generateTheme } from './theme'
+export type { ThemeCustomizations } from './theme'
+export * from './helpers'
+export { pickSpecifiableColors } from './color/pickSpecifiableColors'
+export {
+  intentUIBlend,
+  uiTransparencyBlend,
+  generateIntentShade,
+} from './color/blend'
+export { generatePressed, generateInteractive } from './color/stateful'
```
lukelooker, line 40: 👍

packages/design-tokens/src/tokens/breakpoints.ts
dbchristopher: Sure sounds good! I definitely wasn't sure where the best place for that would be

packages/components/src/Breakpoint/Breakpoint.story.tsx
dbchristopher: D'oh! Yes of course. No external dependencies in a snapshot test

packages/components/src/Breakpoint/Breakpoint.tsx
dbchristopher: It depends. Looks like the FC type sets `children` as optional. So if I remove children as a required prop and remove the fragment, it complains that this might return `undefined` which is not a valid react node. 

I think it makes sense to enforce children as required, so I think I'll remove the fragment wrapper but leave the required `children: ReactNode` prop on line 38

dbchristopher: oh wait, nevermind. the `ReactNode` definition itself allows for `undefined` 🤦 

I'll remove line 38, but leave the fragment wrapper here. Here's the typescript error I get when I remove the wrapper:

```
const Breakpoint: React.FC<BreakpointProps>
Type '({ children, show }: PropsWithChildren<BreakpointProps>) => string | number | ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<...>)> | null' is not assignable to type 'FC<BreakpointProps>'.
  Type 'string | number | ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<...>)> | null' is not assignable to type 'ReactElement<any, any> | null'.
    Type 'string' is not assignable to type 'ReactElement<any, any> | null'.
```

packages/components/src/Dialog/Layout/DialogContent.tsx
dbchristopher: oh yeah, good call!

packages/design-tokens/src/tokens/breakpoints.ts
dbchristopher: Yeah that's a good point. I realized that if I name it `BreakpointRamp` it really needs to be a named hash to be consistent with other ramps. That's a more useful data structure anyways, so went ahead and updated

packages/design-tokens/src/GlobalStyle/GlobalStyle.tsx
dbchristopher: Just to ensure a consistent value between the global styles and `convertRemToPx`. But I can just hard-code that number in the helper function as you suggested

packages/components/src/utils-storybook/viewportMap.ts
```diff
@@ -0,0 +1,69 @@
...
+      ...map,
+      [name as string]: {
+        name: startCase(name),
+        styles: { height, width },
+        type: name,
+      },
+    }
+  },
+  {}
+)
```
dbchristopher, line 69: It really seems like storybook-config is not a friendly place to store helper functions like this. When I try to import anything from there using the relative path, it seems to break all our package dependencies:

Screen Shot 2020-12-16 at 10 49 25 AM

But I also get that this isn't really isn't useful to people outside our team, and shouldn't be delivered with the design tokens. So I just added this new `utils-storybook` directory within the components src, which won't be exported as part of any package. Does this seem like a good location to you @lukelooker?

lukelooker: Ah - that's annoying. Your solve works for me :)

www/src/documentation/components/overlays/popover.mdx
```diff
@@ -101,6 +101,10 @@ Placement can be adjusted with the `placement` prop. Valid positions are `top`,
 }
 ```
 
+## Alternate Mobile UX
+
+As an avaliable mobile optimization, we provide the `renderMobileDialog` prop. Setting this prop will render a [full-screen Dialog](/components/dialogs/) on screen sizes less than 480px wide.
+
```
dbchristopher, line 107: I can't really render an example here because this is what you see when you resize the window to mobile breakpoint:

Screen Shot 2020-12-21 at 3 53 55 PM

I tried @mdodgelooker's suggestion for rendering the component in an iFrame, but found the breakpoint component still relying on the parent `document.body` to measure the screen size. I think it was because the styled components get injected through a portal, so it maintained the parent context

packages/components/src/Dialog/SurfaceBase.tsx
```diff
@@ -52,6 +52,7 @@ export const surfaceTransition = () => css`
 `
 
 export const SurfaceBase = styled(SurfaceLayout).attrs(() => ({
+  'aria-labelledby': 'dialog-header',
```
dbchristopher, line 55: I wanted to add the aria label as per the [dialog accessibility specs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role#Description). 

I feel like there really shouldn't ever be two DialogHeaders on the page at any given time, so hopefully this sort of naive implementation should be sufficient. What do you think @lukelooker?

packages/components/src/Popover/Popover.tsx
```diff
@@ -55,14 +57,39 @@ export interface PopoverProps extends UsePopoverProps {
...
+  /**
+   * Convert popover to full-screen dialog on mobile
+   * @default false
+   */
+  renderMobileDialog?: boolean
+  /**
+   * Aria label string used to describe the popover content.
+   * It is also rendered as dialog header content when renderMobileDialog is true
+   */
+  label?: string
```
dbchristopher, line 69: So just to clarify @lukelooker, do we have plans to render label text on normal popovers in the future?

www/src/documentation/components/dialogs/index.mdx
```diff
@@ -79,12 +79,12 @@ Dialogs break out of the standard application flow and UI to present new informa
 
 ```jsx
 ;() => {
-  const { open, dialog } = useDialog({ content: 'My Neat Dialog' })
+  const { dialog, setOpen } = useDialog({ content: 'My Neat Dialog' })
```
dbchristopher, line 82: Previous `useDialog` example was broken in master

dbchristopher:
Whew ok that was a lot! 😅  @lukelooker this should all be ready for final review

packages/components/src/Popover/Popover.tsx
lukelooker: I believe so - I think it'll be visually hidden in many / most cases for "Desktop" presentation but it's needed for a11y compliance

www/src/documentation/components/dialogs/index.mdx
lukelooker: 🤦

packages/components/src/InputDateRange/InputDateRange.tsx
```diff
@@ -439,15 +439,25 @@ const HyphenWrapper = styled.div<{ hasInputValues: boolean }>`
...
+  @media screen and (max-width: ${({ theme }) => theme.breakpoints[1]}) {
+    display: grid;
+    justify-items: center;
+  }
 `
 
 const MultiCalendarLayout = styled.div`
   display: inline-grid;
   grid-column-gap: ${({ theme }) => theme.space.large};
+  grid-row-gap: ${({ theme }) => theme.space.medium};
```
lukelooker, line 455: `grid-gap` doesn't have IE11 support. Probably not a particularly tragic rendering in that case?

packages/design-tokens/src/system/color/backgroundColor.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+import { system } from 'styled-system'
+
+export const backgroundColor = system({
+  backgroundColor: {
+    property: 'backgroundColor',
+    scale: 'colors',
+  },
+})
```
lukelooker, line 34: Styled-system has a built-in helper for this, you should be able to add it to `packages/design-tokens/src/system/index.ts` and forego adding this.

```suggestion
```

packages/design-tokens/src/system/color/index.ts
```diff
@@ -51,3 +51,5 @@ export type { ColorProps } from 'styled-system'
 
 export type { TextColorProps } from './textColor'
 export { textColor } from './textColor'
+
+export { backgroundColor } from './backgroundColor'
```
lukelooker, line 55:
```suggestion
```

lukelooker: I'm also completely baffled why I added `textColor` instead of using the built-in type/helper. I think I was playing with "locking" out non-system colors but never actually landed the full change.

Feel free to remove the `textColor` as well if you feel like it.

packages/storybook-config/src/storyshotsConfig.ts
```diff
@@ -36,6 +38,25 @@ export const storyshotsConfig = (pkg: string) => {
...
+            width: parseInt(defaultViewport.styles.width, 10),
+          })
+          await page.waitForTimeout(500)
+        } else if (
+          pageViewport.width !== STORYBOOK_DEFAULT_VIEWPORT.width ||
+          pageViewport.height !== STORYBOOK_DEFAULT_VIEWPORT.height
+        ) {
+          await page.setViewport(STORYBOOK_DEFAULT_VIEWPORT)
+          await page.waitForTimeout(500)
+        }
```
lukelooker, line 58: Neat!

lukelooker (commented):
Some nits on a few changes (don't want to add more props to `DialogContent` but in general this is super-close. LMK if anything doesn't make sense.

packages/components/src/Dialog/Layout/DialogContent.tsx
```diff
@@ -26,18 +26,22 @@
...
 } from '@looker/design-tokens'
 import React, { FC, useRef, useState, useEffect } from 'react'
 import ReactResizeDetector from 'react-resize-detector'
-import styled from 'styled-components'
+import styled, { css } from 'styled-components'
 import omit from 'lodash/omit'
 
 export interface DialogContentProps
   extends LayoutProps,
+    PaddingProps,
```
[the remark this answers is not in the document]
dbchristopher: oh but I do need external control of DialogContent padding for this new case where DialogContent will be wrapped in PopoverContent

Screen Shot 2020-12-23 at 11 51 54 AM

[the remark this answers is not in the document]
lukelooker: Oh bummer... unwinding all the variations on `DialogHeader` took a lot of time / design negotiation a while back so I'm really reticent to re-open the flood gates on this. 

We also have an existing issue around properly introducing padding to `DialogContent` when it doesn't have adjacent `DialogHeader` / `DialogFooter` and introducing padding as props will make that a lot more complex / brittle. 

Two thoughts occur:

1. What if when `renderMobileDialog` we automatically wrap the `content` in `PopoverContent` when it's in a non-mobile breakpoint. Then we could leave the composition for the mobile-ready Dialog "pure" as it were?
2. On looking at this change to `Popover` in general I'm realizing that we've just made `Popover`'s a lot more expensive to render in all cases (even when they haven't `enabled` the `renderMobileDialog` feature) - I'm thinking that perhaps we want to make these changes into a separate purpose-built component `ResponsivePopover` (name could definitely be better) so that existing use-cases don't end up with a significant complexity boost / bundle-size impact.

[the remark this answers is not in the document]
dbchristopher: > On looking at this change to Popover in general I'm realizing that we've just made Popover's a lot more expensive to render in all cases (even when they haven't enabled the renderMobileDialog feature)

Hmm, yeah that is a good point. I guess to me it comes down to how much investment we want to make in mobile UX, and how often we'll use this feature. Would you feel comfortable deciding that all popovers should just convert to this new pattern? In that case I think the added complexity would be justified.

But if not, maybe this shouldn't even be a feature of popover. Maybe what we actually deliver is just the Breakpoint component, and then just build this logic at the call site. I.E. it might make sense if it was just a part of the timeframe filter in our internal repo.

What do you think, should I just strip `renderMobileDialog` out altogether from this PR?

packages/components/src/InputDateRange/InputDateRange.tsx
dbchristopher: yeah that seems like an acceptable level of IE11 jank to me 😄

packages/design-tokens/src/system/color/index.ts
dbchristopher: Ok I just tried removing these in favor what styled system provides. Looks like they don't provide specific helpers for specific `textColor` and `backgroundColor` props. They only provide a single `color` helper which consolidates BackgroundColor, TextColor, and Opacity attributes. 

I assume we do want to be able to maintain this level of specificity, yeah?

packages/components/src/Dialog/SurfaceBase.tsx
dbchristopher: ah dang, yeah that's tough! that does seem out of scope from this PR, so maybe I'll just revert this change and add a ticket for the Dialog aria changes

packages/components/src/Dialog/Layout/DialogContent.tsx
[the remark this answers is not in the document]
dbchristopher: On further reflection, I think this does just feel like a one-off feature we should add to the Timeframe filter itself. I went ahead and reverted those changes, but let me know if you think we should in fact ship that as part of a popover component!

packages/components/src/Dialog/SurfaceBase.tsx
dbchristopher: dialog accessibility ticket here: https://looker.atlassian.net/browse/LENS-1448

PR #1704, opened 2020-12-11
Standardize hover and selected states

ElliotTheHuman:
### :sparkles: Changes
Standardized the background colors on various components when in a hovered and/or selected state according to the chart below:
Screen Shot 2020-12-10 at 2 48 23 PM

**Always use "keyColor"**
- ComboBox / Select (also removed default checked icon when item is selected)
- ComboboxMulti / SelectMulti (also removed default checked icon when item is selected)
- DataTable

**Default to "standard"**
- TreeItem / Tree

**Always use "standard"**
- Menu (Note: no "selected" state)

Relevant JIRA ticket:
https://looker.atlassian.net/browse/LENS-1293

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


packages/components/src/Form/Inputs/Combobox/ComboboxOption.tsx
```diff
@@ -135,7 +139,22 @@ export const ComboboxOptionWrapper = forwardRef(
...
+
+  &:hover {
+    background-color: ${({ isSelected, theme }) =>
+      isSelected ? theme.colors.keyAccent : theme.colors.ui1};
+  }
+
+  /** Essentially on hover and when "selected" */
+  /* &[aria-selected='true'] {
+    background-color: ${(props) => props.theme.colors.ui1};
+  } */
```
mdodgelooker, line 156: Right now keyboard nav doesn't get any feedback on the "active" option. If you make this change, I think it looks good for both keyboard nav and hover, although when you first open a `Select` that has a current value, it'll be `keyAccent` because it is both "active" and "selected". If that's not what we want, we'll probably have to think of something fancy here.
```suggestion
  &[aria-selected='true'] {
    background-color: ${({ isSelected, theme }) =>
      isSelected ? theme.colors.keyAccent : theme.colors.ui1};
  }
```

mdodgelooker (changes requested):
Aside from the `Combobox` keyboard nav suggestion, the only other thing I see is that `MenuItem` does have sort of a "selected" state if you add the `current` prop – it uses `aria-current="true"` on the dom. And I think that might need a bg color update.

Random thought, am I greyscale colorblind, or are ui1 & ui2 pretty much the same? 
image

mdodgelooker:
Oh and I know we can't do image snapshots of hover state, but what do you think about adding / updating some stories to capture selected/current on `TreeItem` and `MenuItem`?

packages/components/src/Form/Inputs/Combobox/ComboboxOption.tsx
ElliotTheHuman: 👍 Good catch. 

Regarding the opening a `Select` with a current value and that value having a `keyAccent` background, I think I'd defer to @jhardy.

ElliotTheHuman: Also you're not blind! I imagine those look really similar because of the opacity we apply to `TreeItem` backgrounds. We do this because we don't want to blot out the optional vertical `border`:
Screen Shot 2020-12-11 at 2 55 35 PM

ElliotTheHuman:
> Oh and I know we can't do image snapshots of hover state, but what do you think about adding / updating some stories to capture selected/current on `TreeItem` and `MenuItem`?

I think this makes total sense to track that. I'll get some additional stories included in this PR.

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -258,7 +258,7 @@ interface TreeItemLabelProps {
 export const TreeItemLabel = styled(Space)<TreeItemLabelProps>`
   align-items: center;
   background-color: ${({ hovered, selected }) =>
-    hovered ? uiTransparencyBlend(2) : selected && uiTransparencyBlend(1)};
+    selected ? uiTransparencyBlend(2) : hovered && uiTransparencyBlend(1)};
```
mdodgelooker, line 261: I realize now why the `selected` bg color looked too close to the `hovered` bg color. I think `uiTransparencyBlend(2)` is not equivalent to `ui2` – the latter is much darker. Would using `ui2` and `ui1` here break anything?

mdodgelooker (commented):
The changes look great! On closer look at `TreeItem`, I had a question about the colors there. And I just wanted to make extra sure the new "selected" bg color on `SelectMulti` is cool w/ @jhardy :
image

mdodgelooker:
FYI the keyboard nav is active on "Friday" in my screenshot above, which is why it's darker.

packages/components/src/Tree/TreeItem.tsx
ElliotTheHuman: Ah good call. In taking a closer look, it looks like `uiTransparencyBlend` uses that number arg to alter opacity and the base color is `colors.text`.
```
export const uiTransparencyBlend = (level: UIColorLevels) =>
  css`
    ${({ theme: { colors } }) => blendColorTransparency(colors.text, level)}
  `
```

For context, we couldn't use vanilla ui2 and ui1 because the optional vertical borders would be covered by non-opaque colors. I'm not sure what the most color-math conscious solution is here; @lukelooker any suggestions?

packages/components/src/Form/Inputs/Combobox/ComboboxOption.tsx
ElliotTheHuman: Spoke to @jhardy and he approves of this!

packages/components/src/Tree/TreeItem.tsx
ElliotTheHuman: Spoke to @jhardy about this as well. He's cool with using vanilla `ui1` and `ui2` for consistency and revisiting `border`-friendly solutions in the future (specifically when folks in HT land start using the `border` prop).

mdodgelooker (approved):
LGTM!

PR #1712, opened 2020-12-15
Fix Tooltip for autoFocus input

mdodgelooker:
### :sparkles: Changes

- `Tooltip`'s `onFocus` handler may be triggered before the callback ref on the trigger is updated, for example an `autoFocus` input receiving focus just as it mounts.
- A recent change for the [Focus Trap refactor](https://github.com/looker-open-source/components/pull/1644) added a check if the element exists and does NOT have `data-notooltip`. In the `autoFocus` scenario, the element does not exist yet (it will soon) but the focus event should not be igored.
- The updated check is for _either_ the element has no `data-tooltip` or it doesn't exist (yet, hopefully). 

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `Tooltip` shows on an input with `autoFocus` on first mount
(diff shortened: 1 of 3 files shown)


lukelooker (approved):
LGTM

PR #1713, opened 2020-12-15
MenuList windowing fixes

mdodgelooker:
### :sparkles: Changes

- Fixed: `variable` windowing was being incorrectly applied to any `MenuList` where the first item was not exactly `MenuItem`, even with <100 children. Now `variable` must be explicitly set – the detection logic for `MenuGroup` vs `MenuItem` was faulty – these components are often wrapped, which would fail a check like `child.type === MenuGroup`.
- Fixed: sticky `MenuGroup` label stacking under item labels. A later change for the focus ring style added `position: relative` to the button, which makes it win the stacking order against the `position: sticky` group label. Moving `position: relative` inside `:focus` where it's needed fixes this. (It would still stack on top of the group label if it has focus, but I think that's ok)

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/Menu/Menu.story.tsx
+++ b/packages/components/src/Menu/Menu.story.tsx
@@ -448,6 +448,22 @@ export const LongMenus = () => {
         </Menu>
         <Menu>
           <MenuDisclosure>
+            <Button>No windowing (groups)</Button>
+          </MenuDisclosure>
+          <MenuList width={100}>
+            {groups.slice(0, 5).map(({ label, items }, index) => (
+              <MenuGroup key={`${label}-${index}`} label={label}>
+                {items.map((item, index2) => (
+                  <MenuItem key={`${item.label}-${index2}`}>
+                    {item.label}
+                  </MenuItem>
+                ))}
+              </MenuGroup>
+            ))}
+          </MenuList>
+        </Menu>
+        <Menu>
+          <MenuDisclosure>
             <Button>Fixed Windowing (3k)</Button>
           </MenuDisclosure>
           <MenuList width={100} windowing={!value ? 'none' : undefined}>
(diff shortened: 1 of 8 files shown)


packages/components/src/Menu/Menu.story.tsx
```diff
@@ -446,6 +446,22 @@ export const LongMenus = () => {
             ))}
           </MenuList>
         </Menu>
+        <Menu>
```
mdodgelooker, line 449: To make sure a `Menu` with <100 groups will not have windowing.

packages/components/src/Menu/Menu.story.tsx
```diff
@@ -472,7 +488,7 @@ export const LongMenus = () => {
           <MenuDisclosure>
             <Button>Variable Windowing (groups)</Button>
           </MenuDisclosure>
-          <MenuList width={300} windowing={!value ? 'none' : undefined}>
+          <MenuList width={300} windowing={!value ? 'none' : 'variable'}>
```
mdodgelooker, line 491: Gotta explicitly set `variable` now – the auto-detect was faulty and there's no good way (that I can think of) to do it right.

packages/components/src/Menu/MenuItemLayout.tsx
```diff
@@ -85,14 +85,14 @@ export const MenuItemLayout = styled(MenuItemWrapper)`
...
     }) => (compact ? `${xxsmall} ${medium}` : `${xsmall} ${medium}`)};
-    position: relative;
     text-align: left;
     text-decoration: none;
     width: 100%;
 
     &:hover,
     &:focus {
       color: inherit;
+      position: relative;
```
mdodgelooker, line 95: Fixes the layering on the sticky group label. Somehow the group label sticky behavior isn't happening at all without `windowing` and hasn't been for a while – which is why this issue only got caught today. A future PR can address why it's not working in non-windowed menus.

lukelooker (approved):
LGTM!

PR #1716, opened 2020-12-18
Deprecate defaultProps library-wide

lukelooker:

- [ ] Update documentation
- [ ] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/Tooltip/TooltipContent.tsx
+++ b/packages/components/src/Tooltip/TooltipContent.tsx
@@ -24,24 +24,24 @@
 
  */
 import styled from 'styled-components'
-import { Paragraph } from '../Text'
+import { Paragraph, ParagraphProps } from '../Text'
 import { TooltipProps } from './Tooltip'
-export const TooltipContent = styled(Paragraph).attrs(
-  (props: TooltipProps) => ({
+
+export const TooltipContent = styled(Paragraph).attrs<TooltipProps>(
+  ({ textAlign = 'center', width }) => ({
     color: 'inherit',
     fontSize: 'xsmall',
     lineHeight: 'xsmall',
     m: 'none',
-    maxWidth: props.width,
+    maxWidth: width,
     p: 'xsmall',
-    width: 'auto',
+    textAlign,
+    width: width || '16rem',
   })
-)`
+)<ParagraphProps>`
   hyphens: auto;
   overflow-wrap: anywhere;
   text-transform: none;
   white-space: normal;
   word-break: break-word;
 `
-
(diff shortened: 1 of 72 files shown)


www/src/Layout/Documentation.tsx
```diff
@@ -85,7 +91,7 @@ const DocumentationLayout = (props: DocQuery) => {
           <Heading as="h1" fontSize="xxxxxlarge">
             {title}
           </Heading>
-          <Status status={status || 'stable'} />
```
ElliotTheHuman, line 88: We might still need this. In doing some research, destructuring a property with a false/null value will not lead to the default value replacing said false/null value:
https://stackoverflow.com/questions/39522014/destructuring-falsey-and-null-with-default-parameters

Note: Gatsby is currently failing because it receives "null" here and the `Status` component tries to call `toUpperCase()` on that null.

ElliotTheHuman (commented):
Things look good! However, we're still hitting a Gatsby error (due to some destructuring shenanigans afaik).

Once that's fixed, I can give a 👍

ElliotTheHuman (changes requested):
Pressed "Comment" instead of "Request Changes" last time.

www/src/Layout/Documentation.tsx
lukelooker: Yeah - that's really annoying - wish I could find the case where it's getting supplied `null` but I made the gate more universal to deal with this now (realized the switch was possibly also suffering from the bad value so moved it up a smidge).

ElliotTheHuman (approved):
LGTM!

packages/components/src/Tooltip/TooltipContent.tsx
```diff
@@ -24,24 +24,24 @@
...
     color: 'inherit',
     fontSize: 'xsmall',
     lineHeight: 'xsmall',
     m: 'none',
-    maxWidth: props.width,
+    maxWidth: width,
     p: 'xsmall',
-    width: 'auto',
+    textAlign,
+    width: width || '16rem',
```
mdodgelooker, line 39: This is confusing but I think at some point we decided the width of this `<p>` should always be `'auto'` and if someone was passing in a `width`, it should be set as the `maxWidth` instead. So this should be:
```
    maxWidth: width || '16rem',
    p: 'xsmall',
    textAlign,
    width: 'auto',
```
image
I'll open a PR...

mdodgelooker: Fix merged: https://github.com/looker-open-source/components/pull/1732

PR #1722, opened 2020-12-22
Fix Combobox jumpy scrolling

mdodgelooker:
### :sparkles: Changes

- When the user hovers over or keyboard navigates to an option, it becomes the "navigated-to" option in Combobox state.
- If the navigated-to option is outside of the scroll window, it is auto-scrolled into view at either the top or bottom of the list, as appropriate.
- This logic is meant for keyboard nav, but targets the navigated-to state, and thus is run for hovering as well.
- If the list is smaller when it first opens and grows as options are added, the height is not getting appropriately updated, so the check for being outside of the scroll window can be incorrectly true.
- As the user scrolls, and the mouseenter event fires when scrolling pauses, the hovered option is unnecessarily auto-scrolled to the top or bottom of the list.
- Fixing the logic that keeps track of the list's height fixes the issue.

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots
Updated Combobox story shows the issue, pre-fix:
9f75b794-b4a7-41ed-9542-5d3ce9812840

### DIFF
--- a/packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
+++ b/packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
@@ -41,6 +41,7 @@ import {
 import React, {
   forwardRef,
   Ref,
+  useCallback,
   useContext,
   useLayoutEffect,
   useEffect,
@@ -49,6 +50,7 @@ import styled, { css } from 'styled-components'
 import once from 'lodash/once'
 import throttle from 'lodash/throttle'
 import { usePopover } from '../../../Popover'
+import { useResize } from '../../../utils'
 import { ComboboxOptionIndicatorProps } from './ComboboxOptionIndicator'
 import { ComboboxContext, ComboboxMultiContext } from './ComboboxContext'
 import { useBlur } from './utils/useBlur'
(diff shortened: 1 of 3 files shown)


packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
```diff
@@ -212,6 +214,12 @@ const ComboboxListInternal = forwardRef(
       popperInstanceRef.current && popperInstanceRef.current.update()
     }, [popperInstanceRef, valueLength])
 
+    const resizeListener = useCallback(() => {
+      setListClientRect?.(contentContainer?.getBoundingClientRect())
+    }, [setListClientRect, contentContainer])
+
+    useResize(contentContainer, resizeListener)
```
mdodgelooker, line 221: `useResize` is a utility hook that calls a callback using both `ResizeObserver` on the specified element, and the window resize event.

packages/components/src/Form/Inputs/Combobox/stories/Combobox.story.tsx
```diff
@@ -47,6 +47,16 @@ const CustomIndicator: FC<OptionIndicatorProps> = ({
...
 export const ComboboxDemo = () => {
+  const [loading, setLoading] = useState(true)
+  useEffect(() => {
+    const t = window.setTimeout(() => {
+      setLoading(false)
+    }, 4000)
+    return () => {
+      window.clearTimeout(t)
+    }
+  }, [])
```
mdodgelooker, line 58: Updating the story to show a list that changes height after opened. (Also added a bunch more options to make it scroll.)

lukelooker (approved):
LGTM. Tested an things appear to be working well, list scrolls smoothly

lukelooker (approved):
👍

PR #1724, opened 2020-12-23
Tree: "brand" color prop

ElliotTheHuman:
### :sparkles: Changes

- Created a `getBackgroundColor` util function
- `Tree`, `TreeItem` support `brand` prop to change default state coloring

- [x] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### :camera: Screenshots
Screen Shot 2020-12-22 at 5 12 39 PM


www/src/documentation/components/content/tree.mdx
```diff
@@ -69,17 +69,29 @@ Use the `border` prop if you would like to have vertical lines expand from your
...
+    </Tree>
+  </Tree>
+  <Divider my="large" />
+  <Tree brand selected label="Orders" icon="Table" defaultOpen>
+    <Tree disabled label="Created" defaultOpen>
+      <TreeItem>Created Date</TreeItem>
+      <TreeItem>Created Month</TreeItem>
+      <TreeItem>Created Year</TreeItem>
+      <TreeItem>Created Quarter</TreeItem>
+    </Tree>
```
jhardy, line 93:
```suggestion
    </Tree>
    </>
```

Need to close fragment

jhardy: One small think I noticed is the selected color for the top level `Tree` and a `TreeItem` are slightly difference
image

jhardy: One other small nit is that the hover state when branded switches to a darker color, but it does not do this on the non branded color application. I think we should probably consistent here 
image

jhardy: I lied, one other small nit, in this example the nested tree items are disabled, yet you can still toggle them open and close, is that the correct behavior? 

https://user-images.githubusercontent.com/170681/103008370-e1206f00-44e9-11eb-9819-d54ed3cfb7d0.mp4

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -266,19 +275,23 @@ export const TreeItemSpace = styled(Space)<TreeItemSpaceProps>`
...
-      : hovered && colors.ui1};
+  background-color: ${({ brand, disabled, hovered, selected, theme }) => {
+    if (selected && !brand) return itemSelectedColor(theme.colors.ui2)
+
+    return getBackgroundColor(
+      Boolean(brand),
+      { disabled, hovered, selected },
+      theme
+    )
+  }};
```
lukelooker, line 294: Probably worth pulling this whole thing out as a helper function as the CSS block here is a pretty epic read :)

```
const treeItemBackground = (props: TreeItemLabelProps) => css`background:  ...;`
```

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -266,19 +275,23 @@ export const TreeItemSpace = styled(Space)<TreeItemSpaceProps>`
...
-    disabled
-      ? 'none'
-      : selected
-      ? itemSelectedColor(colors.ui2)
-      : hovered && colors.ui1};
+  background-color: ${({ brand, disabled, hovered, selected, theme }) => {
+    if (selected && !brand) return itemSelectedColor(theme.colors.ui2)
+
+    return getBackgroundColor(
+      Boolean(brand),
```
lukelooker, line 290: `Boolean(brand),` should be handled inside `getBackgroundColor` rather than requiring type-casting in multiple places.

packages/components/src/Tree/utils/getBackgroundColor.ts
```diff
@@ -0,0 +1,57 @@
...
+
+ */
+
+import { Theme } from '@looker/design-tokens'
+
+interface TreeState {
+  disabled?: boolean
+  hovered?: boolean
+  selected?: boolean
+}
```
lukelooker, line 33: Move this up to `Tree` and consume it from there so that this function is leveraging an interface shared across the actual components and this helper function.

lukelooker (commented):
Add some nits around function structure (I'll leave design decisions to people like @jhardy who are actually qualified) :)

mdodgelooker (changes requested):
Similar thoughts to @lukelooker on the `getBackgroundColor` utility but maybe went a bit further :) and put my suggestions in a PR: https://github.com/looker-open-source/components/pull/1727

www/src/documentation/components/content/tree.mdx
ElliotTheHuman: Good calls @jhardy. Tried to organize my responses to each below:

```
One small think I noticed is the selected color for the top 
level Tree and a TreeItem are slightly difference
```
Ah good point. To clarify, I _believe_ we made the compromise with the Explore team to use a lightened version of `ui2` for `Tree`/`TreeItem`. Do we want to use that lightened version (using `itemSelectedColor(props.theme.colors.ui2)`) or just stick with `ui2`?

@lukelooker Might have more insight on the `itemSelectedColor` function we're using here.

```
One other small nit is that the hover state when branded switches to a darker color, 
but it does not do this on the non branded color application. I think we should probably consistent here
```
Understood. Was following the specs laid out in [LENS-1293](https://looker.atlassian.net/browse/LENS-1293) which had a color switch on selected-hovered for keyColor but not for standard.

Thoughts on whether we should remove the color switch from the keyColor suite or add the color switch to the standard suite?

```
another small nit, in this example the nested tree items are disabled, 
yet you can still toggle them open and close, is that the correct behavior?
```
This was intentional. For context (afaik), DiscoNav wants `Tree`s to be selectable like `TreeItem`s. However, there's a world where a folder (represented by a `Tree`) shouldn't be selectable but a child of that folder is, if that makes sense.
Screen Shot 2020-12-23 at 2 50 17 PM

ElliotTheHuman: Discussed with Jared over a meet. Here's what we came up with:

```
One small think I noticed is the selected color for the top 
level Tree and a TreeItem are slightly difference
```
`Tree` and `TreeItem` should both use that ui1.5 (itemSelectedColor(ui2))

```
One other small nit is that the hover state when branded switches to a darker color, 
but it does not do this on the non branded color application. I think we should probably consistent here
```
For `Tree` in particular, we'll have keySubtle for both selected and selected+hovered to create consistency with the standard color suite. This might extend to other `keyColor` suite components (i.e. Combobox and Select) but we'll revisit this later.

```
another small nit, in this example the nested tree items are disabled, 
yet you can still toggle them open and close, is that the correct behavior?
```
@jhardy will follow up with the designer on `DiscoNav` to figure out their use case with `disabled`. That should hopefully inform how the prop should act / what it should be named.

packages/components/src/Tree/TreeItem.tsx
ElliotTheHuman: Resolved in 6f36b33

packages/components/src/Tree/TreeItem.tsx
ElliotTheHuman: Resolved in 6f36b33

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -145,6 +151,8 @@ const TreeItemLayout: FC<TreeItemProps> = ({
     'icon',
   ])
 
+  const brand = undefinedCoalesce([props.brand, treeContext.brand])
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 42c1032

packages/components/src/Tree/utils/getBackgroundColor.ts
ElliotTheHuman: Resolved in eed935c

ElliotTheHuman:
Poking @lukelooker and @mdodgelooker for re-review!

packages/components/src/Tree/Tree.tsx
```diff
@@ -39,8 +39,32 @@ import { TreeContext } from './TreeContext'
...
+   * If true, the Tree/TreeItem will have a darker background color
+   * @default false
+   */
+  selected?: boolean
+}
+
 export interface TreeProps
-  extends Omit<AccordionProps, 'indicatorGap' | 'indicatorSize'> {
+  extends Omit<AccordionProps, 'indicatorGap' | 'indicatorSize'>,
+    Omit<TreeBackgroundStyleProps, 'hovered' | 'theme'> {
```
mdodgelooker, line 67:
```suggestion
    Omit<TreeBackgroundStyleProps, 'hovered'> {
```

lukelooker (dismissed):
LGTM - core logic function is an easy read now. Thank you!

mdodgelooker (changes requested):
Just a small nit on the `Omit` usage.

mdodgelooker (commented):
Also `border` doesn't seem to be working but I can't tell why. Maybe it's not from this PR? Was it an intentional change in a previous PR? I vaguely remember a tree border discussion coming up at some point...

packages/components/src/Tree/utils/treeBackgroundColor.ts
```diff
@@ -0,0 +1,60 @@
...
+    selected: itemSelectedColor(colors.ui2),
+  }
+  const stateColors = brand ? brandColors : defaultColors
+  let renderedColor
+
+  if (disabled) renderedColor = colors.background
+  else if (selected && hovered) renderedColor = stateColors.all
+  else if (selected) renderedColor = stateColors.selected
+  else if (hovered) renderedColor = stateColors.hovered
+  else renderedColor = colors.background
```
mdodgelooker, line 55: Aha, my bad. My earlier suggestion to add `colors.background` as the fallback is covering up the border style.

ElliotTheHuman: Ah good catch. I'm a bit stumped as to what the ideal solution is here. `colors.background` plays nice with dark backgrounds, but not with border. And "none" plays nice with border but not dark backgrounds.

Would accepting `border` as a part of the parameter object and using `colors.background` or "none" depending on if `border` is true be the right answer? The downside to this is that using border would be unsupported for dark background afaik.

ElliotTheHuman: Tagging @lukelooker and @jhardy if interested

mdodgelooker: @ElliotTheHuman I'd say for this PR go back to `none` like it was before. Has anyone complained about `Tree` on a dark bg? If/when they do, a new PR can figure it out.

mdodgelooker: Random thought – I wonder why the `yarn image-snapshots` didn't complain? It looks like on master, it isn't showing the borders: https://github.com/looker-open-source/components/blob/master/packages/components/snapshots/Tree/Border-snap.png
Perhaps the renderer can't handle the gradient...

ElliotTheHuman: Eek. That's a bummer 😞 

For future reference, I will say that one of my monitors has had trouble rendering the `border` unless I zoom in. No issues when it comes to my Macbook Pro's screen, but my external monitor appears as if it isn't rendering the `border` (or does it so faintly that the naked eye can't see it).

packages/components/src/Tree/utils/treeBackgroundColor.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+  const defaultColors = {
+    all: itemSelectedColor(colors.ui2),
+    hovered: colors.ui1,
+    selected: itemSelectedColor(colors.ui2),
+  }
+  const stateColors = brand ? brandColors : defaultColors
+  let renderedColor
+
+  if (disabled) renderedColor = 'none'
```
lukelooker, line 51: You could just `return` here instead:

```suggestion
  if (disabled) return
```

packages/components/src/Tree/utils/treeBackgroundColor.ts
```diff
@@ -0,0 +1,60 @@
...
+  else if (selected && hovered) renderedColor = stateColors.all
+  else if (selected) renderedColor = stateColors.selected
+  else if (hovered) renderedColor = stateColors.hovered
+  else renderedColor = 'none'
```
lukelooker, line 55:
```suggestion
  else return
```

lukelooker (changes requested):
Minor nit on function return (no CSS req'd to do "nothing" in these cases.

packages/components/src/Tree/utils/treeBackgroundColor.ts
ElliotTheHuman: 👍  Resolved in 7b76233

packages/components/src/Tree/utils/treeBackgroundColor.ts
ElliotTheHuman: 👍  Resolved in 7b76233

lukelooker (approved):
LGTM (border works on high and low res monitors now)

PR #1725, opened 2020-12-23
Circular Progress variants

dbchristopher:
@jhardy I'm working through the circular progress PR at the moment. It's looking really great! One piece of feedback I have is around the way it's handling the size variants. I found myself tripped up by how the parent `CircularProgress` component would take in the string size and convert it to an object before passing it on to the child components. I find it a lot easier to just pass the prop down as-is, and have all three components access the helpers directly.

In addition, we have a standard way to handle component variants through styled system's `variant` helper. I wanted to convert it to that pattern as well, for consistency and readability, as well as the nice integration into styled components. 

It was too much to try to explain in the original branch without just changing a bunch of files, so I went ahead and opened up this new PR with those suggested changes. Take a look and let me know if this all makes sense!


packages/components/src/ProgressCircular/ProgressCircular.tsx
```diff
@@ -55,37 +51,32 @@ export interface ProgressCircularProps {
...
+        <IndeterminateProgress size={size} />
       )}
     </ProgressContainter>
   )
 }
 
-const ProgressContainter = styled.div<{ size: ProgressCircularSizeProps }>`
-  direction: ltr;
+const ProgressContainter = styled.div<{ size: ProgressCircularSizes }>`
+  ${progressCircularSize}
```
dbchristopher, line 79: styled-system variants integrate cleanly here

jhardy (approved):
@dbchristopher this is great, thanks for taking the time to do this. I always forget about styled system variants and this cleans it up nicely!

PR #1734, opened 2020-12-30
Fix InputText mousedown issue

mdodgelooker:
### :sparkles: Changes

- A recent change in https://github.com/looker-open-source/components/pull/1692 added a `preventDefault()` on `InputText`'s wrapper `<div>` mousedown event if the internal input was already focused. That change fixed a bug in `Select`, where `InputText` is an underlying component.
- The aim was to avoid triggering a blur and having to re-focus the input when the caret icon (or summary text, etc) is clicked.
- The unintended side effect was that it broke moving the cursor the input and selecting text with the mouse.
- This PR checks if the mousedown originated in the input and doesn't do anything if so.
- I tried to add a test to guard against this issue but apparently jsdom doesn't simulate moving the cursor around or selecting text on mousedown.

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

#### Issue:
391519f0-44fe-41b1-87bd-4688bd8a3da4

#### Fixed:
fd05e100-0236-4690-910b-364c498ad1e4

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `InputText` issue moving cursor and selecting text when input is already focused
 - `TooltipContent` default width is back to `'auto'`
 - Erratic scrolling after dynamic list resize in all `Combobox`-based components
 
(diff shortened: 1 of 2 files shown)


dbchristopher (approved):
LGTM. In general I wonder if it's even a helpful abstraction to have ComboboxInput build on top of InputText? Select is such a special case it might make things simpler in the future to build a one-off text input component for use in combobox

PR #1743, opened 2021-01-06
Menu: Updated structure

mdodgelooker:
### :sparkles: Changes

- Phase 1 of a 3-phase `Menu` refactor*
  - Phase 2: changing keyboard nav to "save" focus location
  - Phase 3: adding `submenu` to `MenuItem`
- Simplifies the structure for improved ease of use
- Matches `Popover` structure, but wraps `content` in a `MenuList` and adds 2 extra DOM props to the child (`aria-controls` and `disabled`)
  - Due to the 2 extra DOM props I wasn't able to support render-props structure, but hopefully that's ok since I think we're only supporting that in `Popover` for backward compatibility?
- `MenuDisclosure` is deprecated
  - Instead of the `tooltip` prop on `MenuDisclosure`, `Tooltip` can be nested inside `Menu`
- Removes the need for `MenuContext`, which has been known to cause bugs
- Before, any `Dialog`s triggered by a `MenuItem` had to be placed outside `MenuList` – now they will have to be placed outside of `Menu` (see new story example)

- [x] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC


packages/components/src/Menu/MenuItem.tsx
```diff
@@ -93,7 +94,7 @@ const MenuItemInternal: FC<MenuItemProps> = (props) => {
     onBlur && onBlur(event)
   }
 
-  const { setOpen } = useContext(MenuContext)
+  const { closeModal } = useContext(DialogContext)
```
mdodgelooker, line 97: For closing the menu's `Popover` now that `MenuContext` is deprecated. I just added a row to the Tech Debt table for replacing `Popover`'s use of `DialogContext` with a new `PopoverContext`.

packages/components/src/Menu/Menu.tsx
```diff
@@ -24,57 +24,87 @@
...
+    const id = useID(propsID)
+    const list = content && (
+      <MenuList id={id} {...props} ref={listRef}>
+        {content}
+      </MenuList>
+    )
+    children = cloneElement(children, {
+      'aria-controls': id,
+      disabled,
+    })
```
mdodgelooker, line 86: These two DOM props are unique to `Menu` and not present in `Popover` so we need to add them to the child here before `Popover` adds its DOM props.

packages/components/src/DataTable/Item/DataTableAction.tsx
```diff
@@ -45,13 +45,6 @@ export interface DataTableActionProps extends CompatibleHTMLProps<HTMLElement> {
...
  * allows us to adapt to any changes to MenuItem or its interface.
  * */
-export const DataTableAction = (props: DataTableActionProps) => {
-  const { setOpen } = useContext(MenuContext)
-  const handleActionClick = (
-    event: React.MouseEvent<HTMLElement, MouseEvent>
-  ) => {
-    setOpen && setOpen(false)
-    props.onClick && props.onClick(event)
-  }
```
mdodgelooker, line 55: This "close menu on click" behavior was added here before the same behavior was added to `MenuItem` itself, so it's been duplicated for a while.

packages/components/src/Button/IconButton.tsx
```diff
@@ -163,8 +164,9 @@ const IconButtonComponent = forwardRef(
       },
       tooltip,
     } = useTooltip({
-      content: label,
-      disabled: tooltipDisabled || hasOuterTooltip,
+      // ariaExpanded indicates an open Popover or Menu – don't show the tooltip
+      content:
+        tooltipDisabled || hasOuterTooltip || ariaExpanded ? undefined : label,
```
mdodgelooker, line 169: I was having an issue nesting `Tooltip` inside a `Menu` (previously `MenuDisclosure` had a `tooltip` prop, but with the new structure, nesting is the preferred way). When a `Menu` is disabled, the `Tooltip` was intercepting the `disabled` prop, using it to not render the tooltip, but not passing it down to the child, so the button was not getting disabled.

So I changed how `disabled` works on `Tooltip`. I removed the `disabled` prop from `useTooltip` and instead look for `content` being undefined to prevent the tooltip from rendering. `Tooltip` still accepts `disabled` but merely passes it on to the child, to support disabling a `Menu`. See changes in `Tooltip.tsx` and `useTooltip.tsx` below.

The alternative to these changes would be to not pass down `disabled` to the `Menu` child and instead require the developer to add the `disabled` prop to the menu button themselves. Thoughts?

www/src/documentation/components/overlays/menu.mdx
```diff
@@ -88,218 +114,76 @@ Control the state of the menu with the React `useState` hook.
...
+        >
+          <IconButton icon="DotsVert" label="More Options" />
         </Menu>
       </Flex>
     </Card>
   )
 }
 ```
 
-## MenuList
```
mdodgelooker, line 106: I moved `MenuList`, `MenuGroup`, and `MenuItem` to their own docs pages. Eventually `MenuList` will become `List` and `ListItem` will hold nearly all the functionality of `MenuItem` except closing the menu on click and (coming soon) the `submenu` prop.

packages/components/src/Button/IconButton.tsx
lukelooker: > Tooltip still accepts disabled but merely passes it on to the child, to support disabling a Menu

I think this is unintuitive. I'd prefer to explicitly name it – "disableChild" or, preferably just leave it to the developer to handle.

packages/components/src/utils/mergeHandlers.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export const mergeHandlers = <E extends SyntheticEvent>(
+  newHandler?: (e: E) => void,
+  existingHandler?: (e: E) => void
+) => (event: E) => {
+  existingHandler?.(event)
+  if (!event.defaultPrevented) {
+    newHandler?.(event)
+  }
+}
```
lukelooker, line 42: Neat :)

packages/components/src/Button/IconButton.tsx
mdodgelooker: Thanks @lukelooker ! Reworked the `disabled` prop to be more intuitive, per our convo.

lukelooker (approved):
LGTM ;)

PR #1772, opened 2021-01-12
fix: MenuList windowing initial render performance

mdodgelooker:
### :sparkles: Changes

- Previously the initial state for `useWindow` was to render all items in the list, then the height of the list container element is measured, and the state updates to only render the necessary items.
- This basically defeated the purpose of windowing, making the initial render just as slow as it would have been without `useWindow`.
- This PR updates the initial state to only render the first list item until the correct window size is calculated immediately after.

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC

### DIFF
--- a/packages/components/src/utils/useWindow.tsx
+++ b/packages/components/src/utils/useWindow.tsx
@@ -55,18 +55,18 @@ interface WindowHeightPayloadObject {
 }
 
 interface WindowHeightAction {
-  type: 'CHANGE' | 'RESET'
-  payload: number | WindowHeightPayloadObject
+  type: 'CHANGE'
+  payload: WindowHeightPayloadObject
 }
 
-const initialState = (length: number) => ({
+const initialState = {
   afterHeight: 0,
   beforeHeight: 0,
-  end: length - 1,
+  end: 0,
   scrollBottom: 0,
   scrollTop: 0,
   start: 0,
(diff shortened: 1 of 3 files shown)


packages/components/src/utils/useWindow.tsx
```diff
@@ -55,18 +55,18 @@ interface WindowHeightPayloadObject {
 }
 
 interface WindowHeightAction {
-  type: 'CHANGE' | 'RESET'
-  payload: number | WindowHeightPayloadObject
+  type: 'CHANGE'
```
mdodgelooker, line 58: `RESET` is no longer needed since it was just used for when `enabled` changes to `false`. But now whenever `enabled` is false, the window state is not even used and instead the entire `childArray` is simply returned.

packages/components/src/utils/useWindow.tsx
```diff
@@ -55,18 +55,18 @@ interface WindowHeightPayloadObject {
...
+  payload: WindowHeightPayloadObject
```
mdodgelooker, line 59: `number` here was only being used to calculate the `end` as `length - 1` in `initialState` but now it's statically set to `0`.

lukelooker (approved):
LGTM

PR #1773, opened 2021-01-12
feat: List and ListItem

ElliotTheHuman:
### :sparkles: Changes

- Added `List` and `ListItem`
  - Also added `ListItemContext`, `ListItemLayout`, `ListLabel` (used if `List` is given optional "label' prop value)
  - `getListItemDimensions` util function
  - story and test files

### 📓  Notes

- `Menu`-related image snapshots were affected by the `ListItemDetail` spacing change (which is used in both `ListItem` and `MenuItem`)
  -  Spoke to Jared about this and he gave me the 👍  to have consistency between these two

- [ ] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### :camera: Screenshots
`List`
Screen Shot 2021-01-13 at 3 23 13 PM

`ListItem`
Screen Shot 2021-01-13 at 3 23 40 PM


packages/components/src/List/utils/dimensions.ts
```diff
@@ -0,0 +1,68 @@
...
+  iconSize: 'xsmall',
+  py: 'xxsmall',
+}
+export const densityN3: ListItemDimensions = {
+  ...densityN2,
+  height: 24,
+  iconSize: 'xxsmall',
+  labelFontSize: 'xxsmall',
+  labelLineHeight: 'xsmall',
+}
```
ElliotTheHuman, line 68: Wondering if it's just better to repeat out the specs for each density rather than spread them onto each other.

If we avoid spreading them, we could create an object where keys are density values and values are dimension objects, which could be more elegant!

packages/components/src/List/List.tsx
```diff
@@ -0,0 +1,217 @@
...
+        console.warn(
+          `Define a width when using windowing to avoid width fluctuations during scrolling.`
+        )
+      }
+    }, [windowing, props.width])
+
+    const { content, ref } = useWindow({
+      childHeight: childArray[0]
+        ? getListItemHeight(childArray[0] as ReactChild, itemDimensions.height)
+        : 0,
```
ElliotTheHuman, line 169: We removed the use of `useMemo` here since we're not supporting 'variable' windowing at the moment. That being said, it could still be worth it to memoize that child height value.

@mdodgelooker @lukelooker Thoughts?

mdodgelooker: @ElliotTheHuman yep that should be fine – a function (for variable windowing) would be newly instantiated on every render, triggering too many changes downstream, hence the `useMemo` (can't do `useCallback` b/c it might be a number). Since it'll only ever be a number here, no need to memoize.

packages/components/src/List/utils/dimensions.ts
lukelooker: Couldn't you get both by using these to create said object:

```
const densities = {
  1: densityP1,
  0: density0,
 -1: densityN1,
...
}

File: packages/components/src/List/utils/getListItemDimensions.ts [hunk 3]
```diff
@@ -0,0 +1,51 @@
...
+    case -2:
+      return { ...densityN2 }
+    case -3:
+      return { ...densityN3 }
+    case 1:
+      return { ...densityP1 }
+    default:
+      return { ...density0 }
+  }
+}
```
lukelooker, line 51: With the object above this could become a simple object lookup:


```suggestion
export const getListItemDimensions = (
  density: DensityRamp
): ListItemDimensions => densities[density]
```

packages/components/src/List/ListItem.tsx
```diff
@@ -0,0 +1,214 @@
...
+        artwork={iconArtwork}
+        color="text1"
+        name={icon}
+        size={itemDimensions.iconSize}
+        mr={itemDimensions.iconGap}
+      />
+    ) : (
+      renderIconPlaceholder && (
+        <IconPlaceholder
+          data-testid={`list-item-${renderedIconID}-icon-placeholder`}
```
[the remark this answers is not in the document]
ElliotTheHuman: We had this here strictly for help testing the ID iirc.

That being said, image snapshotting would probably work even better for testing that the placeholder behavior works on a branch. I'm for scrapping this (and the `useId` call) if you're cool with that.

packages/components/src/List/List.tsx
```diff
@@ -0,0 +1,217 @@
...
+   * By default List will reposition itself if they overflow the widow.
+   * You can use the pin property to override this behavior.
+   */
+  pin?: boolean
+
+  /**
+   * Allow the overlay to break out of the scroll parent
+   */
+  escapeWithReference?: boolean
+
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 8ac36fd

packages/components/src/List/List.tsx
```diff
@@ -0,0 +1,217 @@
...
+  ${reset}
+
+  ${height}
+  ${minHeight}
+  ${maxHeight}
+  ${minWidth}
+  ${maxWidth}
+  ${width}
+
+  border-radius: inherit;
```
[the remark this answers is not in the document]
ElliotTheHuman: Carry over from `MenuList`. I can scrap it here if it's unneeded.

Resolved in 8ac36fd

packages/components/src/List/List.tsx
```diff
@@ -0,0 +1,217 @@
...
+  list-style: none;
+  outline: none;
+  overflow: auto;
+  user-select: none;
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 8ac36fd

packages/components/src/List/ListItem.tsx
```diff
@@ -0,0 +1,214 @@
...
+  detail?: ReactNode
+  icon?: IconNames
+  /**
+   * Sets the correct accessible role for the ListItem:
+   * - Use **'link'** for items that navigation to another page
+   * - Use **'button'** for items that trigger in page interactions, like displaying a dialog
+   * @default 'button'
+   *
+   */
+  itemRole?: 'link' | 'button'
```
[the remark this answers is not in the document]
ElliotTheHuman: You are correct :( In this case, would the div w/ role="button" button trick be sufficient to get around this? Or is there a different solution that seems more appropriate?

packages/components/src/List/ListItem.tsx
```diff
@@ -0,0 +1,214 @@
...
+  tooltip?: string
+  tooltipPlacement?: Placement
```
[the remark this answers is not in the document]
ElliotTheHuman: Yes we can!

Resolved in 8ac36fd

packages/components/src/List/ListItem.tsx
```diff
@@ -0,0 +1,214 @@
...
+        name={icon}
+        size={itemDimensions.iconSize}
+        mr={itemDimensions.iconGap}
+      />
+    ) : (
+      renderIconPlaceholder && (
+        <IconPlaceholder
+          data-testid={`list-item-${renderedIconID}-icon-placeholder`}
+          size={itemDimensions.iconSize}
+          iconGap={itemDimensions.iconGap}
```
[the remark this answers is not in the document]
ElliotTheHuman: Not that I can think of. We don't currently support just `MarginProps` though in design-tokens so I wasn't sure if bringing all of `SpaceProps` was the call. Is there a more appropriate interface or are we okay with giving it the whole space enchilada?

packages/components/src/List/ListItem.tsx
```diff
@@ -0,0 +1,214 @@
...
+   * `target="_blank" can be used to reverse tab-nab
+   * https://owasp.org/www-community/attacks/Reverse_Tabnabbing
+   */
+  const noTabNab = 'noopener noreferrer'
+  const rel =
+    target === '_blank'
+      ? props.rel
+        ? `${props.rel} ${noTabNab}`
+        : noTabNab
+      : props.rel
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 248f502 / 4ae2bb0

packages/components/src/List/ListLabel.tsx
```diff
@@ -0,0 +1,67 @@
...
+        we detect when this 0-height element disappears from the page and then
+        render the shadow.
+      */}
+      <div ref={labelShimRef} style={{ height: '0' }} />
+      <Heading as="h2" fontSize="small" fontWeight="semiBold" pl="medium">
+        {children}
+      </Heading>
+    </ListLabelWrapper>
+  )
+}
```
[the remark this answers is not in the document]
ElliotTheHuman: Good call. Got into a mix up between the remote and my local branch, and the label removal commit got lost in the crossfire.

Resolved in 03f544e

packages/components/src/List/utils/getListItemDimensions.ts
ElliotTheHuman: That works for me! Honestly, my biggest thing with the object route was the naming convention I whipped up, but if it's passable there's definitely no downside afaik to having these available.

ElliotTheHuman: Resolved in ab951f7

packages/components/src/List/List.tsx
```diff
@@ -0,0 +1,217 @@
...
+import { DensityRamp } from './types'
+
+export interface ListProps
+  extends Omit<CompatibleHTMLProps<HTMLUListElement>, 'label'>,
+    MaxHeightProps,
+    MinHeightProps,
+    HeightProps,
+    MaxWidthProps,
+    MinWidthProps,
+    WidthProps {
```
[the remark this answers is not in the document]
ElliotTheHuman: Makes sense for `List` to fill its container. I'm on board with removing these.

packages/components/src/List/List.tsx
```diff
@@ -0,0 +1,217 @@
...
+        <ul tabIndex={-1} {...omitStyledProps(props)} {...navProps}>
+          {label && <ListLabel>{label}</ListLabel>}
+          {content}
+        </ul>
+      </ListItemContext.Provider>
+    )
+  }
+)
+
+export const List = styled(ListInternal).attrs(({ minWidth = '12rem' }) => ({
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in a025132

packages/components/src/List/List.tsx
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in a025132

packages/components/src/List/utils/dimensions.ts
ElliotTheHuman: Resolved in a025132

packages/components/src/List/List.tsx
```diff
@@ -0,0 +1,217 @@
...
+      onFocus,
+      onKeyDown,
+      ref,
+    })
+
+    const context = {
+      itemDimensions,
+      renderIconPlaceholder,
+      setRenderIconPlaceholder,
+    }
```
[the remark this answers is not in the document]
ElliotTheHuman: I think children-injection behavior makes sense here since no one outside of the direct `ListItem` children needs the current context info.

That being said, would there be any problems with the children-injection behavior if we introduced a `ListHeader` component that could also be a child of `List`? That's the only potential argument that favors context in my mind (that being said, might not be much of an argument).

ElliotTheHuman:
Addressed most of the conversations, but some seemed to require more back-and-forth so left those open!

packages/components/src/List/ListItem.tsx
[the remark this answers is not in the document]
ElliotTheHuman: Will remove useId call here

packages/components/src/List/ListItem.tsx
```diff
@@ -0,0 +1,191 @@
...
+      )
+    )
+
+  if (disabled && itemRole === 'link') {
+    // eslint-disable-next-line no-console
+    console.warn(
+      'itemRole="link" and disabled cannot be combined - use itemRole="button" if you need to offer a disabled ListItem'
+    )
+  }
+  const Component = !disabled && itemRole === 'link' ? 'a' : 'button'
```
lukelooker, line 136: Need "listItem"

lukelooker, line 136:
```suggestion
  const Component = !disabled && itemRole === 'link' ? 'a' : itemRole === 'button' ? 'button' : Fragment
```

packages/components/src/List/ListItem.tsx
```diff
@@ -0,0 +1,191 @@
...
+      {listItemContent}
+    </ListItemLayout>
+  )
+}
+
+export const ListItem = styled(ListItemInternal)`
+  ${Icon} {
+    align-self: ${({ description }) => (description ? 'flex-start' : 'center')};
+  }
+`
```
lukelooker, line 191: Look to combine w/ `ListItemLayout`

packages/components/src/List/List.tsx
```diff
@@ -0,0 +1,138 @@
...
+    const navProps = useArrowKeyNav({
+      onBlur,
+      onFocus,
+      onKeyDown,
+      ref,
+    })
+
+    const context = {
+      itemDimensions,
+      renderIconPlaceholder,
```
lukelooker, line 120: `iconGutter` instead - explicitly set on `List` default to false

packages/components/src/List/utils/dimensions.ts
```diff
@@ -0,0 +1,77 @@
...
+  ...density0,
+  height: 32,
+  py: '0.375rem',
+}
+export const densityN2: ListItemDimensions = {
+  ...densityN1,
+  height: 28,
+  iconGap: 'xsmall',
+  iconSize: 'xsmall',
+  py: 'xxsmall',
```
lukelooker, line 60: css`calc(${theme.xxsmall} + ${theme.xxxsmall})`

packages/components/src/List/ListItem.tsx
ElliotTheHuman: Resolved in bc3c64a

packages/components/src/List/utils/dimensions.ts
```diff
@@ -0,0 +1,77 @@
...
+  labelFontSize: 'medium',
+  labelLineHeight: 'medium',
+  py: 'small',
+}
+
+// Negative density values
+export const densityN1: ListItemDimensions = {
+  ...density0,
+  height: 32,
+  py: '0.375rem',
```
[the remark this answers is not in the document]
ElliotTheHuman: IIRC, the plan was to leave the density objects alone and to leave a comment with this calc for now (in case we wanted to revisit how density worked and make these functions that can accept theme).

packages/components/src/List/ListItem.tsx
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in f1595fe

packages/components/src/List/List.tsx
ElliotTheHuman: Resolved in f1595fe

packages/components/src/List/ListItem.tsx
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 8438ef7

packages/components/src/List/ListItem.tsx
ElliotTheHuman: Bad news, good news.

Bad news: Setting `Component` to a Fragment breaks the item layout (`ListItemLayout` has selectors that give child `a` / `button` elements the necessary layout CSS)

Good news: If we set `Component` to a `div` and then add that `div` to the necessary selectors, we _should_ get the behavior we want (i.e. a container type that won't bork if the detail includes a button element).

I pushed up a commit b3d77be that shows what this `div` approach looks like. If it doesn't work here, we can definitely revert it and explore other options.

ElliotTheHuman: Follow-up: Commit 39a3882b removes a test that assumes `ListItem` is a `button` when disabled. With the commit described above, `ListItem` will only be a button if `itemRole` is explicitly "button".

packages/components/src/Icon/IconPlaceholder.tsx
```diff
@@ -24,16 +24,25 @@
...
-export type IconPlaceholderProps = CompatibleHTMLProps<HTMLDivElement> &
-  SizeProps
+export interface IconPlaceholderProps
+  extends CompatibleHTMLProps<HTMLDivElement>,
+    SizeProps,
+    SpaceProps {}
 
 export const IconPlaceholder = styled((props: IconPlaceholderProps) => (
-  <div aria-hidden {...props} />
+  <div aria-hidden {...omit(props, 'iconGap')} />
```
lukelooker, line 44: Don't need this `omit` any more

packages/components/src/List/ListItem.tsx
```diff
@@ -0,0 +1,179 @@
...
+  /**
+   * Indicates the ListItem is checked
+   */
+  current?: boolean
+  /*
+   * optional extra description
+   */
+  description?: ReactNode
+  detail?: ReactNode
+  icon?: IconNames
```
lukelooker, line 49: All props should have JSDoc descriptions

packages/components/src/List/ListItem.tsx
```diff
@@ -0,0 +1,179 @@
...
+  /**
+   * Sets the correct accessible role for the ListItem:
+   * - Use **'link'** for items that navigation to another page
+   * - Use **'button'** for items that trigger in page interactions, like displaying a dialog
+   * - Use **'none'** for all other items
```
lukelooker, line 54: Thought... do we need `'none'` or is `undefined` basically inferring that already?

Another thought, should assigning `onClick` infer `button` or `link` and if so, which one? (`button` feels right but I'm not 100% sure of that)

packages/components/src/List/ListItem.tsx
```diff
@@ -0,0 +1,179 @@
...
+   * @default 'none'
```
lukelooker, line 55: `undefined` doesn't have a default really (see comment above)

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
```diff
@@ -84,7 +84,11 @@ const OptionLayout = ({ option, scrollIntoView }: OptionLayoutProps) => {
         data-testid="option-icon"
       />
     ) : hasIcons ? (
-      <IconPlaceholder size="small" data-testid="option-icon-placeholder" />
+      <IconPlaceholder
+        size="small"
+        data-testid="option-icon-placeholder"
+        mr="xsmall"
```
lukelooker, line 90: `mr` should be a default value on `IconPlaceholder` I think (both use cases assign it so it seems a safe default)

packages/components/src/Icon/IconPlaceholder.tsx
lukelooker: Also `aria-hidden` could be assigned via `attrs`

This would end up looking something like:

```jsx
export const iconPlaceholder = styled.div.attrs<IconPlaceholderProps>(
  ({ mr = 'xsmall' }) => ({ 'aria-hidden': true, mr })
  )<IconPlaceholderProps>`
  ${size}
  ${space}
`
```

lukelooker (changes requested):
Ok, sorry, dug deeper and found a few more minor things to add.

We'll also need to add documentation but that can be a follow-up PR.

I did push a commit to change the density story to make cross-comparison easier and add a few variations I wanted to see for comparison as well:

image

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
ElliotTheHuman: Resolved in 4b86540

packages/components/src/Icon/IconPlaceholder.tsx
ElliotTheHuman: Resolved in 4b86540

packages/components/src/List/ListItem.tsx
ElliotTheHuman: Resolved in 4b86540

packages/components/src/List/ListItem.tsx
ElliotTheHuman: Resolved in 4b86540

packages/components/src/List/ListItem.tsx
ElliotTheHuman: As discussed, the plan is to do a follow-up PR, bringing in the "detailAccessory" pattern from `Tree` (along with "hover disclosure" and "truncate").

This means that the item label will always be a button / link, but devs can use "detailAccessory" if they want to have a button in the detail element (to get around the button within a button problem).

lukelooker (approved):
LGTM.

PR #1776, opened 2021-01-13
feat(DataTable): Accessibility keyboard navigation

laraharrow:
### :sparkles: Changes

- Add keyboard accessibility to DataTable

- [ ] Update documentation
- [x] Updated CHANGELOG
- [x] PR is ideally < ~400LOC


packages/components/src/DataTable/DataTable.tsx
```diff
@@ -45,6 +47,10 @@ export const DataTableLayout: FC<DataTableProps> = (props) => {
     onSort,
     select,
   } = props
+
+  const ref = useCallback((node) => {
+    node?.focus()
+  }, [])
```
mdodgelooker, line 53: Not needed anymore:
```suggestion
```

packages/components/src/DataTable/getNextFocus.ts
```diff
@@ -0,0 +1,78 @@
...
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+export const getTabStops = (ref: HTMLElement): HTMLElement[] =>
+  Array.from(ref.querySelectorAll('td[tabindex="-1"],a,button'))
+
```
mdodgelooker, line 29: Some helper functions known as [type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) to make TS happy:
```suggestion
const isTableCell = (
  element: Element
): element is HTMLTableCellElement | HTMLTableHeaderCellElement => {
  return ['TD', 'TH'].includes(element.tagName)
}
const isTableRow = (
  element: Element | null
): element is HTMLTableRowElement => {
  return element ? element.tagName === 'TR' : false
}
```

packages/components/src/DataTable/getNextFocus.ts
```diff
@@ -0,0 +1,78 @@
...
+        if (rowBelow != null) {
+          return rowBelow.cells[cellIndex]
+        }
+      }
+    }
+
+    if (
+      document.activeElement &&
+      tabStops.includes(document.activeElement as HTMLElement)
+    ) {
```
mdodgelooker, line 70: I think it actually makes sense to move the "vertical" logic inside the `document.activeElement` check below. And then the type guard functions will fix the TS errors:

```suggestion
    if (
      document.activeElement &&
      tabStops.includes(document.activeElement as HTMLElement)
    ) {
      const element = document.activeElement
      if (vertical && isTableCell(element)) {
        const cellIndex = element.cellIndex
        if (direction === -1) {
          // Up Arrow
          const rowAbove =
            element &&
            element.parentElement &&
            element.parentElement.previousElementSibling
          if (isTableRow(rowAbove)) {
            return rowAbove.cells[cellIndex]
          }
        } else if (direction === 1) {
          // Down Arrow
          const rowBelow =
            element &&
            element.parentElement &&
            element.parentElement.nextElementSibling
          if (isTableRow(rowBelow)) {
            return rowBelow.cells[cellIndex]
          }
        }
      }
```

mdodgelooker (changes requested):
I haven't spent much time playing around with the keyboard nav yet, but here are some suggestions to make TS happy. :)

mdodgelooker (changes requested):
Playing around with keyboard nav a bit and noticed a few things:
1. Using just the `Tab` key, focus move from the "name" link to the actions menu in each row, which I don't think is what we want. Hitting tab once (when focused on something above the table – the Filters story is good for testing this) should move focus to the 1st column header, or select-all checkbox, if it's there, then hitting tab again should move focus out of the table to whatever's after it. I think any `<a>`, `<button>` or `<input>` need to have `tabIndex={-1}` to make this work.
 2. Focus is not moving into the column headers. Hitting the up arrow from the top row moves focus to the cell to the left, if there is one, but it should move focus to the header cell above.
3. Hitting the down arrow from the bottom row is moving focus to the cell to the right, if there is one, but it shouldn't move focus at all.
4. Hitting up or down arrow when the actions menu button is focused moves to the previous cell in the same row or the 1st cell in the next row but it should move to the actions button above or below.
5. Hitting the left arrow from the 1st cell in a row moves focus to the last cell in the previous row and hitting the right arrow from the last cell in a row moves to the 1st cell in the next row, but focus shouldn't move for either case.

laraharrow:
1. `<a>`, `<button>` or `<input>`  now have `tabIndex={-1}` but I'm not sure how to make only the fist element in the header to be focusable using `Tab` key
2. Focus now is going to the `header` elements but it is not moving up and down only sideways - not sure why that is.
3. fixed
4. fixed
5. not sure if that is something that is automatic happen from the keyboard arrows code that but I'm not quite sure how to fix this.

mdodgelooker:
1. I just realized `{...tabProps}` should go on the `<table>` instead of the `<div>`. This PR fixes that and some related stuff: https://github.com/looker-open-source/components/pull/1793
2. For the `<th>` it's not finding any `element.parentElement.nextElementSibling` because it's wrapped in `<thead>` so you'll have to write some custom custom logic for the header row.
5. I think you should be able to check if `tabStops[next]` is in the same table row as `document.activeElement` and return `null` if it isn't.

mdodgelooker:
1. Try to pass tabIndex={-1} to actionPrimary using cloneElement
2. Go between links (not focusing that table cell) Try adding a ref on <td> and useEffect that looks inside ref.current DOM node to find a, button, input and do element.setAttribute(‘tabindex’, ‘-1’)

packages/components/src/DataTable/Column/DataTableCell.tsx
```diff
@@ -40,46 +41,64 @@ export interface DataTableCellProps
...
+
+    const ref = useRef<HTMLTableDataCellElement>(null)
+    const forkedRef = useForkedRef(ref, forwardedRef)
 
-    if (indicator) {
+    useEffect(() => {
+      const element = ref?.current
+      const button = element?.querySelector('button')
+      const link = element?.querySelector('a')
+      const input = element?.querySelector('input')
```
lukelooker, line 58: Could this be made more efficient by looking for them in a single `querySelector`? e.g.:

```suggestion
      const tabIndexElement = element?.querySelector('button, a, input')
```

What would happen if there were more than one button, link or input? 

```suggestion
      const tabIndexElements = element?.querySelectorAll('button, a, input')
```

packages/components/src/DataTable/Column/DataTableCell.tsx
```diff
@@ -40,46 +41,64 @@ export interface DataTableCellProps
...
+
+      button?.setAttribute('tabIndex', '-1')
+      link?.setAttribute('tabIndex', '-1')
+      input?.setAttribute('tabIndex', '-1')
```
lukelooker, line 62: If above is accepted than this would need to be a loop over `tabIndexElements`

lukelooker (commented):
This is super cool... found a few items:

- Focus is visible when the user clicks a cell with a mouse. I think it's only supposed to show focus when using the keyboard?
- Focus ring is blue instead of purple (keyColor)
- `tab` should move focus out of the entire `DataTable`, it should not focus within the table

packages/components/src/DataTable/Column/DataTableCell.tsx
```diff
@@ -93,4 +112,7 @@ export const DataTableCell = styled(DataTableCellLayout)`
       text-decoration: underline;
     }
   }
+  :focus {
+    border: 1px solid ${({ theme }) => theme.colors.key};
+  }
```
mdodgelooker, line 117: I think the browser default `outline` might be the blue ring that we're seeing. And for the "focus visible" behavior (keyboard vs clicking), take a look at how `Tab.tsx` does it – the border style you have here should only be on when `focusVisible` is true.

(Don't worry about adding focus visible behavior to `DataTableHeaderCell` – it should already be on there from `useClickable`.)

```suggestion
  :focus {
    outline: none;
  }
```

packages/components/src/DataTable/Header/DataTableHeaderCell.tsx
```diff
@@ -78,9 +80,10 @@ const DataTableHeaderCellLayout = forwardRef(
             : 'none'
         }
         className={className}
-        onClick={handleClick}
         ref={ref}
         style={{ cursor: canSort ? 'pointer' : undefined }}
+        tabIndex={-1}
+        {...clickableProps}
```
mdodgelooker, line 86: I forgot that `clickableProps` contains `tabIndex: 0`. This change should fix the tab moving focus around the header issue:
```suggestion
        {...clickableProps}
        tabIndex={-1}
```

packages/components/src/DataTable/Column/DataTableCell.tsx
mdodgelooker, line 62: I agree with @lukelooker 's suggestion. This should be:
```suggestion
      if (element) {
        const focusableElements = Array.from(element.querySelectorAll('button,a,input'))
        focusableElements.forEach((el) => el.setAttribute('tabindex', '-1'))
      }
```

mdodgelooker (changes requested):
The Action and Primary Action example shows the awkwardness of up/down arrow navigation with multiple focusable elements inside a single cell. But I worry that any attempt to improve this would break down if a there were different numbers of focusable elements between cells in a single column. To me, the current awkwardness is acceptable, but perhaps others have ideas. 🙂

jhardy:
We have small issue with the table headers, the `role="button"` being applied is interfering with tying the heads to the cells. 

image

I get this was to make sure our headers announced as sortable, but I think breaking the semantic tie between the column and the cell is probably the wrong call. What we need to do is have the sort control be a button inside of the `th` so that it can be focused on its own. Not sure its worth doing that in this PR or addressing it in a follow up.

jhardy:
another thing I think we will need to think about is correct a11y handling of interactive controls with a description, here the link gets focused and read, but not the description. You can still get to the description using keyboard controls but wondering if there is a better pattern to implement 
image

packages/components/src/DataTable/Column/DataTableCell.tsx
```diff
@@ -24,73 +24,133 @@
...
+)
+
+DataTableCellLayout.displayName = 'DataTableCellLayout'
 
 export const DataTableCell = styled(DataTableCellLayout)`
   ${columnSize}
 
+  ${({ focusVisible, theme: { colors } }) =>
+    focusVisible && `border: 1px solid ${colors.key};`}
+
```
mdodgelooker, line 142: The focus ring isn't working right now b/c the `focusVisible` prop is applied to the internal `<td>`, not to `DataTableCell`, the external component, so this style block won't see it. You'll need to replace the `<td>` with a styled component, and move this style there.

My suggestion would be to create a new file, `FocusableCell` in the `Column` folder, something like the following. And then you could use this styled component instead of the `<td>` above and the `<th>` in `DataTableHeaderCell` (with `as="th"`) and also remove `tabIndex={-1}` from both since it's included with `FocusableCell`.
```
import { FocusVisibleProps } from '../../utils'

const focusVisibleStyle = css<FocusVisibleProps>`
  &:focus {
    outline: ${({ focusVisible, theme: { colors } }) => {
      return focusVisible ? `1px solid ${colors.key}` : 'none'
    }};
  }
`

export interface FocusableCellProps extends FocusVisibleProps {
  as?: 'th'
}

export const FocusableCell = styled.td
  .withConfig({
    shouldForwardProp,
  })
  .attrs<FocusableCellProps>(() => ({
    tabIndex: -1,
  }))<FocusableCellProps>`
  ${focusVisibleStyle}
`
```

jhardy:
Another small nit, is that the current focus indicator lives inside the cell, which causes the cell content to jump, its small but noticeable  

https://user-images.githubusercontent.com/170681/105767594-10266800-5f10-11eb-88bf-3534acc0e22b.mp4

A couple ideas we could relative position the cell, and use an absolute positioned pseudo content, I believe we do this on `Menu` , or we could probably use an inset box shadow.

packages/components/src/DataTable/Header/DataTableHeaderCell.tsx
```diff
@@ -53,11 +54,12 @@ const DataTableHeaderCellLayout = forwardRef(
   ) => {
     const { onSort } = useContext(DataTableContext)
 
-    const handleClick = () => {
+    const onClick = () => {
       if (onSort && canSort) {
         onSort(id, sortDirection === 'asc' ? 'desc' : 'asc')
       }
     }
+    const clickableProps = useClickable({ onClick })
```
mdodgelooker, line 62: To @jhardy 's point about `role="button"`, you can exclude it like so:
```suggestion
    const {role, ...clickableProps} = useClickable({ onClick })
```

mdodgelooker (changes requested):
A few suggestions for the `focusVisible` feature.

mdodgelooker:
> Another small nit, is that the current focus indicator lives inside the cell, which causes the cell content to jump, its small but noticeable
I could be wrong but I think if you change the focus ring to use outline instead of border, it won't jump.

mdodgelooker:
Also I noticed a bug in the `useClickable` hook where `focusVisible` wouldn't work with arrow keys. I just pushed a fix.

packages/components/CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 ### Features
 
+- Added keyboard accessibility to `DataTable`
+- `Tabs` and `Menu` arrow key navigation persists the last focused item
```
lukelooker, line 22: CHANGELOG is now maintained by automation so no edits should be made via PR

lukelooker (changes requested):
Remove CHANGELOG update and then GTG

lukelooker (approved):
LGTM!

PR #1780, opened 2021-01-14
feat(components): nestedMenu on MenuItem

mdodgelooker:
### :sparkles: Changes

- Adds `submenu` prop to `MenuItem`

#### Design questions
- Google Material leaves the submenu open when the mouse hovers off of it (but not back onto another menu item). Should we follow this? Currently I've got the submenu closing on hover out.
- The shadows don't match the spec. Should I update the shadow values in `design-tokens` in this PR?

- [x] Update documentation
- [x] Updated CHANGELOG
- [ ] PR is ideally < ~400LOC

### :camera: Screenshots
image


packages/components/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `Menu` nesting using the `submenu` prop on `MenuItem`
```
lukelooker, line 12: This should get removed :)

packages/components/src/Menu/Menu.tsx
```diff
@@ -75,7 +75,7 @@ export const Menu = forwardRef(
   ) => {
     const id = useID(propsID)
     const list = content && (
-      <MenuList id={id} {...props} ref={listRef}>
+      <MenuList id={id} {...props} ref={listRef} data-autofocus="true">
```
lukelooker, line 78: How'd this work before?

packages/components/src/Menu/MenuItem.tsx
```diff
@@ -171,24 +198,32 @@ const MenuItemInternal: FC<MenuItemProps> = (props) => {
...
+      <MenuItemLayout
+        aria-current={current && 'true'}
+        compact={compact}
+        disabled={disabled}
+        focusVisible={isFocusVisible}
+        onBlur={handleOnBlur}
+        onClick={disabled ? undefined : handleOnClick}
+        onKeyUp={handleOnKeyUp}
+        className={className}
+        {...props}
```
lukelooker, line 211: `props` weren't spread before - is there something in there we think needs spreading specifically or just a previous oversight?

packages/components/src/Overlay/OverlaySurface.tsx
```diff
@@ -109,12 +109,21 @@ export const OverlaySurface = styled(OverlaySurfaceLayout)`
 
 export const OverlaySurfaceContentArea = styled.div`
   background: ${({ theme }) => theme.colors.background};
-  border: 1px solid ${({ theme }) => theme.colors.ui2};
```
lukelooker, line 112: `border` was removed here. Intentional?

packages/components/src/utils/useResize.ts
```diff
@@ -41,7 +41,7 @@ export const useResize = (element: HTMLElement | null, handler: () => void) => {
       return
     }
 
-    throttledHandler()
+    handler()
```
lukelooker, line 44: Will this have adverse affects elsewhere? (think not but just pondering)

packages/components/src/Menu/MenuItem.tsx
```diff
@@ -65,35 +68,60 @@ export interface MenuItemProps extends CompatibleHTMLProps<HTMLElement> {
...
+    onMouseEnter,
+    onMouseLeave,
+    submenu,
+  })
+
+  detail = submenu ? (
+    <Icon color="text1" name="ArrowRight" size={compact ? 'small' : 'medium'} />
+  ) : (
+    detail
+  )
```
lukelooker, line 123: Will people want to be able to use `detail` & `submenu`? (I don't think it's terrible to say no but we might want to call this out in prop description and/or docs.

lukelooker (commented):
A few questions inline but functionality works great!

packages/components/src/Menu/MenuItem.tsx
lukelooker: Throw console warning if both are set

packages/components/src/Menu/MenuItem.tsx
lukelooker: answer: Allows `id` (if explicitly set, to pass-through)

lukelooker (commented):
- jsDoc for "only supports 1 level nesting"
- move NestedSurface to private spot
- Name "nestedX" consistently
- Console warn about detail + nestedMenu

mdodgelooker (commented):
Whoops, I never clicked submit on all my self-review comments. Here they are!

lukelooker (commented):
Everything looks really close but there's some complaint from Jest.

mdodgelooker:
Whoops, fixed the test. I'm gonna hold off on merging till @jhardy has a chance to weigh in on the design questions 🙂

packages/components/src/Menu/MenuItem.tsx
```diff
@@ -152,7 +183,7 @@ const MenuItemInternal: FC<MenuItemProps> = (props) => {
   const menuItemContent = (
     <Component
       href={href}
-      rel={createSafeRel(props.rel, props.target)}
+      rel={createSafeRel(props.rel, target)}
```
jhardy, line 186: Small a11y thing, `MenuItem`'s with submenus should have `aria-haspopup="menu"` and `aria-expanded="{false|true}"` depending if its open ore closed.

jhardy:
This is really nice, I only have a few small nits. 

1. Is it possible to close he parent menu when clicking on a submenu item? I feel like that would be the desired behavior most of the time.
2. When closing a submenu with the esc key or clicking outside, focus does not get placed back on the trigger element. For a11y and keyboard purposes we will want focus to return to trigger element if possible.

mdodgelooker:
- The nested menu now stays open when you hover out of it (but still closes when you hover onto a different item in the parent menu).
- Reverted back to `medium` for the surface border-radius.
- Keeping `has-popup="true"` for now but created a follow-up ticket for `usePopover` to support `has-popup="menu"` (as well as other values).

mdodgelooker:
Oops @jhardy just seeing your last 2 comments now. 
1. If the parent menu item has its own `onClick` action, both menus will close on click. If it doesn't, the click event opens the nested menu for interactions that don't involve hovering, i.e. `Enter` & `Spacebar` key for keyboard nav (in addition to `ArrowRight`) and for mobile screens, or keeps it open if it was already opened via hover. Shall I close it in the cases where it's already open via hover?
2. This is a known issue for which the fix is planned during fix-it week. :)

mdodgelooker:
@jhardy after my latest change the menus will close on click if the nested menu was already opened via hover.

lukelooker (approved):
Did a once-over of the code again and tried things out in Storybook - works really nicely and moving around with keyboard and/or mouse feels "right" in every case I could think to try. 

I say ship it :)

PR #1785, opened 2021-01-16
chore: Correct mis-configuration with Github Actions setup

lukelooker:
Remove Node 10 installation (just use pre-installed Node v12 that works just fine in core `.github/workflows/test.yml`


lukelooker:
First attempt crashed and burned due to (erroneously) hard-coded Node version:

image

PR #1800, opened 2021-01-21
feat: List & ListItem detail props

ElliotTheHuman:
- Added detailAccessory and detailHoverDisclosure props to `List` and `ListItem`
- Added truncate prop to `ListItem`
- Changed padding on `ListItem` to support detailAccessory prop (i.e. padding will be on child containers or parent container depending on detailAccessory)
- Added detailAccessory and detailHoverDisclosure-related tests 
- Added typography and textColor styled functions to `Truncate`

- Note: The image snapshots changes _should_ be from changing the label for `ListItem` stories. Was originally "Menu Item" 🤦‍♂️  Outside of that, those old snapshots _should_ be identical to previous snapshots. Thinking it might be best to keep the "Menu Item" label just so we can be absolutely sure the layout / padding changes to `ListItem` didn't change anything visually.

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts


packages/components/src/List/ListItemLayout.tsx
```diff
@@ -65,18 +68,20 @@ export const ListItemLayoutGrid = styled.div``
 export const ListItemLayout = styled(ListItemWrapper)`
   align-items: center;
   color: ${({ theme: { colors } }) => colors.text5};
+  display: flex;
   font-size: ${({ theme: { fontSizes } }) => fontSizes.small};
   font-weight: ${({ theme: { fontWeights } }) => fontWeights.normal};
+  height: ${({ description, height }) =>
+    description ? height + 16 : height}px;
```
ElliotTheHuman, line 75: Debating whether this should be min-height or height.

min-height would mean that the item container could expand vertically to properly fit content, but that would throw off the specific measurements we had for each density value.

height would keep density values mapped to specific heights, but putting taller items as content / detail could lead to weirdness.

This choice could affect Explore's Field Picker since we're currently using no vertical padding on `TreeItem`, which allows them to cleanly fit a 24px Icon into a 24px TreeItem.

image


@jhardy

packages/components/src/List/ListItemLayout.tsx
```diff
@@ -65,18 +68,20 @@ export const ListItemLayoutGrid = styled.div``
...
   text-decoration: none;
   transition: ${({ theme: { easings, transitions } }) =>
     `background ${transitions.quick}ms ${easings.ease},
     color ${transitions.quick}ms ${easings.ease}`};
 
-  button,
-  a,
-  & > div {
+  & > button,
+  & > a {
```
ElliotTheHuman, line 84: Adding a button or link element in the detail was affected by these selectors previously. Making them more specific should solve that.

packages/components/src/List/ListItemLayout.tsx
```diff
@@ -132,7 +144,9 @@ export const ListItemLayout = styled(ListItemWrapper)`
   }
 
   &:hover {
-    background: ${({ theme: { colors } }) => colors.ui1};
+    /* stylelint-disable */
```
ElliotTheHuman, line 147: Hit an odd stylelint issue involving space before the bang operator. Definitely open to suggestions on this one if the disabling isn't ideal.

packages/components/src/List/ListItemDetail.tsx
```diff
@@ -24,11 +24,15 @@
...
 
+import { padding, PaddingProps } from '@looker/design-tokens'
 import styled from 'styled-components'
 
-export const ListItemDetail = styled.div`
+export const ListItemDetail = styled.div.attrs<PaddingProps>((props) => ({
+  ...props,
+  pl: props.pl || 'xsmall',
+}))<PaddingProps>`
+  ${padding}
```
ElliotTheHuman, line 34: I needed to be able to have padding on the `ListItemDetail` when detailAccessory === true, but NOT when it was false (i.e. the parent layout container takes care of padding if false).

Giving this padding props seemed like the best solution, but could be wrong.

packages/components/src/Truncate/Truncate.tsx
```diff
@@ -26,18 +26,31 @@
...
+  CompatibleHTMLProps,
+} from '@looker/design-tokens'
 import { useIsTruncated } from '../utils/useIsTruncated'
 import { useTooltip } from '../Tooltip'
 
-export interface TruncateProps extends WidthProps {
+export interface TruncateProps
+  extends TypographyProps,
+    WidthProps,
+    CompatibleHTMLProps<HTMLSpanElement> {
```
ElliotTheHuman, line 41: Felt like `Truncate` could use TypographyProps. My thought was that since it was used in place of Paragraph depending on X prop, it should have similar props to Paragraph.

packages/components/src/List/List.tsx
```diff
@@ -52,6 +52,19 @@ export interface ListProps
...
+   * In addition, if true, events will not propagate from the detail container
+   * @default false
+   */
+  detailAccessory?: boolean
+
+  /**
+   * If true, child ListItems' detail elements will only appear on hover
+   * @default false
+   */
+  detailHoverDisclosure?: boolean
```
lukelooker, line 66: I'm _tempted_ to say that we shouldn't support these on list and only support them directly on `listItem`

packages/components/src/List/ListItem.tsx
```diff
@@ -48,6 +56,17 @@ export interface ListItemProps extends CompatibleHTMLProps<HTMLElement> {
...
+   * If true, the detail elements will appear outside of the item's grey background on hover
+   * In addition, if true, events will not propagate from the detail container
+   * @default false
+   */
+  detailAccessory?: boolean
+  /**
+   * If true, the detail element will only appear on hover
+   * @default false
+   */
+  detailHoverDisclosure?: boolean
```
lukelooker, line 69: Crazy idea... what if `detail` took one of two type signatures:

```tsx
interface DetailOptions {
  /**
   * If true, child ListItems' detail will appear outside of the item's grey background on hover
   * In addition, if true, events will not propagate from the detail container
   * @default false
   */
  accessory?: boolean
  /**
   * If true, child ListItems' detail will only appear on hover
   * @default false
   */
  hoverDisclosure?: boolean
}

interface ListItem {
  ...
  detail: ReactNode | { content: ReactNode, options: DetailOptions }
}
```

packages/components/src/List/ListItem.tsx
```diff
@@ -125,60 +169,93 @@ const ListItemInternal: FC<ListItemProps> = (props) => {
...
+          {children}
+        </Truncate>
+      ) : (
+        <Paragraph
+          fontSize={itemDimensions.labelFontSize}
+          lineHeight={itemDimensions.labelLineHeight}
+        >
+          {children}
+        </Paragraph>
+      )
```
lukelooker, line 190: Use dynamic component reference to simplify this some...

```
     const Component = truncate ? Truncate : Paragraph

     (<Component
        fontSize={itemDimensions.labelFontSize}
        lineHeight={itemDimensions.labelLineHeight}
      >
        {children}
      </Component>
      )
```

packages/components/src/List/ListItem.tsx
```diff
@@ -125,60 +169,93 @@ const ListItemInternal: FC<ListItemProps> = (props) => {
...
+  const renderedDescription =
+    typeof children === 'string' ? (
+      truncate ? (
+        <Truncate color="text2" fontSize="xsmall">
+          {description}
+        </Truncate>
+      ) : (
+        <Paragraph color="text2" fontSize="xsmall">
+          {description}
+        </Paragraph>
```
lukelooker, line 204: Dynamic JSX can be used here too.

packages/components/src/List/ListItemDetail.tsx
lukelooker: I'd be tempted to make the solution narrower by having an `accessory` prop that just toggles the padding value as needed instead of making `PaddingProps` available any upstream `ListItemDetail` consumers.

packages/components/src/List/ListItemLayout.tsx
lukelooker: Hmmm...

packages/components/src/List/ListItemLayout.tsx
```diff
@@ -87,20 +92,28 @@ export const ListItemLayout = styled(ListItemWrapper)`
...
-    padding: ${({ px, py, theme }) =>
-      `${py === '0.375rem' ? py : theme.space[py]} ${theme.space[px]}`};
+    padding: ${({ detailAccessory, px: propsPx, py: propsPy, theme }) => {
+      const pt = propsPy === '0.375rem' ? propsPy : theme.space[propsPy]
+      const pr = detailAccessory ? '0' : theme.space[propsPx]
+      const pb = pt
+      const pl = theme.space[propsPx]
+
+      return `${pt} ${pr} ${pb} ${pl}`
+    }};
```
lukelooker, line 107: Probably worth lifting this out into a separate function given the level of complexity within.

packages/components/src/List/ListItemLayout.tsx
lukelooker: I wonder if you could do something like `detailAccessory === false && colors.ui1` to not upset stylelint?

packages/components/src/Truncate/Truncate.tsx
lukelooker: Probably fair. It's worth noting that `Truncate` uses a `span` underneath but also has `display: block` so it's an odd fit but I can live with it.

lukelooker (commented):
When `ListItem` has a `detail` & is `selected` with `detailAccesssory` the hover behavior doesn't look quite as-expected:

image

packages/components/src/List/ListItemLayout.tsx
ElliotTheHuman: Discussed over Meet, but the plan is to try a css grid implementation that allows detail children to be as big as their parent container.

packages/components/src/List/List.tsx
ElliotTheHuman: Resolved in c82b064

packages/components/src/Truncate/Truncate.tsx
ElliotTheHuman: Resolved in 49909b9

ElliotTheHuman: Resolved in 49909b9

packages/components/src/List/utils/getItemBackgroundColor.ts
```diff
@@ -48,11 +48,11 @@ export const treeBackgroundColor = ({
   const stateColors = brand ? brandColors : defaultColors
   let renderedColor
 
-  if (disabled) return
+  if (disabled) renderedColor = 'transparent'
   else if (selected && hovered) renderedColor = stateColors.all
   else if (selected) renderedColor = stateColors.selected
   else if (hovered) renderedColor = stateColors.hovered
-  else return
+  else renderedColor = 'transparent'
```
ElliotTheHuman, line 55: disabled and else needed to output a valid "background-color" CSS value. Otherwise, the browser's default button CSS takes over in `ListItemLayout`.

packages/components/src/List/ListItemLayout.tsx
ElliotTheHuman: Resolved in 66bbd9c

packages/components/src/List/ListItem.tsx
ElliotTheHuman: Resolved in f838b1b

packages/components/src/List/ListItem.tsx
ElliotTheHuman: Resolved in f838b1b

packages/components/src/List/ListItemLayout.tsx
ElliotTheHuman: Resolved in 66d4330

packages/components/src/List/ListItem.tsx
ElliotTheHuman: Resolved in eaa1187

packages/components/src/List/ListItemLayout.tsx
ElliotTheHuman: Resolved in 66d4330

packages/components/src/List/ListItemDetail.tsx
ElliotTheHuman: The one problem is that a `ListItem`'s density determines the size of the conditional "padding-right" on `ListItemDetail`. In which case we need two props (i.e. "density" and "accessory") to replace `PaddingProps` props.

I can do that if that's preferable here!

ElliotTheHuman:
Hey Luke! Got the first round of List & ListItem feedback done. Everything was straightforward except the potential grid implementation (see commit 66d4330).

TLDR is that I couldn't get grid to solve for a lot of the confusing pieces, like the conditional padding on the `ListItemDetail` or having the detail sit within or outside of the internal "button" / "a" element. Apologies ahead of time I'm ignorant to any grid-related properties / magic :( It did help organize some of the JSX structure in the `ListItem` component (but did require the addition of more containers within).

packages/components/src/List/ListItem.tsx
```diff
@@ -62,13 +104,20 @@ export interface ListItemProps extends CompatibleHTMLProps<HTMLElement> {
    * - Use **'button'** for items that trigger in page interactions, like displaying a dialog
    */
   itemRole?: 'link' | 'button'
+  /**
+   * Indicates the ListItem is checked
+   */
+  selected?: boolean
```
lukelooker, line 110: `current` and `selected` may need to both exist as `current` has a specific ARIA meaning (current page) vs. `selected` - not exactly sure the right answer here but we should discuss / research what the best practice here needs to be. It may be the case that `selected` and/or `current` needs to either-or perhaps? Anyway, topic to ponder a bit.

packages/components/src/List/types.ts
```diff
@@ -49,3 +49,26 @@ export const itemDimensionKeys = [
...
+   * @default false
+   */
+  disabled?: boolean
+  /**
+   * If true, the Tree/TreeItem will have a light background color
+   * @default false
+   */
+  hovered?: boolean
+  /**
+   * If true, the Tree/TreeItem will have a darker background color
```
lukelooker, line 70:
```suggestion
   * If true, the ListItem will have a darker background color
```

packages/components/src/List/types.ts
```diff
@@ -49,3 +49,26 @@ export const itemDimensionKeys = [
...
+   * @default false
+   */
+  brand?: boolean
+  /**
+   * If true, the Tree/TreeItem will have a "disabled" presentation.
+   * @default false
+   */
+  disabled?: boolean
+  /**
+   * If true, the Tree/TreeItem will have a light background color
```
lukelooker, line 65:
```suggestion
   * If true, the ListItem will have a light background color
```

packages/components/src/List/types.ts
```diff
@@ -49,3 +49,26 @@ export const itemDimensionKeys = [
...
 ]
+
+export interface ItemBackgroundColorProps {
+  /**
+   * Replace the normal grey selected and selected + hovered color with brand colors
+   * @default false
+   */
+  brand?: boolean
+  /**
+   * If true, the Tree/TreeItem will have a "disabled" presentation.
```
lukelooker, line 60:
```suggestion
   * If true, the ListItem will have a "disabled" presentation.
```

packages/components/src/List/types.ts
```diff
@@ -49,3 +49,26 @@ export const itemDimensionKeys = [
   'labelLineHeight',
   'detailFontSize',
 ]
+
+export interface ItemBackgroundColorProps {
+  /**
+   * Replace the normal grey selected and selected + hovered color with brand colors
+   * @default false
+   */
+  brand?: boolean
```
lukelooker, line 58: It occurs to me that `brand` isn't used anywhere else but `keyColor` is supported by `Link` so it'd be more consistent to use that and create an association the the `theme.colors.key`

```suggestion
  /**
   * Replace the normal uiX color for selected and selected + hovered color with key colors
   * @default false
   */
  keyColor?: boolean
```

packages/components/src/List/utils/getItemBackgroundColor.ts
```diff
@@ -26,15 +26,15 @@
 
 import { itemSelectedColor, Theme } from '@looker/design-tokens'
 import { css } from 'styled-components'
-import { TreeBackgroundStyleProps } from '../types'
+import { ItemBackgroundColorProps } from '../types'
```
lukelooker, line 29: Admittedly, it's annoyingly long but I think for consistency's sake we should title this file and it's interface/functions `ListItemBackgroundColorProps`

packages/components/src/List/ListItem.story.tsx
```diff
@@ -96,6 +126,16 @@ export const Link = () => {
...
+const lorem =
+  'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.'
+export const Truncate = Template.bind({})
+Truncate.args = {
+  ...Basic.args,
+  children: lorem,
+  description: lorem,
+  truncate: true,
+}
+
```
lukelooker, line 138: Would be good to add an example of `truncate` + `detail` & `icons` to make sure layout plays nice in a worst-case scenario.

lukelooker (commented):
Let's chat tomorrow morning about grid challenges here - overall this is looking really great and I think I have an idea or two about how we can perhaps make managing the "accessory" implementation less painful.

packages/components/src/List/ListItem.story.tsx
ElliotTheHuman: Resolved in c886cb4!

packages/components/src/List/types.ts
ElliotTheHuman: Resolved in 4b6f7ad

packages/components/src/List/types.ts
ElliotTheHuman: Resolved in c6c8885

packages/components/src/List/types.ts
ElliotTheHuman: Resolved in c6c8885

packages/components/src/List/types.ts
ElliotTheHuman: Resolved in c6c8885

packages/components/src/List/utils/getItemBackgroundColor.ts
ElliotTheHuman: 👍  7b0516b

ElliotTheHuman: Resolved in 0a34bfa!

However, what do you think of the name `ListItemStateProps`? I know that set of props functionally exists to determine the background color of the `ListItem`, but they are reflective of the "state" of the `ListItem` (outside of `keyColor`).

packages/components/src/List/ListItemDetail.tsx
ElliotTheHuman: Accidentally minimized Luke's previous comment (thought it was a dupe). Added it back here:
```
I'd be tempted to make the solution narrower by having an accessory prop that just toggles the 
padding value as needed instead of making PaddingProps available any upstream ListItemDetail consumers.
```

packages/components/src/List/ListItem.tsx
ElliotTheHuman: Resolved in 9352c1e

packages/components/src/List/ListItem.tsx
```diff
@@ -125,60 +163,96 @@ const ListItemInternal: FC<ListItemProps> = (props) => {
...
+        color="text2"
+        fontSize={itemDimensions.descriptionFontSize}
+        lineHeight={itemDimensions.descriptionLineHeight}
+        truncate={truncate}
+      >
+        {description}
+      </Paragraph>
+    ) : (
+      description
+    )
```
ElliotTheHuman, line 194: Ended up going with `Paragraph` and conditionally toggling it's `truncate` prop (didn't realize `Paragraph` had a truncate prop in retrospect 🤦 ).

Considering we check the type of the child / description, it felt like wrapping it in a `Paragraph` was the simplest way to get the right dimensions / truncation behavior.

packages/components/src/List/ListItemLayout.tsx
```diff
@@ -24,133 +24,46 @@
...
+  })
+}
 
-    &:hover {
-      background: ${({ theme: { colors } }) => colors.background};
-      color: ${({ theme: { colors } }) => colors.text1};
-    }
-  }
+export const ListItemLayout = styled(ListItemLayoutInternal)`
+  ${(props) => listItemPadding({ accessory: false, ...props })}
```
ElliotTheHuman, line 68: I thought having the Layout components handle the padding  made the most readability sense (i.e. packaging the conditional padding with the different JSX layouts).

That being said, I could see an argument for moving this up to the `ListItemWrapper` component so that we're a bit more DRY.

ElliotTheHuman:
Summary of the layout changes!

- Added `ListItemWrapper` which is essentially the same as the old `ListItemLayout` (i.e. renders the `li` and handles (most of the) list styling 
- Added `ListItemLayout` and `ListItemLayoutAccessory` which represent the different layout options when accessory is off or on, respectively.
- `ListItem`'s primary purpose to distribute props to either `Layout` or `Wrapper` and also construct the "rendered" children (i.e. icon, children, description, detail)

lukelooker (changes requested):
Ok - a small pile of what I think is mostly minor nits.

If `codecov` complaints look tricky we can go ahead and defer those for later.

Super excited to enable `Tree`/`TreeItem` density support soon! :)

packages/components/src/List/types.ts
```diff
@@ -49,3 +53,32 @@ export const itemDimensionKeys = [
   'labelLineHeight',
   'detailFontSize',
 ]
+
+export interface ListItemBackgroundColorProps {
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 0a279378e!

packages/components/src/List/ListItem.tsx
```diff
@@ -125,60 +153,105 @@ const ListItemInternal: FC<ListItemProps> = (props) => {
...
+    <Layout
+      // eslint-disable-next-line react/no-children-prop
+      children={renderedChildren}
+      containerCreator={LabelContainerCreator}
+      description={renderedDescription}
+      detail={renderedDetail}
+      icon={renderedIcon}
+      px={itemDimensions.px}
+      py={itemDimensions.py}
+    />
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 0a279378e

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -53,12 +53,12 @@ import {
...
 import { Truncate } from '../Truncate'
+import { listItemBackgroundColor } from '../List/utils'
+import { ListItemBackgroundColorProps } from '../List/types'
 import { TreeContext } from './TreeContext'
-import { treeBackgroundColor } from './utils'
-import { TreeBackgroundStyleProps } from './types'
 
 export interface TreeItemProps
-  extends TreeBackgroundStyleProps,
+  extends ListItemBackgroundColorProps,
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 0a279378e!

packages/components/src/List/types.ts
```diff
@@ -49,3 +53,36 @@ export const itemDimensionKeys = [
...
+}
+
+export interface ListItemStatefulWithHoveredProps
+  extends ListItemStatefulProps {
+  /**
+   * If true, the ListItem will have a light background color
+   * @default false
+   */
+  hovered?: boolean
+}
```
ElliotTheHuman, line 88: I don't love this name, but can't come up with anything more concise.

I _could_ do something like `ListItemFullStateful` but figured being explicit was probably more important. This component _should_ only be used by things like `TreeStyle` and the `listItemBackground` util since nothing that devs should touch should have a "hovered" prop.

packages/components/src/Tree/Tree.tsx
```diff
@@ -94,16 +97,16 @@ const TreeLayout: FC<TreeProps> = ({
...
         detailAccessory: hasDetailAccessory,
         detailHoverDisclosure: hasDetailHoverDisclosure,
+        keyColor: hasKeyColoring,
       }}
     >
       <TreeStyle
         className={className}
         border={hasBorder}
-        brand={hasBrandColoring}
+        keyColor={hasKeyColoring}
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 8861783

packages/components/src/Tree/Tree.tsx
```diff
@@ -94,16 +97,16 @@ const TreeLayout: FC<TreeProps> = ({
     <TreeContext.Provider
       value={{
         border: hasBorder,
-        brand: hasBrandColoring,
         depth: depth + 1,
         detailAccessory: hasDetailAccessory,
         detailHoverDisclosure: hasDetailHoverDisclosure,
+        keyColor: hasKeyColoring,
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 8861783

packages/components/src/List/types.ts
```diff
@@ -49,3 +53,32 @@ export const itemDimensionKeys = [
...
+  /**
+   * If true, the ListItem will have a "disabled" presentation.
+   * @default false
+   */
+  disabled?: boolean
+  /**
+   * If true, the ListItem will have a light background color
+   * @default false
+   */
+  hovered?: boolean
```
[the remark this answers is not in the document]
ElliotTheHuman: Ended up creating another interface since `hovered` is necessary in >1 spots.

Don't love the name I came up with though. Added a comment covering the new interface below.

packages/components/src/List/ListItem.tsx
```diff
@@ -125,60 +153,105 @@ const ListItemInternal: FC<ListItemProps> = (props) => {
...
       </Paragraph>
     ) : (
       children
     )
 
-  const listItemContent = (
-    <Component
+  const renderedDescription =
+    typeof description === 'string' ? (
+      <Paragraph
```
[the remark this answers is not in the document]
ElliotTheHuman: Should be resolved in b1eb581

packages/components/src/List/ListItemWrapper.tsx
```diff
@@ -0,0 +1,151 @@
...
+    &:focus {
+      color: inherit;
+      position: relative;
+      text-decoration: none;
+    }
+  }
+
+  ${({ focusVisible, theme: { colors } }) =>
+    focusVisible &&
+    `&:focus-within button:after,
```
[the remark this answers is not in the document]
ElliotTheHuman: Left a note about an attempted test at this behavior. See comment below.

packages/components/src/List/ListItem.test.tsx
```diff
@@ -59,19 +70,33 @@ describe('ListItem', () => {
...
+    configure({ computedStyleSupportsPseudoElements: true })
+    const { getByRole } = renderWithTheme(<ListItem>Item</ListItem>)
+    const item = getByRole('listitem')
+    fireEvent.keyUp(item, {
+      key: 'Enter',
+    })
+    expect(window.getComputedStyle(item, ':after')).toHaveAttribute(
+      'border',
+      'solid 2px #9785F2'
+    )
```
ElliotTheHuman, line 87: Based on the [RTL docs](https://testing-library.com/docs/dom-testing-library/api-configuration/#computedstylesupportspseudoelements), the pseudo-selector parameter in getComputedStyle is not supported by JSDom. Afaik, I don't think there's another way to test the ":after" pseudo-element that creates the purple border when focusVisible === true to make codecov happy.

Could be wrong though (hopefully)! For now, leaving an xtest for future development felt like the right call.

lukelooker: Ah... yes, this is a case where we need to use `jest-image-snapshot` to drive a `puppeteer` instance, move focus and then take a picture. 

Suffice it to say a problem for another day :)

packages/components/src/List/types.ts
lukelooker: 👍  - Perhaps someday the perfect name will pop into my head but I can live with this as-is for now :)

lukelooker (approved):
Such `List`, much wow!

List

PR #1814, opened 2021-01-25
chore(ci): Enable yarn cache restoration on all CI workflows

lukelooker:
### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -26,13 +26,11 @@ jobs:
           ref: main
           fetch-depth: 0
 
-      - name: Install dependencies
-        run: |
-          yarn install
+      - uses: c-hive/gha-yarn-cache@v1
+      - run: yarn install
 
       - name: Build packages
-        run: |
-          yarn build
+        run: yarn build
 
       - name: Create canary release
         run: |
--- a/.github/workflows/release-pull-request.yml
+++ b/.github/workflows/release-pull-request.yml
@@ -24,16 +24,13 @@ jobs:
           fetch-depth: 0
 
       - name: Fetch all tags
-        run: |
-          git fetch --depth=1 origin +refs/tags/*:refs/tags/*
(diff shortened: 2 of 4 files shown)


.github/workflows/canary.yml
```diff
@@ -26,13 +26,11 @@ jobs:
           ref: main
           fetch-depth: 0
 
-      - name: Install dependencies
-        run: |
-          yarn install
```
ElliotTheHuman, line 31: Noob question, but does deleting the pipe character change anything here?

.github/workflows/canary.yml
```diff
@@ -26,13 +26,11 @@ jobs:
...
+      - uses: c-hive/gha-yarn-cache@v1
```
ElliotTheHuman, line 29: I imagine we don't need `- name: Install dependencies` here since the yarn cache restoration glides through that step.

Food for thought, but would something like `- name: Use yarn cache` make sense for consistency with the other steps?

ElliotTheHuman (approved):
LGTM

Added some wandering thoughts I had, but things look good as far as I can tell.

.github/workflows/canary.yml
lukelooker: So a `uses` like this will be basically silent in CI output which is the desire - developers don't really need to know this is happening as it's a CI-only flow to speed things up.

.github/workflows/canary.yml
lukelooker: Correct - the pipe allows for multi-line scripts but in this case it works fine to put it on a single line and removes some "noise" in the workflow config.

.github/workflows/canary.yml
ElliotTheHuman: Neat! Makes sense to keep this silent. Thanks for elaborating.

PR #1824, opened 2021-01-28
chore: Fixed test:image-snapshots-update command

ElliotTheHuman:
The `test:image-snapshots-update` command referenced `yarn image-snapshots` as opposed to the new name `yarn test:image-snapshots`. Should've caught that in previous PR; my mistake.


ElliotTheHuman:
CI:TEST

lukelooker (approved):

PR #1826, opened 2021-01-29
fix: DataTable browser sticky cells

laraharrow:
### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts

### DIFF
--- a/packages/components/src/DataTable/Table.tsx
+++ b/packages/components/src/DataTable/Table.tsx
@@ -171,12 +171,8 @@ const stickyColumns = css<TableProps>`
 `
 
 export const Table = styled(TableLayout)`
-  /* the * is set so the border-bottom displays correctly on Firefox on sticky items */
-  *border-collapse: collapse;
   border-spacing: 0;
   font-size: ${({ theme }) => theme.fontSizes.small};
-  /* stylelint-disable-next-line value-no-vendor-prefix */
-  height: -webkit-fill-available;
   line-height: 1;
   width: 100%;
   td,
--- a/packages/components/src/DataTable/Table.tsx
+++ b/packages/components/src/DataTable/Table.tsx
@@ -175,6 +175,9 @@ export const Table = styled(TableLayout)`
   *border-collapse: collapse;
   border-spacing: 0;
   font-size: ${({ theme }) => theme.fontSizes.small};
+  /* stylelint-disable value-no-vendor-prefix, property-no-vendor-prefix */
+  height: -webkit-fill-available;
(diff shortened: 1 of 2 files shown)


packages/components/src/DataTable/Table.tsx
```diff
@@ -171,12 +171,15 @@ const stickyColumns = css<TableProps>`
 `
 
 export const Table = styled(TableLayout)`
-  border-collapse: collapse;
+  /* the * is set so the border-bottom displays correctly on Firefox on sticky items */
+  *border-collapse: collapse;
```
lukelooker, line 175: This is a browser hack `*` causes Firefox to just ignore this property - in general I'm leery of browser hacks because they can be hard to maintain / work inconsistently. 

It _appears_ that just removing this property altogether solves the problem.

```suggestion
```

packages/components/src/DataTable/Table.tsx
```diff
@@ -171,12 +171,15 @@ const stickyColumns = css<TableProps>`
...
   border-spacing: 0;
   font-size: ${({ theme }) => theme.fontSizes.small};
+  /* stylelint-disable value-no-vendor-prefix, property-no-vendor-prefix */
+  height: -webkit-fill-available;
+  /* stylelint-enable */
```
lukelooker, line 180:
```suggestion
  /* stylelint-disable-next-line value-no-vendor-prefix */
  height: -webkit-fill-available;
```

lukelooker (changes requested):

lukelooker (commented):
I also noticed when running image-snapshots that two moved. Can you investigate why?

packages/components/src/DataTable/Table.tsx
laraharrow: I thought of that but I didn't know if that was doing something else that I wasn't aware. I'll fix it right away

packages/components/src/DataTable/Table.tsx
```diff
@@ -171,12 +171,11 @@ const stickyColumns = css<TableProps>`
 `
 
 export const Table = styled(TableLayout)`
-  border-collapse: collapse;
   border-spacing: 0;
+  /* display: flex; */
```
lukelooker: I'm pretty sure this means this PR does nothing - the line won't even be emitted to the browser.

laraharrow: oh yeah that was just a fail attempt to fix the safari issue

PR #1831, opened 2021-01-29
chore: Enable image-snapshots generation via Github Actions CI

lukelooker:
- Enables image-snapshots check for CI
- Adds Github Actions workflow to automatically open a PR with any updated snapshots
- Includes all snapshots rendered via Github Actions (so this PR can pass CI itself)

### Requirements

- [x] Documentation updated


ElliotTheHuman (dismissed):
Two things:
1. I imagine we won't see a super high number of snapshots showing changes after this PR merges in, but to be sure, this is just our cloudtop snapshots bumping against the first of the GH generated ones, correct?
2. Those generated image-snapshot PRs won't create noise in our auto-generated changelog right (since they're being not being merged directly onto main)?

Assuming the answer is "Yes" to both those questions, LGTM!

lukelooker:
> Two things:
> 
> 1. I imagine we won't see a super high number of snapshots showing changes after this PR merges in, but to be sure, this is just our local snapshots bumping against the first of the GH generated ones, correct?

Correct! From here-on out you should only expect to see snapshots that were altered as part of the scope-of-work within the PR.

> 2. Those generated image-snapshot PRs won't create noise in our auto-generated changelog right (since they're being not being merged directly onto main)?

You're correct. Also worth noting, that if by some chance one of those commits survived and made it to `main` because they use the `chore:` keyword they still won't be reflected in the CHANGELOG - only `feat` and `fix` tagged commits are denoted in the publish CHANGELOG.

dbchristopher (approved):
this all looks great to me!

lukelooker:
@dbchristopher, @mdodgelooker, @ElliotTheHuman - I put this through it's paces tonight and I think the image-snapshots is ready to merge. If I can get a ✅  from one of you I'll go ahead and do so but _not_ make it required yet until we've had a chance to watch it do it's thing for a day or two.

Let me know if you have any questions. Thanks - Luke

dbchristopher (approved):
✅

PR #1846, opened 2021-02-02
feat: Tree refactor with List pieces

ElliotTheHuman:
- Refactored `TreeItem` to utilize `ListItem` underneath.
- Also took the opportunity to refactor `TreeGroup` to make indentation selectors in `TreeStyle` more streamlined

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts


packages/components/src/List/ListItemContext.tsx
```diff
@@ -25,21 +25,16 @@
...
 
 import { createContext } from 'react'
-import { ListItemDimensions } from './types'
-import { density0 } from './utils'
+import { DensityRamp } from './types'
 
 export interface ListItemContextProps {
-  itemDimensions: ListItemDimensions
+  density: DensityRamp
   iconGutter: boolean
```
ElliotTheHuman, line 32: Because `ListItem` now accepts `density` directly, I thought it'd make sense to have the `ListItemContext` provide the density value instead of the `ItemDimensions` object (which was previously looked up in a parent `List` but now is looked up in each item).

Code readability-wise, I think this is cleaner (since the `ListItem` code reads as "If you gave me a density, use that density. Otherwise, use the context density"). However, we might not like the performance hit of each item doing its own dimension lookup.

Thoughts? Happy to retool `ListItemContext` to hold an itemDimensions object if that's the better route.

packages/components/src/Tree/Tree.tsx
```diff
@@ -25,68 +25,117 @@
...
+        ref={detailRef}
+      >
+        {content}
+      </ListItemInnerDetail>
+    ),
+    options: {
+      accessory,
+      hoverDisclosure,
+    },
+  }
```
ElliotTheHuman, line 118: All this dark magic is used to prevent the Accordion from opening when clicking a `Tree`'s detail element and `accessory === true`.

There's an argument to have this behavior baked into `List`, but I figure that this is specifically a `Tree` problem and so any dark magic belongs within `Tree`'s implementation.

packages/components/src/Tree/stories/Tree.story.tsx
```diff
@@ -30,10 +30,10 @@ import { Tree, TreeProps, TreeItem } from '..'
 
 export * from './BorderRadius.story'
 export * from './ColorfulTree.story'
+export * from './Density.story'
 export * from './DisabledAndSelected.story'
 export * from './FieldPicker.story'
 export * from './FileTree.story'
-export * from './Flat.story'
```
ElliotTheHuman, line 36: Will create a `List` story with a header since that seems more appropriate than "Flat" `Tree`.

packages/components/src/Tree/TreeHeading.tsx
```diff
@@ -25,63 +25,39 @@
...
-  ${TreeItemLabel}, ${TreeGroupLabel}, ${AccordionDisclosure} {
-    ${color}
-  }
+}) => {
+  return (
+    <Paragraph
+      className={className}
+      fontSize="xxsmall"
+      fontWeight="semiBold"
+      lineHeight="xsmall"
```
ElliotTheHuman, line 50: While it's possible to derive `fontSize` and `lineHeight` from the `Tree`'s density, the Explore team had specific desires about the sizing of the `Tree` group headers.

Keeping this set in stone for now made sense, or at least until someone requests for these to be scalable with density / customizable.

packages/components/src/Tree/TreeStyle.tsx
```diff
@@ -49,46 +52,51 @@ const dividersCSS = css`
...
-        border && generateTreeBorder(depth, theme)}
+      ${({ border, depth, indicatorSize, theme }) =>
+        border && generateTreeBorder(depth, indicatorSize, theme)}
     }
 
     & > ${AccordionDisclosureStyle} {
+      ${ListItem} {
+        font-weight: ${({ branchFontWeight, theme: { fontWeights } }) =>
+          branchFontWeight ? fontWeights.normal : fontWeights.semiBold};
+      }
```
ElliotTheHuman, line 70: `font-weight` was being clobbered by the browser's default button CSS, so making this selector more specific does the trick (Note: this may not be the ideal solution, happy to discuss better alternatives).

packages/components/src/Tree/TreeStyle.tsx
```diff
@@ -49,46 +52,51 @@ const dividersCSS = css`
...
-    border-width: 0;
-    height: 100%;
-
-    & > ${TreeItemLabel} {
-      background-color: transparent;
-      padding-left: ${({ theme }) => theme.space.none};
-    }
+  & > ${Accordion} > ${AccordionContent} > ${List} > ${TreeHeading},
+  & > ${Accordion} > ${AccordionContent} > ${List} > ${ListItem} > button,
+  & > ${Accordion} > ${AccordionContent} > ${List} > ${ListItem} > a {
```
ElliotTheHuman, line 89: These selectors (and the ones below) boil down to "Indent my direct child `TreeHeadings` and `ListItem`s

packages/components/src/Tree/TreeHeading.tsx
```diff
@@ -25,63 +25,36 @@
...
-      {truncate ? <Truncate>{label}</Truncate> : label}
-    </TreeGroupLabel>
-    {children}
-  </div>
-)
+}) => {
+  return (
+    <Paragraph className={className} truncate={truncate}>
+      {children}
+    </Paragraph>
```
lukelooker, line 48: Does this need to be a `li` underneath to keep DOM happy? (immediate descendants of `<ul>` / `<ol>` is usually a `li`)

lukelooker: Ok - poked around, appears to not be a problem.

packages/components/src/Tree/TreeItem.tsx
```diff
@@ -24,266 +24,51 @@
...
+      keyColor={keyColor}
+      onBlur={onBlur}
+      onClick={onClick}
+      onKeyDown={onKeyDown}
+      onKeyUp={onKeyUp}
+      selected={selected}
+      {...props}
+    >
+      {children}
+    </ListItem>
```
lukelooker, line 70: This file refactor is awesome 😃

packages/components/src/Tree/TreeStyle.tsx
```diff
@@ -58,37 +77,44 @@ export const TreeStyle = styled.div<TreeStyleProps>`
...
```
lukelooker, line 107: Nesting could help make this easier to read:


```suggestion
  & > ${Accordion} > ${AccordionContent} > ${List} {
    & > ${TreeHeading}, & > ${ListItem} > button, ${ListItem} > a {
```

packages/components/src/Tree/stories/BorderRadius.story.tsx
```diff
@@ -39,7 +44,8 @@ export const BorderRadiusOverride = () => (
 )
 
 const BorderRadiusOverrideTree = styled(Tree)`
-  ${AccordionDisclosureStyle}, ${TreeItemLabel} {
+  ${AccordionDisclosureStyle},
+  ${Accordion} > ${AccordionContent} > ${List} > ${ListItem} > button {
```
lukelooker, line 48: Would be nice to find a way to make the old CSS work so upstream changes aren't needed / provide a stable styling "contract"

packages/components/src/Tree/types.ts
```diff
@@ -24,53 +24,31 @@
...
-  detailAccessory?: boolean
-  /**
-   * Icon element that appears between the Tree indicator and the Tree label
-   */
-  icon?: IconNames
-  /**
-   * Text label of the Tree
-   * Note: This is a required prop
-   */
-  label: ReactNode
```
lukelooker, line 64: woohoo!

packages/components/src/Tree/types.ts
```diff
@@ -24,53 +24,31 @@
...
    */
   branchFontWeight?: boolean
   /**
-   * Prevent text wrapping on long labels and instead render truncated text
-   **/
-  truncate?: boolean
+   * Determines the sizing and spacing of child TreeItems
+   * @default 0
+   */
+  density?: DensityRamp
```
lukelooker, line 51: I wonder if we should just add this to `ListItem` since I could see cases where folks want / need to specify explicitly instead of at the `List` / `Tree` level?

lukelooker (commented):
Some minor questions and comments but overall this looks really great!

packages/components/src/Tree/TreeStyle.tsx
lukelooker: Honestly I'm not sure my refactor here really helps - seems more complex if anything. AKA: Ignore me.

packages/components/src/Tree/TreeHeading.tsx
ElliotTheHuman: Good call. Technically speaking, this should probably be an `li` based on MDN:
Screen Shot 2021-02-04 at 11 35 17 AM

I read on an SO post that browsers can handle it, but it's pretty simple to get this wrapped in an `li` (Note: it does slightly change the selectors in `TreeStyle` but only by a smidge).

ElliotTheHuman: Resolved in d581d9b

packages/components/src/List/ListItem.tsx
```diff
@@ -90,7 +88,9 @@ const ListItemInternal: FC<ListItemProps> = (props) => {
   const {
     children,
     className,
+    color,
```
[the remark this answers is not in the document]
ElliotTheHuman: In this case, the default is just for `icon` (i.e. if given a color, use that color, otherwise, use "text1" for the icon). It does make sense though to not have the color logic sit in the prop where it's hard to parse out. I tried to make things a little more readable (as well as make the color prop work on `Tree`s in [this commit](d581d9b)).

Let me know what you think / how I can make this better!

packages/components/src/List/ListItem.tsx
```diff
@@ -133,7 +143,7 @@ const ListItemInternal: FC<ListItemProps> = (props) => {
     icon || iconArtwork ? (
       <Icon
         artwork={iconArtwork}
-        color="text1"
+        color={color || 'text1'}
```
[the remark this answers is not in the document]
ElliotTheHuman: See comment above for additional thoughts

packages/components/src/Tree/types.ts
ElliotTheHuman: We did actually! We needed a way to pass the density value to the parent-most `Tree`'s inner item, and adding the prop onto `ListItem` made the most sense (but it's a @private so hopefully shouldn't be out in the wild too much).

packages/components/src/Tree/stories/BorderRadius.story.tsx
ElliotTheHuman: That would be ideal... would making it a prop be a viable option? Then it could just be on us to manage the border-radius'ing.

Alternatively, we _could_ try exporting the entire selector 😝  Haven't played around with it much, but in that world, users would just have to worry about importing the "direct child item selector + disclosure selector" and not have to maintain it in anyway.

ElliotTheHuman:
Note: did some reworking of how colors are distributed from Tree and TreeItem / ListItem.

Some of the logic gets hairy, but it's (mostly) all managed via props and not within the styled selectors which seems a little easier to follow. Default color "text5" of `Tree` structures is still set by a selector though.

packages/components/src/Tree/TreeHeading.tsx
```diff
@@ -26,11 +26,15 @@
...
 import { Paragraph } from '../Text'
 
-export interface TreeHeadingProps extends TextColorProps {
+export interface TreeHeadingProps {
+  children: string
   className?: string
+  /**
+   * Determines the text color
+   */
+  color?: string
```
lukelooker, line 37: Probably ideal to use `TextColorProps`

packages/components/src/Tree/stories/BorderRadius.story.tsx
lukelooker: `ListItemLabel`

packages/components/src/List/ListItem.tsx
```diff
@@ -86,11 +84,26 @@ export interface ListItemProps
...
+  itemRole?: 'link' | 'button'
+}
+
+export const ListItemLabel = styled.div
+  .withConfig<ListItemLabelProps>({
+    shouldForwardProp: (prop) => prop !== 'itemRole',
+  })
+  .attrs<ListItemLabelProps>(({ disabled, itemRole }) => ({
+    as: !disabled && itemRole === 'link' ? 'a' : 'button',
+  }))<ListItemLabelProps>``
```
ElliotTheHuman, line 98: This _should_ keep `itemRole` from getting down to the actual `div` element, correct? I'm not seeing any errors in the console, but wanted to make sure I was using "shouldForwardProp" properly here.

packages/components/src/Tree/stories/BorderRadius.story.tsx
ElliotTheHuman: Resolved in 8e727fb

packages/components/src/Tree/TreeHeading.tsx
ElliotTheHuman: As discussed, we cut `TreeHeading` down to its bare bones (i.e. so that it's only role is to get proper indentation from parent `Tree` and contain children). Now known as `TreeBranch`.

Resolved in a8838ce

packages/components/src/Truncate/Truncate.tsx
```diff
@@ -25,19 +25,26 @@
...
   children: ReactNode
+  className?: string
 }
 
 /**
  * Prevent text wrapping on long labels and instead render truncated text.
  * Renders a tooltip to view the entire text content on hover.
  **/
-export const Truncate: FC<TruncateProps> = ({ children, width = '100%' }) => {
+const TruncateLayout: FC<TruncateProps> = ({
```
ElliotTheHuman, line 43: I needed to be able to change line height on the `Truncate` container in `ListItem` since default button browser styling was clobbering / increasing `ListItem` height by 1px. 

However, couldn't change it directly on `Truncate`'s CSS because that would change the layout on existing `Truncate` use cases (i.e. `FieldCheckboxGroup`).

Decided that exporting `Truncate` as a styled component made sense, but let me know if that's something we totally do not want and I can rethink how we get the proper line-height for `ListItem` here.

packages/components/src/List/utils/listItemPadding.ts
```diff
@@ -42,8 +41,8 @@ export const listItemPadding = ({
    * Without it, density = -1 ListItems would be at 44px.
    */
   const pt = propsPy === '0.375rem' ? propsPy : space[propsPy]
-  const pr = accessory ? '0' : space[propsPx]
   const pb = pt
+  const pr = space[propsPx]
```
ElliotTheHuman, line 45: We want that inner button / a to always have right padding (afaik). If the button content / label stretches to the edge in the `accessory` case, it would make sense for it to have some right padding rather than reach the edge of the button.

ElliotTheHuman:
Changes since last time:
- `Truncate` is now exported as a styled component for styling purposes within `ListItem`
- Added `ListItemLabel` styled component for easy border radius targeting
- Fixed some minor CSS issues (i.e. border on `TreeArtifical` being weird, `Truncate` + `accessory` not playing nicely because flex)
- `listItemPadding` always has right padding since  properly aligns unhovered, hoverDisclosure, accessory items

packages/components/src/List/ListItem.tsx
lukelooker: Yep - that should swat it away :)

packages/components/src/List/utils/listItemPadding.ts
ElliotTheHuman: Previous implementation:
`const pr = accessory ? '0' : space[propsPx]`

packages/components/src/List/ListItemLayoutAccessory.tsx
```diff
@@ -26,31 +26,31 @@
...
-import { FlexItem } from '../Layout'
+import { Flex } from '../Layout'
 import { ListItemLayoutProps } from './ListItemLayout'
 import { listItemPadding } from './utils'
 
 export const ListItemLayoutAccessoryInternal: FC<ListItemLayoutProps> = ({
   children,
   className,
-  containerCreator,
+  labelCreator,
```
lukelooker, line 36: What's the reason to do this as a function instead of as a component?

packages/components/src/List/ListItemLayoutAccessory.tsx
```diff
@@ -26,31 +26,31 @@
...
         {description}
-      </FlexItem>
+      </Flex>
     </>
   )
 
   return (
     <>
-      {containerCreator({
+      {labelCreator({
```
lukelooker, line 53: Same as above - but why not do this as `<LabelCreator className={className}>content</LabelCreator>`?

packages/components/src/Truncate/Truncate.tsx
```diff
@@ -58,10 +65,17 @@ export const Truncate: FC<TruncateProps> = ({ children, width = '100%' }) => {
     width: 'auto',
   })
 
+  const className = join([domProps.className, propsClassName], ' ')
```
lukelooker, line 68: I'd just use built-in join here. 

```suggestion
  const { className, otherDomProps } = domProps
  const className = [className, propsClassName].join(' ')
```

packages/components/src/Truncate/Truncate.tsx
```diff
@@ -58,10 +65,17 @@ export const Truncate: FC<TruncateProps> = ({ children, width = '100%' }) => {
...
+
   return (
     <>
       {tooltip}
-      <TextStyle {...domProps} ref={textRef} width={width}>
+      <TextStyle
+        {...omit(domProps, 'className')}
```
lukelooker, line 74:
```suggestion
        {...otherDomProps}
```

packages/components/src/Truncate/Truncate.tsx
```diff
@@ -25,19 +25,26 @@
  */
 import React, { FC, useState, useCallback, ReactNode } from 'react'
 import styled from 'styled-components'
+import join from 'lodash/join'
+import omit from 'lodash/omit'
```
lukelooker, line 29: Down with `lodash` (see below)

packages/components/src/Truncate/Truncate.tsx
lukelooker: I'm cool with exporting this but we should log a ticket to eventually fix `Truncate` to have `line-height: 1` (fixing the locations that depend on that) because it really does seem more "correct" that way and then we could unwind `Tree`'s need to reach-in as it were.

lukelooker (approved):

packages/components/src/List/ListItemLayoutAccessory.tsx
```diff
@@ -26,31 +26,31 @@
...
     <>
       {icon}
-      <FlexItem>
+      <Flex flexDirection="column" minWidth={0}>
         {children}
         {description}
-      </FlexItem>
+      </Flex>
     </>
   )
```
lukelooker, line 50: const LabelCreator = labelCreator

packages/components/src/List/ListItemLayoutAccessory.tsx
ElliotTheHuman: Discussed over meet, but because `LabelCreator` needs to wrap around the content in the `XLayout` components, we would need to pass in the component via a prop, then set it to a variable with a capital name so that we can create JSX elements off of it.

IIRC, the plan is to keep things as they are for now, but might be worth revisiting this assuming there's no performance loss from passing in a component vs. a vanilla function.

ElliotTheHuman:
Excuse the latest commit name 🤦  I set the density to -3 by default, the commit was just named incorrectly by yours truly.

packages/components/src/List/ListItem.tsx
```diff
@@ -86,11 +97,26 @@ export interface ListItemProps
...
+
 const ListItemInternal: FC<ListItemProps> = (props) => {
   const {
     children,
     className,
+    color,
     current,
+    density: propsDensity,
```
lukelooker, line 119:
```suggestion
    density: propsDensity = -3,
```

lukelooker: Can you switch the default density back down to match the previous setup so we can compare screenshots?

lukelooker (changes requested):
Push density down so we can do one last careful comparison of snapshot changes :)

packages/components/src/List/ListItem.tsx
ElliotTheHuman: Ah I set the default density back to `-3` on `Tree` in particular since those stories were at that small "Developer" density.

`ListItem` and `List` default to `0` on main, but I can bump those down to `0` as well if you'd like! Just wanted to make sure.

lukelooker: Oh - I'm a dork and misread here. Ignore me!

lukelooker (approved):
Looking at `File Tree Closed` is appears the icon size and spacing changed a little bit:

image

image

That said the Field Picker appears to be rock-solid so I think we're good to go!

PR #1850, opened 2021-02-02
fix: Radio validation error

laraharrow:
### Requirements

- [x] Documentation updated
- [x] i18n impacts
- [x] a11y impacts


packages/components/src/Form/Inputs/Radio/Radio.tsx
```diff
@@ -28,21 +28,29 @@ import React, { forwardRef, Ref } from 'react'
...
+    const { className, validationType, ...restProps } = props
 
     return (
       <div className={className}>
-        <input type="radio" {...pickInputProps(restProps)} ref={ref} />
+        <input
+          type="radio"
+          aria-invalid={validationType === 'error' ? 'true' : undefined}
+          {...pickInputProps(restProps)}
+          ref={ref}
```
lukelooker, line 52: Since we don't want the spread `...` to overwrite any of the key settings on the radio input it should be done before those attributes get assigned explicitly.

```suggestion
          {...pickInputProps(restProps)}
          type="radio"
          aria-invalid={validationType === 'error' ? 'true' : undefined}
          ref={ref}
```

packages/components/src/Form/Inputs/Radio/Radio.tsx
```diff
@@ -69,13 +77,25 @@ export const Radio = styled(RadioLayout)`
...
 
+  input + ${FauxRadio} {
+    ${inputTextValidation}
+  }
+
   input:checked + ${FauxRadio} {
-    color: ${({ theme }) => theme.colors.key};
+    border-color: ${({ theme, validationType }) =>
+      validationType === 'error'
+        ? theme.colors.criticalBorder
```
lukelooker, line 87: Codedov notes for this line - basically a specific test is needed that puts `Radio` into an error state and then verifies that the `border-color` is set as-expected.

packages/components/src/Form/Inputs/Radio/Radio.tsx
```diff
@@ -69,13 +77,25 @@ export const Radio = styled(RadioLayout)`
...
+    color: ${({ readOnly, theme }) =>
+      readOnly ? 'transparent' : theme.colors.key};
   }
 
   input:not(:checked) + ${FauxRadio} {
     background: ${({ theme }) => theme.colors.field};
-    border-color: ${({ theme }) => theme.colors.ui2};
+    border-color: ${({ theme, validationType }) =>
+      validationType === 'error'
+        ? theme.colors.criticalBorder
```
lukelooker, line 97: Same as above here.

packages/components/src/Form/Inputs/Radio/Radio.tsx
```diff
@@ -69,13 +77,25 @@ export const Radio = styled(RadioLayout)`
     z-index: 1;
   }
 
+  input + ${FauxRadio} {
+    ${inputTextValidation}
+  }
+
   input:checked + ${FauxRadio} {
-    color: ${({ theme }) => theme.colors.key};
+    border-color: ${({ theme, validationType }) =>
```
lukelooker, line 85: I believe `border-color: ...` can be specified on the core selector `input + ${FauxRadio} {` just once and then not repeated for all the different states.

packages/components/src/Form/Inputs/Radio/Radio.tsx
```diff
@@ -69,13 +77,25 @@ export const Radio = styled(RadioLayout)`
...
+      readOnly ? 'transparent' : theme.colors.key};
   }
 
   input:not(:checked) + ${FauxRadio} {
     background: ${({ theme }) => theme.colors.field};
-    border-color: ${({ theme }) => theme.colors.ui2};
+    border-color: ${({ theme, validationType }) =>
+      validationType === 'error'
+        ? theme.colors.criticalBorder
+        : theme.colors.ui2};
```
lukelooker, line 98: I believe this can be removed and replaced with a specification on the default (non-pseudo state) as noted above

packages/components/src/Form/Inputs/Radio/Radio.tsx
```diff
@@ -85,11 +105,14 @@ export const Radio = styled(RadioLayout)`
   }
 
   input:disabled + ${FauxRadio} {
+    border-color: ${({ theme }) => theme.colors.ui2};
```
lukelooker, line 108: See note above.

```suggestion
```

packages/components/src/Form/Inputs/Radio/Radio.tsx
```diff
@@ -85,11 +105,14 @@ export const Radio = styled(RadioLayout)`
...
     color: ${({ theme }) => theme.colors.text1};
   }
 
   input:disabled:not(:checked) + ${FauxRadio} {
     background: ${({ theme }) => theme.colors.ui1};
+    border-color: ${({ theme }) => theme.colors.ui2};
```
lukelooker, line 114: See note above.

```suggestion
```

lukelooker (changes requested):
New appearance looks good.

I think the CSS can be simplified a bit (specifically `border-color` assignment)

Since we're making changes here in general this would be a good chance to remove Jest snapshot tests. Instead we should be using jest to verify behavior – [Checkbox.test.tsx](https://github.com/looker-open-source/components/blob/main/packages/components/src/Form/Inputs/Checkbox/Checkbox.test.tsx) is a really good template for what the `Radio.test.tsx` should look more like (basically testing every prop and in so doing exercising all the code within the component).

Specifically, the complaints from codecov can be seen at  [code coverage report for Radio.txt](https://codecov.io/gh/looker-open-source/components/pull/1850/diff?src=pr&el=tree#diff-cGFja2FnZXMvY29tcG9uZW50cy9zcmMvRm9ybS9JbnB1dHMvUmFkaW8vUmFkaW8udHN4) – basically what it's complaining about is that coverage % has gone down - this is because there are new, un-tested lines (the lines red-blocks next to them - 87 & 97)

image

lukelooker:
Also, the Ci has opened up a pull request against this pull request with snapshot updates: https://github.com/looker-open-source/components/pull/1851

packages/components/src/Form/Inputs/Radio/Radio.tsx
laraharrow: That was my first solution too. But some how it didn't quite work for radio. So I ended up finding this way around to get the error color to show up.

packages/components/src/Form/Inputs/Radio/Radio.tsx
laraharrow: I can't really test the `border-color` given that it is not available on tests. instead I tried testing `area-invalid` that also gets triggered by the `validationType`

packages/components/src/Form/Inputs/Radio/Radio.test.tsx
```diff
@@ -26,32 +26,70 @@
...
+
+  expect(RadioInput as HTMLInputElement).toHaveAttribute(
+    'aria-describedby',
+    'some-id'
+  )
+})
+
+test('Renders Radio with error', () => {
+  const { getByRole } = renderWithTheme(<Radio validationType="error" />)
+  const RadioInput = getByRole('radio')
```
lukelooker, line 81: Typically `PascalCase` is reserved for component and class names

```suggestion
  const radioInput = getByRole('radio')
```

packages/components/src/Form/Inputs/Radio/Radio.test.tsx
```diff
@@ -26,32 +26,70 @@
...
+
+  expect(RadioInput as HTMLInputElement).toHaveAttribute('aria-invalid', 'true')
+})
+
+test('Renders Radio checked with error', () => {
+  const { getByRole } = renderWithTheme(
+    <Radio defaultChecked validationType="error" />
+  )
+  const RadioInput = getByRole('radio')
```
lukelooker, line 90:
```suggestion
  const radioInput = getByRole('radio')
```

packages/components/src/Form/Fields/FieldRadio/FieldRadio.test.tsx
```diff
@@ -30,10 +30,46 @@ import { renderWithTheme } from '@looker/components-test-utils'
...
+  expect(getByText('0/50')).toBeInTheDocument()
+})
+
+test('FieldRadio renders disabled', () => {
+  const { getByRole } = renderWithTheme(
     <FieldRadio disabled id="FieldRadioID" label="FM" name="thumbsUp" />
   )
+  const RadioInput = getByRole('radio')
+
+  expect(RadioInput as HTMLInputElement).toBeDisabled()
```
lukelooker, line 59: The `as HTMLInputElement` shouldn't be needed (and doesn't appear to be). In general the test can be made more terse like so:

```suggestion
  renderWithTheme(
    <FieldRadio disabled id="FieldRadioID" label="FM" name="thumbsUp" />
  )

  expect(screen.getByRole('radio')).toBeDisabled()
```

lukelooker: We've also tried to switch to using `screen.X` instead of destrucuring helpers from RTL's `render` method.

packages/components/src/Form/Inputs/Radio/Radio.test.tsx
```diff
@@ -26,32 +26,70 @@
...
+  jest.resetAllMocks()
 })
 
-test('Radio checked set to true', () => {
-  assertSnapshot(<Radio checked={true} id="radioID" />)
+test('Renders Radio', () => {
+  const { getByRole } = renderWithTheme(<Radio defaultChecked />)
+  const RadioInput = getByRole('radio')
+
+  expect(RadioInput as HTMLInputElement).toBeInTheDocument()
```
lukelooker, line 41: All of the tests here-in can be made a lot more terse :)

```suggestion
  renderWithTheme(<Radio defaultChecked />)
  expect(screen.getByRole('radio')).toBeInTheDocument()
```

Also would be good to wrap all of the tests inside a `describe('Radio', () => { ...` to keep them all grouped together in the output.

packages/components/src/Form/Fields/FieldRadio/FieldRadio.test.tsx
```diff
@@ -30,10 +30,46 @@ import { renderWithTheme } from '@looker/components-test-utils'
...
+      label="FM"
+      name="thumbsUp"
+    />
+  )
+  expect(getByText('describe something here.')).toBeInTheDocument()
+  expect(getByText('0/50')).toBeInTheDocument()
+})
+
+test('FieldRadio renders disabled', () => {
+  const { getByRole } = renderWithTheme(
```
lukelooker, line 54: All of the `FieldRadio` tests should be wrapped in a top-level

```describe('FieldRadio', () => {``` 

To group all of the tests together when output

lukelooker (commented):
Really minor feedback on code style patterns within tests but functionally this looks great.

lukelooker (approved):
LGTM

PR #1865, opened 2021-02-04
fix: DataTable Simplify onClick behavior

laraharrow:

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts

### DIFF
--- a/packages/components/src/DataTable/Item/DataTableItem.tsx
+++ b/packages/components/src/DataTable/Item/DataTableItem.tsx
@@ -74,13 +74,7 @@ const DataTableItemLayout: FC<DataTableItemProps> = ({
   const ref = useRef<HTMLTableRowElement>(null)
   const { select } = useContext(DataTableContext)
 
-  const handleOnSelect = () => select && select.onSelect(id)
-
-  const handleClick = disabled
-    ? undefined
-    : select
-    ? handleOnSelect
-    : onClick || undefined
+  const handleClick = disabled ? undefined : onClick || undefined
 
   const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
     const isEventFromChild = event.currentTarget !== event.target
@@ -110,7 +104,6 @@ const DataTableItemLayout: FC<DataTableItemProps> = ({
   const onChange = select ? () => select.onSelect(id) : undefined
 
   const checked = select && select.selectedItems.includes(id)
-
   return (
     <DataTableRow
       checked={checked}
--- a/packages/components/src/DataTable/Item/DataTableItem.tsx
+++ b/packages/components/src/DataTable/Item/DataTableItem.tsx
@@ -78,7 +78,7 @@ const DataTableItemLayout: FC<DataTableItemProps> = ({
 
   const handleClick = disabled
     ? undefined
-    : select && select.onClickRowSelect
+    : select
     ? handleOnSelect
     : onClick || undefined
 
--- a/packages/components/src/DataTable/Item/DataTableRow.tsx
+++ b/packages/components/src/DataTable/Item/DataTableRow.tsx
@@ -35,7 +35,7 @@ import React, {
   Ref,
   useContext,
 } from 'react'
-import styled from 'styled-components'
+import styled, { DefaultTheme } from 'styled-components'
 import { DataTableContext } from '../DataTableContext'
 import {
   DataTableCheckbox,
@@ -119,6 +119,18 @@ const DataTableRowLayout = forwardRef(
 
 DataTableRowLayout.displayName = 'DataTableRowLayout'
 
+const getRowHoverColor = (
+  theme: DefaultTheme,
+  hasOnClick: boolean,
+  isHeader: boolean | undefined,
+  isSelected: boolean | 'mixed' | undefined
+) => {
+  if (!isHeader || !hasOnClick) return undefined
+  if (isSelected && hasOnClick) return theme.colors.keyAccent
+  if (hasOnClick) return theme.colors.ui1
+  return undefined
+}
+
 export const DataTableRow = styled(DataTableRowLayout)`
   td,
   th {
@@ -139,12 +151,8 @@ export const DataTableRow = styled(DataTableRowLayout)`
 
     td,
     th {
-      background: ${({ checked, isHeaderRow, theme: { colors } }) =>
-        checked && !isHeaderRow
-          ? colors.keyAccent
-          : isHeaderRow
-          ? undefined
-          : colors.ui1};
+      background: ${({ checked, isHeaderRow, onClick, theme }) =>
+        getRowHoverColor(theme, onClick !== undefined, isHeaderRow, checked)};
     }
   }
 
--- a/packages/components/src/DataTable/Item/DataTableRow.tsx
+++ b/packages/components/src/DataTable/Item/DataTableRow.tsx
@@ -87,10 +87,12 @@ const DataTableRowLayout = forwardRef(
       }
     })
 
-    const handleOnClick = (event: React.MouseEvent<HTMLElement>) =>
-      event.target instanceof HTMLAnchorElement
+    const handleOnClick = (event: React.MouseEvent<HTMLElement>) => {
+      return event.target instanceof HTMLAnchorElement ||
+        event.target.nodeName === 'INPUT'
         ? undefined
         : onClick && onClick(event)
+    }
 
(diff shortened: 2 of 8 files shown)


laraharrow:
[ ] DataTableItem should only have a “hover” behavior (color or cursor change) when on onClick callback handler is specified
- line 136 of DataTableRow.tsx
```
  &:hover {
    cursor: ${({ onClick }) => onClick && 'pointer'};
    outline: none;
```
    
[x] Remove support for onClickRowSelect - developers can use onClick to emulate this behavior themselves and simplify usage / implementation

[ ] Clicking select Checkbox within DataTableRow should not trigger the onClick assigned to the full row (it should only toggle the checkbox)
- where can I find this behavior?

lukelooker:
> [ ] DataTableItem should only have a “hover” behavior (color or cursor change) when on onClick callback handler is specified
> 
> * line 136 of DataTableRow.tsx
> 
> ```
>   &:hover {
>     cursor: ${({ onClick }) => onClick && 'pointer'};
>     outline: none;
> ```

The row also gets a grey background color on hover - that shouldn't happen unless `onClick` is assigned

> [ ] Clicking select Checkbox within DataTableRow should not trigger the onClick assigned to the full row (it should only toggle the checkbox)
> 
> * where can I find this behavior?

Here's an example of the DataTableRow w/ `onClick`: https://codesandbox.io/s/datatable-example-ypeh5?file=/src/index.tsx - the "alert" shouldn't happen when you click on the checkbox, only on the remainder of the row.

lukelooker (changes requested):
Added a comment on specific questions you asked. 

Also noticed that Jest and Lint CI checks have failures.

packages/components/src/DataTable/stories/items.tsx
```diff
@@ -74,6 +74,8 @@ export const itemBuilder = (
         disabled={disabled}
         id={id}
         key={id}
+        // will be removed jsut for testing reasons
+        onClick={() => alert("This shouldn't happen when clicking checkbox")}
```
laraharrow, line 78: so adding this here, my assumption is that we would repeat the bug behavior as in: https://codesandbox.io/s/datatable-example-ypeh5?file=/src/index.tsx:985-1003
But actually the behavior on storybook is:
* on Basic: clicking anywhere on the row triggers the `onClick`
* on (any) Select: clicking anywhere on the row just triggers the checkbox but not the `onClick`

lukelooker (changes requested):
Tried this out and found the following issues:

- [x] Only the `checkbox` should trigger selection. Clicking else anywhere in the row should not toggle the checkbox value
- [ ] Rows with `onClick` assigned are now missing `:hover` effect (the row should darken when the user hovers and the `DataTableItem` has an `onClick` callback assigned)

These changes are done: (✅ )

- [x] The `checkbox` is not trigger `DataTableItem` onClick

packages/components/src/DataTable/Item/DataTableItem.tsx
```diff
@@ -74,13 +74,7 @@ const DataTableItemLayout: FC<DataTableItemProps> = ({
...
   const { select } = useContext(DataTableContext)
 
-  const handleOnSelect = () => select && select.onSelect(id)
-
-  const handleClick = disabled
-    ? undefined
-    : select && select.onClickRowSelect
-    ? handleOnSelect
-    : onClick || undefined
+  const handleClick = disabled ? undefined : onClick || undefined
```
laraharrow, line 77: this removes the checkbox selecting when clicking anywhere in the row unless it is specific to the checkbox itself.

packages/components/src/DataTable/Item/DataTableRow.tsx
```diff
@@ -87,10 +87,12 @@ const DataTableRowLayout = forwardRef(
       }
     })
 
-    const handleOnClick = (event: React.MouseEvent<HTMLElement>) =>
-      event.target instanceof HTMLAnchorElement
+    const handleOnClick = (event: React.MouseEvent<HTMLElement>) => {
+      return event.target instanceof HTMLAnchorElement ||
+        event.target.nodeName === 'INPUT'
```
laraharrow, line 92: Not sure if this is the correct solution. That adds `nodeName` to a file that I'm not familiar with `lib.dom.d.ts`. Basically we need to confirm that event.target is not a checkbox this will prevent the `onClick` to propagate iself to the checkbox

packages/components/src/DataTable/Item/DataTableRow.tsx
```diff
@@ -119,6 +121,18 @@ const DataTableRowLayout = forwardRef(
...
+  hasOnClick: boolean,
+  isHeader: boolean | undefined,
+  isSelected: boolean | 'mixed' | undefined
+) => {
+  if (!isHeader || !hasOnClick) return undefined
+  if (isSelected && hasOnClick) return theme.colors.keyAccent
+  if (hasOnClick) return theme.colors.ui1
+  return undefined
+}
+
```
laraharrow, line 135: somehow the hasOnClick is returning undefined even when the row has the prop. I compared with line 151 that also uses onClick in that instance the behavior is as expected.

packages/components/src/DataTable/Item/DataTableRow.tsx
laraharrow: this is causing lint errors

packages/components/src/DataTable/Item/DataTableRow.tsx
```diff
@@ -119,6 +121,18 @@ const DataTableRowLayout = forwardRef(
 
 DataTableRowLayout.displayName = 'DataTableRowLayout'
 
+const getRowHoverColor = (
+  theme: DefaultTheme,
```
lukelooker, line 125: Use `Theme` from `@looker/design-tokens`

```suggestion
  theme: Theme,
```

packages/components/src/DataTable/Item/DataTableRow.tsx
```diff
@@ -119,6 +121,18 @@ const DataTableRowLayout = forwardRef(
...
+  hasOnClick: boolean,
+  isHeader: boolean | undefined,
+  isSelected: boolean | 'mixed' | undefined
+) => {
+  if (!isHeader || !hasOnClick) return undefined
```
lukelooker, line 130: Logic for `isHeader` here is backwards :) 


```suggestion
  if (isHeader || !hasOnClick) return undefined
```

packages/components/src/DataTable/Item/DataTableRow.tsx
```diff
@@ -139,12 +153,8 @@ export const DataTableRow = styled(DataTableRowLayout)`
...
     td,
     th {
-      background: ${({ checked, isHeaderRow, theme: { colors } }) =>
-        checked && !isHeaderRow
-          ? colors.keyAccent
-          : isHeaderRow
-          ? undefined
-          : colors.ui1};
+      background: ${({ checked, isHeaderRow, onClick, theme }) =>
+        getRowHoverColor(theme, onClick !== undefined, isHeaderRow, checked)};
```
lukelooker, line 157: You can tighten up the function parameters by casting everything exactly here:

```suggestion
        getRowHoverColor(theme, Boolean(onClick), Boolean(isHeaderRow), Boolean(checked))};
```

packages/components/src/DataTable/Item/DataTableRow.tsx
```diff
@@ -119,6 +121,18 @@ const DataTableRowLayout = forwardRef(
 
 DataTableRowLayout.displayName = 'DataTableRowLayout'
 
+const getRowHoverColor = (
+  theme: DefaultTheme,
+  hasOnClick: boolean,
+  isHeader: boolean | undefined,
+  isSelected: boolean | 'mixed' | undefined
```
lukelooker, line 128: See note below about tightening up inputs, it's a lot easier to test when a functions parameters are all required:

```suggestion
  isHeader: boolean,
  isSelected: boolean
```

lukelooker (changes requested):
Added some notes about the challenging spots.

packages/components/src/DataTable/stories/items.tsx
lukelooker, line 78: This should get removed now :)

lukelooker (changes requested):
Looks great - just need to remove the `alert(...` added for testing and then it'll be GTG! :)

lukelooker (approved):
LGTM

PR #1433, opened 2020-09-15
InputSearch options

mdodgelooker:
### :sparkles: Changes

- Use `Combobox` in `InputSearch` to support `options`
- Refactor `AdvancedInputControls` to support `InputSearch`
- Add `isLoading` and options `detail` on `Select`/`SelectMulti`/`InputSearch`

### :white_check_mark: Requirements

- [x] Includes test coverage for all changes
- [x] Build and tests are passing
- [x] Update documentation
- [x] Updated CHANGELOG
- [ ] Checked for i18n impacts
- [ ] Checked for a11y impacts
- [ ] PR is ideally < ~400LOC

PR #1873, opened 2021-02-05
fix(components): Unable to filter SelectMulti with 1 or more controlled values

mdodgelooker:
Issue repros: 
1. Unnecessary `SELECT_SILENT` due to filtering: https://codesandbox.io/s/wispy-wildflower-p2kn3
2. Legit `SELECT_SILENT` due to external state change: https://codesandbox.io/s/zealous-jepsen-jjiyh

This issue (2 issues really) was exposed by this fix https://github.com/looker-open-source/components/pull/1433/files#r491123741 for a different issue, but there were actually 2 separate underlying causes:
1. The `SELECT_SILENT` action exists for controlled use cases in order to sync up the `values` prop with the values in context, if they are found to me out of sync (for example the values are updated externally via an API call response). But the logic that mapped `SelectMulti` values (strings) to `ComboboxMulti` values (objects) was incorrectly setting the `label` to an empty string when there were no matching options (see `getComboboxText`). So in the 1st example above, the context has `[{ label: 'Cheddar', value: 'Cheddar' }]` as the current values, but once Cheddar gets filtered out, the `values` prop is coming in as `[{ label: '', value: 'Cheddar' }]`. The mismatch (now "better" detected after the change in #1433 above) was triggering `SELECT_SILENT`, which then was resetting the `inputValue` to `''`.
2. Event for a legit `SELECT_SILENT`, there's no need to reset the `inputValue`. I don't think I did that intentionally, just copied from `SELECT_WITH_CLICK` without too much thought.

So this issue should be double fixed now. 🙃

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts

### DIFF
--- a/packages/components/src/Form/Inputs/Combobox/utils/getComboboxText.ts
+++ b/packages/components/src/Form/Inputs/Combobox/utils/getComboboxText.ts
@@ -34,7 +34,9 @@ export function getComboboxText(
   if (typeof value === 'string') {
     if (options && options.length > 0) {
       const currentOption = options.find((option) => option.value === value)
-      return getComboboxText(currentOption)
+      if (currentOption) {
+        return getComboboxText(currentOption)
+      }
     }
     return value
   }
--- a/packages/components/src/Form/Inputs/Select/SelectMulti.test.tsx
+++ b/packages/components/src/Form/Inputs/Select/SelectMulti.test.tsx
@@ -26,6 +26,7 @@
 
 import { firePasteEvent, renderWithTheme } from '@looker/components-test-utils'
 import { cleanup, fireEvent, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
 import React from 'react'
 
 import { SelectMulti } from './SelectMulti'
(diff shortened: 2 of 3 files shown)


packages/components/src/Form/Inputs/Combobox/utils/getComboboxText.ts
```diff
@@ -34,7 +34,9 @@ export function getComboboxText(
   if (typeof value === 'string') {
     if (options && options.length > 0) {
       const currentOption = options.find((option) => option.value === value)
-      return getComboboxText(currentOption)
+      if (currentOption) {
+        return getComboboxText(currentOption)
+      }
```
mdodgelooker, line 39: If no matching option was found (i.e. `currentOption` is `undefined`) it was returning `''`. Now it will fall back to the `value` if no match is found.

packages/components/src/Form/Inputs/Select/SelectMulti.test.tsx
```diff
@@ -80,20 +81,81 @@ describe('SelectMulti', () => {
     expect(getAllByRole('button')).toHaveLength(2)
   })
 
-  test('defaultValues', () => {
-    const options = [
-      { label: 'Foo', value: 'FOO' },
-      { label: 'Bar', value: 'BAR' },
-      { label: 'Baz', value: 'BAZ' },
-      { label: 'Qux', value: 'QUX' },
```
mdodgelooker, line 88: This was an exact duplicate of the test above it. Whoops!

lukelooker (approved):
lol - that's way down there. Great catch, great issue description :)

PR #1878, opened 2021-02-08
fix: InlineInputText updated test coverage

laraharrow:

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts

### DIFF
--- a/packages/components/src/Form/Inputs/InlineInputText/InlineInputText.test.tsx
+++ b/packages/components/src/Form/Inputs/InlineInputText/InlineInputText.test.tsx
@@ -25,10 +25,26 @@
  */
 
 import { renderWithTheme } from '@looker/components-test-utils'
-import { screen } from '@testing-library/react'
-import React from 'react'
+import { screen, fireEvent } from '@testing-library/react'
+import React, { useState, ChangeEvent } from 'react'
+import { Button } from '../../../Button'
 import { InlineInputText, InlineInputTextBase } from './InlineInputText'
 
+const InlineInputTextOnChange = () => {
+  const [value, setValue] = useState('Type here...')
+  const onChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setValue(event.currentTarget.value)
+  }
+  const onClick = () => {
+    setValue('You clicked the button')
+  }
+  return (
+    <>
+      <InlineInputText value={value} onChange={onChange} />
+      <Button onClick={onClick}>Click Me</Button>
+    </>
+  )
+}
 describe('InlineInputText', () => {
   test('renders and displays placeholder', () => {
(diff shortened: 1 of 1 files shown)


packages/components/src/Form/Inputs/InlineInputText/InlineInputText.test.tsx
```diff
@@ -46,10 +62,54 @@ describe('InlineInputText', () => {
     expect(value).toBeInTheDocument()
   })
 
+  test('renders and displays when passed prop simple', () => {
+    const { getByDisplayValue } = renderWithTheme(
+      <InlineInputText simple value="this is the value" />
+    )
+    const value = getByDisplayValue('this is the value')
```
dbchristopher, line 69: Kent C. Dodds [recommends always using `screen.[whatever]`](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen) for everything, rather than pulling the individual selectors from the render function. 

So you could change to `screen.getByDisplayValue` just for future developer ergonomics. I've been making the change for all the newer tests I've been writing as well

packages/components/src/Form/Inputs/InlineInputText/InlineInputText.test.tsx
```diff
@@ -46,10 +62,54 @@ describe('InlineInputText', () => {
...
+    expect(value).toHaveStyleRule('border-bottom-color: transparent')
```
dbchristopher, line 70: I think I would avoid hard-coding style rules in tests now that we have real image snapshots. I'd just do the more basic `toBeInTheDocument()` for everything, just so we don't have to manually update all these tests if the default style ever does change

dbchristopher (changes requested):
This all looks great! I just suggested a couple tiny changes just to make sure these tests are as simple to maintain in the future as possible

packages/components/src/Form/Inputs/InlineInputText/InlineInputText.test.tsx
```diff
@@ -25,10 +25,26 @@
...
 
 import { renderWithTheme } from '@looker/components-test-utils'
-import { screen } from '@testing-library/react'
-import React from 'react'
+import { screen, fireEvent } from '@testing-library/react'
+import React, { useState, ChangeEvent } from 'react'
+import { Button } from '../../../Button'
 import { InlineInputText, InlineInputTextBase } from './InlineInputText'
 
+const InlineInputTextOnChange = () => {
```
dbchristopher, line 33: I think it would clarify intent to name this `ControlledInputText` instead

dbchristopher (approved):
👍

PR #1879, opened 2021-02-08
docs: Add List docs and update Tree docs

ElliotTheHuman:
- Added `List` and `ListItem` docs
- Updated `Tree` docs to align with `List` / `ListItem` refactor

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts


ElliotTheHuman:
Tagging @mariebeau for visibility

www/src/documentation/components/content/demos/ListItemPropTable.tsx
```diff
@@ -0,0 +1,174 @@
...
+        </TableHead>
+        <TableBody>
+          <TableRow>
+            <TableDataCell pr="large">color</TableDataCell>
+            <TableDataCell pr="large">
+              <Code>string</Code>
+            </TableDataCell>
+            <TableDataCell>
+              If child is of type string, <Code>color</Code> determines the
+              color of the child element and icon, if present.
```
dbchristopher, line 60: SUPER minor nitpick: I notice that a few of these descriptions end in period, but most don't. So maybe just for consistency's sake treat them as full grammatical sentences?

www/src/documentation/components/content/demos/ListItemPropTable.tsx
```diff
@@ -0,0 +1,174 @@
...
+          <TableRow>
+            <TableDataCell pr="large">keyColor</TableDataCell>
+            <TableDataCell pr="large">
+              <Code>boolean</Code>
+            </TableDataCell>
+            <TableDataCell>
+              Replace the normal uiN(1-5) color for selected and selected +
+              hovered color with key colors
+            </TableDataCell>
+          </TableRow>
```
dbchristopher, line 136: Maybe I'm not putting it in the right place, but it doesn't seem like the `keyColor` prop does anything? It also seems weird that there would be both `color` and `keyColor` props anyways

dbchristopher (changes requested):
This looks good @ElliotTheHuman! The only real issue I found is that the `keyColor` prop in the prop table doesn't seem to do anything, so maybe it was included by mistake?

www/src/documentation/components/content/demos/ListItemPropTable.tsx
ElliotTheHuman: Ah the need for consistency makes total sense here. Good call.

www/src/documentation/components/content/demos/ListItemPropTable.tsx
ElliotTheHuman: The `keyColor` prop doesn't really do anything unless the item is selected or hovered + selected. It's here as a result of this ticket a while back where we unified the hovered and selected states of a various list-like components:
https://looker.atlassian.net/browse/LENS-1293

Screenshot:
Screen Shot 2021-02-09 at 10 41 08 AM

It definitely is odd to have `color` and `keyColor` since they sound so similar. I could add a note on `keyColor` that directs people to `color` if they're looking to change text color! Changing `color` to a different name might be an option too.

Thoughts?

ElliotTheHuman: Also, I noticed that adding `keyColor` to a parent `List` doesn't actually spread it down to child props like `Tree` does 🤦 I can patch that up since it's a pretty easy context fix.

www/src/documentation/components/content/demos/ListItemPropTable.tsx
ElliotTheHuman: Resolved in 383b79b

dbchristopher (approved):
Awesome, looks good @ElliotTheHuman!

PR #1881, opened 2021-02-09
feat: i18n support

mdodgelooker:
### TODO:

- [x] Localize all text in `/components` (just English for now)

#### In a future PR:
- [ ] Story to demo changing languages, adding translations
- [ ] Documentation – how to use i18n
- [ ] Documentation – which strings are localized in each component

### Requirements

- [ ] Documentation updated
- [x] i18n impacts
- [ ] a11y impacts


packages/components/src/Chip/Chip.tsx
```diff
@@ -120,6 +121,8 @@ const ChipJSX = forwardRef(
     }: ChipProps,
     ref: Ref<HTMLSpanElement>
   ) => {
+    const { t } = useTranslation('Chip')
```
mdodgelooker, line 124: 1. Get the `t` function from the `useTranslation` hook, with the component name as the namespace argument.

packages/components/src/Chip/Chip.tsx
```diff
@@ -152,7 +155,7 @@ const ChipJSX = forwardRef(
             <IconButton
               disabled={disabled}
               icon="Close"
-              label="Delete"
+              label={t('Delete')}
```
mdodgelooker, line 158: 2. Replace the string with a call to `t` function with the old string as the "key" argument.

packages/components-providers/src/I18n/resources.ts
```diff
@@ -0,0 +1,35 @@
...
+
+ */
+
+import { Resource } from 'i18next'
+
+export const i18nResources: Resource = {
+  en: {
+    Chip: {
+      Delete: 'Delete',
+    },
```
mdodgelooker, line 33: 3. For each component, add an object here containing all localized strings (see below for instructions 1 & 2).
- The component name (a.k.a. the "namespace") has to match what you pass to `useTranslation`, for example `const { t } = useTranslation('ComponentName')`
- The "key" has to match what you pass to the `t` function, and for now it can be exactly what the existing text was, for example `label={t('Some text')}`
- The value should be the existing text, so same as the "key": 
```
ComponentName: {
  'Some text': 'Some text'
}
```

laraharrow:
Hey @mdodgelooker if there is anything else that you think I could help you with let me know. I'm here to help!
I'm so excited to get this working 🤩

mdodgelooker:
Thanks @laraharrow! Nothing at the moment but I'll let you know 👍

laraharrow:
ok I think we are good to go on i18n coverage around our components 🎉
there were 2 warnings that I didn't fix, because I think they are unrelated with this work:
Screen Shot 2021-02-12 at 10 14 08 AM

packages/components/src/PageSize/PageSize.tsx
```diff
@@ -66,19 +67,21 @@ export const PageSizeLayout: FC<PageSizeProps> = ({
...
   options = defaultPageSizes,
 }) => {
+  const { t } = useTranslation('PageSize')
+
   const handleOnChange = (newValue: string) => onChange(Number(newValue))
 
   return Math.min(...options) < total ? (
     <div className={className}>
-      Display
+      {t('Display')}
```
lukelooker, line 76: This is an interesting case where I wonder if some languages' pluralization rules would merit recomposition. 

`PageSize` is slated for a larger refactor soon so I don't think we need to address it now - just occurred to me while reading through.

packages/components/src/Status/__snapshots__/Status.test.tsx.snap
```diff
@@ -43,7 +43,7 @@ exports[`critical Status 1`] = `
     xmlns="http://www.w3.org/2000/svg"
   >
     <title>
-      Error
+      Error
```
lukelooker, line 46: ew trailing white-space (j/k - no biggie really)

lukelooker (approved):
Looks great! Super excited to have this in :)

packages/components/src/PageSize/PageSize.tsx
mdodgelooker: Oh dang good catch. And it would be interesting to try to render the `Select` in the right spot. Definitely something to work on in the `PageSize` refactor.

packages/components/src/Status/__snapshots__/Status.test.tsx.snap
mdodgelooker: Ha oops! Not totally sure how that's getting in there, but oh well for now. :)

PR #1882, opened 2021-02-09
test(ProgressCircular): Improve test coverage

laraharrow:

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts

### DIFF
--- a/packages/components/src/ProgressCircular/ProgressCircular.test.tsx
+++ b/packages/components/src/ProgressCircular/ProgressCircular.test.tsx
@@ -1,0 +1,85 @@
+/*
+
+ MIT License
+
+ Copyright (c) 2020 Looker Data Sciences, Inc.
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
+ to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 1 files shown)


mdodgelooker (approved):
LGTM

mdodgelooker (changes requested):
Ok I'm stumped on where the 88.89% on `DeterminateProgress.tsx` is coming from. The only tiny piece that's not being tested is the `progress = 0` default value. In reality, `progress` is never undefined, so I think you could remove the ` = 0` as the `?` in `progress?: number` in the props.

packages/components/src/ProgressCircular/ProgressCircular.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  test('renders progress of 25', () => {
+    renderWithTheme(<ProgressCircular progress={25} />)
+    expect(screen.queryByRole('progressbar')).toHaveAttribute(
+      'aria-valuenow',
+      '25'
+    )
+  })
+
+  test('renders progress of 0', () => {
```
laraharrow, line 46: @mdodgelooker here is the test I wrote for testing when progress = 0

mdodgelooker (approved):
@laraharrow because the progress prop is never undefined by the time it renders DeterminateProgess, the default in that component is never used. That's the only place I can see that isn't covered but I don't think it really matters anyway.

PR #1883, opened 2021-02-09
feat: adding i18n to our components

laraharrow:

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts


packages/components-providers/src/I18n/resources.ts
```diff
@@ -0,0 +1,120 @@
...
+ SOFTWARE.
+
+ */
+
+import { Resource } from 'i18next'
+
+export const i18nResources: Resource = {
+  en: {
+    AdvancedInputControls: {
+      Label: 'Clear Field',
```
mdodgelooker, line 32: Ah looks like you're mostly calling `t()` with in the value, like `t('Clear Field')` in `AdvancedInputControls.tsx`, but using a different name as the key here in resources ("Label" in this case). For all translations except ones that contain `{{}}`, you can use the value as the key, which is how it's done in the core product, so might as well keep it consistent for sanity's sake. You need to add single quotes if the key contains spaces, so this would be:
```suggestion
      'Clear Field': 'Clear Field',
```
Anytime `t` is called with a key that's not found, an error is thrown and you'll see something like the screenshot below, so to check if everything is properly translated, you can just go to a story or docs page containing that component.
image

packages/components-providers/src/I18n/resources.ts
```diff
@@ -0,0 +1,120 @@
...
+    },
+    GetIntentLabel: {
+      Error: 'Error',
+      Inform: 'inform',
+      Positive: 'positive',
+      Warn: 'Warning',
+    },
+    InputDate: {
+      Placeholder:
+        'Date {{formatDateString(new Date(Date.now())}}, {{dateStringLocale}}',
```
mdodgelooker, line 82: I think this one can break the rule we chatted about earlier of not breaking up phrases because of languages like Japanese – since it's not really a phrase, more like a label (see suggestion part 2 in `InputDate.tsx`):
```suggestion
      Date: 'Date',
```
image

packages/components/src/InputDate/InputDate.tsx
```diff
@@ -170,10 +173,10 @@ export const InputDate: FC<InputDateProps> = forwardRef(
...
           aria-describedby={ariaDescribedby}
           aria-labelledby={ariaLabelledby}
-          placeholder={`Date (${formatDateString(
-            new Date(Date.now()),
-            dateStringLocale
-          )})`}
+          placeholder={t('Date', {
+            dateStringLocale: String(dateStringLocale),
+            formatDateString: String(formatDateString),
+          })}
```
mdodgelooker, line 179: Suggestion part 2:
```suggestion
          placeholder={`${t('Date')} (${formatDateString(
            new Date(Date.now()),
            dateStringLocale
          )})`}
```

mdodgelooker (changes requested):
Making so much progress! 🎉 Are you able to change the PR to compare against my `mdodge/i18n` branch so I can keep making updates in there to the setup code (`useI18n`, etc)?

laraharrow:
ok @mdodgelooker  I looked at every file that got `i18ned` and everything seems to be working properly now. Except for `MessageBar.tsx` the error that I'm getting is that `t` is not a function but also on VCCode its complaining that `intent` on line 135 could be undefined and `Type 'undefined' is not assignable to type 'TFunction'` but I'm not using the `t` method on `intent`. Very confusing.

mdodgelooker:
@laraharrow this is awesome! I've merged your `lara/i18n` branch into my `mdodge/i18n`– I'll take a look at `MessageBar` and then add some more features like eslint, context utilities, etc.

PR #1886, opened 2021-02-10
fix(FocusTrap): Return focus behavior on nested focus traps

mdodgelooker:
The [Focus Trap refactor](https://github.com/looker-open-source/components/pull/1644) introduced a bug in the "return focus" behavior. When a trap was activated while a another trap was already active – e.g. a nested menu – the reference to original return focus element was lost. When both focus traps were deactivated (pressing the escape key on a nested menu) focus just landed on `<body>` instead of the menu button.

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [x] a11y impacts


lukelooker (approved):
Works perfectly. Code-comments we're really helpful in understanding what parts go where :)

PR #1888, opened 2021-02-10
test: Combobox useKeyDown coverage

mdodgelooker:
Improves coverage for `Combobox/utils/useKeyDown` via new & updated tests in `Combobox.test.tsx` & `ComboboxMulti.test.tsx`

And, what do you know, because coverage was bad, there was a bug! Arrow Up should open a closed, focused `Combobox`, but it was broken. You can test the issue on a [basic Select here](https://looker-open-source.github.io/components/latest/components/forms/input-select) and the fix on the same page locally. To get a closed, focused `Select`, tab navigate to it or click it then hit Escape.

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts

### DIFF
--- a/packages/components/src/Form/Inputs/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Form/Inputs/Combobox/Combobox.test.tsx
@@ -383,6 +383,11 @@ describe('<Combobox/> with children', () => {
 })
 
 describe('Keyboard navigation', () => {
+  const onChangeMock = jest.fn()
+  beforeEach(() => {
+    onChangeMock.mockClear()
+  })
+
   const arrowDown = {
     key: 'ArrowDown',
   }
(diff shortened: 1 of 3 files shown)


packages/components/src/Form/Inputs/Combobox/Combobox.test.tsx
```diff
@@ -455,9 +473,27 @@ describe('Keyboard navigation', () => {
...
+    expect(items[0]).toHaveAttribute('aria-selected', 'true')
+    expect(items[1]).toHaveAttribute('aria-selected', 'false')
+
     fireEvent.keyDown(input, enter)
     expect(queryByRole('listbox')).not.toBeInTheDocument()
 
+    expect(onChangeMock).toHaveBeenCalledTimes(1)
+    const value = { label: 'Foo', value: '101' }
+    const onChangeValue = name === 'Combobox' ? value : [value]
+    expect(onChangeMock).toHaveBeenCalledWith(onChangeValue)
```
dbchristopher, line 487: I find it convenient to do `expect(onChangeMock.mock.calls).toMatchInlineSnapshot()` for stuff like this. Then you don't have to hard-code the value, and it automatically logs how many times it's been called. Though I suppose inline snapshots probably wouldn't work inside an `each` test block? Might be an appropriate use case for normal snapshots still though

dbchristopher (approved):
LGTM! I made the suggestion that tracking mock calls are a great use of snapshots, but definitely not a required change. Happy to re-approve if you want to make that change

ElliotTheHuman (commented):
Changes look good from my end as well!

Note: I saw Daniel left a comment regarding using snapshot testing for the `onChangeMock` checks in `Combobox.test.tsx`; going to hold off from approving for now.

packages/components/src/Form/Inputs/Combobox/Combobox.test.tsx
mdodgelooker: Yes, sadly `toMatchInlineSnapshot` did complain about the `test.each`.

PR #1889, opened 2021-02-10
test: Increase Tree-related coverage

ElliotTheHuman:
- Added tests for `TreeArtificial` and detail key down behavior in `Tree`
- Added ignore to `generateTreeBorder` since A) We cover it using visual snapshotting B) linear-gradient and jest-dom don't seem to like each other very much

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts


packages/components/src/Tree/Tree.test.tsx
```diff
@@ -124,6 +128,38 @@ describe('Tree', () => {
     expect(onClose).toHaveBeenCalledTimes(0)
   })
 
+  test('Key presses on detail do not open the Tree or trigger callbacks when accessory === true', () => {
+    const onOpen = jest.fn()
+    const onClose = jest.fn()
+    const { getByText, queryByText } = renderWithTheme(
```
dbchristopher, line 134: Minor nit: we should switch to using `screen.[whatever]` for all these selectors! [Kent C Dodds recommends that](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen) as a best practice, and I have to say it is much more convenient

packages/components/src/Tree/Tree.test.tsx
```diff
@@ -124,6 +128,38 @@ describe('Tree', () => {
...
+    fireEvent.keyDown(detail, {
+      key: 'Enter',
+    })
+    expect(queryByText('Hello World')).not.toBeInTheDocument()
+    expect(onOpen).toHaveBeenCalledTimes(0)
+    fireEvent.keyDown(detail, {
+      key: 'Enter',
+    })
+    expect(queryByText('Hello World')).not.toBeInTheDocument()
+    expect(onClose).toHaveBeenCalledTimes(0)
```
dbchristopher, line 160: Do you actually need multiple Enter key presses here? Seems like this would work just as well?

```suggestion
    fireEvent.keyDown(detail, {
      key: 'Enter',
    })
    expect(queryByText('Hello World')).not.toBeInTheDocument()
    expect(onOpen).toHaveBeenCalledTimes(0)
    expect(onClose).toHaveBeenCalledTimes(0)
```

packages/components/src/Tree/TreeArtifical.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+  test('Renders children', () => {
+    const { getByText } = renderWithTheme(
+      <TreeArtificial density={-3}>
+        <TreeItem>Tree Item</TreeItem>
+      </TreeArtificial>
+    )
+
+    getByText('Tree Item')
+  })
+})
```
dbchristopher, line 41: Spelling mistake in the file name: `TreeArtifical.test.tsx`

packages/components/src/Tree/TreeArtifical.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+describe('TreeArtificial', () => {
+  test('Renders children', () => {
+    const { getByText } = renderWithTheme(
+      <TreeArtificial density={-3}>
+        <TreeItem>Tree Item</TreeItem>
+      </TreeArtificial>
+    )
+
+    getByText('Tree Item')
```
dbchristopher, line 39: needs an `expect`?

packages/components/src/Tree/utils/generateTreeBorder.ts
```diff
@@ -24,11 +24,23 @@
...
 import { indicatorDefaults } from './indicatorDefaults'
 
+/**
+ * Ideally, we could use toHaveStyle on the internal AccordionContent container to
+ * check if a linear-gradient background was added when border === true on <Tree>.
+ *
+ * However, toHaveStyle (as well as window.getComputedStyle) return undefined when
+ * trying to check that container's background style. Visual snapshotting currently
+ * accounts for the border styling / prop so ignoring this file for Codecov purposes.
+ *
```
dbchristopher, line 40: Honestly I'm not sure we ever need to do `toHaveStyle` assertions, as that will more effectively covered by visual snapshotting. So I'd just go leave the `/* istanbul ignore file */` and make a note along the lines of "style rules validated by storyshots"

dbchristopher (changes requested):
Looks good! Left a couple small best practice suggestions, but nothing major

packages/components/src/Tree/Tree.test.tsx
ElliotTheHuman: Makes sense! It's way more convenient to use screen in retrospect.

Went ahead and switched to `screen.[selector name]` in `Tree.test` in 5b74c4e

packages/components/src/Tree/TreeArtifical.test.tsx
ElliotTheHuman: :doh: Good catch. Resolved in 34a598065

packages/components/src/Tree/Tree.test.tsx
ElliotTheHuman: We probably don't. The intent is to make sure that both `onOpen` and `onClose` don't trigger (i.e. on the first and second click of a closed `Tree`, respectively), but if the `Tree` "opening" its accordion on the first clicks it's probably safe to assume that it's not "closing" on the second one / firing it's `onClose`.

I do this in `Clicks on detail do not open the Tree or trigger callbacks when accessory === true` as well and I think it's correct to remove the extra click / keypress there as well.

Resolved in 4e225d48d

packages/components/src/Tree/utils/generateTreeBorder.ts
ElliotTheHuman: 👍  Makes sense. That original comment is a bit unwieldy hah.

Resolved in 4db2a31ef

packages/components/src/Tree/TreeArtifical.test.tsx
ElliotTheHuman: I usually just have a `getByText` call (without an `expect`) when checking for something in the document since the "getX" query will fail the test if it doesn't find any matches. That being said, I could see including `toBeInTheDocument()` to be more readable!

Added that in cb8cd2c.

dbchristopher (approved):
Looks good!

PR #1890, opened 2021-02-10
test: Box component snapshot replacements

dbchristopher:
### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts


packages/components/src/Layout/Box/Box.test.tsx
```diff
@@ -36,42 +36,6 @@ import {
...
-    assertSnapshot(<Box mt={['large', null, 'medium']}>🥑</Box>)
-  })
-
-  test('Box with SizeNone is valid', () => {
-    assertSnapshot(<Box mt="none">🥑</Box>)
-  })
-
-  test('Box supports background-color', () => {
-    assertSnapshot(<Box bg="key" />)
-  })
```
lukelooker, line 57: Might be good to use RTL to test that these are assigned as-expected.

packages/components/src/Layout/Box/Box.test.tsx
```diff
@@ -80,37 +44,6 @@ describe('Box', () => {
...
-      assertSnapshot(<Box border="1px solid black" />)
-    })
-
-    test('supports borderRadius', () => {
-      assertSnapshot(<Box borderRadius="medium" />)
-    })
-
-    test('supports borderColor', () => {
-      assertSnapshot(<Box borderColor="ui2" />)
-    })
```
lukelooker, line 99: Since these are doing derived-theme behaviors I think simple RTL tests would be helpful here.

packages/components/src/Layout/Box/Box.test.tsx
```diff
@@ -80,37 +44,6 @@ describe('Box', () => {
     assertSnapshot(<Box bottom="1rem" />)
```
lukelooker, line 44: I think this test could be replaced with RTL as well - poking at the `hasStyledRule`

packages/components/src/Layout/Box/Box.test.tsx
```diff
@@ -80,37 +44,6 @@ describe('Box', () => {
...
-
-  describe('fonts', () => {
-    test('font helpers', () => {
-      assertSnapshot(<Box fontFamily="brand" />)
-      assertSnapshot(<Box fontSize="small" />)
-      assertSnapshot(<Box fontWeight="bold" />)
-      assertSnapshot(<Box letterSpacing="10" />)
-      assertSnapshot(<Box lineHeight="medium" />)
-      assertSnapshot(<Box textAlign="right" />)
-      assertSnapshot(<Box color="ui4" />)
```
lukelooker, line 110: I think there's probably value in keeping the `fontFamily`, `fontSize`, `lineHeight` and `color` as RTL tests (although they could totally be done as one big one) to prove that theme values are pushing down as expected.

lukelooker (commented):
Happened across this PR - realized after digging in that perhaps it's not complete so feel free to ignore my comments as you see appropriate :)

packages/components/src/Layout/Box/Box.test.tsx
dbchristopher: Oh I just replaced these with storyshots, with screenshots taken at different breakpoints. That's probably sufficient to confirm that key color and responsive values are assigned as expected, right?

packages/components/src/Layout/Box/Box.test.tsx
dbchristopher: I'm going to add more another storyshot to confirm absolute positioning

packages/components/src/Layout/Box/Box.test.tsx
dbchristopher: Same here -- I moved the border settings to storyshots. In general, anything style related we'll want to do in storyshots, right?

Screen Shot 2021-02-11 at 9 28 49 AM

packages/components/src/Layout/Box/Box.test.tsx
lukelooker: My only concern is verifying that Box is getting fully "exercised" but I guess `Box` really is completely logic-less so there's nothing to test beyond "it doesn't crash". 

Ignore me :)

packages/components/src/Layout/Box/Box.test.tsx
lukelooker: 👍

packages/components/src/Layout/Box/Box.test.tsx
lukelooker: I think you're right to dump these - really adding tests here would just be verifying that Styled System is working properly - maybe not a bad idea but not really `Box`'s responsibility

dbchristopher:
ok @lukelooker this should be ready for final review! I moved all the style tests into storybook, and updated the UX tests to use RTL. 

As an aside, I have no idea why it triggered these <1 — 10 byte changes in some of the field components. Weird 🤷 

Screen Shot 2021-02-11 at 3 50 37 PM

packages/components/src/Layout/Box/Box.test.tsx
```diff
@@ -25,146 +25,57 @@
...
-      )
-    })
-  })
-
   describe('as=', () => {
     test('allows Box to render as any HTML tag', () => {
-      const boxAsButton = mountWithTheme(<Box as="button" />)
-      expect(boxAsButton.find('button').exists()).toBeTruthy()
+      renderWithTheme(<Box as="button">Press Me</Box>)
+      expect(screen.getByText('Press Me').tagName).toEqual('BUTTON')
```
lukelooker, line 38: Could you use `getByRole` here?

```suggestion
      expect(screen.getByRole('button')).toBeInTheDocument()
```

packages/components/src/Layout/Box/Box.test.tsx
```diff
@@ -25,146 +25,57 @@
...
     })
   })
 
   describe('core HTML attributes', () => {
     test('Box allows autoFocus', () => {
-      assertSnapshot(<Box autoFocus>Autofocus?</Box>)
+      renderWithTheme(
+        <Box as="input" type="text" defaultValue="Autofocus" autoFocus />
+      )
+      expect(screen.getByDisplayValue('Autofocus')).toHaveFocus()
```
lukelooker, line 52: Neat :)

lukelooker (approved):
LGTM. 

Made one minor pedantic suggestion for a jest-helper (feel free to ignore)

The snapshots with the 0 - 5 bytes changes are caused by some kind of "jitter" that happens when taking the snapshots in Github actions. I have a task to explore ignore those in the future because they are, admittedly annoying.

packages/components/src/Layout/Box/Box.test.tsx
dbchristopher: oh yep, that's better! good call

lukelooker (approved):
✅

PR #1896, opened 2021-02-11
test: Deprecate snapshots for Avatar, Badge, Checkbox, InputText, Radio & TextArea

lukelooker:
Removes approximately 50 snapshot tests. Most are replaced with react-testing-library tests and also duplicated as Storybook snapshot tests as well.

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts


packages/components/src/Badge/Badge.test.tsx
```diff
@@ -23,21 +23,18 @@
...
-  assertSnapshot(<Badge intent="critical">critical</Badge>)
+test('Badge', () => {
+  renderWithTheme(
+    <Badge size="small" intent="positive">
+      Good!
+    </Badge>
+  )
+  const badge = screen.getByText('Good!')
+  expect(badge).toHaveStyle('background: rgb(228, 245, 235)')
+  expect(badge).toHaveStyle('line-height: 16px')
```
dbchristopher, line 39: Relevant to the discussion on my Box PR -- what are your thoughts on when we'd want a visual snapshot vs testing for hard-coded style rules in RTL? 

I've been approaching this from the assumption that every style-related test is better off as a storyshot, so RTL should only be used to validate behavior and accessibility

packages/components/src/Form/Inputs/InputHidden/InputHidden.test.tsx
```diff
@@ -24,19 +24,23 @@
...
+    expect(input).toHaveAttribute('name', 'Bob')
+    expect(input).toHaveAttribute('id', 'Bobby')
+  })
+
+  test('with a value', () => {
+    const { container } = renderWithTheme(
+      <InputHidden defaultValue="Some value" />
+    )
+    const input = container.querySelector('input')
+    expect(input).toHaveValue('Some value')
```
dbchristopher, line 44: Probably better to just use `getByDisplayValue` to verify it rendered the input correctly?

packages/components/src/Form/Inputs/InputText/InputText.test.tsx
```diff
@@ -178,12 +182,11 @@ describe('InputText', () => {
...
-    let counter = 0
-    const handleChange = () => counter++
+    const onChange = jest.fn()
 
-    const wrapper = mountWithTheme(<InputText onChange={handleChange} />)
+    renderWithTheme(<InputText onChange={onChange} />)
 
-    wrapper.find('input').simulate('change', { target: { value: '' } })
-    expect(counter).toEqual(1)
+    userEvent.type(screen.getByRole('textbox'), 'Hello world')
```
dbchristopher, line 189: oh I don't think I'm familiar with the `userEvent` actions! That's really nice.

packages/components/src/Form/Inputs/TextArea/TextArea.test.tsx
```diff
@@ -24,42 +24,48 @@
...
+
+  test('TextArea resizes with prop resize = none', () => {
+    renderWithTheme(<TextArea id="TextAreaID" resize="none" />)
+    expect(screen.getByRole('textbox')).toHaveStyle('resize: none')
+  })
+
+  test('TextArea resizes with prop resize = vertical', () => {
+    renderWithTheme(<TextArea id="TextAreaID" resize="vertical" />)
+    expect(screen.getByRole('textbox')).toHaveStyle('resize: vertical')
+  })
```
dbchristopher, line 70: probably none of these actually need the `id` attribute, do they? not a huge deal either way lol

dbchristopher (dismissed):
Left a few comments, but no major blockers! It would be nice to think through some general guidelines on when it's best to do Storysnaps vs `toHaveStyle` in RTL

packages/components/src/Badge/Badge.test.tsx
lukelooker: I think the argument here is that we want to be able to use coverage to verify that all the code is getting exercised - in this case I wanted to make sure that the mapping of intent to color was working. 

Basically I would like to make sure that we're confident that every branch is getting exercise and while Storybook is definitely better for understanding _how_ was presented it can't tell us if every permutation was actually run. 

In an ideal world we'd use `jest-image-snapshot` in some of these so that we had both. Perhaps it's worth putting that down as a Q2 quality goal.

packages/components/src/Form/Inputs/InputHidden/InputHidden.test.tsx
lukelooker: 💯  (I struggled at getting hold of that component via RTL - 🤦 )

packages/components/src/Form/Inputs/TextArea/TextArea.test.tsx
lukelooker: 👍  - copypasta in my part. Reads much better without :)

packages/components/src/Form/Inputs/InputText/InputText.test.tsx
lukelooker: @mdodgelooker FTW! (I got learned)

dbchristopher (approved):

PR #1901, opened 2021-02-11
fix(MessageBar): Use intent accent colors for MessageBar background

jhardy:

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [x] a11y impacts


packages/components/src/MessageBar/MessageBar.tsx
```diff
@@ -247,10 +266,9 @@ export const MessageBar = styled(MessageBarLayout).attrs(
   })
 )`
   ${simpleLayoutCSS}
-
   align-items: center;
-  background: ${({ intent, theme: { colors } }) =>
-    intent === 'critical' ? colors.criticalAccent : colors.neutralAccent};
+  ${({ intent, theme: { colors } }) =>
+    intent ? backgroundColor : `background: ${colors.informAccent};`}
```
lukelooker, line 271: Rather than having the default fallback here we should do it in `attrs` above:

```suggestion
  ${background}
```

packages/components/src/MessageBar/MessageBar.tsx
```diff
@@ -239,6 +240,24 @@ const MessageBarContent = styled.div`
...
+    },
+    warn: {
+      backgroundColor: 'warnAccent',
+    },
+  },
+})
+
 export const MessageBar = styled(MessageBarLayout).attrs(
   ({ px = 'medium', py = 'small', width = '100%' }) => ({
     px,
```
lukelooker, line 263: You can use Styled Components `attrs` method to inject default `intent` value:

```suggestion
  ({ intent = 'inform', px = 'medium', py = 'small', width = '100%' }) => ({
    intent,
    px,
```

packages/components/src/MessageBar/MessageBar.tsx
```diff
@@ -239,6 +240,24 @@ const MessageBarContent = styled.div`
...
+      backgroundColor: 'informAccent',
+    },
+    positive: {
+      backgroundColor: 'positiveAccent',
+    },
+    warn: {
+      backgroundColor: 'warnAccent',
+    },
+  },
+})
```
lukelooker, line 259: I've been tending towards using `background` over `background-color`. No huge argument for it besides that it's slightly more terse and if other CSS specifices `background` it can be confusing to have interactions between `background` and `backgroundColor`

```suggestion
const background = variant({
  prop: 'intent',
  variants: {
    critical: {
      background: 'criticalAccent',
    },
    inform: {
      background: 'informAccent',
    },
    positive: {
      background: 'positiveAccent',
    },
    warn: {
      background: 'warnAccent',
    },
  },
})
```

packages/components/src/MessageBar/MessageBar.test.tsx
```diff
@@ -219,9 +219,9 @@ describe('MessageBar', () => {
         'color',
         theme.colors.critical
       )
-      // MesasgeBar background
+      // MessageBar background
       expect(getByRole('status')).toHaveStyleRule(
-        'background',
+        'background-color',
```
lukelooker, line 224: See comment below

```suggestion
        'background',
```

lukelooker (changes requested):
Functionality looks great, minor suggestions around Styled Components usage

packages/components/src/MessageBar/MessageBar.tsx
```diff
@@ -247,10 +266,9 @@ export const MessageBar = styled(MessageBarLayout).attrs(
   })
 )`
   ${simpleLayoutCSS}
-
   align-items: center;
-  background: ${({ intent, theme: { colors } }) =>
-    intent === 'critical' ? colors.criticalAccent : colors.neutralAccent};
+  ${({ intent, theme: { colors } }) =>
+    intent ? backgroundColor : `background: ${colors.informAccent};`}
   border-radius: ${({ theme: { radii } }) => radii.medium};
```
lukelooker, line 272: Since you're working here anyway it'd be great to fix another problem I've seen:

```suggestion
  border-radius: ${({ theme: { radii } }) => radii.medium};
  color: ${({ theme }) => theme.color.text };
```

Currently color is inherited instead of explicit so when themes are inverted things look sad.

lukelooker (changes requested):
One more note to address a nearby issue.

packages/components/src/MessageBar/MessageBar.tsx
jhardy: I originally tried background, but the color does not get applied.  The theme values name is set as the value of `background`. Any thoughts?
image

packages/components/src/MessageBar/MessageBar.tsx
jhardy: image

packages/components/src/MessageBar/MessageBar.tsx
lukelooker: Oh - weird, I forgot that `variant` is doing `styled-system` magic here. In testing it turns out that `bg` works too but also outputs `backgroundColor` so 🤷

lukelooker (approved):

PR #1902, opened 2021-02-11
test(Breakpoint): Add basic coverage

lukelooker:
Simple coverage for `Breakpoint` component

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts

### DIFF
--- a/packages/components/src/Breakpoint/Breakpoint.test.tsx
+++ b/packages/components/src/Breakpoint/Breakpoint.test.tsx
@@ -1,0 +1,72 @@
+/*
+
+ MIT License
+
+ Copyright (c) 2020 Looker Data Sciences, Inc.
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
+ to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ copies of the Software, and to permit persons to whom the Software is
+ furnished to do so, subject to the following conditions:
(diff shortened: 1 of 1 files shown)


dbchristopher (dismissed):
Very nice! When I coded the Breakpoint component I set the [global jest viewport](https://github.com/looker-open-source/components/blob/main/jest.setup.js#L50-L58) to be 800x600. Since these tests rely on that specific setting, do you think it's worth hard-coding that into a `beforeEach` block here? Just to make sure some other test doesn't change it and break things.

ElliotTheHuman:
> Very nice! When I coded the Breakpoint component I set the [global jest viewport](https://github.com/looker-open-source/components/blob/main/jest.setup.js#L50-L58) to be 800x600. Since these tests rely on that specific setting, do you think it's worth hard-coding that into a `beforeEach` block here? Just to make sure some other test doesn't change it and break things.

I think this would help confused folks like me who didn't know that the global jest viewport width was set to 800px in the config file 😆 

Alternatively, having a comment that points to the `jest.setup` file could be helpful. Feel free to disregard this if this is mostly common sense for others though.

dbchristopher:
@ElliotTheHuman lol a comment at least is a great idea! Anything that's dependent on jest config is never very intuitive

lukelooker:
@ElliotTheHuman / @dbchristopher - Great point on the viewport issue. Added a note to the test. TY!

ElliotTheHuman (approved):
Thanks for adding the note in the test suite. LGTM!

PR #1906, opened 2021-02-12
feat(Menu): Menu Refactor

ElliotTheHuman:
- Refactored `Menu` (`MenuList`) to use `List` as foundation
- Refactored `MenuItem` to use `ListItem` as foundation
- Removed `MenuItemContext`

### To-Do
- Add `MenuDivider` and `MenuHeading` to replace `MenuGroup` use case
  - That follow-up PR will be a good opportunity to remove the `MenuGroup` stories
  - Tangential note, but all `MenuGroup`s being used that DON'T have a parent `MenuList` / `Menu` should be replaced `List` + `ListItem` (probably)

NOTE: As per conversation with @lukelooker, we don't plan on merging this in until after the next release (which includes the `Tree` refactor).

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts


packages/components/src/Menu/MenuGroup.test.tsx
```diff
@@ -1,72 +0,0 @@
...
-        <MenuItem>Gouda</MenuItem>
-        <MenuItem id="cheddar">Cheddar</MenuItem>
-      </MenuGroup>
-    )
-
-    expect(
-      queryByTestId('menu-item-cheddar-icon-placeholder')
-    ).not.toBeInTheDocument()
-  })
-})
```
ElliotTheHuman, line 72: The plan is to remove `MenuGroup` so preemptively removing test suite.

Also, these tests check for icon placeholder behavior, which is more suitably tested by visual snapshotting (I think).

packages/components/src/Menu/MenuList.test.tsx
```diff
@@ -28,25 +28,13 @@ import { screen } from '@testing-library/react'
...
-    const { getByTestId } = renderWithTheme(
-      <MenuList>
-        <MenuItem icon="Calendar">Gouda</MenuItem>
-        <MenuItem id="cheddar">Cheddar</MenuItem>
-      </MenuList>
-    )
-
-    getByTestId('menu-item-cheddar-icon-placeholder')
-  })
-
```
ElliotTheHuman, line 49: Tested via visual snapshotting.

packages/components/src/Menu/Menu.tsx
```diff
@@ -44,62 +53,43 @@ export interface MenuProps
...
-      triggerToggle,
-
-      // List props to pass through
-      ...props
-    }: MenuProps,
+    { children, content, id: propsID, listRef, ...props }: MenuProps,
     ref: Ref<any>
   ) => {
+    const popoverProps = pick(props, popoverPropKeys)
+    const listProps = omit(props, popoverPropKeys)
```
ElliotTheHuman, line 81: Note to self: Should use vanilla JS here instead of lodash.

packages/components/src/List/ListItem.tsx
```diff
@@ -115,190 +125,207 @@ export const ListItemLabel = styled.div
     as: !disabled && itemRole === 'link' ? 'a' : 'button',
   }))<ListItemLabelProps>``
 
-const ListItemInternal: FC<ListItemProps> = (props) => {
```
ElliotTheHuman, line 118: A lot of these changes are from introducing `forwardRef` into `ListItemInternal`. The `mouseEnter` / `mouseLeave` handlers + the new state to track hover state is new to this PR though!

packages/components/src/Menu/Menu.tsx
mdodgelooker: Is there a reason you switched from destructuring to pick/omit (other than fewer total lines of code)? I think the latter is less performant, and also will need to be vanilla-JS-ified.

packages/components/src/Menu/MenuGroup.test.tsx
mdodgelooker: 👍 And actually icon placeholder behavior no longer works on a standalone `MenuGroup` – it relies on a `MenuList`.

packages/components/src/Menu/MenuItem.tsx
```diff
@@ -24,215 +24,99 @@
...
-    setFocusVisible(false)
-    // nestedMenuOnClick wraps onClick from props
-    nestedMenuOnClick(event)
-    // Close the Menu unless event has preventDefault
-    if (closeModal && !event.defaultPrevented) {
-      closeModal()
+  extends ListItemProps,
+    Pick<UseNestedMenuProps, 'nestedMenu'> {}
+
+const MenuItemInternal: FC<MenuItemProps> = forwardRef(
```
mdodgelooker, line 41: Remove this `MenuItemProps` here and add it below...
```suggestion
const MenuItemInternal = forwardRef(
```

packages/components/src/Menu/MenuItem.tsx
```diff
@@ -24,215 +24,99 @@
...
+    {
+      children,
+      detail,
+      onClick,
+      onKeyDown,
+      onMouseEnter,
+      onMouseLeave,
+      nestedMenu,
+      ...props
+    },
```
mdodgelooker, line 52: ...here
```suggestion
    }: MenuItemProps,
```

packages/components/src/Menu/MenuItem.tsx
```diff
@@ -24,215 +24,99 @@
...
+      onClick,
+      onKeyDown,
+      onMouseEnter,
+      onMouseLeave,
+    })
+
+    const ref = useForkedRef<HTMLLIElement>(nestedMenuRef, forwardedRef)
+
+    const { density } = useContext(ListItemContext)
+    const { iconSize } = listItemDimensions(density)
```
mdodgelooker, line 76: Nice!

packages/components/src/Menu/MenuItem.tsx
```diff
@@ -24,215 +24,99 @@
...
+          ref={ref}
+          role="menuitem"
+          {...props}
+          {...nestedMenuProps}
+        >
+          {children}
+        </ListItem>
+        {/* Keep nestedMenu popover outside of MenuItemLayout to prevent its events
        from bubbling up to the MenuItem (especially onClick)
        due to React Portal event bubbling */}
```
mdodgelooker, line 113:
```suggestion
        {/* Keep nestedMenu popover outside of ListItem to prevent its events
       from bubbling up (especially onClick) due to React Portal event bubbling */}
```

packages/components/src/Menu/MenuList.tsx
```diff
@@ -24,228 +24,42 @@
...
+import { MenuGroup } from '.'
 
-export interface MenuListProps
-  extends CompatibleHTMLProps<HTMLUListElement>,
-    MaxHeightProps,
-    MinHeightProps,
-    HeightProps,
-    MaxWidthProps,
-    MinWidthProps,
-    WidthProps {
```
mdodgelooker, line 71: The core product does use `width` and occasionally `minWidth` props on `Menu` (which previously was coming from here). Should we add support in `List`, or just not support it, and you get the width that you get? @lukelooker thoughts?

packages/components/src/utils/useHovered.ts
```diff
@@ -28,8 +28,9 @@ import { RefObject, useEffect, useState } from 'react'
 import { getCurrentNode } from './getCurrentNode'
 import { useCallbackRef } from './useCallbackRef'
 
-// Get the current hover state of an element
-// Recommended to be used with useCallbackRef
+/**
+ * Get the current hover / focus state of a element of which component has no control
```
mdodgelooker, line 32: Thank you so much for adding this. 🙏 Such a confusing hook.
```suggestion
 * Get the hover / focus state of an element over which the current component has no control
```

www/src/documentation/components/overlays/menuitem.mdx
```diff
@@ -59,12 +59,13 @@ Use the `detail` and/or `description` prop to provide extra explanatory informat
...
 ## Tooltip
 
-To implement Tooltip on MenuItem, pass a string value to the `tooltip` prop.
-You can change the tooltip position by setting `tooltipPlacement` to one of `bottom`, `left`, `right` or`top`. Tooltips will render to the`left` by default.
+To implement a tooltip on a `MenuItem`, wrap the `MenuItem` in a `Tooltip` element.
 
 ```jsx
 <MenuList>
-  <MenuItem tooltip="It is gouda!">Gouda</MenuItem>
+  <Tooltip content="It is gouda!">
```
mdodgelooker, line 66: Should we use `"left"` in the example b/c that usually looks good? Feel free to disregard though.
```suggestion
  <Tooltip content="It is gouda!" placement="left">
```

mdodgelooker (changes requested):
Great work @ElliotTheHuman ! And thanks for your patience with my slow reviewing. 😉

packages/components/src/Menu/Menu.tsx
ElliotTheHuman: The props object that was spread onto `MenuList` included props from `PopoverProps` like "portalElement" and "scrollLock" and so giving those props to `Menu` could result in some "React does not recognize X prop on a DOM element" errors.

Having a pick / omit here seemed a bit sturdier since `Menu` wouldn't require its props to be updated if `PopoverProps` was every updated (the `popoverPropKeys` array would just need to be updated instead). That being said, if the performance gain from destructuring the props matters more, I could definitely go that route instead and include the unaccounted for `Popover` props.

packages/components/src/Menu/Menu.story.tsx
```diff
@@ -60,7 +60,6 @@ const menuItems = (
         detail="detail"
         description="this is the description"
         icon="LogoRings"
-        tooltip="Some tooltip"
```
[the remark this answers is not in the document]
lukelooker: If I'm reading this right we removed the the `tooltip` prop on `MenuItem` specifically bit it can still be accomplished via composition. E.g.:

```<Tooltip ...><MenuItem>Stuff</MenuItem></Tooltip>```

File: packages/components/src/Menu/MenuList.tsx (hunk 9 above)
Comment (lukelooker) [reply packages/components/src/Menu/MenuList.tsx:71]: I'm _tempted_ to say you get what you get. Which is to say we specify a `minWidth: 12rem` of on `MenuItem`s themselves and otherwise things just grow to fit.

File: packages/components/src/Menu/Menu.story.tsx (hunk 12 above)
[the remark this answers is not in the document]
Comment (ElliotTheHuman) [reply packages/components/src/Menu/Menu.story.tsx:63]: Correct, we removed the `tooltip` prop on `MenuItem` and instead support that pattern using composition.

I'm not too sure on what the right call is here, but doing this keeps `ListItem` from taking on `tooltip`-related props and keeps it simple.

It _shouldn't_ be hard to add a `tooltip`-related props to `ListItem`, so happy to do that if the convenience to the user outweighs any potential tech debt here.

File: packages/components/src/Menu/Menu.tsx [hunk 13]
```diff
@@ -44,62 +53,43 @@ export interface MenuProps
...
+  placement?: Placement
+  /**
+   * By default MenuList will reposition itself if they overflow the widow.
+   * You can use the pin property to override this behavior.
+   */
+  pin?: boolean
+  /**
+   * Allow the overlay to break out of the scroll parent
+   */
+  escapeWithReference?: boolean
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 4c6420e

packages/components/src/Menu/MenuItem.tsx
ElliotTheHuman: Ah I figured I was doing something wrong with Props here since the typing of `ref` on `ListItem` kept erroring out.

Good call! Resolved in af3e87fdf.

packages/components/src/Menu/MenuItem.tsx
ElliotTheHuman: Resolved in af3e87fdf

packages/components/src/List/ListItem.tsx
```diff
@@ -96,6 +102,10 @@ export interface ListItemProps
    * - Use **'button'** for items that trigger in page interactions, like displaying a dialog
    */
   itemRole?: 'link' | 'button'
+  /**
+   * Optional ref prop
+   */
+  ref?: Ref<HTMLLIElement>
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in af3e87f

packages/components/src/List/ListItem.tsx
```diff
@@ -115,190 +125,207 @@ export const ListItemLabel = styled.div
...
-        color={iconColor}
-        name={icon}
-        size={itemDimensions.iconSize}
-        mr={itemDimensions.iconGap}
-      />
-    ) : (
-      iconGutter && (
-        <IconPlaceholder
+const ListItemInternal: FC<ListItemProps> = forwardRef(
+  (props, ref: Ref<HTMLLIElement>) => {
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in af3e87f

packages/components/src/Menu/Menu.story.tsx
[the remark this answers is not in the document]
mdodgelooker: Oh whoops, I added this comment and then later realized what the change was. Meant to delete it. 🙂

packages/components/src/Menu/MenuList.tsx
mdodgelooker: Works for me... anyone wanna do an audit to see if any menus will look just terrible without being able to specify the width? 😉

ElliotTheHuman: I'll poke around HT to see if anything looks broken by not being able to set the width. Stay tuned!

packages/components/src/Menu/Menu.tsx
mdodgelooker: Oh interesting! I neglected to destructure those props, and I blame typescript. 😉 Apparently it doesn't complain if you spread extra props that are not valid. Then I think it is safer to do what you've done, but to add a type on the array. I'll add an inline suggestion.

packages/components/src/Popover/Popover.tsx
```diff
@@ -45,6 +45,26 @@ const isRenderProp = (
   children: ReactNode | PopoverRenderProp
 ): children is PopoverRenderProp => typeof children === 'function'
 
+export const popoverPropKeys = [
```
mdodgelooker, line 48: If you move this array down below `PopoverProps` and add this type, it'll be slightly safer, though it could still be missing props. I've asked for help on deriving a type that's an exhaustive array, but it's tricky so this I think is good enough for now.

Also I think `escapeWithReference` should be there. My bad I missed it in my `Menu` restructure, but I could think of a situation where you'd want to use it.
```suggestion
export const popoverPropKeys: Array<keyof PopoverProps> = [
  'escapeWithReference',
```

www/src/documentation/components/overlays/menu.mdx
```diff
@@ -61,12 +61,11 @@ To add a tooltip to the menu button, nest a `Tooltip` inside `Menu`.
 ## Placement
 
 Use the `placement` prop to position the overlay that contains the list.
-The `escapeWithReference` and `pin` props allow you to enforce the placement to varying degrees.
+The `pin` props allow you to enforce the placement to varying degrees.
 
 ```jsx
 <Menu
   placement="left"
-  escapeWithReference
```
mdodgelooker, line 69: I think we should add `escapeWithReference` back in – my oversight for removing it.

mdodgelooker (changes requested):

www/src/documentation/components/overlays/menu.mdx
ElliotTheHuman: Discussed over chat, but it looks like `escapeWithReference` was deprecated from `Popover`. Keeping it deleted from `MenuProps` seems like the right call.

packages/components/src/Popover/Popover.tsx
ElliotTheHuman: Discussed over chat, but it looks like `escapeWithReference` was deprecated from `Popover`. Keeping it deleted from `MenuProps` seems like the right call.

Did move down the keys array though to make use of the interface + keyof!

Resolved in ef164e1

packages/components/src/Menu/Menu.tsx
```diff
@@ -46,60 +52,38 @@ export interface MenuProps
...
+    { children, content, id: propsID, listRef, ...props }: MenuProps,
     ref: Ref<any>
   ) => {
+    const popoverProps = popoverPropKeys.reduce((object, key) => {
+      if (props[key] !== undefined) {
+        object[key] = props[key]
+      }
+      return object
+    }, {})
+    const listProps = omitPopoverProps(props, popoverPropKeys)
```
ElliotTheHuman, line 75: Vanilla JS-ifying `popoverProps` was pretty straightforward, but getting `listProps` was a bit trickier. I'm definitely not beholden to my current `omitPopoverProps` implementation so open to any cleaner approaches out there!

packages/components/src/Menu/Menu.tsx
ElliotTheHuman: Cool cool! I left a comment on my recent vanilla JS-ifying attempt. It probably could be better, but couldn't think of anything cleaner for `omitPopoverProps` 🤷 

https://github.com/looker-open-source/components/pull/1906#discussion_r578013729

packages/components/src/Menu/MenuList.tsx
ElliotTheHuman: Discussed over Meet, but adding a `min-width: 12rem` to `MenuList` in particular and removing width-related props was the call iirc.

Resolved in 974aaf7.

packages/components/src/Menu/Menu.tsx
```diff
@@ -46,60 +52,46 @@ export interface MenuProps
...
+    if (props[key] !== undefined) {
+      popoverProps[key] = props[key]
+    }
+    delete allProps[key]
+  })
+
+  const listProps = allProps
+
+  return [popoverProps, listProps]
+}
```
mdodgelooker, line 73: Nice. I think when we do more work on removing/minimizing lodash we'll want to make this a util. I can think of at least all the `Field-` components that would need it (or similar). 🙂

mdodgelooker (approved):
🚢!!!

packages/components/src/Menu/Menu.tsx
```diff
@@ -46,60 +52,46 @@ export interface MenuProps
...
-      ...props
-    }: MenuProps,
+    { children, content, id: propsID, listRef, ...restProps }: MenuProps,
     ref: Ref<any>
   ) => {
+    const [popoverProps, listProps] = partitionMenuProps(
+      restProps,
+      popoverPropKeys
+    )
+
```
ElliotTheHuman, line 84: I really wanted to be able to produce the listProps and popoverProps simultaneously since they're mutually exclusive, so whipped up a function to do so. 

Note: There's technically no reason I need to copy the props object, but it just feels weird to me to delete properties on the `props` object (or mutate it in general) 🤷  I could just be paranoid though.

PR #1910, opened 2021-02-12
fix(Tree): Removed default value on Tree density prop

ElliotTheHuman:
Setting a default value on Tree's `density` doesn't allow context value to trickle down to child Trees (i.e. the `undefinedCoalesce` call will always use the default prop value rather than try to pull from context for a child `Tree`).

### Requirements

- [ ] Documentation updated
- [ ] i18n impacts
- [ ] a11y impacts


lukelooker (approved):
Sold!

PR #1914, opened 2021-02-15
chore: Add JSDoc annotation to props that should be i18n'd

lukelooker:
This PR adds JSDoc call-out to all props should probably be i18n'd upstream.

Also removes a few explicit children`instances that in implicitly added through interface extension of FC generic wrapper.

### DIFF
--- a/config/i18nIgnoredProps.js
+++ b/config/i18nIgnoredProps.js
@@ -25,6 +25,7 @@
  */
 
 const ours = [
+  'align',
   'icon',
   'iconBefore',
   'iconAfter',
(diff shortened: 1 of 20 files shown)


mdodgelooker (changes requested):
The @i18n not isn't not showing up for me in the Storybook docs tab:
image

Also, the hyphen might not be needed?
image

And you've got a few `@i18n-recommended` vs `@i18n`.

lukelooker:
Went with a simpler approach (no `@`).

mdodgelooker (approved):
LGTM! Sadly the line breaks are not respected but what can ya do.
image

PR #1915, opened 2021-02-15
fix(design-tokens): generateTheme now tolerates poorly encoded font stack

lukelooker:
During integration found an issue where `generateTheme` didn't interact well when given more than a single font-face. This change updates `generateTheme` to leverage `sanitizeFontFamily` to make `generateTheme` more tolerant.

This commit also adds `'Noto Sans'` to default fallback font-stack.

### Requirements

- [ ] Documentation updated
- [x] i18n impacts
- [x] a11y impacts


packages/design-tokens/src/utils/theme.test.ts
```diff
@@ -99,13 +99,24 @@ describe('generateTheme', () => {
...
   })
 
+  test('fontFamilies - unquoted faces', () => {
+    const generated = generateTheme(theme, {
+      fontFamilies: {
+        body: 'Open Sans, Noto Sans',
+      },
+    })
+    expect(generated.fonts.body).toMatchInlineSnapshot(
+      `"'Open Sans', 'Noto Sans', 'Noto Sans', 'Noto Sans JP', 'Noto Sans CJK KR', 'Noto Sans Arabic UI', 'Noto Sans Devanagari UI', 'Noto Sans Hebrew', 'Noto Sans Thai UI', Helvetica, Arial, sans-serif"`
```
mdodgelooker, line 116: Is it intentional / ok for `'Noto Sans'` to be in there twice?

packages/design-tokens/src/utils/typography/sanitizeFont.test.ts
```diff
@@ -43,6 +43,18 @@ describe('sanitizeFontFaces', () => {
     expect(sanitizeFontFamily("'Helvetica neue'")).toEqual("'Helvetica neue'")
   })
 
+  it('Several, improperly quoted', () => {
```
mdodgelooker, line 46: Should this be "properly"?
```suggestion
  it('Several, properly quoted', () => {
```

mdodgelooker (changes requested):
Looks good! Just have one question and one test to possibly rename. 🙂

packages/design-tokens/src/utils/theme.test.ts
lukelooker: In this case yes, (they asked for Noto Sans and it's part of the fallback set by default).

I suppose I could try to de-duplicate the array but it's technically valid and I was aiming to make the fewest possible operations on the input to keep things fast so I thin

packages/design-tokens/src/utils/typography/sanitizeFont.test.ts
lukelooker: 🤦  - yes, yes it should be. Fixed!

mdodgelooker (approved):
LGTM!

PR #1922, opened 2021-02-17
feat: Add label to Dialog

kvrmd:
### Requirements

- [x] Documentation updated
- [x] i18n impacts
- [x] a11y impacts

Storybook DialogHeader inherits Dialog id:
image

Screen reader reads labelled Dialog:
image

### DIFF
--- a/packages/components/src/Dialog/DialogContext.ts
+++ b/packages/components/src/Dialog/DialogContext.ts
@@ -29,10 +29,12 @@ import noop from 'lodash/noop'
 
 export interface DialogContextProps {
   closeModal: () => void
+  dialogId: string
 }
 
 const dialogContext: DialogContextProps = {
   closeModal: () => noop,
+  dialogId: '',
 }
 
 export const DialogContext = createContext(dialogContext)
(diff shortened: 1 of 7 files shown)


packages/components/src/Dialog/Layout/DialogHeader.tsx
```diff
@@ -66,12 +66,13 @@ const DialogHeaderLayout: FC<DialogHeaderProps> = ({
   ...props
 }) => {
   const { t } = useTranslation('DialogHeader')
-  const { closeModal } = useContext(DialogContext)
+  const { closeModal, dialogId } = useContext(DialogContext)
   const { id } = props
 
   return (
-    <header {...omitStyledProps(props)}>
+    <div {...omitStyledProps(props)}>
```
kvrmd, line 73: Changed `header` to be `div` here, because the screen reader was using this element as the title of the dialog. Alternatively, I could keep the `header` tag and do something like this:

```tsx
<header aria-labelledby={id ? `${id}-header-heading` : ''}>
  <Heading id={id ? `${id}-header-heading` : ''} />
</header>
```

packages/components/src/Dialog/DialogContext.ts
```diff
@@ -29,10 +29,12 @@ import noop from 'lodash/noop'
 
 export interface DialogContextProps {
   closeModal: () => void
+  dialogId: string
 }
 
 const dialogContext: DialogContextProps = {
   closeModal: () => noop,
+  dialogId: '',
```
kvrmd, line 37: Should I use `id` or `dialogId` here?

PR #1924, opened 2021-02-17
feat: Add label to Dialog

kvrmd:
### Requirements

- [x] Documentation updated
- [x] i18n impacts
- [x] a11y impacts

Storybook DialogHeader inherits Dialog id:
image

Screen reader reads labelled Dialog:
image

### DIFF
--- a/packages/components/src/Dialog/DialogContext.ts
+++ b/packages/components/src/Dialog/DialogContext.ts
@@ -29,10 +29,12 @@ import noop from 'lodash/noop'
 
 export interface DialogContextProps {
   closeModal: () => void
+  dialogId: string
 }
 
 const dialogContext: DialogContextProps = {
   closeModal: () => noop,
+  dialogId: '',
 }
 
 export const DialogContext = createContext(dialogContext)
--- a/packages/components/src/Dialog/DialogContext.ts
+++ b/packages/components/src/Dialog/DialogContext.ts
@@ -29,12 +29,16 @@ import noop from 'lodash/noop'
 
 export interface DialogContextProps {
   closeModal: () => void
-  dialogId: string
+  /**
+   * Used to connect the DialogHeader's title to the Dialog's `aria-labelledby`
+   * @private
+   */
+  id: string
 }
 
 const dialogContext: DialogContextProps = {
   closeModal: () => noop,
-  dialogId: '',
+  id: '',
 }
 
 export const DialogContext = createContext(dialogContext)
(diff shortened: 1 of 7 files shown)


packages/components/src/Dialog/Layout/DialogHeader.tsx
```diff
@@ -66,12 +66,13 @@ const DialogHeaderLayout: FC<DialogHeaderProps> = ({
   ...props
 }) => {
   const { t } = useTranslation('DialogHeader')
-  const { closeModal } = useContext(DialogContext)
+  const { closeModal, dialogId } = useContext(DialogContext)
   const { id } = props
 
   return (
-    <header {...omitStyledProps(props)}>
+    <div {...omitStyledProps(props)}>
```
kvrmd, line 73: Changed `header` to be `div` here, because the screen reader was using this element as the title of the dialog. Alternatively, I could keep the `header` tag and refer to the `Heading` for the `header`'s label:

```tsx
<Dialog aria-labelledby={`${dialogId}-header`} />

<header
  id={dialogId ? `${dialogId}-header` : ''}
  aria-labelledby={id ? `${id}-header-heading` : ''}
>
  <Heading id={id ? `${id}-header-heading` : ''} />
</header>
```

packages/components/src/Dialog/DialogContext.ts
```diff
@@ -29,10 +29,12 @@ import noop from 'lodash/noop'
 
 export interface DialogContextProps {
   closeModal: () => void
+  dialogId: string
 }
 
 const dialogContext: DialogContextProps = {
   closeModal: () => noop,
+  dialogId: '',
```
kvrmd, line 37: Should I use `id` or `dialogId`?

packages/components/src/Dialog/Layout/DialogHeader.tsx
kvrmd: I updated this to keep the `header` element for semantics.

packages/components/src/Dialog/DialogContext.ts
lukelooker: I think I'd prefer `id` since the context consumer can alias if name collisions are a concern and the terse version is nice

lukelooker, line 37: We haven't been super consistent about prop-docs but we're _trying_ to get better. I'd also like to mark this as `@private` since I don't (at this point at least) want to encourage more use of this outside this since use-case.

```suggestion
  /**
   * Used to connect the DialogHeader's title to the Dialog's `aria-labelledby`
   * @private
   */
  id: string
```

lukelooker (changes requested):
Very minor change about `dialogId` to `id` and then this is good to go.

My apologies on being slow to review this - I'll be pushing through lots of reviews today!

lukelooker (approved):
LGTM!

PR #1927, opened 2021-02-18
feat: Focus on first tabbable node when Dialog opens

kvrmd:
### Requirements

- [x] Documentation updated
- [x] i18n impacts
- [x] a11y impacts

In this PR, I update FocusTrap utils to focus on the first tabbable element, instead of the dialog element.

Few things to note:
1. VoiceOver default arrow-key navigation gets stuck when the dialog element is focused. This seems to be because dialog element is treated as the last focusable element -> you'd have to navigate backwards to read the dialog content.
2. The [w3c aria-practices page](https://w3c.github.io/aria-practices/#dialog_modal) says to focus on the first focusable element when the Dialog is opened, with some exceptions. It's worth mentioning that there are conflicting opinions on when the exceptions apply.

For 2, I'm thinking we could add a `initialFocusStrategy` prop to `Dialog` to specify the fallback-logic when determining what element to focus when the Dialog opens. By default it could be: `['autofocus', 'firstTabbableNode', 'header', 'dialog']`.

```
  initialFocusStrategy: Array<'autofocus' | 'firstTabbableNode' | 'header' | 'dialog'>
```


packages/components/src/Portal/Portal.tsx
```diff
@@ -78,7 +78,7 @@ export const Portal = forwardRef(
       }
     }, [el])
 
-    const content = <InvisiBox ref={ref} tabIndex={-1} {...props} />
```
kvrmd, line 81: @lukelooker Is there a reason we add `tabIndex={-1}` to this component? This is causing VoiceOver to navigate to this element.

lukelooker: 👍  This was a legacy thing (pre-`*Surface` and should go away)

packages/components-providers/src/FocusTrap/utils.ts
```diff
@@ -81,11 +81,15 @@ export const activateFocusTrap = ({
         '[data-autofocus="true"]'
       ) as HTMLElement
 
-      // In the absence of autofocus, the surface will have initial focus
+      // Without autofocus, fallback to the first tabbable node.
+      const tabbableNodes = tabbable(element)
```
lukelooker, line 85: So @mdodgelooker discussed this some. We _used_ to have a strategy very much like this and it led to some angst when things were unexpectedly or annoyingly focused. We've mitigated many of the key issues there (introducing `:focus-visible` pattern, making `DialogHeader`'s close icon `tabIndex={-1]}` etc...

That said we want to avoid focusing on a button within the DialogContent unless it's our only option so we want to go with a slightly more complex strategy here. `firstTabbableNode` should be (in this order)

1. `inputs[0]` (`input, textarea, select`) - find the first input
2. `footer > tabbable[0]` - if that fails, find the first tabbable item in the footer
3. `tabbable[0]` - if all else fails, find the first tabbable thing

lukelooker (changes requested):
Wrote up notes on the `firstTabbableNode` strategy

packages/components-providers/src/FocusTrap/utils.ts
```diff
@@ -81,11 +81,25 @@ export const activateFocusTrap = ({
...
-      // In the absence of autofocus, the surface will have initial focus
+      // Without autofocus, fallback to a tabbable node by priority, if one exists.
+      const firstInputElement = element.querySelector('input, textarea, select')
+      const footerElement = element.querySelector('footer')
+      const firstTabbableFooterElement = footerElement
+        ? tabbable(footerElement)[0]
+        : null
+      const firstTabbableElement = tabbable(element)[0]
+      const prioritizedTabbableElement =
+        firstInputElement || firstTabbableFooterElement || firstTabbableElement
```
kvrmd, line 92: @lukelooker + @mdodgelooker Here's the meat of the initial focus logic change.

packages/components/src/utils/useFocusTrap.test.tsx
```diff
@@ -84,7 +84,7 @@ describe('useFocusTrap', () => {
   describe('initial focus', () => {
     test('focus starts on surface', async () => {
       render(
-        <FocusTrapComponent>
+        <FocusTrapComponent hideClose>
```
kvrmd, line 87: Added `hideClose` to prevent the close button from being initially focused. Alternatively, I could update the test to `expect` the close button being focused as opposed to the surface.

packages/components/src/utils/useFocusTrap.test.tsx
```diff
@@ -98,16 +98,79 @@ describe('useFocusTrap', () => {
       renderWithTheme(
         <FocusTrapComponent>
           <Surface>
-            <FieldText label="Text field" autoFocus />
+            <FieldText label="Text field A" />
+            <FieldText label="Text field B" autoFocus />
```
kvrmd, line 102: Added a 2nd input to ensure `autoFocus` overrides the initial tabbable element logic.

kvrmd (commented):
I couldn't find any docs to update. Please let me know if I missed it 🙏

packages/components-providers/src/FocusTrap/utils.ts
```diff
@@ -81,11 +81,25 @@ export const activateFocusTrap = ({
...
       ) as HTMLElement
 
-      // In the absence of autofocus, the surface will have initial focus
+      // Without autofocus, fallback to a tabbable node by priority, if one exists.
+      const firstInputElement = element.querySelector('input, textarea, select')
+      const footerElement = element.querySelector('footer')
+      const firstTabbableFooterElement = footerElement
+        ? tabbable(footerElement)[0]
+        : null
+      const firstTabbableElement = tabbable(element)[0]
```
lukelooker, line 90: To otimize performance we probably want to check if a condition before has returned a result and only keep querying if a tabbable element hasn't already been found.

packages/components/src/utils/useFocusTrap.test.tsx
lukelooker: I think this actually points out a minor defect in the `FocusTrapComponent` where it doesn't emulate the structure of our Dialog layout components

I'd actually propose restructuring the `Inner` to like the example below - basically introducing `tabIndex={-1}` to the "early" close option and introducing a `footer` wrapper.

```
const Inner: FC<TestProps> = ({ children, clickOutsideDeactivates }) => {
  const [, ref] = useFocusTrap({ options: { clickOutsideDeactivates } })
  const { value, setOff, toggle } = useToggle()
  return (
    <>
      {value && (
        <div ref={ref}>
          {children}
          <button tabIndex={-1} onClick={setOff}>
            Close
          </button>
        </div>
      )}
      <footer>
        <button onClick={toggle}>toggle</button>
        <button onClick={setOff}>Another button</button>
      </footer>
    </>
  )
}
```

lukelooker (changes requested):
I think there's a fairly straightforward performance optimization that could be done on the `prioritizedTabbableElement` assignment and some thoughts on the associated test.

packages/components-providers/src/FocusTrap/utils.ts
```diff
@@ -70,31 +70,58 @@ export const activateFocusTrap = ({
...
+    if (surfaceElement) {
+      return surfaceElement
+    }
+
+    // default to element
+    return element
+  }
+
+  const getInitialFocusNode = () => {
+    const node = getInitialFocusNodeByPriority()
```
kvrmd, line 118: I refactored this a bit to call a helper function that does early-returns to avoid unnecessary queries.

packages/components/src/utils/useFocusTrap.test.tsx
```diff
@@ -42,22 +42,19 @@ afterEach(() => {
...
   const { value, setOff, toggle } = useToggle()
   return (
     <>
       {value && (
         <div ref={ref}>
           {children}
-          {!hideClose && <button onClick={setOff}>Close</button>}
+          <button tabIndex={-1} onClick={setOff}>
+            Close
+          </button>
```
kvrmd, line 57: @lukelooker fyi I didn't add a tabbable button here, so the test for focusing the surface would still work.

lukelooker (approved):
Thanks so much for taking this on - this updated behavior is a really nice improvement. :)

Ship it!

PR #1931, opened 2021-02-18
chore(List): Exported listItemDimensions util function

ElliotTheHuman:
If your `ListItem` / `TreeItem` / `MenuItem` has string children, those strings are automatically be styled according to our density dimensions objects. However, if a non-string ReactNode is passed in, those stylings don't take affect (since we don't wrap non-string ReactNode children in a Truncate + Text wrapper).

Exposing the `listItemDimensions` function provides a way for non-string ReactNode children to stay consistent with a given density's intended dimensions.

### DIFF
--- a/packages/components/src/List/index.ts
+++ b/packages/components/src/List/index.ts
@@ -26,3 +26,4 @@
 
 export * from './List'
 export * from './ListItem'
+export { listItemDimensions } from './utils'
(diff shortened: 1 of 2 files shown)


lukelooker (approved):
Sold!

lukelooker (changes requested):
ack! Approved too quick. The `listItemDimensions` should have a JSDod especially now that it's exported for external use

lukelooker (approved):
LGTM

PR #1933, opened 2021-02-18
fix(ListItem): Removed reset from ListItemWrapper

ElliotTheHuman:
Currently, the `${reset}` CSS is removing vertical and right padding, which makes the internal button / a element within `ListItem` much shorter than it should be. Removing the `${reset}` from `ListItemWrapper` should resolve this issue and allow us to proceed with the HT upgrade PR.

**Note:** The relevant snapshot changes to `List` and `Menu` have to do with the "detail" string fonts changing in various snapshots from what I can tell.

### Requirements

  - [ ] yes
  - [x] not applicable

  - [ ] yes
  - [x] not applicable


lukelooker (dismissed):
I removed the snapshot updates because they all looked spurious and we'll see if CI passes without those. Assuming it does 👍  from me.

lukelooker (approved):
Woot! LGTM

PR #1938, opened 2021-02-19
fix(ListItem): Set background of disabled ListItemWrapper to transparent

ElliotTheHuman:
In the even that `theme.colors.background` does not match the actual background color of a ListItem's container (like with the "Copy my dash" modal), hovering over a disabled item produces a hover background:
Screen Shot 2021-02-18 at 7 03 01 PM

Setting the `background-color` to "transparent" when a `ListItem` is disabled and hovered should get the look we want (i.e. no hover background when an item is disabled).

### Requirements

  - [ ] yes
  - [ ] not applicable

  - [ ] yes
  - [ ] not applicable

### DIFF
--- a/packages/components/src/List/ListItemWrapper.tsx
+++ b/packages/components/src/List/ListItemWrapper.tsx
@@ -77,9 +77,6 @@ export const ListItemWrapper = styled(ListItemWrapperInternal)`
   min-height: ${({ height }) => height}px;
   outline: none;
   text-decoration: none;
-  transition: ${({ theme: { easings, transitions } }) =>
-    `background ${transitions.quick}ms ${easings.ease},
-  color ${transitions.quick}ms ${easings.ease}`};
 
   & > button,
   & > a {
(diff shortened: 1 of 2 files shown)


packages/components/src/List/ListItemWrapper.tsx
```diff
@@ -157,7 +157,7 @@ export const ListItemWrapper = styled(ListItemWrapperInternal)`
     }
 
     &:hover {
-      background: ${({ theme: { colors } }) => colors.background};
+      background: transparent;
```
lukelooker, line 160: Pondering if just removing this althogether would work?

```suggestion
```

ElliotTheHuman: Tried deleting it on my local HT branch in the node_modules directory and it seemed to work just fine! Opting to delete it since that's probably the better solution that explicitly leaving `background: transparent;`

lukelooker (approved):
LGTM

PR #1939, opened 2021-02-19
docs: updated documentation to display on validation message

laraharrow:

### Requirements

  - [ ] yes
  - [x] not applicable

  - [x] yes
  - [ ] not applicable


laraharrow:
there was a question on ticket [1598](https://looker.atlassian.net/browse/LENS-1598) about the validation not being triggered at first render.
I believe that it shouldn't happen as it should be triggered when it get focus and the behavior passed is validated.

www/src/documentation/components/fields/field-text.mdx
```diff
@@ -18,7 +18,7 @@ The `<FieldText />` component is composed of an `<InputText />` component and a
 
 The `<FieldText />` component accepts `validationMessage` as a property. This is an object with properties `type` and `message`.
 
-See `ValidationMessage` [here](/#!/ValidationMessage)
+See `ValidationMessage` [here](https://github.com/looker-open-source/components/blob/main/packages/components/src/Form/ValidationMessage/ValidationMessage.tsx)
```
dbchristopher, line 21: What is the intent of linking to the component source here? Honestly I think you could just delete this line, as I don't think it clarifies anything

dbchristopher (changes requested):
Looks like `components/forms/demos/index.ts` needs to be deleted now. Gatsby won't start because it's still trying to export that removed demo:

Screen Shot 2021-02-19 at 10 30 07 AM

Good call moving the demo source inline though!

www/src/documentation/components/fields/field-text.mdx
laraharrow: So yeah that is not the final solution. But it solves the problem of a broken link. I'm going to create a new ticket to specifically create a docs with the appropriated documentation to `ValidationMessage`.

dbchristopher: works for me!

dbchristopher (approved):

PR #1950, opened 2021-02-22
feat(DataTable): Allow filters to be specified by composition rather than complex prop object

lukelooker:
### Requirements

  - [x] yes
  - [ ] not applicable

  - [ ] yes
  - [x] not applicable

### DIFF
--- a/packages/components/src/DataTable/types.ts
+++ b/packages/components/src/DataTable/types.ts
@@ -24,8 +24,8 @@
 
  */
 
-import { ReactNode } from 'react'
-import { FieldFilter } from '../Form/Inputs/InputFilters'
+import { ReactElement, ReactNode } from 'react'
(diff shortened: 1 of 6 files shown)


packages/components/src/DataTable/types.ts
```diff
@@ -56,7 +56,7 @@ export interface DataTableProps {
    * Object containing list of filters the user can select from
    * and an onFilter callback to update the filters
    **/
-  filterConfig?: FilterConfig
+  filters?: ReactElement<InputFiltersProps>
```
roninjin10, line 59: This is interesting.  I normally would specify this as a React.ReactNode.  I have never considered specifying a speicific element like this.  I find it interesting because it is kinda using typescript to still guarantee the right type of component gets passed in here.

roninjin10 (approved):

laraharrow (approved):
LGTM🚢

PR #1956, opened 2021-02-23
fix: Combobox aria-owns targets ul instead of input

kvrmd:
Referring to the [w3 listbox specs](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html), the `aria-owns` prop of the `[role="combobox"]` element should target the `[role="listbox"]` element. Previously, the input was being targetted.

This change actually fixes the screen reader not reading out the options using VoiceOver.

### Requirements

  - [ ] yes
  - [x] not applicable

  - [ ] yes
  - [x] not applicable


mdodgelooker (changes requested):
This looks good! I think you also need to update the `id` in `ComboboxMultiInput` though.

mdodgelooker (approved):

lukelooker (approved):
LGTM

PR #1957, opened 2021-02-23
refactor: Remove react-resize-detector dependency

dbchristopher:
Use useResize hook to toggle the overflow styling on `DialogContent`, and remove react-resize-detector as a dependency.

### Requirements

  - [ ] yes
  - [x] not applicable

  - [ ] yes
  - [x] not applicable

### DIFF
--- a/packages/components/src/Dialog/Layout/DialogContent.tsx
+++ b/packages/components/src/Dialog/Layout/DialogContent.tsx
@@ -35,9 +35,8 @@ import {
   pickStyledProps,
 } from '@looker/design-tokens'
 import React, { FC, useRef, useState, useEffect } from 'react'
-import ReactResizeDetector from 'react-resize-detector'
 import styled from 'styled-components'
-import omit from 'lodash/omit'
+import { useResize } from '../../utils'
 
 export interface DialogContentProps
   extends LayoutProps,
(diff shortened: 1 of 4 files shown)


packages/components/src/Dialog/Layout/DialogContent.tsx
```diff
@@ -56,76 +56,63 @@ export interface DialogContentProps
...
+    }
+  }
+
+  useResize(internalRef.current, handleResize)
 
   useEffect(() => {
-    /**
-     * Once you overflow, you never go back (tough luck chuck)
-     */
-    if (!overflow) {
```
dbchristopher, line 78: I wasn't sure the reason for this limitation. I couldn't think of any technical reason to avoid disabling the overflow state if the content changes and can fit within available area. Happy to reinstate that behavior though!

packages/components/src/Dialog/Layout/DialogContent.tsx
```diff
@@ -56,76 +56,63 @@ export interface DialogContentProps
...
-    <div
-      className={overflow ? `overflow ${className}` : className}
+    <InnerDialogContent
+      hasOverflow={hasOverflow}
       ref={internalRef}
-      {...omit(omitStyledProps(props), ['renderedHeight'])}
+      px={['medium', 'xlarge']}
+      pb={hasOverflow || !!hasFooter ? 'large' : 'xxxsmall'}
+      pt={hasOverflow || !!hasHeader ? 'large' : 'xxxsmall'}
+      {...pickStyledProps(props)}
```
dbchristopher, line 92: I couldn't figure out any reason to have the wrapper div and the `Inner` div be two different components. I merged them to simplify the markup

lukelooker: Hunting through git blame history I can't find a _good_ reason we did this – I remember we added it explicitly at one point but I think we've since evolved out of the need for this.

lukelooker (approved):
Works great in all the evergreen browsers and code looks good. 

I noticed that's it slightly quirky in IE11 but it does appear to be any worse than before (if you resize the window to trigger overflow and then resize it again to remove overflow the `useResize` hook doesn't seem to detect the change:

image
 
That said, this seems like a very edgy-edge case and it also appears to be a problem in the previous implementation.

PR #1959, opened 2021-02-24
fix(FocusTrap): Overly aggressive return focus behavior

mdodgelooker:
Classic `useEffect` dependency mis-use.

[This PR](https://github.com/looker-open-source/components/pull/1886) added a `returnFocusRef` option to `useFocusTrap` in such a way that the `useEffect` in `useTrapStack` would run on every render. For a nested `disabled` focus trap (e.g. `Select`* inside a `Popover`) it would re-activate the a previous focus trap too early (and repeatedly). This PR adds a `useMemo` to fix that.

For the life of me I could not reproduce the bug in jest. But to test on https://looker-open-source.github.io/components/latest/storybook/?path=/story/fieldselect--select-demo try opening the 2nd `Select` as soon as you open the `Popover`. The 1st `Select` will keep focus as the focus trap repeatedly deactivates & re-activates. (See screen captures below.)

*`Select` has a "disabled" focus trap – i.e. it doesn't trap focus to the list because focus remains on the input. But it does disable any existing focus trap, in case focus _does_ need to move to the list, like for dragging the scrollbar thumb, and re-enables a previous focus trap when it closes.

#### Issue:
75f646a6-655e-45ec-b7e2-d36761833ca5
#### Fixed:
456541dd-c682-4c8f-8857-6ff1c9246725

### Requirements

  - [ ] yes
  - [x] not applicable

  - [ ] yes
  - [x] not applicable

### DIFF
--- a/packages/components-providers/src/TrapStack/TrapStackProvider.tsx
+++ b/packages/components-providers/src/TrapStack/TrapStackProvider.tsx
@@ -48,7 +48,7 @@ export const TrapStackProvider = <O,>({
   const registeredTrapsRef = useRef<TrapMap<O>>({})
   // Stores the current trap (element) where scrolling is allowed
   // null if no trap is active
-  const activeTrapRef = useRef<HTMLElement | null>(null)
+  const activeTrapRef = useRef<HTMLElement>()
   // Stores the callback to remove the trap behavior
   const deactivateRef = useRef<() => void>()
 
@@ -64,7 +64,7 @@ export const TrapStackProvider = <O,>({
       if (newTrap?.element !== activeTrapRef.current) {
         // Disable the existing trap and update the activeTrapRef
         // (whether there's a new trap or not)
-        activeTrapRef.current = newTrap?.element || null
+        activeTrapRef.current = newTrap?.element
         deactivateRef.current?.()
         deactivateRef.current = undefined
         // If there's a new trap, activate it and
@@ -78,7 +78,7 @@ export const TrapStackProvider = <O,>({
     const disableCurrentTrap = () => {
       deactivateRef.current?.()
       deactivateRef.current = undefined
-      activeTrapRef.current = null
+      activeTrapRef.current = undefined
     }
 
     const addTrap = (id: string, trap: Trap<O>) => {
--- a/packages/components-providers/src/TrapStack/TrapStackProvider.tsx
+++ b/packages/components-providers/src/TrapStack/TrapStackProvider.tsx
@@ -47,16 +47,16 @@ export const TrapStackProvider = <O,>({
   // (map of ids to elements that have traps)
   const registeredTrapsRef = useRef<TrapMap<O>>({})
   // Stores the current trap (element) where scrolling is allowed
-  // null if no trap is active
+  // undefined if no trap is active
   const activeTrapRef = useRef<HTMLElement>()
   // Stores the callback to remove the trap behavior
   const deactivateRef = useRef<() => void>()
 
   // Create the context value
   const value = useMemo(() => {
-    const getTrap = (id?: string): Trap<O> | null => {
+    const getTrap = (id?: string): Trap<O> | undefined => {
       const registeredTraps = registeredTrapsRef.current
-      return id ? registeredTraps[id] || null : getActiveTrap(registeredTraps)
+      return id ? registeredTraps[id] : getActiveTrap(registeredTraps)
     }
 
     const enableCurrentTrap = () => {
--- a/packages/components-providers/src/TrapStack/TrapStackProvider.tsx
+++ b/packages/components-providers/src/TrapStack/TrapStackProvider.tsx
@@ -24,7 +24,6 @@
 
  */
 
-import noop from 'lodash/noop'
 import React, { Context, ReactNode, useRef, useMemo } from 'react'
 import { Trap, TrapStackContextProps, TrapMap } from './types'
 import { getActiveTrap } from './utils'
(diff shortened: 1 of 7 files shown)


packages/components-providers/src/TrapStack/TrapStackProvider.tsx
```diff
@@ -62,11 +61,12 @@ export const TrapStackProvider = <O,>({
...
     const enableCurrentTrap = () => {
       const newTrap = getTrap()
-      if (newTrap !== activeTrapRef.current) {
+      if (newTrap?.element !== activeTrapRef.current) {
         // Disable the existing trap and update the activeTrapRef
         // (whether there's a new trap or not)
         activeTrapRef.current = newTrap?.element || null
-        deactivateRef.current()
+        deactivateRef.current?.()
+        deactivateRef.current = undefined
```
mdodgelooker, line 69: Part of the issue was the deactivate function getting called repeatedly. I realized once it's called, it should never get called again until & unless there's a new instance from a trap getting activated. (I also fixed the `useEffect` issue that was causing the repeated calls – double fix!)

packages/components/src/Dialog/useDialog.tsx
```diff
@@ -161,9 +161,7 @@ export const useDialog = ({
       ? controlledSetOpen
       : setUncontrolledIsOpen
 
-  const [, focusRef] = useFocusTrap({
-    options: { clickOutsideDeactivates: true },
-  })
+  const [, focusRef] = useFocusTrap({ clickOutsideDeactivates: true })
```
mdodgelooker, line 164: Less nesting is better for the `useEffect` dependency array (see changes in `useFocusTrap`).

packages/components/src/utils/useFocusTrap.test.tsx
```diff
@@ -108,7 +108,7 @@ describe('useFocusTrap', () => {
       )
     })
 
-    describe('focus starts on tabbable element by priority', async () => {
+    describe('focus starts on tabbable element by priority', () => {
```
mdodgelooker, line 111: I was getting a jest error from this.

packages/components/src/utils/useFocusTrap.ts
```diff
@@ -24,22 +24,26 @@
...
  */
 
-import { useRef } from 'react'
+import { useMemo, useRef } from 'react'
 import {
   FocusTrapContext,
   FocusTrapOptions,
 } from '@looker/components-providers'
-import { useTrapStack, UseTrapStackProps } from './useTrapStack'
+import { useTrapStack, UseTrapStackBaseProps } from './useTrapStack'
```
mdodgelooker, line 32: Some type changes to support less nesting on the props.

packages/components/src/utils/useFocusTrap.ts
```diff
@@ -24,22 +24,26 @@
...
   const returnFocusRef = useRef<Element>(null)
+  const options = useMemo(() => ({ clickOutsideDeactivates, returnFocusRef }), [
+    returnFocusRef,
+    clickOutsideDeactivates,
+  ])
   return useTrapStack<E, FocusTrapOptions>({
     context: FocusTrapContext,
     // If options.returnFocusRef is set, it will override this one
-    options: { returnFocusRef, ...options },
+    options,
```
mdodgelooker, line 46: `options` was in the `useEffect` dependency array in `useTrapStack` and since this line was making it new every render, the `useEffect` was getting run every render.

packages/components-providers/src/TrapStack/TrapStackProvider.tsx
```diff
@@ -76,8 +76,8 @@ export const TrapStackProvider = <O,>({
     }
 
     const disableCurrentTrap = () => {
-      deactivateRef.current()
-      deactivateRef.current = noop
+      deactivateRef.current?.()
+      deactivateRef.current = undefined
       activeTrapRef.current = null
```
kvrmd, line 81: Any reason you set `deactiveRef.current` differently from `activeTrapRef.current`? Can both be set to null/undefined?

packages/components-providers/src/TrapStack/TrapStackProvider.tsx
```diff
@@ -62,11 +61,12 @@ export const TrapStackProvider = <O,>({
 
     const enableCurrentTrap = () => {
       const newTrap = getTrap()
-      if (newTrap !== activeTrapRef.current) {
+      if (newTrap?.element !== activeTrapRef.current) {
```
kvrmd, line 64: This seems highly unlikely, but I'll still ask. Is it possible for both `newTrap?.element` and `activeTrapRef.current` to be equally falsy (i.g. both undefined/null). If so, would that cause issues when `deactivateRef.current()` is called?

kvrmd (approved):
Tested and fix looks good

lukelooker (dismissed):
Jez... that's fun to repro. 

Code looks good and I _think_ I saw the repro / fix in the Storybook example :)

packages/components-providers/src/TrapStack/TrapStackProvider.tsx
mdodgelooker: Good call @kvrmd – I'm not sure why I originally had so many `null`s in here, but they're not necessary and I've converted them all to `undefined`s.

lukelooker (approved):

kvrmd (approved):

packages/components-providers/src/TrapStack/TrapStackProvider.tsx
mdodgelooker: Good q – I forgot to add a comment here explaining. Before this change it was super bad, like you're saying – the condition was only false when both `newTrap` and `activeRef.current` were `null` but it was supposed to be false anytime the elements themselves were the same (or both `null`). So `deacivate.current` was running way too much. 

Now the condition will be true and call `deactivate.current` if `newTrap` is `undefined` _or_ if `activeTrapRef.current` is undefined (but then `deactivate.current` will also be undefined) _or_ if the elements are not a match.

PR #1961, opened 2021-02-24
fix(DataTable): Sticky "columns" should display a bottom-border (firefox)

laraharrow:
**problem:** if page has `border-collapse: collapse;` as a global css rule. It was causing the border-bottom of the sticky first column to not display.

To see the error on PDT open the link below on Firefox
https://lenslatest.dev.looker.com/admin/pdts?connection=All%20Connections

To see the problem locally add this to: `storybook/storybook/preview-head.html`
this will make the `border-collapse` rule be placed as global to storybook.
```
<style>e
table { border-collapse: collapse }
</style>
```
then open storybook on Firefox and open the `DataTable - select-active-items` example.

### Requirements

  - [ ] yes
  - [x] not applicable

  - [ ] yes
  - [x] not applicable

### DIFF
--- a/packages/components/src/DataTable/Table.tsx
+++ b/packages/components/src/DataTable/Table.tsx
@@ -178,9 +178,9 @@ const stickyColumns = css<TableProps>`
 `
 
 export const Table = styled(TableLayout)`
+  border-collapse: initial;
   border-spacing: 0;
   font-size: ${({ theme }) => theme.fontSizes.small};
-  height: initial;
   line-height: 1;
   width: 100%;
 


laraharrow:
Here is the fix:
https://user-images.githubusercontent.com/23616264/109071940-3a9f4880-76a9-11eb-8ae5-d8451a904a34.mov

packages/components/src/DataTable/Table.tsx
```diff
@@ -178,9 +178,9 @@ const stickyColumns = css<TableProps>`
 `
 
 export const Table = styled(TableLayout)`
+  border-collapse: initial;
   border-spacing: 0;
   font-size: ${({ theme }) => theme.fontSizes.small};
-  height: initial;
```
lukelooker, line 183: This is related to the Safari fix (#1965), correct? If so, can it get moved over there to keep git-history linked up correctly?

lukelooker (changes requested):

packages/components/src/DataTable/Table.tsx
laraharrow: this is out of date

lukelooker (approved):
LGTM! Ship it!

PR #1962, opened 2021-02-24
refactor(Tree): Replaced useHovered with simpleHandlers

ElliotTheHuman:
`useHovered` was designed specifically for the "hidden element revealed when hovering over some other element" pattern. In `Tree`'s case, we just need simple mouseEnter and mouseLeave handlers to track hover state for styling purposes.

**Note:** Needed to update some code in `AccordionDisclosure` since it had CompatibleHTMLProps in its interface, but didn't actually pass those down to underlying DOM elements.

### Requirements

  - [ ] yes
  - [ ] not applicable

  - [ ] yes
  - [ ] not applicable

### DIFF
--- a/packages/components/src/Accordion/AccordionDisclosure.tsx
+++ b/packages/components/src/Accordion/AccordionDisclosure.tsx
@@ -27,6 +27,7 @@
 import React, {
   FC,
   KeyboardEvent,
+  MouseEvent,
   useContext,
   Ref,
   forwardRef,
@@ -39,7 +40,6 @@ import {
   CompatibleHTMLProps,
   padding,
   PaddingProps,
-  pickStyledProps,
   shouldForwardProp,
   TextColorProps,
   color as colorStyleFn,
@@ -59,7 +59,7 @@ export interface AccordionDisclosureProps
 }
 
 const AccordionDisclosureInternal: FC<AccordionDisclosureProps> = forwardRef(
-  ({ children, className, ...props }, ref) => {
+  ({ children, className, onClick, onKeyDown, onKeyUp, ...props }, ref) => {
     const [isFocusVisible, setFocusVisible] = useState(false)
     const {
       accordionContentId,
@@ -79,19 +79,23 @@ const AccordionDisclosureInternal: FC<AccordionDisclosureProps> = forwardRef(
     }
 
     const handleKeyDown = (event: KeyboardEvent<HTMLElement>) => {
-      if (event.keyCode === 13) {
-        handleToggle()
-      }
+      if (event.key === 'Enter') handleToggle()
+
+      onKeyDown && onKeyDown(event)
     }
 
     const handleKeyUp = (event: KeyboardEvent<HTMLElement>) => {
-      if (event.keyCode === 9 && event.currentTarget === event.target)
+      if (event.key === 'Tab' && event.currentTarget === event.target)
         setFocusVisible(true)
+
+      onKeyUp && onKeyUp(event)
     }
 
-    const handleClick = () => {
+    const handleClick = (event: MouseEvent<HTMLElement>) => {
       setFocusVisible(false)
       handleToggle()
+
+      onClick && onClick(event)
     }
 
     const handleBlur = () => {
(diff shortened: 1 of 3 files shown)


packages/components/src/Accordion/AccordionDisclosure.tsx
```diff
@@ -79,19 +79,23 @@ const AccordionDisclosureInternal: FC<AccordionDisclosureProps> = forwardRef(
     }
 
     const handleKeyDown = (event: KeyboardEvent<HTMLElement>) => {
-      if (event.keyCode === 13) {
-        handleToggle()
-      }
+      if (event.key === 'Enter') handleToggle()
+
+      onKeyDown && onKeyDown(event)
     }
```
lukelooker, line 85: Could `useWrapEvent` be used here with? E.g.:

```
const handleKeyDown = (event: KeyboardEvent<HTMLElement>) => 
  useWrapEvent((event) => { event.key === 'Enter' && handleToggle() }, onKeyDown)
```

lukelooker (commented):
Two questions but behavior / code looks good overall.

packages/components/src/Accordion/AccordionDisclosure.tsx
```diff
@@ -79,19 +79,23 @@ const AccordionDisclosureInternal: FC<AccordionDisclosureProps> = forwardRef(
...
 
     const handleKeyUp = (event: KeyboardEvent<HTMLElement>) => {
-      if (event.keyCode === 9 && event.currentTarget === event.target)
+      if (event.key === 'Tab' && event.currentTarget === event.target)
         setFocusVisible(true)
+
+      onKeyUp && onKeyUp(event)
```
lukelooker, line 91: Would be good to have a test that assigns `onKeyUp` & onClick` and `onBlur` and verifies that the passed in callbacks are getting called properly.kj

lukelooker (changes requested):
Realized this begs for a quick test coverage pass of `on*` handlers getting passed through.

packages/components/src/Accordion/AccordionDisclosure.tsx
```diff
@@ -112,7 +116,7 @@ const AccordionDisclosureInternal: FC<AccordionDisclosureProps> = forwardRef(
         onKeyUp={handleKeyUp}
         ref={ref}
         tabIndex={0}
-        {...pickStyledProps(props)}
+        {...props}
```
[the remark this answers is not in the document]
ElliotTheHuman: All of the CompatibleHTMLProps land on the `AccordionDisclosureStyle` div, so (iiuc) using `pickStyledProps` here will keep HTML props from reaching the an underlying div.

That being said, I could see an argument for renaming `AccordionDisclosureStyle` to `AccordionDisclosureWrapper` since it's technically doing more than styled. Thoughts?

ElliotTheHuman:
Future TODO for myself: Add CompatibleHTMLProps to `Accordion` during the `AccordionDisclosure` refactor since disclosure will no longer be exposed to users.

They'll have to pass any HTML props for the underlying disclosure div to `Accordion`, who then passes it to the internal `AccordionDisclosure.

packages/components/src/Accordion/AccordionDisclosure.tsx
ElliotTheHuman: Good call. Resolved in 8779501.

packages/components/src/Accordion/AccordionDisclosure.tsx
ElliotTheHuman: We could! And should.

Resolved in 8779501.

packages/components/src/Accordion/AccordionDisclosure.tsx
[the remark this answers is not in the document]
lukelooker: 👍  That makes sense. I guess this change has the possibility of introducing breakage downstream if a developer was assigning a prop that we previously ignored but in general I think that's acceptable since the previous behavior (ignoring domProps) wasn't intuitive and the typing made it look like it _should_ work to assign those.

As to renaming - we're not super consistent with our internal component names so I wouldn't worry about it within this PR. It probably would be worth doing an audit of our naming conventions and seeing if we can come up with some better conventions standards at some point (I'll make a note on our tech-debt log)

lukelooker (approved):
LGTM! 

Great work :)

PR #1964, opened 2021-02-25
fix: InputDate & InputDateRange accessibility improvements

kvrmd:
I'm addressing 2 concerns related to a11y with date picker component. Both InputDate and InputRangeDate are affected.

1. Read out the Month+Year, currently displayed on the calendar when it's changed (either via Previous/Next month buttons or Up/Down to change year).
2. Add labels to input texts.

InputDate:
1. Added `aria-live="polite"` to `CalendarNav` makes VO read the new Month+Year nicely.
2. No change. Already labelled.

InputRangeDate:
1. Multiple `aria-live` doesn't seem to work with VO. Only the last Month+Year is mentioned. I need some suggestions on how to address this.
2. Added `aria-labelledby` to both the Start and End date input texts.

### Requirements

  - [ ] yes
  - [x] not applicable

  - [ ] yes
  - [x] not applicable

### DIFF
--- a/packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
+++ b/packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
@@ -40,6 +40,7 @@ import max from 'lodash/max'
 import isEmpty from 'lodash/isEmpty'
 import isEqual from 'lodash/isEqual'
 import values from 'lodash/values'
+import { useTranslation } from 'react-i18next'
 import { VisuallyHidden } from '../../../VisuallyHidden'
 import { Icon } from '../../../Icon'
 import { ValidationType } from '../../ValidationMessage'
@@ -65,6 +66,7 @@ import { useID } from '../../../utils/useID'
 import { useReadOnlyWarn } from '../../../utils/useReadOnlyWarn'
 
 export interface InputDateRangeProps {
+  'aria-labelledby'?: string
   disabled?: boolean
   dateStringLocale?: Locales
   defaultValue?: Partial<RangeModifier>
@@ -139,6 +141,7 @@ const isDateRangeInView = (
 export const InputDateRange: FC<InputDateRangeProps> = forwardRef(
   (
     {
+      'aria-labelledby': ariaLabelledby,
       dateStringLocale,
       defaultValue = {},
       disabled,
@@ -354,6 +357,10 @@ export const InputDateRange: FC<InputDateRangeProps> = forwardRef(
       setViewMonth(transformMonth(month, viewMonthDiff))
     }
 
+    const { t } = useTranslation('InputDateRange')
+    const startDateLabelledby = `startDate-labelledby-${id}`
+    const endDateLabelledby = `startDate-labelledby-${id}`
+
     const _formatMonthTitle = formatMonthTitle(localization)
     const monthTitle = `${_formatMonthTitle(viewMonth)} ${_formatMonthTitle(
       viewNextMonth
@@ -369,6 +376,9 @@ export const InputDateRange: FC<InputDateRangeProps> = forwardRef(
           }
         >
           <InputTextWrapper inputLength={inputs.from.value.length}>
+            <VisuallyHidden id={startDateLabelledby}>
+              {t('Start date')}
+            </VisuallyHidden>
             <InlineInputTextBase
               placeholder={`${formatDateString(
                 new Date(Date.now()),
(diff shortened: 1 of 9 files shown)


packages/components/src/Calendar/CalendarNav.tsx
```diff
@@ -97,6 +97,7 @@ export const CalendarNav: FC<NavbarElementProps> = ({
             as={headingSizeMap(size)}
             fontWeight="semiBold"
             fontFamily="body"
+            aria-live="polite"
```
kvrmd, line 100: I need some suggestions here. This works fine for InputDate. However, for InputDateRange, the screen reader only reads the Month+Year for the last Calendar. Here's a video:

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
```diff
@@ -359,6 +368,9 @@ export const InputDateRange: FC<InputDateRangeProps> = forwardRef(
           }
         >
           <InputTextWrapper inputLength={inputs.from.value.length}>
+            <VisuallyHidden id={startDateLabelledby}>
+              {t('Start date')}
+            </VisuallyHidden>
```
kvrmd, line 373: Not sure if this is good practice. However, I wanted the SR to read the original label from the Field component and then read "Start date"/"End date"

image

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
```diff
@@ -373,12 +385,19 @@ export const InputDateRange: FC<InputDateRangeProps> = forwardRef(
               onFocus={partial(handleTextInputFocus, 'from')}
               readOnly={readOnly}
               value={inputs.from.value}
+              aria-labelledby={`${ariaLabelledby} ${startDateLabelledby}`}
             />
           </InputTextWrapper>
-          <HyphenWrapper hasInputValues={!isEmpty(dateRange)}>
+          <HyphenWrapper
+            hasInputValues={!isEmpty(dateRange)}
+            aria-hidden="true"
```
kvrmd, line 393: Ignore the hyper here.

packages/components/src/Calendar/CalendarNav.tsx
mdodgelooker: I never would've thought of this, but Expedia.com puts both months in a totally separate element just for the screen reader. Is that the kind of behavior you're looking for?
image

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
mdodgelooker: I don't see any problem with that. 👍

packages/components/src/Calendar/CalendarNav.tsx
kvrmd: Ahh yeah I tried this, but I couldn't find an element with both data. The month value looks like it's in each individual calendar? Let me take another look. Thanks for confirming that using `VisuallyHidden` is an acceptable strategy.

dbchristopher: Oh nice, that's a great idea! Yeah I think you'll have to add a new element to summarize the selected date range

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
```diff
@@ -373,12 +385,19 @@ export const InputDateRange: FC<InputDateRangeProps> = forwardRef(
...
+          >
             &ndash;
           </HyphenWrapper>
           <InputTextWrapper inputLength={inputs.to.value.length}>
+            <VisuallyHidden id={startDateLabelledby}>
```
dbchristopher, line 398: this id should probably be `endDateLabelledby`?

packages/components/src/Form/Inputs/InputDateRange/InputDateRange.tsx
```diff
@@ -349,6 +353,11 @@ export const InputDateRange: FC<InputDateRangeProps> = forwardRef(
       setViewMonth(transformMonth(month, viewMonthDiff))
     }
 
+    const { t } = useTranslation('InputDateRange')
+
+    const startDateLabelledby = `startDate-labelledby-${id}`
+    const endDateLabelledby = `startDate-labelledby-${id}`
```
dbchristopher, line 359:
```suggestion
    const endDateLabelledby = `endDate-labelledby-${id}`
```

dbchristopher (approved):
This all looks great to me!

PR #1966, opened 2021-02-25
fix(MenuList): Windowing

mdodgelooker:
This PR puts back `overflow: hidden` on `MenuList` that got lost in [this change](https://github.com/looker-open-source/components/pull/1906/files#diff-d6c17c6698ceeea0bf50ca79fad14f8e6b235b2feb1ae07b0cd6b353a2b5380aL248).

### Requirements

  - [ ] yes
  - [x] not applicable

  - [ ] yes
  - [x] not applicable

### DIFF
--- a/packages/components/src/Menu/MenuList.tsx
+++ b/packages/components/src/Menu/MenuList.tsx
@@ -44,4 +44,5 @@ MenuListInternal.displayName = 'MenuListInternal'
 
 export const MenuList = styled(MenuListInternal)`
   min-width: 12rem;
+  overflow: auto;
 `
--- a/packages/components/src/Menu/MenuList.tsx
+++ b/packages/components/src/Menu/MenuList.tsx
@@ -24,13 +24,19 @@
 
  */
 
+import { width, WidthProps } from '@looker/design-tokens'
 import React, { forwardRef, Ref } from 'react'
 import styled from 'styled-components'
 import { List, ListProps } from '../List'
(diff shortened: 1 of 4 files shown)


packages/components/src/Menu/MenuList.tsx
```diff
@@ -44,4 +44,5 @@ MenuListInternal.displayName = 'MenuListInternal'
 
 export const MenuList = styled(MenuListInternal)`
   min-width: 12rem;
+  overflow: auto;
```
mdodgelooker, line 47: I realized belatedly that windowing requires a fixed width on the list, otherwise you get a funky bounce as you scroll. Any suggestion other than adding `WidthProps` and `${width}` back?
f343107c-85c2-4f0a-97cb-035974929320

ElliotTheHuman: I can't think of any creative solutions outside of adding them back in :( 

Some blue sky ideas:
- Find a way to calculate the max width among children (which probably isn't the best idea when you have thousands of elements)
- Automatically add truncate to all child elements (which is probably too invasive, and probably won't work for every use case)

Hoping @lukelooker sees a solution that I can't, but otherwise it probably just makes sense to add `WidthProps` to `MenuList`. 🤞

mdodgelooker:
Thanks @ElliotTheHuman , added `WidthProps` back on `MenuList` (and as a result `Menu` too).

ElliotTheHuman (approved):
LGTM!

PR #1971, opened 2021-02-26
fix(useArrowKeyNav): Focus lost when item is removed

mdodgelooker:
If `useArrowKeyNav`'s currently focused element is removed from the DOM, it becomes impossible to get focus back into the list. But as long as the list itself hasn't been removed, we want focus to stay inside until the user moves it elsewhere.

React does not trigger `onBlur` in this scenario. Some browsers do trigger a `focusout` event, but not all. The only reliable way to catch it is using a `MutationObserver` that checks if an element being removed was the focused item.

### Requirements

  - [ ] yes
  - [x] not applicable

  - [ ] yes
  - [x] not applicable

### DIFF
--- a/packages/components/src/List/List.story.tsx
+++ b/packages/components/src/List/List.story.tsx
@@ -104,11 +104,12 @@ export const ExpandingList = () => {
       <List>
         {showMore2 ? (
           <>
-            <ListItem>Cheddar</ListItem>
-            <ListItem>Swiss</ListItem>
-            <ListItem>Gouda</ListItem>
-            <ListItem>American</ListItem>
+            <ListItem key="0">Cheddar</ListItem>
+            <ListItem key="1">Swiss</ListItem>
+            <ListItem key="2">Gouda</ListItem>
+            <ListItem key="3">American</ListItem>
             <ListItem
+              key="4"
               onClick={() => setShowMore2(false)}
               description="Replaces all items"
             >
@@ -117,9 +118,10 @@ export const ExpandingList = () => {
           </>
         ) : (
           <>
-            <ListItem>Cheddar</ListItem>
-            <ListItem>Gouda</ListItem>
+            <ListItem key="5">Cheddar</ListItem>
+            <ListItem key="6">Gouda</ListItem>
             <ListItem
+              key="7"
               onClick={() => setShowMore2(true)}
               description="Replaces all items"
             >
--- a/packages/components/src/List/List.story.tsx
+++ b/packages/components/src/List/List.story.tsx
@@ -26,7 +26,7 @@
 
 import React, { FC, useState } from 'react'
 import { Story } from '@storybook/react/types-6-0'
-import { Grid } from '../Layout'
+import { Grid, Space } from '../Layout'
 import { DensityRamp } from './types'
 import { List, ListProps } from './List'
 import { ListItem } from './ListItem'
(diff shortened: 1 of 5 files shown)


packages/components/src/utils/useArrowKeyNav.ts
```diff
@@ -74,7 +83,7 @@ export const useArrowKeyNav = <E extends HTMLElement = HTMLElement>({
   onKeyDown,
 }: UseArrowKeyNavProps<E>) => {
   const internalRef = useRef<E>(null)
-  const [focusedItem, setFocusedItem] = useState<HTMLElement | null>(null)
+  const focusedItemRef = useRef<HTMLElement>()
```
mdodgelooker, line 86: Realized this doesn't need to be a state value. We only need the latest, and no need to trigger a re-render.

packages/components/src/utils/useArrowKeyNav.ts
```diff
@@ -134,6 +154,26 @@ export const useArrowKeyNav = <E extends HTMLElement = HTMLElement>({
...
   }
 
+  useEffect(() => {
+    const element = internalRef.current
+
+    // Create an observer to check if the focused element is removed
+    const observer = new MutationObserver((mutationsList) => {
+      if (checkElementRemoved(mutationsList, focusedItemRef.current)) {
+        // If it was, start focus back at the beginning
+        placeInitialFocus()
```
kvrmd, line 164: Jared and I were just talking about this yesterday about this Date preset menu.

image

mdodgelooker: That's the one! Terrible experience for keyboard nav right now. 😣

packages/components/src/List/List.story.tsx
```diff
@@ -72,6 +72,39 @@ LongList.parameters = {
...
+        <ListItem
+          onClick={() => setShowMore(true)}
+          description="To test keyboard nav"
+        >
+          Show More
+        </ListItem>
+      )}
+    </List>
+  )
+}
```
kvrmd, line 102: In dashboards-next Date preset filter, the entire list actually changes when More is clicked. 1 item in particular is added in between previously visible options. See the "Last 28 days" option below. Could we add a case when the menu list is completely reorganized?

image

Maybe something like this?
```
return (
  <List>
    {showMore ? (
      <ListGroup>
        <ListItem>Cheddar</ListItem>
        <ListItem>Swiss</ListItem>
      </ListGroup>
      <ListGroup>
        <ListItem>Gouda</ListItem>
        <ListItem>American</ListItem>
      </ListGroup>
      <ListItem>Show less</ListItem>
    ) : (
      <ListItem>Cheddar</ListItem>
      <ListItem>Gouda</ListItem>
      <ListItem>Show more</ListItem>
  )
)
```

kvrmd (dismissed):
There's still some small issues as discussed with @mdodgelooker . But this is an overall improvement so +1.

packages/components/src/utils/useArrowKeyNav.ts
```diff
@@ -113,27 +121,59 @@ export const useArrowKeyNav = <E extends HTMLElement = HTMLElement>({
...
-        const toFocus = getNextFocus(1, internalRef.current)
-        if (toFocus) {
-          // No need to update focusedItem with this since it's the default
-          toFocus.focus()
-        }
+        // Otherwise focus the first item
+        placeInitialFocus()
       }
+    } else {
+      focusedItemRef.current = e.target
```
mdodgelooker, line 149: I was updating this in 2 other spots before, under the assumption that the arrow key handlers included here were the only way to move focus on the items. But screen readers have their own method to move focus (ctrl+option+arrow-right, in the `List` examples), so this is the obvious spot to keep track of the currently focused item. Thanks @kvrmd for catching that!

kvrmd (approved):
LGTM. Thank you

PR #1974, opened 2021-02-27
fix: Slider & RangeSlider properly support floating step value

lukelooker:
Corrects behavior in `Slider` to treat steps as floats.
Corrects incrementPoint / decrementPoint in `RangeSlider` to address JS floating point eccentricities

### Requirements

  - [ ] yes
  - [x] not applicable

  - [ ] yes
  - [x] not applicable


lukelooker:
@dbchristopher - I suspect the `RangeSlider` fix I did could be done in a more clever way perhaps leveraging other helpers already within that code so before I chase down codecov's complaint (needs a smidge more test-coverage) I figured I'd wait to see your general feedback here.

dbchristopher (changes requested):
This looks good to me! I think you picked the best place to handle the floating point math. I did see this when I was testing in storybook, but it was after trying to tweak a couple of things on my end. So I'm not sure if it's a real issue or not, but my cloudtop has suddenly become so mind bogglingly slow that I can't even back out of my changes and try again 😳 

Screen Shot 2021-03-01 at 10 57 06 AM

So anyways, just want to make sure you can double check that it's working well! And then I suggested a couple of more readable approaches to those precision utility functions

dbchristopher:
Oh actually, the incrementPoint/decrementPoint helpers are only part of the keyboard navigation ux 🤦‍♂️  That was why I was still seeing the floating point issue in the UI. I think it could go in the roundToStep helper. I think I have an idea of how it could be integrated more thoroughly..

packages/components/src/Form/Inputs/RangeSlider/RangeSlider.tsx
```diff
@@ -119,10 +120,15 @@ const createNewValue = (
 const roundToStep = (
   min: number,
   max: number,
-  newPoint: number,
-  step: number
+  step: number,
+  newPoint: number
```
dbchristopher, line 124: swapped this argument order to make it easier to partially apply those first three values which shouldn't change over the life of this component

packages/components/src/Form/Inputs/RangeSlider/RangeSlider.tsx
```diff
@@ -231,13 +239,11 @@ export const InternalRangeSlider = forwardRef(
...
-      return Math.min(point + step * stepMultiplier, max)
-    }
+    const incrementPoint = (point: number, stepMultiplier = 1) =>
+      point + step * stepMultiplier
 
-    const decrementPoint = (point: number, stepMultiplier = 1) => {
-      return Math.max(point - step * stepMultiplier, min)
-    }
+    const decrementPoint = (point: number, stepMultiplier = 1) =>
+      point - step * stepMultiplier
```
dbchristopher, line 246: Remove math.min and math.max because that happens in `roundToStep` now

packages/components/src/Form/Inputs/RangeSlider/RangeSlider.tsx
```diff
@@ -248,7 +254,9 @@ export const InternalRangeSlider = forwardRef(
             e.key === 'ArrowUp' || e.key === 'ArrowRight'
               ? incrementPoint
               : decrementPoint
-          const newPoint = mutationFn(value[focusedThumb], e.shiftKey ? 10 : 1)
+          const newPoint = roundSliderValue(
+            mutationFn(value[focusedThumb], e.shiftKey ? 10 : 1)
+          )
```
dbchristopher, line 259: now both keyboard and mouse events use the same `roundToStep` helper which makes more sense anyways

dbchristopher (approved):
Ok this looks good to me! While I almost always prioritize readability over performance optimizations, I'm ok with leaving those precision helpers in their mathematical forms. 

I did some testing at https://jsbench.me/ and found your versions to be ~80% faster. In this case there are potentially so many mouse event updates that it would be nice to squeeze out a few more ms to keep things snappy

packages/components/src/Form/Fields/FieldRangeSlider/FieldRangeSlider.story.tsx
```diff
@@ -54,7 +54,7 @@ Steps.parameters = {
 }
 
 export const Floating = Template.bind({})
-Floating.args = { max: 1.1, min: 0.1, step: 0.1 }
+Floating.args = { max: 3.7, min: 0.1, step: 0.1 }
```
dbchristopher, line 57: added a larger slider range to more easily test the shift+arrow interactions (for skipping 10x steps per keypress)

PR #1985, opened 2021-03-02
fix(useClickable): Improve and export onClick type

mdodgelooker:
Added a missing `MouseEvent` to the generic `onClick` type in `useClickable`. I also exported `GenericOnClick` from there to slightly improve the awkwardness of matching it up with a "regular" `onClick`. This hook calls `onClick` from `onKeyDown` for the space & enter keys, which is why `| KeyboardEvent<E>` is needed.

### Requirements

  - [ ] yes
  - [x] not applicable

  - [ ] yes
  - [x] not applicable

### DIFF
--- a/packages/components/src/utils/useClickable.ts
+++ b/packages/components/src/utils/useClickable.ts
@@ -25,15 +25,26 @@
  */
 
 import { CompatibleHTMLProps } from '@looker/design-tokens'
-import { FocusEvent, KeyboardEvent, MouseEvent, useState, useMemo } from 'react'
+import {
+  FocusEvent,
+  KeyboardEvent,
+  MouseEvent as ReactMouseEvent,
+  useState,
+  useMemo,
+} from 'react'
 
 // These 2 are helper interfaces for components using this hook
 export interface FocusVisibleProps {
   focusVisible: boolean
 }
+
+export type GenericOnClick<E extends HTMLElement> = (
+  e: ReactMouseEvent<E, MouseEvent> | KeyboardEvent<E>
+) => void
+
 export interface GenericClickProps<E extends HTMLElement>
   extends Omit<CompatibleHTMLProps<E>, 'onClick'> {
-  onClick?: (e: MouseEvent<E> | KeyboardEvent<E>) => void
+  onClick?: GenericOnClick<E>
 }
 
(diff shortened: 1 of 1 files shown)


lukelooker (approved):
LGTM!

packages/components/src/utils/useClickable.ts
```diff
@@ -25,15 +25,26 @@
...
+} from 'react'
 
 // These 2 are helper interfaces for components using this hook
 export interface FocusVisibleProps {
   focusVisible: boolean
 }
+
+export type GenericOnClick<E extends HTMLElement> = (
+  e: ReactMouseEvent<E, MouseEvent> | KeyboardEvent<E>
+) => void
```
mdodgelooker, line 43: @patricksmithlooker unfortunately `as MouseEventHandler | KeyboardEventHandler` still couldn't be used for the spot in the core product where you had previously called this out. The same handler gets called for an actual click as for enter/space keydown, so it's just the event argument that has to combine both types.

But by having this exported, at least `as GenericOnClick<HTMLAnchorElement>` would be a little less painful than `as (e: React.MouseEvent<HTMLAnchorElement> | React.KeyboardEvent<HTMLAnchorElement>) => void`

mdodgelooker: But as I type this, I wonder if `useClickable` is even necessary in that spot anymore, now that it's an anchor tag?

mdodgelooker: For reference, this `usePress` hook from Adobe's React Aria was kind of the inspiration behind `useClickable`. They go way beyond though and have created their own event typing system. We're not quite there. 😅

patricksmithlooker: If I remember correctly, useClickable was still adding some focus states that weren't there otherwise.

patricksmithlooker: Though I suppose those styles could just be updated to not rely on props added by useClickable. 🤷‍♂️

mdodgelooker: Oh good point @patricksmithlooker – I think it's probably time to extract a separate `useFocusVisible` hook. 🎣

PR #1986, opened 2021-03-02
refactor(design-tokens): Consolidate `color` and `theme` attributes

lukelooker:
### Requirements

  - [ ] yes
  - [x] not applicable

  - [ ] yes
  - [x] not applicable


packages/design-tokens/src/color/defaults.ts
```diff
@@ -0,0 +1,52 @@
...
+  key: purple400,
+}
+
+export const defaultIntentColors: IntentColors = {
+  link: blue600,
+  critical: red500,
+  warn: yellow500,
+  positive: green500,
+  inform: blue500,
+}
```
ElliotTheHuman, line 52: super minor nit: But we can probably reorder these and get rid of the eslint-disable right?

packages/design-tokens/src/color/utils/blendColorTransparency.test.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  test('named color - 4', () => {
+    expect(blendColorTransparency('purple', 2)).toEqual('rgba(128,0,128,0.12)')
+  })
+
+  test('rgb color - 5', () => {
+    expect(blendColorTransparency('rgb(0,20,50)', 2)).toEqual(
+      'rgba(0,20,50,0.12)'
+    )
+  })
```
ElliotTheHuman, line 45: Not sure what the "- X" mean here in these last two test names (I thought X corresponded to the uiBlend level but I guess not)? Ignore if this is pretty known and I'm just ignorant 😆

packages/design-tokens/src/color/utils/blendColorTransparency.test.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { colors } from '../colors'
+import { blendColorTransparency } from './blendColorTransparency'
+
+const { text } = colors
+
+describe('blendColorTransparency', () => {
+  test('text - 2', () => {
+    expect(blendColorTransparency(text, 2)).toEqual('rgba(38,45,51,0.12)')
+  })
```
ElliotTheHuman, line 35:
```suggestion
describe('blendColorTransparency', () => {
  const alphaLevel = 2
  const targetUiBlend = uiBlends[alphaLevel - 1]
  test('text - 2', () => {
    expect(blendColorTransparency(text, alphaLevel)).toEqual(`rgba(38,45,51,${targetUiBlend}`)
  })
```

Could be helpful to derive the '0.12' in the `toEqual` call using the `uiBlends` array. Otherwise, it sort of feels like '0.12' is a magic number.

My code suggestion though might not be very readable, in which case, maybe a comment explaining where 0.12 comes from would be more helpful!

packages/design-tokens/src/color/utils/generateIntentShade.ts
```diff
@@ -0,0 +1,45 @@
...
+
+ */
+
+import darken from 'polished/lib/color/darken'
+import getLuminance from 'polished/lib/color/getLuminance'
+import lighten from 'polished/lib/color/lighten'
+import { css } from 'styled-components'
+
+// Returns a tint or shade of an intent color, used to for text text that sits on top of a intentUIBlend
+// Adjust amount of lightening or darkening based on colors luminance as well as background colors luminance
```
ElliotTheHuman, line 33:
```suggestion
// Returns a tint or shade of an intent color, used for text that sits on top of an intentUIBlend
// Adjust amount of lightening or darkening based on color's luminance as well as background color's luminance
```

packages/design-tokens/src/color/utils/intentUIBlend.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+      background-color: ${intentUIBlend('critical', 2)};
+      color: ${intentUIBlend('critical', 5)};
+    `
+
+    renderWithTheme(<Test>Find me</Test>)
+
+    const test = screen.getByText('Find me')
+    expect(test).toHaveStyleRule('background', '#f8e4e6')
+    expect(test).toHaveStyleRule('background-color', '#f3cbd0')
+    expect(test).toHaveStyleRule('color', 'rgba(NaN,NaN,NaN,NaN)')
```
ElliotTheHuman, line 46: Clarifying, is the intent of this part of the test to check that there are only 5 `uiBlends`? (i.e. index 5 fails since 0-4 are the only valid indices for the `uiBlends` array?)

packages/design-tokens/src/theme/ThemeCustomizations.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export interface ThemeCustomizations {
+  /**
+   * Override default color specifications
+   */
+  colors?: Partial<SpecifiableColors>
+  /**
+   * Override default font-family specifications. Specified fonts will have out built-in
+   * font-stack appended. Built-in font stacks are designed to provide i18n character
+   * support and fallbacks for browsers that can't load web fonts.
```
ElliotTheHuman, line 38:
```suggestion
   * Override default font-family specifications. Specified fonts will have our built-in
   * font-stack appended. Built-in font stacks are designed to provide i18n character
   * support and fallbacks for browsers that can't load web fonts.
```

ElliotTheHuman (commented):
I left a few questions mostly regarding testing and a few nits about various JSDoc comments, but things are looking good. Should be able to approve after this round of feedback :)

packages/design-tokens/src/color/defaults.ts
lukelooker: Yep. ✅

packages/design-tokens/src/color/utils/blendColorTransparency.test.ts
lukelooker: Ironically after digging into this file for a while it occured to me that I'd never seen it used. Upon checking this doesn't get used _anywhere_ so I went ahead and just removed it. 🎉

packages/design-tokens/src/color/utils/blendColorTransparency.test.ts
lukelooker: See above. 🤦

packages/design-tokens/src/color/utils/intentUIBlend.test.tsx
lukelooker: Actually this is just a total mistake (classic off-by-1 error). I went ahead and tightened up the interface to restrict the inputs to valid values and updated the test.

ElliotTheHuman (approved):
LGTM!!

PR #1995, opened 2021-03-03
Child ref setter

dbchristopher:
copy

### DIFF
--- a/packages/components/src/Button/MultiFunctionButton.tsx
+++ b/packages/components/src/Button/MultiFunctionButton.tsx
@@ -44,12 +44,16 @@ export interface MultiFunctionButtonProps {
     }
   >
   children: ReactElement<ButtonProps | IconButtonProps>
+  setChildrenRef?: any
+  setAlternateRef?: any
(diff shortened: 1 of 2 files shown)


packages/components/src/Button/MultiFunctionButton.tsx
```diff
@@ -84,8 +88,20 @@ export const MultiFunctionButton: FC<MultiFunctionButtonProps> = ({
...
+          aRef.current = node
+        },
+      })}
+      {cloneElement(alternate, {
+        'aria-hidden': !swap,
+        ref: (node: HTMLButtonElement) => {
+          setAlternateRef?.(node)
+          bRef.current = node
+        },
+      })}
```
dbchristopher, line 104: The ultimate solution I came up with was to store the ref in state, and then pass `setButtonRef` down to the MultiFunctionButton to keep the refs in sync between the two components. It's kind of gross, but I couldn't figure out another way to set a new ref without overwriting the old one entirely

laraharrow (approved):

PR #1997, opened 2021-03-04
fix(List): Fixed List / Menu windowing

ElliotTheHuman:
I noticed that scrolling in a long `Menu` seemed to have a bounce / rubber band effect on `main`. See the "Windowing (3k) Menu" on `main`'s storybook:
https://looker-open-source.github.io/components/latest/storybook/?path=/story/menu--long-menus

Turns out, I had forgotten to pass the `ref` returned from `useWindow` to `List`'s underlying `ul` element 🤦 . From what I can tell, this meant that the scroll position event handlers produced by `useScrollPosition` weren't attaching to the container `ul`, which meant that scroll position wasn't being calculated properly on scroll (which caused the bounce effect).

### Requirements

  - [ ] yes
  - [x] not applicable

  - [ ] yes
  - [x] not applicable

### DIFF
--- a/packages/components/src/List/List.tsx
+++ b/packages/components/src/List/List.tsx
@@ -137,6 +137,7 @@ export const ListInternal = forwardRef(
           role={role || 'list'}
           {...omitStyledProps(props)}
           {...navProps}
+          ref={ref}
         >
           {content}
         </ul>
@@ -149,4 +150,5 @@ export const List = styled(ListInternal)`
   ${reset}
 
   list-style: none;
+  overflow: auto;
 `
--- a/packages/components/src/List/List.tsx
+++ b/packages/components/src/List/List.tsx
@@ -148,6 +148,7 @@ export const ListInternal = forwardRef(
 export const List = styled(ListInternal)`
   ${reset}
 
+  height: 100%;
   list-style: none;
   overflow: auto;
 `
--- a/packages/components/src/List/List.tsx
+++ b/packages/components/src/List/List.tsx
@@ -137,7 +137,6 @@ export const ListInternal = forwardRef(
           role={role || 'list'}
           {...omitStyledProps(props)}
           {...navProps}
-          ref={ref}
         >
           {content}
         </ul>
--- a/packages/components/src/utils/getNextFocus.ts
+++ b/packages/components/src/utils/getNextFocus.ts
@@ -37,11 +37,20 @@ const getFallbackElement = (
   containerElement: HTMLElement,
   tabStops: HTMLElement[]
 ) => {
-  const firstVisibleChild = tabStops.find((childElement) => {
-    return childElement.offsetTop >= containerElement.scrollTop
-  })
+  let fallback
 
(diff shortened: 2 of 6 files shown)


mdodgelooker (changes requested):
Good catch on the bounce, but `navProps` contains a `ref`, merged with the one used for arrow key navigation. This change would break the navigation, but it does fix the bounce. So I think there must be something in `useArrowKeyNav` causing this. I'll look into it too.

mdodgelooker:
Looks like the bounce is being caused by repeated calls to `placeInitialFocus` inside `useArrowKeyNav` as the focused element keeps getting removed from the dom as you scroll – and then moving focus to the top rendered item makes the browser try to scroll that into view, I think. Somehow we need to disable that behavior until after scrolling has stopped. Let's chat about it mañana?

ElliotTheHuman:
Heads up! This PR ended up covering two separate issues related to windowing. The two broad areas of code change are:
1. Created a `getFallbackElement` function that returns the first visible fallback element (if possible)
2. The `LongList` story in `MenuList` and `List` did not have a container with explicit height, which led to all child items trying to render. I added a `Box` container to each and left a note in the `Windowing` sections of both respective docs.

packages/components/src/utils/getNextFocus.ts
```diff
@@ -31,6 +31,28 @@ export const getTabStops = (ref: HTMLElement): HTMLElement[] =>
...
+) => {
+  let fallback
+
+  if (direction === 1) {
+    const firstVisibleChild = tabStops.find((childElement) => {
+      return childElement.offsetTop >= containerElement.scrollTop
+    })
+
+    if (firstVisibleChild) fallback = firstVisibleChild
+    else fallback = tabStops[0]
```
ElliotTheHuman, line 48: I don't know exactly what case you'd want this else, but...
1. Having it prevents the return type from including "undefined"
2. It probably makes sense to return the first tabbable element should anything weird happen

packages/components/src/List/List.tsx
```diff
@@ -148,5 +148,7 @@ export const ListInternal = forwardRef(
 export const List = styled(ListInternal)`
   ${reset}
 
+  height: 100%;
```
mdodgelooker, line 151: Hopefully this won't have any unintended consequences? I can't think of any, and I know it's a good way to make sure windowing is easy to accomplish – otherwise you have to do a bunch of flex-y styles on the container – but I wouldn't want it to give us any fun surprises...

mdodgelooker (approved):
Great fix @ElliotTheHuman ! As long as you're feeling good about `height: 100%` on list, I say ship it.

packages/components/src/List/List.tsx
ElliotTheHuman: Yeah, this seemed like the simplest way to prevent `List` from growing outside the confines of its parent. Definitely open to revisiting this solution should we run into a fun surprise with an edge case.

PR #2002, opened 2021-03-04
fix(MenuHeading): Fix MenuHeading style overrides

ElliotTheHuman:
Added `className` to `MenuHeading` props to allow for style overrides. We need style overrides specifically as a part of the HT `Menu` refactor work, where we're swapping `MenuHeading`s in for existing `MenuGroup`s which can specific label styling. 

### Requirements

  - [ ] yes
  - [ ] not applicable

  - [ ] yes
  - [ ] not applicable


dbchristopher (changes requested):
The code looks fine, but is there any way we can avoid doing this? We've said in the past that once HT developers start overriding css it takes us down a dark path. How about if we update the MenuHeading to accept `TypographyProps` and pass those down directly?

dbchristopher:
@ElliotTheHuman I opened up a PR into this branch to show what I mean. I think we should try to keep people on rails as much as possible and prevent them from introducing arbitrary style overrides. Otherwise it can make future upgrades a lot more brittle and painful! https://github.com/looker-open-source/components/pull/2004

ElliotTheHuman:
Hey Daniel. Merged in your props-oriented suggestion; it definitely makes sense to maintain customization control / keep people on the happy path via props.

Also added in "ColorProps" after doing a brief look over what other style overrides are happening in HT.

dbchristopher (approved):
Ok cool! Looks great 👍

PR #2004, opened 2021-03-05
Menu Heading Styled System

dbchristopher:
An alternate way to allow menu heading overrides while still keeping HT developers in the happy path


ElliotTheHuman:
Keeping devs on the happy path is definitely preferable, so adding the necessary props makes sense!

PR #2012, opened 2021-03-09
refactor: List and Menu quality of life updates

ElliotTheHuman:
- Export `ListItemContext` so external components can utilize context + `listItemDimensions` util function to pull from our density scale (i.e. if they're using a non-text child for their `ListItem`)
- Change `MenuHeading` child to `ReactNode` (the `MainNav` has `<Highlight/>` children for search in the Explore dropdown, and the heading can be included in the search results so it can't just be a text child afaik)

### Requirements

  - [ ] yes
  - [ ] not applicable

  - [ ] yes
  - [ ] not applicable

### DIFF
--- a/packages/components/src/Menu/MenuHeading.tsx
+++ b/packages/components/src/Menu/MenuHeading.tsx
@@ -27,9 +27,9 @@
 import React, { FC, useContext, useRef, ReactNode, RefObject } from 'react'
 import styled from 'styled-components'
 import {
-  ColorProps,
+  TextColorProps,
   TypographyProps,
-  SpaceProps,
+  PaddingProps,
   pickStyledProps,
 } from '@looker/design-tokens'
 import { Heading } from '../Text/Heading'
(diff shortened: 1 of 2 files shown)


packages/components/src/Menu/MenuHeading.tsx
```diff
@@ -24,25 +24,30 @@
...
+    TypographyProps,
+    PaddingProps {
+  children: ReactNode
+  className?: string
 }
 
-export const MenuHeading: FC<MenuHeadingProps> = ({
+const MenuHeadingInternal: FC<MenuHeadingProps> = ({
   children,
+  className,
```
ElliotTheHuman, line 50: Discussed with Luke, but it probably makes sense to allow for styled overrides in case we don't include a needed prop via the interface. While harder to manage, styled overrides are relatively easy to find / track in HT and keep components from blocking progress on the dev side (i.e. they don't need to wait for us to add a needed prop for styling).

lukelooker (approved):
LGTM

PR #2014, opened 2021-03-09
fix(Divider): Remove Dividers's default vertical margin

lukelooker:
Also modernized test-suite

### Requirements

  - [x] yes
  - [ ] not applicable

  - [ ] yes
  - [x] not applicable


packages/components/src/Divider/Divider.tsx
```diff
@@ -70,6 +70,7 @@ export const DividerBase = styled.hr
     ({ appearance = 'default', customColor, size = '1px' }) => ({
       appearance,
       bg: customColor,
+      my: 'none',
```
ElliotTheHuman, line 73: The hard-coded `my: 'none'` seems to suppress any user-provided `my / mb / mt` prop.

Adding `margin: 0;` above the `space` styled function in the CSS-in-JS should allow us to set a default 0 margin, while also allowing for space props to work!

ElliotTheHuman, line 73: The hard-coded `my: 'none'` seems to suppress any user-provided `my / mb / mt` prop.

Adding `margin: 0;` above the `space` styled function in the CSS-in-JS should allow us to set a default 0 margin, while also allowing for space props to work!

packages/components/src/Divider/Divider.story.tsx
```diff
@@ -0,0 +1,50 @@
...
+export const Appearance = Template.bind({})
+Appearance.args = {
+  appearance: 'onDark',
+}
+
+export const Custom = Template.bind({})
+Custom.args = {
+  customColor: 'turquoise',
+  size: '20px',
+}
```
ElliotTheHuman, line 50: I might add a story that uses margin-y since I imagine that's how most people will space out their dividers. Could be wrong though, in which case, ignore me!

ElliotTheHuman (changes requested):
Just that `margin`-props related issue. Once that's patched, we should be good to go!

packages/components/src/Divider/Divider.tsx
lukelooker: Nice catch / also doh!

packages/components/src/Divider/Divider.story.tsx
lukelooker: We actually want to encourage folks to use `SpaceVertical` for that use-case. This fix actually comes from confusion about how `SpaceVertical` interacts (poorly) with the built-in margin on `hr` / `Divider`

ElliotTheHuman (dismissed):
LGTM!

ElliotTheHuman (approved):
LGTM-er

PR #2025, opened 2021-03-11
fix: Export all Date & Time-related components from a sub-location to prevent bundle-size impacts

lukelooker:
For reasons I don't totally understand Date-related components cause massive bundle-size impacts - I suspect it's caused by one of the two upstream dependencies (moment, react-date-picker or both). The easiest solve in the short-term is to take those components out of the default indices to prevent them from bloating the bundle for use cases where they're not needed.

### Requirements

  - [ ] yes
  - [ ] not applicable

  - [ ] yes
  - [ ] not applicable


www/src/MDX/Pre/allComponents.ts
```diff
@@ -182,6 +172,19 @@ import {
...
+import {
+  Calendar,
+  DateFormat,
+  DateTimeFormat,
+  FieldDate,
+  FieldDateRange,
+  InputDate,
+  InputDateRange,
+  InputTime,
+  InputTimeSelect,
```
mdodgelooker, line 184: I think you're just missing `FieldTime` and `FieldTimeSelect` from this list.

mdodgelooker (changes requested):

mdodgelooker (approved):
👍

PR #2028, opened 2021-03-11
feat(Layout): add position fixed to Header and Footer

laraharrow:

### Requirements

  - [ ] yes
  - [ ] not applicable

  - [ ] yes
  - [ ] not applicable


packages/components/src/Layout/Semantics/Header.tsx
```diff
@@ -45,4 +47,13 @@ export const Header = styled.header.withConfig({
   shouldForwardProp,
 })<HeaderProps>`
   ${headerFooterCSS}
+
+  width: 100%;
+
+  ${({ fixed }) =>
+    fixed &&
+    `
+      position: sticky;
```
lukelooker, line 56: `position: sticky` isn't supported in IE11 :/

packages/components/src/Layout/Semantics/Layout.tsx
```diff
@@ -49,9 +49,9 @@ const hasAsideCSS = css`
 
 export const Layout = styled.div.withConfig({ shouldForwardProp })<LayoutProps>`
   ${simpleLayoutCSS}
-
   display: flex;
   flex: 1 0 auto;
+  max-height: 100vh;
```
lukelooker, line 54: This may not work in some scenarios - `100vh` applies to the viewport so if the `Layout` isn't the full viewport height this will mix-calculate.

lukelooker (changes requested):
Let's chat about this tomorrow - I think we should come up with a few specific examples so we can have a "test suite" as it were for possible solutions :)

packages/components/src/Layout/Semantics/Semantics.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+        </Layout>
+      </Page>
+    )
+    expect(screen.getByText("I'm the header")).toBeInTheDocument()
+    expect(screen.getByText("I'm the header").closest('div')).toHaveStyle(
+      'overflow: auto;'
+    )
+  })
+
+  test('has Header and Footer positions fixed when passifn prop fixed', () => {
```
dbchristopher, line 73:
```suggestion
  test('has Header and Footer positions fixed when passing prop fixed', () => {
```

packages/components/src/Layout/Semantics/Semantics.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+              everyday carry.
+            </p>
+          </Section>
+        </Layout>
+      </Page>
+    )
+    expect(screen.getByText("I'm the header")).toBeInTheDocument()
+    expect(screen.getByText("I'm the header").closest('div')).toHaveStyle(
+      'overflow: hidden;'
+    )
```
dbchristopher, line 109: judging by the test description, you probably want to validate the `position: fixed` style rule instead?

dbchristopher (changes requested):
Looks good! Just caught a couple really tiny issues

packages/components/src/Layout/Semantics/Layout.tsx
```diff
@@ -49,9 +54,15 @@ const hasAsideCSS = css`
...
   display: flex;
-  flex: 1 0 auto;
+  flex: 1 1 auto;
+  overflow: auto;
+
+  ${({ fixed }) =>
+    fixed &&
+    `
+      overflow: hidden;
+    `}
```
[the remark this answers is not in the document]
lukelooker: Moving the overflow out of the interpolation saves the need for the `css` helper:

```
overflow:  ${({ fixed }) => fixed ? 'hidden' : 'auto'};
```

packages/components/src/Layout/Semantics/Semantics.test.tsx
laraharrow: Because of the way we are implementing the fixed position we actually does not have `position: fixed` we are using `flex` and all the logic is all on the `Layout` layer and doesn't ever trigger any change on `Header` or `Footer`

dbchristopher (approved):

PR #2050, opened 2021-03-15
fix(Space): Remove snapshot tests, add justify support, default DataTable to 100% width

lukelooker:
### Requirements

  - [ ] yes
  - [ ] not applicable

  - [ ] yes
  - [ ] not applicable


packages/components/src/Layout/Space/Space.test.tsx
```diff
@@ -26,82 +26,82 @@
...
-      <div>🦜</div>
-      <div>🐈</div>
-    </Space>
-  )
-})
+describe('Space', () => {
+  xtest('default', () => {
+    renderWithTheme(<Space data-testid="space">{content}</Space>)
+    // Should test default gap size
+  })
```
ElliotTheHuman, line 46: I think we're already "testing" space with image snapshots.

It might make sense to check the `margin-right` on one of the child divs, but image snapshots probably do a good enough job of covering this. In which case, maybe we should just delete the xtests?

packages/components/src/Layout/Space/Space.tsx
```diff
@@ -72,10 +72,19 @@ export interface SpaceHelperProps extends SimpleLayoutProps, FlexboxProps {
...
    * Align items vertically within `Space`
+   * NOTE: This will by overridden if `stretch=true`
    * @default 'center'
    */
   align?: 'start' | 'center' | 'end'
 
+  /**
+   * Justify items horizontally within `Space`
+   * NOTE: This will by overridden if `stretch=true`
+   * NOTE: Justifcation is based on flex-direction so if `reverse=true` this will be "backwards"
```
ElliotTheHuman, line 83:
```suggestion
   * NOTE: Justification is based on flex-direction so if `reverse=true` this will be "backwards"
```

packages/components/src/Layout/Space/Space.tsx
```diff
@@ -25,7 +25,7 @@
  */
 
 import styled, { css } from 'styled-components'
-import { variant } from 'styled-system'
+import { justifyContent, variant } from 'styled-system'
```
ElliotTheHuman, line 28: I don't _think_ this `justifyContent` function is being used anywhere (the only other place the text `justifyContent` is seen is seen is in the variant call below, but it's just as a property name, i.e. not the styled system function).

ElliotTheHuman (changes requested):
Mostly minor requests! One thing I will say is that it might be worth replacing the `spaceCSS` styled function with `simpleLayoutCSS` (which doesn't include the `flexbox` styled function) and removing `FlexboxProps` outright.

As it stands, you can still add a `justifyContent` content prop that works properly. It's not a problem per say, but it could be confusing to have both the traditional flex-related props and our more targeted suite of flex props (i.e. `justifyContent` vs `justify`).
Screen Shot 2021-03-15 at 4 25 52 PM

I'm cool with supporting both the traditional flex props and our own flex-related props if it's not a huge deal though!

packages/components/src/Layout/Space/Space.test.tsx
lukelooker: Yeah - i'm down with that :)

packages/components/src/Layout/Space/Space.tsx
lukelooker: Oops... accidentally imported this :)

lukelooker:
Your point about `justifyContent` _also_ being available is apt but I _think_ there's value in being explicit about the preferred way to justify content (I've seen folks frequently mixing up flex attributes so providing a "happy path" seems worthwhile. 

I did update the JSDoc to be more explicit about what will conflict with the `justify` setting. Let me know your thoughts.

ElliotTheHuman (approved):
LGTM! Thanks for adding that note about the `justify` prop.

PR #2055, opened 2021-03-16
refactor(ListItem): Always wrap ListItem children in styled span

ElliotTheHuman:
Rather than conditionally apply density-respecting styles to `ListItem` children (i.e. only when `ListItem` children are of type string), it probably makes sense to always _attempt_ to apply styles. Practically speaking, this just means always wrapping the children (and description) with a wrapper `span` that has these styles.

New story that uses nonstring `ListItem` children (still receives density-respecting styles from `span` wrapper)
Screen Shot 2021-03-16 at 3 25 12 PM

### Requirements

  - [ ] yes
  - [ ] not applicable

  - [ ] yes
  - [ ] not applicable


lukelooker (approved):
LGTM.

Codecov complaint seems spurious to me.

PR #2065, opened 2021-03-17
fix(DataTable): update gap space between content and indicator

laraharrow:

### Requirements

  - [x] yes
  - [ ] not applicable

  - [ ] yes
  - [x] not applicable


packages/components/src/DataTable/stories/useIndicator.tsx
```diff
@@ -0,0 +1,81 @@
...
+    </DataTableCell>
+  </DataTableItem>
+))
+
+const Template: Story<DataTableProps> = ({ ...args }) => {
+  return <DataTable {...args} />
+}
+
+export const UseIndicator = Template.bind({})
+UseIndicator.args = {
```
dbchristopher, line 78: Just for the sake of preventing future confusion, I'd probably just rename the story and filename to simply `Indicator`. The convention for `use...` usually indicates a custom hook. But I think in this case it's simply indicating that the indicator prop is being used?

dbchristopher (changes requested):
The code looks good! I just left a comment about naming convention that I think would be good to update so as not to cause any confusion in the future

packages/components/src/DataTable/stories/useIndicator.tsx
laraharrow: oh good point!

dbchristopher (dismissed):
Looks good! 👍

dbchristopher (approved):

PR #2070, opened 2021-03-17
refactor(ListItem): Moved Truncate within Text for proper styling

ElliotTheHuman:
Previously, the `ListItem` child wrapper and description wrapper followed this composition:
```
<Truncate>
  <Text         
       color={labelColor}
       fontSize={itemDimensions.labelFontSize}
       lineHeight={itemDimensions.labelLineHeight}
   >
    {children}
  </Text>
</Truncate>
```

I had mistakenly thought that `Truncate` would only work if its immediate parent wasn't an inline element 🤦  Because of this old composition, the `Truncate` element wasn't receiving the density-respecting styles (i.e. font size and line height) and required a `line-height: 1` style override to get the right heights for each density.

Moving `Truncate` within the `Text` element means that `Truncate` just "absorbs" the density-respecting styles on the `Text` element.

### Requirements

  - [ ] yes
  - [ ] not applicable

  - [ ] yes
  - [ ] not applicable

### DIFF
--- a/packages/components/src/Truncate/Truncate.tsx
+++ b/packages/components/src/Truncate/Truncate.tsx
@@ -87,7 +87,7 @@ const TruncateLayout: FC<TruncateProps> = ({
 
 const TextStyle = styled.span<WidthProps>`
   display: block;
-  overflow: hidden;
+  overflow-x: hidden;
   text-overflow: ellipsis;
   ${widthHelper}
   white-space: nowrap;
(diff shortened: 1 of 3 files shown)


packages/components/src/Truncate/Truncate.tsx
```diff
@@ -87,7 +87,7 @@ const TruncateLayout: FC<TruncateProps> = ({
 
 const TextStyle = styled.span<WidthProps>`
   display: block;
-  overflow: hidden;
+  overflow-x: hidden;
```
ElliotTheHuman, line 90: Minor add since there shouldn't be any content overflowing in the y plane because of `nowrap`.

lukelooker (approved):
Nice! :)

PR #2072, opened 2021-03-17
fix(Listitem): Improved strategy for :focus-visible CSS (remove z-index from MenuHeading)

lukelooker:
Per a conversation earlier today I realized that the strategy I used to display `focus-visible` on `ListItem` to prevent conflicts with `MenuHeading` was probably going to cause issues (either now or in the future). T

his change alters the strategy to `box-shadow` (per @jhardy's suggestion) and removes the problematic usage of `position: relative` abdicating the need for `z-index` on `MenuHeading`

While surveying our other strategies for displaying the "focus ring" I noticed that we were using `outline` on `AccordionDisclosure` and went ahead and pulled it into the same approach used for `Listitem` to reduce the number of strategies in use. 

### Developer Pre-Review Checklist

- [ ] a11y impact
- [ ] Needs new snapshot coverage
- [ ] Needs documentation update

### DIFF
--- a/packages/components/src/Accordion/AccordionDisclosure.tsx
+++ b/packages/components/src/Accordion/AccordionDisclosure.tsx
@@ -144,13 +144,19 @@ export const AccordionDisclosureStyle = styled.div
     color ? colorStyleFn : `color: ${theme.colors.ui5};`}
   cursor: pointer;
   display: flex;
-  outline: 1px solid transparent;
-  outline-color: ${({ focusVisible, theme }) =>
-    focusVisible && theme.colors.keyFocus};
   ${padding}
+  outline: none;
   text-align: left;
   width: 100%;
-`
+
+  ${({ focusVisible, theme }) =>
+    focusVisible &&
+    `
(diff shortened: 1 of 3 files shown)


packages/components/src/Accordion/AccordionDisclosure.tsx
```diff
@@ -144,13 +144,19 @@ export const AccordionDisclosureStyle = styled.div
...
+  outline: none;
   text-align: left;
   width: 100%;
-`
+
+  ${({ focusVisible, theme }) =>
+    focusVisible &&
+    `
+      &:focus {
+        box-shadow: inset 0 0 0 1px ${theme.colors.keyFocus};
```
ElliotTheHuman, line 156: Spoke to Jared. It probably makes sense to set the `AccordionDisclosure` focusVisible "border" to 2px to match `ListItem`. Has a little more pop / focusVisible-ness 😃

ElliotTheHuman (approved):
Yas.

PR #2074, opened 2021-03-18
fix(Grid): Default to 100% width

lukelooker:
Grid now defaults to 100% width so that when it's composed within other layout components (Space / SpaceVertical) it will "fill" its container without developer specify it. Width can be overridden if needed.

Also modernized test coverage

## Developer Checklist [ℹ️](https://github.com/looker-open-source/components/blob/main/CONTRIBUTING.md#developer-checklist)

- [x] 📚 Documentation updated (added JSDoc block)
- [ ] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11


ElliotTheHuman (approved):
LGTM

lukelooker:
Landing this as is but I did notice issues with IE11 + Grid. I'm going to log it as a bug in the backlog to follow-up since it's not a regression specific to this change.

PR #2102, opened 2021-03-22
refactor(ListItem): Added itemRole "none"

ElliotTheHuman:
## Changes
- Added "none" as valid option to `ListItem` `itemRole` prop
- Updated `TreeStyle` selectors to support `<div>` label container
- Added test and story for `itemRole === "none"`
- Updated `FieldPicker` story to match updated specs

📺 Watch a quick video on how to you make your pull requests fast and easy for you and your reviewer(s):
https://threadit.app/thread/k45nyrqugoinw6xub9vf/message/fl8m57exhqg4r0a4hg01eebn

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11


packages/components/src/List/ListItemLayout.tsx
```diff
@@ -50,7 +50,7 @@ const ListItemLayoutInternal: FC<ListItemLayoutProps> = ({
   const content = (
     <>
       {icon}
-      <Flex flexDirection="column" minWidth={0}>
+      <Flex flexDirection="column" minWidth={0} width="100%">
```
ElliotTheHuman, line 53: Need this so that a `<Space between>` child actually takes as much space as possible.

packages/components/src/List/ListItem.tsx
```diff
@@ -288,7 +304,7 @@ const ListItemInternal = forwardRef(
         detail={renderedDetail}
         icon={renderedIcon}
         px={itemDimensions.px}
-        py={itemDimensions.py}
+        py={itemRole === 'none' ? 'none' : itemDimensions.py}
```
ElliotTheHuman, line 307: Not having this means that a default-sized `IconButton` (i.e. of size `24px`) stretches a developer-sized / density = -3 `ListItem`.

It's not ideal, so wondering if you have thoughts on how to allow for default-sized icons within a dev-sized `ListItem` / `TreeItem` without expanding the height.

packages/components/src/List/ListItem.tsx
```diff
@@ -107,17 +111,32 @@ export interface ListItemProps
...
+    } else {
+      domElement = 'button'
+    }
+
+    return {
+      as: domElement,
+      type: itemRole === 'button' || disabled ? 'button' : undefined,
+    }
+  })<ListItemLabelProps>`
+    ${({ height, itemRole }) => itemRole === 'none' && `height: ${height}px;`}
```
ElliotTheHuman, line 138: See the below comment about default-sized icons stretching developer-sized items.

ElliotTheHuman:
Tagging @larissam! (TLDR this PR allows the label container to contain buttons / icon buttons without triggering the "button inside a button" error)

packages/components/src/List/ListItem.tsx
lukelooker: Let's discuss tomorrow in depth... I suspect this fix or a slightly broader approach (using height across the board perhaps?) is probably the right way to go as that composition seems likely.

packages/components/src/List/ListItemLayout.tsx
```diff
@@ -57,7 +57,7 @@ const ListItemLayoutInternal: FC<ListItemLayoutProps> = ({
   const content = (
     <>
       {icon}
-      <Flex flexDirection="column" minWidth={0}>
+      <Flex flexDirection="column" minWidth={0} flexGrow={1}>
```
lukelooker, line 60: Why is this needed? To ensure that the "label" space ends using all unneeded space or ...?

packages/components/src/List/ListItemWrapper.tsx
```diff
@@ -81,8 +81,7 @@ export const ListItemWrapper = styled(
   outline: none;
   text-decoration: none;
 
-  & > button,
-  & > a {
+  & > .list-item-label {
```
lukelooker, line 84: This solves the problem elegantly but I'm slightly concerned about downstream consumers (test suites or developers) hooking to this className and expecting it to be consistently available. An alternative would be to create a Styled Component and use the interpolated variable here and elsewhere. Thoughts?

packages/components/src/List/ListItemLayout.tsx
ElliotTheHuman: Correct! This allows the `Flex` container to take up all remaining space (so that a `<Space between>` element actually stretches its elements right before the detail container.

packages/components/src/List/ListItemWrapper.tsx
ElliotTheHuman: We actually already have a styled component called `ListItemLabel`, but `styled-components` gets into some trouble when trying to reference a styled component that uses `as` (_I think_):
Screen Shot 2021-03-23 at 10 03 11 AM

ElliotTheHuman: We cooooould just create a named `div` `ListItemLabelDiv` and just add to the original selectors like this:
```
& > button, 
& > a,
& > ${ListItemLabelDiv} {
...
}
```

It's definitely less likely to cause headache than the class name approach, but does mean we have to keep the trifecta selector approach. That being said, it probably makes sense to play it safer and prevent people from doing wacky stuff with that class name 🤷

packages/components/src/List/ListItem.tsx
```diff
@@ -107,17 +111,32 @@ export interface ListItemProps
...
+    let domElement
+
+    if (!disabled && itemRole === 'link') {
+      domElement = 'a'
+    } else if (!disabled && itemRole === 'none') {
+      domElement = 'div'
+    } else {
+      domElement = 'button'
+    }
+
```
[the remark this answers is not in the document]
ElliotTheHuman: Discussed over Meet, but it makes sense to allow div labels to be disabled. Resolved in 5c71f1e.

packages/components/src/List/ListItemWrapper.tsx
ElliotTheHuman: Resolved in b72f35f

ElliotTheHuman:
Addressed all comments except the "What do we do with padding when role='none'?" question.

Outside of introducing a "heightWithDescription" property to all densities, I'm not sure what could work to keep div labels at the same height without padding y 😕

lukelooker (changes requested):
Ok - this looks great overall, small nits and ponderings inline.

Only question I didn't call out inline is why did this need snapshot updates outside of the two for `role="none"` that were added? The others appear to just "jitter" from CI? Ideally we'd trim those out so we can just see the intentional changes in this PR :)

packages/components/src/List/ListItem.tsx
```diff
@@ -97,28 +103,16 @@ export interface ListItemProps
    * Sets the correct accessible role for the ListItem:
    * - Use **'link'** for items that navigation to another page
    * - Use **'button'** for items that trigger in page interactions, like displaying a dialog
+   * - Use **'none'** when including buttons as children in the label container (i.e. the label container will be a <div>).
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 856bfc1

packages/components/src/Tree/TreeStyle.tsx
```diff
@@ -33,6 +33,11 @@ import {
 import { listItemBackgroundColor } from '../List/utils'
 import { ListItemStatefulWithHoveredProps } from '../List/types'
 import { List, ListItem } from '../List'
+import {
+  ListItemLabelA,
```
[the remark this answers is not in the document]
ElliotTheHuman: Makes sense! The A at the end of the variable name does look a bit wonky.

Changed to `ListItemLabelLink` in 9e9d46b

packages/components/src/Tree/stories/FieldPicker.story.tsx
```diff
@@ -118,7 +103,30 @@ const PickerItem: FC<{ color?: string; truncate?: boolean }> = ({
...
+            >
+              <IconButton
+                icon={<FilterList />}
+                label="Filter"
+                tooltipPlacement="top"
+              />
+            </Popover>
+          </HoverDisclosure>
+        </div>
+      </Space>
```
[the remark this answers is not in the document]
ElliotTheHuman: Makes sense! I opted to use a `Flex` instead and a `FlexItem` wrapper around children with "flex={1}"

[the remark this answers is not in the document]
ElliotTheHuman: Resolved in a96f70d

packages/components/src/Tree/TreeStyle.tsx
```diff
@@ -100,19 +105,40 @@ export const TreeStyle = styled.div<TreeStyleProps>`
 
   ${({ dividers }) => dividers && dividersCSS}
 
-  & > ${Accordion} > ${AccordionContent} > ${List} > ${TreeBranch},
-  & > ${Accordion} > ${AccordionContent} > ${List} > ${ListItem} > button,
-  & > ${Accordion} > ${AccordionContent} > ${List} > ${ListItem} > a {
-    ${({ depth, indicatorSize, theme }) =>
-      generateIndent(depth + 2, indicatorSize, theme)}
+  & > ${Accordion} > ${AccordionContent} > ${List} {
+    &
```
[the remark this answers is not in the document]
ElliotTheHuman: They aren't! I'm partial to leaving an "&" in because it seemed cleaner, but having them here leads to some real prettier jankiness.

Resolved in 4b4d064

packages/components/src/Tree/TreeStyle.tsx
```diff
@@ -100,19 +105,40 @@ export const TreeStyle = styled.div<TreeStyleProps>`
...
+      > ${ListItem}
+      > ${ListItemLabelButton},
+      &
+      > ${ListItem}
+      > ${ListItemLabelA},
+      &
+      > ${ListItem}
+      > ${ListItemLabelDiv},
```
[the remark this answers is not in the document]
ElliotTheHuman: Resolved in 4b4d064!

It helped keep Prettier from freaking out fwiw

lukelooker (approved):
LGTM!

PR #2127, opened 2021-03-25
chore: Introduce codeql-analysis

lukelooker:
Introduce CodeQL to see what it can offer: https://codeql.github.com/docs/


lukelooker:
@mdodgelooker / @dbchristopher - CodeQL seems like tool that could offer occasional insight into better practices. Adding as a (non-blocking) CI check to see if it does so. Any objections?

dbchristopher (approved):
Oh interesting! Yeah that works for me 👍

PR #2128, opened 2021-03-25
fix(ToggleSwitch): Remove verticalAlign / inline-block to improve layout

lukelooker:
In tracing down micro-variance introduced in reset work Jared & I discovered that `ToggleSwitch`'s height is being slightly mis-assigned due to `vertical-align + inline-block` styling that is unneeded.

Change reduces `FieldToggleSwitch` height from `21px` to `20px` (the latter is the "correct" value per-design system guidance and matches `FieldCheckbox / FieldRadioGroup` vertical cadence

- [ ] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11


mdodgelooker (approved):
LGTM!

PR #2138, opened 2021-03-30
refactor(listItemDimensions): Updated density scale and minor color updates

ElliotTheHuman:
- Updated various density scale values to match Jared's [latest `ListItem` designs](https://www.figma.com/file/ADmrXwJUbKG0eVAhco8X10/LGM2-Transition?node-id=119%3A38939)
- Changed default color of `ListItemDetail` and `ListItem`'s optional icon to `text2` to match latest specs

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/List/ListItemLayout.tsx
```diff
@@ -86,7 +86,8 @@ export const listItemIconCSS = css<ListItemLayoutProps>`
 
 export const ListItemLayout = styled(ListItemLayoutInternal).attrs(
   ({ color, disabled }) => ({
-    color: disabled ? 'text1' : color || 'text1',
+    // Determines icon color
+    color: disabled ? 'text1' : color || 'text2',
   })
```
dbchristopher, line 91: I know the general structure here was already set, but I can see it's approaching a nested ternary in that it's probably easy to write but unnecessarily hard to read. I might just update the destructuring assignment to set the default value there:

```suggestion
 ({ color = 'text2', disabled }) => ({
    color: disabled ? 'text1' : color,
  })
```

CC @lukelooker since we were just talking about trying to remove all our nested ternaries yesterday. This one is kind of a grey area, but I think trying avoid multiple levels of nesting logic in general is probably a good rule of thumb

dbchristopher (changes requested):
Looks good! I just noticed a small tech debt/code style issue that might be good to update while your hands are in the code. Should be a quick change though

packages/components/src/List/ListItemLayout.tsx
ElliotTheHuman: Yeah this suggestion looks a lot cleaner than having both the ternary and the or operator there. Good catch.

ElliotTheHuman: Resolved in e06184b

dbchristopher (approved):
Cool, looks good! 👍

PR #2154, opened 2021-04-01
feat(Tree): Added labelBackgroundOnly prop

ElliotTheHuman:
The updated Field Picker design has background colors start at the `TreeItem` label rather than at the left edge of the `TreeItem` wrapper like default. Also, the background color for `Tree`s should start at the label rather than left of the indicator / `Accordion` wrapper.

Functionally, `labelBackgroundOnly` === true does the following:
- Moves the background color styling for `Tree` to the internal `ListItem`
- Moves any `TreeItem` indentation padding to the wrapper container rather than the label (so that padding isn't included in the background color) 

Screen Shot 2021-04-01 at 3 47 19 PM

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11


lukelooker (approved):
Nice. Ship it!

PR #2156, opened 2021-04-02
feat(Link): Support prop that displays icon indicating link is "external"

lukelooker:
This is a pattern I've seen used several times within the core product and seemed like it'd be a nice add to our built-in `Link` to support presenting this nicely as well as leveraging the `external` link type (https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types) as well `noreferrer`

Added screenshots as well.

image

- [x] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


jhardy (approved):
LGTM

packages/components/src/Link/Link.tsx
```diff
@@ -50,35 +51,76 @@ export interface LinkProps
    * @default false
    */
   underline?: boolean
+
+  /**
+   * Display an icon indicating that the link is to an external resource
+   * Also sets `rel="external noreferrer" on generated link.
+   * @default false
+   */
+  isExternal?: boolean
```
dbchristopher, line 60: Minor nit: my first reaction was that simply naming this `external` would be cleaner and more inline with conventional html attributes. I tend to think of arguments starting with `is` as transitive state that is subject to change, like `isOpen` or `isLoading`. Whereas `external` would describe a variant that is more fixed.

Not a huge deal either way though, and I'm sure I haven't been entirely consistent in that either lol 🤷

packages/components/src/Link/Link.tsx
```diff
@@ -50,35 +51,76 @@ export interface LinkProps
...
+const generateLinkTypes = ({ isExternal, rel, target }: LinkProps) => {
+  const linkTypes = rel ? rel.split(' ') : []
+
+  if (target === '_blank') {
+    linkTypes.push('noopener', 'noreferrer')
+  } else if (isExternal) {
+    linkTypes.push('external', 'noreferrer')
+  }
+
+  return [...new Set(linkTypes)].join(' ')
```
dbchristopher, line 89: nice use of Set 👍

dbchristopher (approved):
Looks good! I left a note about prop naming convention, but I'm honestly not sure we're entirely settled or consistent in that department either way

packages/components/src/Link/Link.tsx
lukelooker: Yep - I was originally going to use `external` but then discovered that that's a `Link Type` (see https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types) and wanted to avoid any confusion with that so did a more explicit name for this case. 

It probably _is_ (see what I did there) worth a longer conversation as a team at some point about prop-naming conventions :)

PR #2159, opened 2021-04-02
test: Specify storyshots body font-family

lukelooker:
Turns out that our storyshots are slightly "incorrect" because we currently don't have a base font-family set in Storybook - this means that the default line-height calculation is based on `Times` (browser default) rather than `Roboto`. Browsers base line-height on "roughly" 1.2 times the font-size but that roughly is determined by the `em` height of the inherited font.

This PR explicitly specifies the base font of our story-shots pages to be `Roboto` so that future updates to the reset can be verified against this corrected value.


packages/components/.storybook/preview-head.html
```diff
@@ -1,4 +1,8 @@
 <style>
+  body {
+    font-family: Roboto;
+  }
```
lukelooker, line 4: This (repeated in 3 duplicate files) is the key change.

jhardy (approved):
LGTM

PR #2172, opened 2021-04-06
fix(MenuHeading): Added spacing to MenuHeadings that follow MenuDividers

ElliotTheHuman:
- Genericized selectors in listPadding to make anything after a `ListDivider` take an `xsmall` margin 

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


ElliotTheHuman:
If making those selectors is too much of a nuclear option, I can definitely just add specific selectors for `MenuHeading` to `listPadding` and revert most of these changes. My thought was that pretty much anything that followed a `ListDivider` should probably have some spacing to avoid feeling too scrunched up.

mdodgelooker:
My only question would be, why doesn't `ListDivider` just have its own vertical margin?

ElliotTheHuman:
@mdodgelooker We used to have `ListDivider` / `MenuDivider` have "xsmall" margin above and below, but we replaced it with the selectors in `listPadding` for some reason. I can't for the life of me remember why though.

@lukelooker Do you happen to remember why we went with those selectors instead of just having vertical margin on `ListDivider`?

Side note, but it looks like we could get the desired spacing by adding vertical margin to `ListDivider` and adding vertical padding to `MenuList` to give the top and bottom breathing room.
Screen Shot 2021-04-06 at 2 53 55 PM

lukelooker:
I _think) we avoided vertical margin on `ListDivider` because we found cases where an "extra" Divider was at the start or end of a list and would cause funny renderings. 

I think this is the PR: https://github.com/looker-open-source/components/commit/ea6e028f75503f16884cafc1ec9acd0e0ae4b5ae but am still poking around to confirm.

lukelooker:
Oh! I remember now, `margin` caused weird scroll edges when `MenuHeaders` get fixed in a scrollable `Menu` - while the CSS is a bit more complex it seemed to balance the various scenarios best to leave it `*List` to place margin on dividers.

Complete(-ish) history at: https://github.com/looker-open-source/components/commits/8f82f7282059ec6c929b784a5cec95c50283d2a9/packages/components/src/Menu/MenuDivider.tsx

lukelooker:
I'm good with this change if @mdodgelooker is as well (I'll leave approval to you) :)

mdodgelooker (approved):
Ah thanks for the context @lukelooker . Looks like codecov is unhappy but I can't remember if that blocks merging or not. If you have to make more commits @ElliotTheHuman would you mind adding a bit in the `listPadding` jsdoc about the sticky header?

ElliotTheHuman:
Heyo! So in doing some testing, padding on the `MenuList` seemed to be the cause of the weird scroll edge issues we saw with sticky `MenuHeading`s. In which case, moving the `ListDivider` margin CSS to within the actual `ListDivider` styled component seems to be the right call for code readability.

I also went ahead and added CSS to handle the adjacent `ListDivider` case as per Luke's request (i.e. two dividers next to each other without anything in between). We ultimately hide the second divider using `display: none`.

packages/components/src/List/ListDivider.tsx
```diff
@@ -37,4 +37,10 @@ const ListDividerLayout: FC<{}> = (props) => (
 
 export const ListDivider = styled(ListDividerLayout)<SpaceProps>`
   ${space}
+
+  margin: ${({ theme }) => theme.space.xsmall} 0;
+
+  & + & {
+    display: none;
+  }
```
lukelooker, line 45: Can you put an explanatory comment on here to explain what this does

packages/components/src/List/ListDivider.tsx
```diff
@@ -37,4 +37,10 @@ const ListDividerLayout: FC<{}> = (props) => (
 
 export const ListDivider = styled(ListDividerLayout)<SpaceProps>`
```
lukelooker, line 38: This component should have a JSDoc block explaining its use and giving a head's up that two co-located dividers will "magically" hide siblings :)

lukelooker (changes requested):
Code looks good - would like clear code-comments about "double entry" behavior.

Also - since this does change something user-facing (margin on headings) it should be a `fix:` type instead of `refactor` (refactors are invisible in CHANGELOG so they need to be used really-sparingly)

lukelooker (approved):
LGTM!

PR #2174, opened 2021-04-06
fix(RangeSlider): Value state reverting if re-rendered before prop is updated

mdodgelooker:
This `RangeSlider` issue was uncovered in the core product Filter component when an intermediate re-render occurs _after_ the user drags a slider thumb but _before_ the prop reflects the change. Because the `value` prop is an array, newly instantiated on each render, the shallow diffing on the `[valueProp]` dependency array in a `useEffect` in `RangeSlider` caused the state value to be reverted to the stale prop value.

This PR removes the `[]` around `valueProp` in the dependency array to allow comparison of the min & max values themselves.

The use of the `RerenderRepro` story in the unit test is a newish pattern for us (used once in `Drawer.test.tsx`), an idea from @lukelooker a while back that I thought was great.

- [ ] 📚 Documentation updated
- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11

### DIFF
--- a/packages/components/src/Form/Inputs/RangeSlider/RangeSlider.story.tsx
+++ b/packages/components/src/Form/Inputs/RangeSlider/RangeSlider.story.tsx
@@ -44,7 +44,6 @@ const NumberFilter = ({
   AST: { value?: number[] }
   onChange: (value: number[]) => void
 }) => {
-  // const rangeValue = React.useMemo(() => getRange(value), [value])
   const rangeValue = getRange(value)
   return (
     <RangeSlider min={0} max={100} value={rangeValue} onChange={onChange} />
@@ -86,5 +85,6 @@ export const RerenderRepro = () => {
 }
 
 RerenderRepro.parameters = {
+  docs: { disable: true },
   storyshots: { disable: true },
 }
--- a/packages/components/src/Form/Inputs/RangeSlider/RangeSlider.story.tsx
+++ b/packages/components/src/Form/Inputs/RangeSlider/RangeSlider.story.tsx
@@ -1,0 +1,90 @@
+/*
+
+ MIT License
+
+ Copyright (c) 2021 Looker Data Sciences, Inc.
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
+ to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ copies of the Software, and to permit persons to whom the Software is
+ furnished to do so, subject to the following conditions:
+
+ The above copyright notice and this permission notice shall be included in all
+ copies or substantial portions of the Software.
+
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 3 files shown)


mdodgelooker:
@ThomasPinellaLooker FYI

packages/components/src/Form/Inputs/RangeSlider/RangeSlider.story.tsx
```diff
@@ -0,0 +1,90 @@
...
+export const RerenderRepro = () => {
+  const [expression, setExpression] = React.useState('0,10')
+  const handleChange = (newValue: string) => {
+    setExpression(newValue)
+  }
+  return <Filter expression={expression} onChange={handleChange} />
+}
+
+RerenderRepro.parameters = {
+  storyshots: { disable: true },
```
lukelooker, line 89: Probably good practice to start hiding these "repro" stories from Docs since they're pretty narrow use-cases and without extensive explanation I don't think they have much value to consumers :)

```suggestion
  docs: { disable: true },
  storyshots: { disable: true },
```

packages/components/src/Form/Inputs/RangeSlider/RangeSlider.story.tsx
```diff
@@ -0,0 +1,90 @@
...
+]
+
+const NumberFilter = ({
+  AST: { value },
+  onChange,
+}: {
+  AST: { value?: number[] }
+  onChange: (value: number[]) => void
+}) => {
+  // const rangeValue = React.useMemo(() => getRange(value), [value])
```
lukelooker, line 47: Remove since unused?

lukelooker (approved):
Couple of (very minor) nits but change looks good overall.

packages/components/src/Form/Inputs/RangeSlider/RangeSlider.tsx
```diff
@@ -343,8 +343,12 @@ export const InternalRangeSlider = forwardRef(
       if (!isEqual(value, boundedValue)) {
         setValue(sort(boundedValue))
       }
+      // Compares of the min & max values (or fallbacks) themselves to avoid
+      // unintentionally reverting state value due to shallow diffing
+      // a newly instantiated but stale valueProp, if [valueProp] were used
+      // (see RerenderRepro story)
       // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [valueProp])
+    }, valueProp || [min, max])
```
dbchristopher, line 351: I'm not sure if I entirely understand the implications of having a fallback dependency array. I wonder if it might be cleaner to merge these like `[min, max, ...valueProp]`? But on the other hand this seems to work as is so... 🤷

dbchristopher (approved):
Wow that's some tricky debugging! Great solution to simply compare each array member directly 👍 👍

packages/components/src/Form/Inputs/RangeSlider/RangeSlider.tsx
dbchristopher: oh nevermind, I guess that's simply saying just use `[min, max]` if the component isn't controlled. I'm just not used to seeing alternate options in the dependency array is all lol

lukelooker (approved):
LGTM

PR #2175, opened 2021-04-07
fix(Truncate): Prevent spurious scrollbar presentation due to browser settings

lukelooker:
In some cases scroll bars may be displayed on `Truncate` components

- If users have checked the "always visible" scrollbar option on MacOS
- Firefox on Windows _sometimes_ triggers (not sure of exact setting)
- Chrome on Linux _sometimes_

- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11

### DIFF
--- a/packages/components/src/Truncate/Truncate.tsx
+++ b/packages/components/src/Truncate/Truncate.tsx
@@ -87,7 +87,7 @@ const TruncateLayout: FC<TruncateProps> = ({
 
 const TextStyle = styled.span<WidthProps>`
   display: block;
-  overflow-x: hidden;
+  overflow: hidden;
   text-overflow: ellipsis;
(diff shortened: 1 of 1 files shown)


lukelooker:
NOTE: Image snapshot coverage only "sort-of" helps here as it seems like the Github Chrome environment doesn't reproduce the issue clearly. Did repro on my local Mac using the "Show scroll bars: Always" option.

ElliotTheHuman (approved):
I'm not able to reproduce any scrollbars when using `overflow-x: hidden` locally, but I don't see anything wrong with this change since we used `overflow: hidden` previously (iirc).

LGTM!

PR #2183, opened 2021-04-13
fix(DialogLayout): Delay rendering dialog content by one frame

dbchristopher:
This PR resolves a dom measurement race condition that could cause the MultiFunctionButton to set style width=0 and height=0 when used inside a Dialog.

Original YAQs issue: https://yaqs.corp.google.com/eng/q/6551937013359050752

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Dialog/Layout/DialogLayout.tsx
+++ b/packages/components/src/Dialog/Layout/DialogLayout.tsx
@@ -24,7 +24,7 @@
 
  */
 
-import React, { FC, ReactNode } from 'react'
+import React, { FC, ReactNode, useState } from 'react'
 import { Spinner } from '../../Spinner'
 import { DialogContent } from './DialogContent'
 import { DialogFooter } from './DialogFooter'
@@ -73,8 +73,16 @@ export const DialogLayout: FC<DialogLayoutProps> = ({
   header,
   headerCloseButton = !footer && true,
   headerDetail,
-  isLoading,
+  isLoading: isContentLoading,
 }) => {
+  // Delay rendering dialog content by one frame.
+  // This resolves a race condition where child content that needs a dom measurement
+  // finds width/height to be zero when it measures before dialog content finishes rendering.
+  const [isInitialLoad, setIsInitialLoad] = useState(true)
+  requestAnimationFrame(() => {
+    setIsInitialLoad(false)
+  })
+
   const dialogHeader = header && (
     <DialogHeader hideClose={!headerCloseButton} detail={headerDetail}>
       {header}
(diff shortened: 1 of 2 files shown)


packages/components/src/Dialog/Layout/DialogLayout.tsx
```diff
@@ -73,8 +73,16 @@ export const DialogLayout: FC<DialogLayoutProps> = ({
...
+  isLoading: isContentLoading,
 }) => {
+  // Delay rendering dialog content by one frame.
+  // This resolves a race condition where child content that needs a dom measurement
+  // finds width/height to be zero when it measures before dialog content finishes rendering.
+  const [isInitialLoad, setIsInitialLoad] = useState(true)
+  requestAnimationFrame(() => {
+    setIsInitialLoad(false)
+  })
+
```
dbchristopher, line 85: Alternatively we could run the `requestAnimationFrame` when MultiFunctionButton takes it's [measurement and stores dimensions in state](https://github.com/looker-open-source/components/blob/main/packages/components/src/Button/MultiFunctionButton.tsx#L69-L76). But I figured the better option would be to implement a universal fix in case any other future child content needs to access clientRect metadata.

packages/components/src/Dialog/stories/Dialog.story.tsx
```diff
@@ -141,6 +144,28 @@ ClickOutside.parameters = {
...
+      }
+    >
+      <Button>Open Dialog</Button>
+    </Dialog>
+  )
+}
+MultiFunctionButton.parameters = {
+  docs: { disable: true },
+  storyshots: { disable: true },
+}
```
dbchristopher, line 167: Fixed:

Screen Shot 2021-04-12 at 5 23 27 PM

lukelooker: 🎉

dbchristopher:
@laraharrow @lukelooker @mdodgelooker here is one potential fix I found for the issue Raymond found when trying to use the CopyToClipboard button inside a Dialog.

This PR is not currently passing CI, but I wanted to check in about whether you all think this is the best approach before spending any more time on it.

mdodgelooker (commented):
I realized the same thing was happening in `Popover`:
image
`Portal` initially renders its children in an unattached `div`, and by the time it appends the `div` to `#modal-root` the measurement in `MultiFunctionButton` has already been done. It looks like updating to `useLayoutEffect` in `Portal` fixes the timing issue: https://github.com/looker-open-source/components/pull/2185

dbchristopher:
Closing in favor of #2185

PR #2185, opened 2021-04-13
fix(Portal): Insert element earlier to support measured components

mdodgelooker:
- `Portal` initially renders its contents into an unattached `div` which it then appends to `div#modal-root` in a `useEffect` hook
- `MultiFunctionButton` contains a `useEffect` hook that measures the height & width of the button & alternate
- When `MultiFunctionButton` is inside a `Portal` (e.g. in a `Dialog` or `Popover`), its measurement `useEffect` occurs before the `Portal`'s dom-attachment `useEffect`, so the measurements return 0
- Changing `Portal`'s `useEffect` to `useLayoutEffect` moves it before the `useEffect`s of any children, so `MultiFunctionButton`'s measurements will happen after the elements have been inserted and return appropriate values

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Dialog/Layout/DialogLayout.tsx
+++ b/packages/components/src/Dialog/Layout/DialogLayout.tsx
@@ -24,7 +24,7 @@
 
  */
 
-import React, { FC, ReactNode, useState } from 'react'
+import React, { FC, ReactNode } from 'react'
 import { Spinner } from '../../Spinner'
 import { DialogContent } from './DialogContent'
 import { DialogFooter } from './DialogFooter'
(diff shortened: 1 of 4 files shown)


dbchristopher (approved):
This is so great @mdodgelooker! I had a feeling there must be a better solution out there. Works like a charm.

lukelooker (approved):
Super clever fix :) - had to go read `useLayoutEffect` docs again - this seems like a perfect use case.

PR #2186, opened 2021-04-13
feat(Layout): scroll use cases

laraharrow:

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/Layout/Semantics/Layout.tsx
```diff
@@ -37,6 +37,12 @@ export interface LayoutProps
    * @default false
    */
   fixed?: boolean
+  /**
+   * scroll patterns for different areas inside the layout:
+   *  independently | x number of areas together | all together
```
dbchristopher, line 42: this seems to describe three different options, when in reality it's just a boolean. Maybe just lose that middle one

```suggestion
   *  independently | all together
```

packages/components/src/Layout/Semantics/Layout.tsx
```diff
@@ -37,6 +37,12 @@ export interface LayoutProps
...
+   * @default false
+   */
+  scrollChildren?: boolean
```
dbchristopher, line 45: I wonder if `scrollGroup` or `scrollTogether` would be any more clear about the behavior this prop controls? I'm just thinking that when `scrollChildren` is false the children _do_ still scroll, they just scroll independently. 🤔  That's definitely a subtle distinction though and I might be overthinking it.

packages/components/src/Layout/Semantics/Layout.tsx
```diff
@@ -58,4 +64,13 @@ export const Layout = styled.div.withConfig({ shouldForwardProp })<LayoutProps>`
   flex: 1 1 auto;
   overflow: ${({ fixed }) => (fixed ? 'hidden' : 'auto')};
   ${({ hasAside }) => (hasAside ? hasAsideCSS : 'flex-direction: column;')}
+  ${({ scrollChildren }) =>
+    scrollChildren &&
+    css`
+      aside,
+      main {
```
dbchristopher, line 71: I will always be nervous when components start overwriting each other's styles. It makes them impure, which makes css harder to debug because you can't just look at one file and know where the styles are coming from. Do you think it would make sense to flip the approach, and add a prop like `scrollWithin` to the `Aside` and `Section` components instead? I imagine something like this:

```
<Layout hasAside>
  <Aside>...</Aside>
  <Section scrollWithin={false}>...</Section>
  <Aside scrollWithin={false}>...</Aside>
</Layout>
```

`scrollWithin` would toggle `height: fit-content;` as you've done here, but default to `true` so as not to break current implementations. And with this approach _I believe_ you could accomplish all three of these new stories with identical markup, without having to create nested Layout groups.

Let me know what you think!

dbchristopher (changes requested):
This works well! Clever use of `fit-content` to accomplish the task, I don't think I would have ever thought of that. I suggested a couple small tweaks to the code as is, but also offered up an alternate approach to structure the API in hopes of keeping the components pure. Let me know what you think!

packages/components/src/Layout/Semantics/Layout.tsx
```diff
@@ -58,4 +64,13 @@ export const Layout = styled.div.withConfig({ shouldForwardProp })<LayoutProps>`
...
+        height: fit-content;
+        overflow: auto;
```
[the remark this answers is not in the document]
laraharrow: yep you are right!

packages/components/src/Layout/Semantics/Layout.tsx
laraharrow: yeah I agree, that was just the name on the example Luke gave me but `scrollTogether` is more appropriated for sure

packages/components/src/Layout/Semantics/Layout.tsx
laraharrow: Oh yeah I do like this a lot. I think you bring up a good point I'm updating the code now to follow your logic!

packages/components/src/Layout/Semantics/Layout.tsx
```diff
@@ -59,3 +59,12 @@ export const Layout = styled.div.withConfig({ shouldForwardProp })<LayoutProps>`
...
 `
+
+// ${({ scrollTogether }) =>
+//     scrollTogether &&
+//     css`
+//       aside,
+//       main {
+//         height: fit-content;
+//       }
+//     `}
```
dbchristopher, line 70: can be removed now!

packages/components/src/Layout/Semantics/Semantics.test.tsx
```diff
@@ -108,4 +109,47 @@ describe('Layout', () => {
...
+          <Layout hasAside>
+            <Section scrollWithin>I'm the main area</Section>
+            <Aside scrollWithin>I'm the Aside after</Aside>
+          </Layout>
+        </Layout>
+      </Page>
+    )
+    expect(screen.getByText("I'm the Aside before")).not.toHaveStyle({
+      height: 'fit-content',
+    })
```
dbchristopher, line 147: this passes if you change to `toHaveStyleRule`

```suggestion
    expect(screen.getByText("I'm the Aside before")).not.toHaveStyleRule(
      'height'
    )
```

packages/components/src/Layout/Semantics/semanticStyledBase.ts
```diff
@@ -34,7 +34,15 @@ import { css } from 'styled-components'
...
   extends PaddingProps,
-    CompatibleHTMLProps<HTMLElement> {}
+    CompatibleHTMLProps<HTMLElement> {
+  /**
+   * scroll patterns for different areas inside the layout:
+   *  independently (default) | all together
+   * this prop has to be used in conjunction with the prop fixed.
+   * @default false
+   */
+  scrollWithin?: boolean
```
dbchristopher, line 44: Naming is so hard 😩 I know this was my suggestion, but seeing it now I think it still isn't quite clear what behavior this controls. What about `scrollWithLayoutGroup`?

packages/components/src/Layout/Semantics/semanticStyledBase.ts
```diff
@@ -34,7 +34,15 @@ import { css } from 'styled-components'
...
 export interface SemanticLayoutBase
   extends PaddingProps,
-    CompatibleHTMLProps<HTMLElement> {}
+    CompatibleHTMLProps<HTMLElement> {
+  /**
+   * scroll patterns for different areas inside the layout:
+   *  independently (default) | all together
+   * this prop has to be used in conjunction with the prop fixed.
+   * @default false
+   */
```
dbchristopher, line 43: Oh that's a good callout that it's meant to be used within a `<Page fixed>`! I think that could be a little more explicit about _which_ fixed prop you're referring to since that's a different component entirely. Maybe this?

```suggestion
  /**
   * To be used within the context of <Page fixed> container. 
   * When true, this removes the inner overflow-y scrolling
   * and allows content within a Layout group to scroll together.
   * @default false
   */
```

packages/components/src/Layout/Semantics/Header.tsx
```diff
@@ -32,8 +32,8 @@ import { borderHelper, SemanticBorderProps } from './semanticBorderHelper'
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface HeaderProps
-  extends SemanticLayoutBase,
-    HeightProps,
+  extends HeightProps,
+    Omit<SemanticLayoutBase, 'scrollWithin'>,
```
dbchristopher, line 36: I think that the fact that this prop can't be used on Header means the prop shouldn't be defined on the shared `SemanticLayoutBase` interface. It's one of those cases where [some duplication is better than DRY code](https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction).  I'd copy it over over to both `AsideProps` and `SectionProps` instead

dbchristopher (changes requested):
This is really close! Thanks for the updates. 

I'm sorry I suggested another name change for that prop lol 🙈  That's a tough one to describe succinctly!

packages/components/src/Layout/Semantics/semanticStyledBase.ts
laraharrow: thanks, naming things and explaining  use cases are so hard some times.

dbchristopher (approved):
Looks good!

packages/components/src/Layout/Semantics/Section.tsx
```diff
@@ -41,4 +49,5 @@ export const Section = styled.section.withConfig({
 })<SectionProps>`
   ${sectionCSS}
   overflow: auto;
+  ${({ scrollWithin }) => scrollWithin && 'height: fit-content;'}
```
lukelooker, line 52: @laraharrow  - `fit-content` isn't supported in IE11

PR #2189, opened 2021-04-13
fix(TreeItem): Clicks on padding trigger onClick when parent Tree labelBackgroundColor is true

ElliotTheHuman:
**Proposed Solution:** Add an additional click handler onto the `ListItem` wrapper that only triggers when clicking on the wrapper itself (i.e. NOT on a child). Theoretically, this should mean that this handler only fires when clicking on the `ListItem` / `TreeItem`'s indent padding when used in a `Tree` structure. 

Clicking on the rest of the `ListItem` would fire the normal handler attached to the `ListItemLabel` or not fire anything at all (i.e. the detail elements when accessory is true).

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/List/ListItem.tsx
```diff
@@ -164,29 +173,23 @@ const ListItemInternal = forwardRef(
...
-    const handleOnMouseEnter = useWrapEvent(
-      () => setHovered(true),
-      onMouseEnter
-    )
-    const handleOnMouseLeave = useWrapEvent(
-      () => setHovered(false),
-      onMouseLeave
-    )
+    const handleMouseEnter = useWrapEvent(() => setHovered(true), onMouseEnter)
+    const handleMouseLeave = useWrapEvent(() => setHovered(false), onMouseLeave)
```
ElliotTheHuman, line 192: Wasn't sure if we wanted the `handleOnX` or `handleX` nomenclature, but going with the latter option meant these handlers were one-liners 🤷 

Happy to change these back if we prefer the former.

packages/components/src/Tree/stories/FieldPicker.story.tsx
```diff
@@ -112,7 +112,10 @@ const PickerItem: FC<{ color?: string; truncate?: boolean }> = ({
             icon={<SubdirectoryArrowLeft />}
             label="Pivot"
             tooltipPlacement="top"
-            onClick={() => alert('Pivot')}
+            onClick={(event) => {
+              event.stopPropagation()
+              alert('Pivot')
+            }}
```
ElliotTheHuman, line 118: The onClick callbacks in the Pivot / Filter icons (both of which live within the item label container)  need a `stopPropagation` call regardless of this PR's other changes.

lukelooker: This seems awkward to me - it feels like it should "just work" or not work at all - this seems to sort-of split the difference (at least in my head) but I think I'd be more confused by this middle track than just saying it's "batteries not included" if you go the `itemRole="none"` route.

I think in an `itemRole="none"` the `onClick` specified on the `ListItem` shouldn't apply to label and we should leave it to the developer to apply it manually to the parts of the label they want it to be activated `onClick` (in this case `<FlexItem flex={1}>{children}</FlexItem>`?)

You could be _very_ explicit and add a `onClickWhitespace` handler that uses a type-gate (see Daniel's clever TS scheme from earlier this week) to gate that callback to only be available when `itemRole="none"`. Along those same lines you could also prevent `onClick` in that case as well to further reduce the potential for mis-steps.

lukelooker (changes requested):

packages/components/src/Tree/stories/FieldPicker.story.tsx
ElliotTheHuman: Makes sense. I went ahead and got those changes in via 44771eb.

Note: I tried getting the type-gating in, but for whatever reason I couldn't get an error to pop up when using `onClick` and `itemRole="none"` 😟 

For reference, I used these types below:
```
type ListItemRoleOther = {
  itemRole?: 'button' | 'link'
  onClick?: (event: React.MouseEvent<HTMLElement>) => void
  onClickWhitespace?: never
}

type ListItemRoleNone = {
  itemRole: 'none'
  onClick?: never
  onClickWhitespace?: (event: React.MouseEvent<HTMLElement>) => void
}

type ListItemRoleProps = ListItemRoleOther | ListItemRoleNone
```

ElliotTheHuman: In the meantime, I added `onClickWhitespace` as just an optional prop on `ListItemProps` and added a `console.warn` when trying to use `itemRole="none"` and `onClick` at the same time.

lukelooker (changes requested):

packages/components/src/List/ListItem.tsx
```diff
@@ -278,13 +294,21 @@ const ListItemInternal = forwardRef(
       </Layout>
     )
 
+    const handleOnClickWhitespace = (event: React.MouseEvent<HTMLElement>) => {
+      if (event.currentTarget === event.target) {
+        setFocusVisible(true)
+        onClickWhitespace && onClickWhitespace(event)
+      }
+    }
```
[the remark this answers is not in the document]
ElliotTheHuman: Technically speaking, it's possible to have a `Tree` with `labelBackgroundOnly` enabled and items that have `itemRole="button"` (or "link"). In which case, clicking the whitespace wouldn't do anything without an `onClickWhitespace` passed in.

Ideally, we could disable this by checking if `labelBackgroundOnly` is enabled, but that's strictly a `Tree` prop (i.e. not a `TreeItem` / `ListItem` prop) since `Tree` is the one that manages the indent logic.

I could try pulling from `TreeContext` to see if `labelBackgroundOnly` is enabled, but tying `ListItem` and `Tree` together in that way seems a bit invasive for `ListItem`.

packages/components/src/List/ListItem.tsx
```diff
@@ -195,6 +204,13 @@ const ListItemInternal = forwardRef(
       )
     }
 
+    if (itemRole === 'none' && onClick) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        'itemRole="none" and onClick cannot be combined - if itemRole="none" is a necessity, assign click behavior directly to ListItem\'s children'
+      )
+    }
```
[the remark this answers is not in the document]
ElliotTheHuman: It's definitely discouraged to use `onClickWhitespace` without `labelBackgroundOnly`, but I don't _think_ there's anything wrong with using `onClickWhitespace` when `itemRole` is "link" or "button" (i.e. you'd need an `onClickWhitespace` in that scenario so that anything left of the link / button is clickable).

The problem is that `labelBackgroundOnly` is a `Tree` prop and thus we'd have to use `useContext(TreeItemContext)` within `ListItem` to figure out if `labelBackgroundOnly` is enabled. Not the end of the world, but my gut seems odd to tie those two together. I could wrong about that though.

ElliotTheHuman:
Jest tests are failing on GH, but tests passing locally 👍 
Screen Shot 2021-04-20 at 10 37 05 AM

packages/components/src/List/ListItem.tsx
```diff
@@ -113,6 +113,11 @@ export interface ListItemProps
    * If true, text children and description will be truncated if text overflows
    */
   truncate?: boolean
+  /**
+   * Should only be passed via TreeItem
+   * @private
```
lukelooker, line 118:
```suggestion
   * Callback to specify onClick handler on item's whitespace.
   * @private May only be passed via TreeItem. This feature may be removed without a breaking change, we STRONGLY discourage the direct use of this property
```

lukelooker (commented):
Pushed a commit to _hopefully_ straighten out Codecov error

lukelooker (approved):
Ship it. :)

PR #2190, opened 2021-04-13
refactor: Single Storybook instance instead of per-package

lukelooker:
This change simplifies our monorepo Storybook setup to use a single instance of Storybook for development and story-shots rather the the previous, per-package configuration.


package.json
```diff
@@ -34,7 +34,7 @@
     "clean:storybook": "rm -Rf packages/*/storybook-static ",
     "clean:docs": "find docs -type f ! -name 'index.html' -delete",
     "clean:gatsby": "rm -Rf www/.cache www/public",
-    "clean:coverage": "rm -rf coverage",
```
mdodgelooker, line 37: This isn't needed anymore?

package.json
```diff
@@ -48,10 +48,10 @@
     "test": "yarn jest",
 
     "storybook": "yarn workspace storybook develop",
-    "storybook-docs": "yarn workspace storybook develop-docs",
+    "storybook-nodocs": "yarn workspace storybook develop",
```
mdodgelooker, line 51: I think you probably meant to remove this
```suggestion
```

mdodgelooker (changes requested):

package.json
lukelooker: Whoops. Mistake

package.json
lukelooker: Yep that's should go away 🤦

lukelooker:
@mdodgelooker Fixed what you caught and went ahead and removed Storybook step from CI since this needed to re-run anyway.

mdodgelooker (approved):
👍

PR #2199, opened 2021-04-14
feat(Accordion): Refactored Accordion interface and composition

ElliotTheHuman:
Updates `Accordion`s composition to match our modern approach to overlay-style components:
```
<Accordion content="I show or hide sometimes">
  "I am disclosure"
</Accordion>
```
Note: The old composition is supported, but will be deprecated in 2.x.

- [x] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11


packages/components/src/Accordion/Accordion.story.tsx
```diff
@@ -115,54 +118,66 @@ Controlled.parameters = {
...
+            Hello World
           </Accordion>
         </li>
       </UnorderedList>
-    </AccordionContent>
+    }
+  >
+    Hello World
   </Accordion>
 )
```
lukelooker, line 182: My brain was having a hard-time groking this - came up with a (I think) simpler to read version:

```jsx
const Branch: FC = ({ children }) => (
  <Accordion
    indicatorPosition="left"
    defaultOpen
    pl="medium"
    content={
      <UnorderedList>
        <li>Cheddar</li>
        <li>Gouda</li>
        <li>Swiss</li>
        {children}
      </UnorderedList>
    }
  >
    Hello World
  </Accordion>
)

export const Nested = () => (
  <Branch>
    <Branch>
      <Branch>
        <Branch />
      </Branch>
    </Branch>
  </Branch>
)
```

lukelooker (commented):
I think you're going the right direction with this - continue on with my blessings!

Also, I have an couple of ideas about how we can solve for the "old" composition style - conversation for tomorrow :)

packages/components/src/Accordion/Accordion.story.tsx
ElliotTheHuman: This is 100000% more readable. Sick composition.

packages/components/src/Accordion/Accordion.tsx
```diff
@@ -92,64 +105,105 @@ export interface AccordionProps
...
+        accordionDisclosureId={accordionDisclosureId}
+        indicatorGap={indicatorGap}
+        indicatorIcons={indicatorIcons}
+        indicatorPosition={indicatorPosition}
+        indicatorSize={indicatorSize}
+        isOpen={isOpen}
+        onClose={onClose}
+        onOpen={onOpen}
+        toggleOpen={toggleOpen}
+        {...props}
```
ElliotTheHuman, line 167: I needed a way to provide HTML props (and things like text color) to the underlying `AccordionDisclosure` since the user can't do it themselves in the new/ideal composition.

My proposed solution is to add the necessary props to `Accordion` and then pass any props that are HTML compatible down to the internal `AccordionDisclosure`.

ElliotTheHuman:
Got the rest of the codebase (i.e. docs and stories and tests) refactored to the new composition!

www/src/documentation/components/content/list.mdx
```diff
@@ -18,7 +18,7 @@ import { ListItemPropTable } from './demos'
   <ListItem icon={<AnalyticsApp />} detail="Italy">
     Mozzarella
   </ListItem>
-  <ListItem icon={<Account />} detail="Switzerland">
+  <ListItem icon={<LogoRings />} detail="Switzerland">
```
ElliotTheHuman, line 21: `<Account />` was causing an error on the docs page so swapped to an icon that wouldn't error out.

packages/components/src/Accordion/AccordionContent.tsx
```diff
@@ -24,23 +24,23 @@
...
-import React, { useContext, FC } from 'react'
+import React, { FC } from 'react'
 import styled from 'styled-components'
 
-import { AccordionContext } from './AccordionContext'
-
 export interface AccordionContentProps {
   className?: string
+  accordionContentId?: string
+  accordionDisclosureId?: string
```
ElliotTheHuman, line 33: Ideally these would be required since the only place `AccordionContent` would be used is within `Accordion`, but to support legacy compositions we make them optional.

packages/components/src/Accordion/AccordionDisclosure.tsx
```diff
@@ -118,7 +132,19 @@ const AccordionDisclosureInternal: FC<AccordionDisclosureProps> = forwardRef(
...
+        <AccordionDisclosureLayout
+          indicatorPosition={
+            indicatorPosition || accordionContextDefaults.indicatorPosition
+          }
+          indicatorSize={
+            indicatorSize || accordionContextDefaults.indicatorSize
+          }
+          indicatorGap={indicatorGap || accordionContextDefaults.indicatorGap}
+          indicatorIcons={
+            indicatorIcons || accordionContextDefaults.indicatorIcons
```
ElliotTheHuman, line 144: It's a bit janky, because we changed the interface around to support the legacy composition, we need to add these fallbacks so that `AccordionDisclosureLayout` is happy (since the indicator related props and `isOpen` are required on the layout component).

packages/components/src/Form/Fieldset/Fieldset.tsx
```diff
@@ -113,11 +111,23 @@ const FieldsetLayout = forwardRef(
...
+      // Accordion Control-related props
+      defaultOpen,
+      isOpen,
+      toggleOpen,
+      onClose,
+      onOpen,
+
       ...restProps
-    } = omit(props, [...AccordionControlPropKeys])
+    } = props
```
ElliotTheHuman, line 123: Seemed like a good time to get rid of the `omit` and `pick` usage I added to `Fieldset` way back when.

packages/components/src/Accordion/AccordionDisclosure.tsx
lukelooker: Maybe add a `@todo` above this to note that the fallbacks should be removed in `2.x`?

lukelooker (changes requested):
Few minor nits on documentation and one adjustment to Styled System usage.

lukelooker (approved):
Failing due to a snapshot difference but otherwise this LGTM. Once CI is green I'm happy to do a fresh 👍

ElliotTheHuman:
🤦  I used the `Accordion > Basic` story to test some weirdness with `Fieldset`. Changed it back. Will re-request review once snapshots are updated / merged into this PR.

lukelooker (approved):
Approved :)

PR #2202, opened 2021-04-15
fix: use DOMRect in lieu of ClientRect

AprilArcus:
Align on DOMRect over ClientRect, for interoperability with ESLint's `browser` [environment global](https://github.com/sindresorhus/globals/blob/548191a277c1b946bbcba6e34d593e00290baf48/globals.json#L480).

### DIFF
--- a/packages/components/src/utils/useMeasuredElement.ts
+++ b/packages/components/src/utils/useMeasuredElement.ts
@@ -29,14 +29,7 @@ import { useResize } from './useResize'
 
 function measureElement(element?: HTMLElement | null) {
   if (!element) {
-    return {
-      bottom: 0,
-      height: 0,
-      left: 0,
-      right: 0,
-      top: 0,
-      width: 0,
-    }
+    return new DOMRect()
   }
 
   return element.getBoundingClientRect()
@@ -44,7 +37,7 @@ function measureElement(element?: HTMLElement | null) {
 
 export const useMeasuredElement = (
   element: HTMLElement | null
-): [ClientRect, () => void] => {
+): [DOMRect, () => void] => {
   const [rect, setRect] = useState(measureElement())
 
   const refreshDomRect = useCallback(() => {
--- a/packages/components/src/utils/useMeasuredElement.ts
+++ b/packages/components/src/utils/useMeasuredElement.ts
@@ -29,7 +29,20 @@ import { useResize } from './useResize'
 
 function measureElement(element?: HTMLElement | null) {
   if (!element) {
-    return new DOMRect()
+    return typeof DOMRect === 'function'
+      ? new DOMRect()
+      : {
+          bottom: 0,
+          height: 0,
+          left: 0,
+          rect: {},
+          right: 0,
+          toJSON: () => null,
+          top: 0,
+          width: 0,
+          x: 0,
+          y: 0,
+        }
   }
 
   return element.getBoundingClientRect()
--- a/packages/components/src/Form/Inputs/RangeSlider/RangeSlider.tsx
+++ b/packages/components/src/Form/Inputs/RangeSlider/RangeSlider.tsx
@@ -138,7 +138,7 @@ const roundToStep = (
  */
 const calculatePointValue = (
   mouseX: number,
-  containerRect: ClientRect,
+  containerRect: DOMRect,
   min: number,
   max: number,
   step: number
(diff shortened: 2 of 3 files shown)


lukelooker:
It appears `DOMRect` [isn't supported by IE11](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect#notes) so I'm a bit concerned that this switch might encourage developers to use aspects of that aren't going to work in that (annoyingly) supported environment.

AprilArcus:
> It appears `DOMRect` [isn't supported by IE11](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect#notes)

Can we link to a polyfill, e.g. https://github.com/jarek-foksa/geometry-polyfill#readme in the documentation?

lukelooker (approved):
Took a closer look at our existing use cases and read through MDN / Microsoft docs more closely on a second pass – we're not using any of the problematic features at this point and it appears we already use DOMRect elsewhere so consolidating on the contemporary interface seems like the best tact.

The polyfill you suggested would be a reasonable workaround but it seems easy enough to just stay away from the delta between ClientRect/DOMRect on our end and save downstream folks the pain :)

Pending fixes to pass CI (DOMRect import is needed it appears) I'm fine with this change.

FYI - Updated PR title so it'd pass CI and match our CHANGELOG formatting.

packages/components/src/utils/useMeasuredElement.ts
```diff
@@ -29,22 +29,15 @@ import { useResize } from './useResize'
...
   if (!element) {
-    return {
-      bottom: 0,
-      height: 0,
-      left: 0,
-      right: 0,
-      top: 0,
-      width: 0,
-    }
+    return new DOMRect()
```
lukelooker, line 32: DOMRect needs to be imported to work in our Jest & Gatsby CI runs.

AprilArcus: imported from where, though? It's [browser global](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect/DOMRect).

lukelooker: Less elegant but this would work and keep the object strongly typed. 

```jsx

function measureElement(element?: HTMLElement | null): DOMRect {
  if (!element) {
    return {
      bottom: 0,
      height: 0,
      left: 0,
      right: 0,
      toJSON: () => undefined,
      top: 0,
      width: 0,
      x: 0,
      y: 0,
    }
  }

  return element.getBoundingClientRect()
}```

For better or worse Gatsby's SSR build has pushed us into a few cases like this where we can't _assume_ the browser APIs are available.

Definitely open to other ideas if you have any :)

Comment (AprilArcus) [reply packages/components/src/utils/useMeasuredElement.ts:32]: I don't like that this would fail `instanceof` checks and might not yield the same key order...

Comment (AprilArcus) [reply packages/components/src/utils/useMeasuredElement.ts:32]: it seems like attempting to measure an element in an SSR context should return an error token rather than an all-zeros DOMRect

Comment (mdodgelooker) [reply packages/components/src/utils/useMeasuredElement.ts:32]: An error token instead of the all-zeros DOMRect would work but would require some refactoring at the call sites. It's not so much for SSR context... this hook was actually designed for [use with a `useState`-style ref](https://github.com/looker-open-source/components/blob/main/packages/components/src/Form/Inputs/RangeSlider/RangeSlider.tsx#L211), so `element` could be `null` if the element it represents is mounted or unmounted after or before the component calling the hook. Ironically, I don't think it's being used that way anywhere, but I did add [one `usage` where I pass `null` when I wanted to turn measuring off](https://github.com/looker-open-source/components/blob/main/packages/components/src/utils/useWindow.tsx#L177). Probably a separate `disabled` prop would be better there.

File: packages/components/src/utils/useMeasuredElement.ts [hunk 2]
```diff
@@ -29,22 +29,24 @@ import { useResize } from './useResize'
...
+    return typeof DOMRect === 'function'
+      ? new DOMRect()
+      : {
+          bottom: 0,
+          height: 0,
+          left: 0,
+          right: 0,
+          top: 0,
+          width: 0,
+        }
```
lukelooker, line 41: @mdodgelooker This isn't my favorite workaround but the latter case should only get triggered in browsers that are missing DOMRect (IE11 & JSDom) and both are already getting this behavior so this seems like a improvement, albeit imperfect, fix.

dbchristopher (approved):
👍

PR #2208, opened 2021-04-16
fix(List): Only define height when windowing

mdodgelooker:
The `height: 100%` style on `List` is necessary for windowing to work, but it can be problematic for non-scrolling, non-windowed lists when the container's height is defined (say 100% of the window height) but `List` is not the only child and is not intended to take up the full height of the container. For example, this page shouldn't need to scroll: https://codesandbox.io/s/stupefied-sanne-9vsm1

This PR removes `height: 100%` when `windowing` is `none`.

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/config/i18nIgnoredProps.js
+++ b/config/i18nIgnoredProps.js
@@ -34,6 +34,7 @@ const ours = [
   'iconVerticalAlign',
   'iconViewBox',
   'textDecoration',
+  'windowing',
 ]
 const aria = [
   'aria-autocomplete',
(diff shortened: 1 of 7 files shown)


lukelooker (dismissed):
LGTM. Thanks for taking this on :)

lukelooker (approved):
👍

PR #2213, opened 2021-04-19
docs(DataTable): Update examples with deprecated widthPercent prop (size is correct)

lukelooker:
- [x] 📚 Documentation updated


jhardy (approved):
LGMT

PR #2215, opened 2021-04-19
chore: Ignore stories in coverage reporting

lukelooker:


lukelooker:
Slightly alters the calculation for coverage but makes it easy to ignore stories. I lean towards merging with the small loss of coverage this indicates to get the simpler configuration win.

packages/components/src/Tree/utils/generateTreeBorder.ts
```diff
@@ -24,7 +24,6 @@
 
  */
 
-/* istanbul ignore file */
```
mdodgelooker, line 27: Was this here because the output is unreasonably difficult to test @ElliotTheHuman ?

@lukelooker are you thinking it's better to live with the coverage loss here and maybe someday revisit when the tools are better? 🙂

mdodgelooker (approved):
Cool! Just a qq about `generateTreeBorder.ts`.

packages/components/src/Tree/utils/generateTreeBorder.ts
lukelooker: Yep - I'd prefer not to disable coverage unless the code in question isn't included in the build artifact (documentation, examples or test fixtures all being cases where I think ignoring is probably fine)

PR #2216, opened 2021-04-19
feat(List): support fontFamily as prop

laraharrow:
- [x] 📚 Documentation updated
- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11


packages/components/src/List/List.story.tsx
```diff
@@ -59,6 +59,12 @@ Basic.args = {
   children: listItems,
 }
 
+export const FontFamily = Template.bind({})
+FontFamily.args = {
+  children: listItems,
+  fontFamily: 'code',
+}
+
```
lukelooker, line 67: Needs a screenshot to be added manually :)

packages/components/src/List/List.tsx
```diff
@@ -57,17 +58,19 @@ export interface ListProps
...
-   */
-  keyColor?: boolean
-
   /**
    * If true, all ListItem children without an icon will reserve space for an icon
    * for alignment purposes.
    */
   iconGutter?: boolean
 
+  fontFamily?: FontFamilies
```
lukelooker, line 67: Needs JSDoc explanation.

www/src/documentation/components/content/list.mdx
```diff
@@ -18,7 +18,26 @@ import { ListItemPropTable } from './demos'
...
+  <ListItem icon={<Favorite />} detail="Switzerland">
+    Swiss
+  </ListItem>
+</List>
+```
+
+Use the `fontFamily` prop to change the font on all text elements in the `List`.
+
+```jsx
+<List fontFamily="Impact">
```
lukelooker, line 30: I'd prefer showing folks a `theme` reference here. Even though they _can_ specify an explicit font-face I'd prefer them to think in a theme world :)

```suggestion
<List fontFamily="code">
```

lukelooker (changes requested):

packages/components/src/List/List.tsx
```diff
@@ -162,12 +165,17 @@ export const ListInternal = forwardRef(
...
   ${reset}
   ${height}
   ${width}
 
   list-style: none;
   overflow: auto;
+
+  button {
+    ${fontFamily}
+  }
```
dbchristopher, line 178: I wonder if this selector is actually too narrow in this case? It seems like users could potentially render any form element in a list which might not inherit the type settings as expected. Maybe it makes sense to use the `*` selector instead?

```suggestion
  * {
    font-family: inherit;
  }
```

What do you think?

lukelooker: This shouldn't be needed if the branch is off the latest `main` - `ListItem` _should_ just inherit the `List` font-family correctly.

lukelooker (changes requested):

packages/components/src/List/List.story.tsx
laraharrow: if you look at the files on gitHub you will see the snapshot file. Not sure if I'm missing something here.

lukelooker: The PR: https://github.com/looker-open-source/components/pull/2217 needs to get merged into this one so that the snapshot PNG is included.

packages/components/src/List/List.tsx
laraharrow: I updated to use `*` instead of `button` great suggestion thanks. As far as removing the code all together I just updated to be in the same place as main but if I remove it the font does not get updated. Not sure if I'm missing something here.

packages/components/src/List/List.story.tsx
laraharrow: I did add the snapshot thou. I see the same file here just a little lower in this page.

packages/components/src/List/List.tsx
lukelooker: I'm not a big fan of having a component reach into another component like this - with the refactor of the `reset` behavior it _should_ be possible to simply leverage CSS inheritance to get this to work.

Checking out branch locally to see if I can understand why it doesn't work without the specific selector here.

laraharrow: I updated to use * instead of button great suggestion thanks. As far as removing the code all together I just updated to be in the same place as main but if I remove it the font does not get updated. Not sure if I'm missing something here.

packages/components/src/List/List.tsx
```diff
@@ -162,12 +168,17 @@ export const ListInternal = forwardRef(
 const ListStyle = styled.ul.withConfig({ shouldForwardProp })<
   HeightProps & WidthProps
 >`
+  ${fontFamily}
   ${reset}
```
lukelooker, line 172: `reset` is overriding `fontFamily` because it specifies `font-family: inherit` _after_ `fontFamily` helper emits its preferred font-family.

I'd recommend re-working `ListStyle` like so:

```jsx
const ListStyle = styled.ul
  .withConfig({ shouldForwardProp })
  .attrs<ListProps>(({ fontFamily = 'inherit' }) => ({
    fontFamily,
  }))<ListProps>`
  ${fontFamily}
  ${height}
  ${width}

  list-style: none;
  margin: 0;
  overflow: auto;
  padding: 0;
`
```

This directly injects the "reset" behaviors into the style block (`margin: 0` & `padding: 0` and leverages a default value on `fontFamily` so that the existing behavior stays intact.

packages/components/src/List/List.tsx
```diff
@@ -57,17 +58,22 @@ export interface ListProps
...
-  keyColor?: boolean
-
   /**
    * If true, all ListItem children without an icon will reserve space for an icon
    * for alignment purposes.
    */
   iconGutter?: boolean
 
+  /**
+   * Use this prop to specify themed fonts for the List text.
```
lukelooker, line 68:
```suggestion
   * Specify font-family. Generally will end up inheriting `theme.fonts.body` but can be specified as `brand`, `code` or `body` to explicitly specify theme-controlled font-family
   * default inherit
```

lukelooker (changes requested):

lukelooker (approved):
LGTM! :)

PR #2220, opened 2021-04-20
chore(deps): Styled Components 5.2.3 (+ @types / jest-styled-components)

lukelooker:


storybook/src/Theme/Theme.story.tsx
```diff
@@ -25,16 +25,16 @@
...
-import React, { FC } from 'react'
+import React from 'react'
 import { Editor, EditorProps } from './Editor'
 
-export const ThemeEditor: FC<EditorProps> = (props) => (
-  <>
-    <Editor {...props} />
-  </>
-)
+export const ThemeEditor = (props: EditorProps) => <Editor {...props} />
```
dbchristopher, line 31: any particular reason not to use `FC<>` anymore?

dbchristopher (approved):

storybook/src/Theme/Theme.story.tsx
lukelooker: Typescript gets angry when attempting to assign the `ThemeEditor.parameters = { storyshots: false }` when `FC` is assigned (because `FC` doesn't support `parameters`). I could've re-cast to `Story` but simply making them pure functions was less lines and didn't seem detrimental in this case.

dbchristopher: Ah yes, that makes sense!

PR #2221, opened 2021-04-20
chore(deps): Typescript 4.2.4

lukelooker:


packages/components/src/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -75,21 +81,20 @@ export const CopyToClipboard: FC<CopyToClipboardProps> = (props) => {
...
-      >
-        {children}
-      </ButtonOutline>
-    ) : (
-      cloneElement(children, {
-        onClick: clickCopyButton,
-        ref: buttonRef,
-      })
-    )
+  const copyButton = isValidElement(children) ? (
```
lukelooker, line 84: TS got smarter and discovers that previous type-guard of `typeof children === 'string'` is insufficient. Using `isValidElement` is a better solution so I inverted logic to leverage that and put it up-front.

packages/components/src/utils/getCurrentNode.ts
```diff
@@ -36,7 +36,6 @@ export function getCurrentNode<E extends HTMLElement = HTMLElement>(
   elementOrRefObject: ElementOrRef<E>
 ): E | null {
   if (!elementOrRefObject) return null
-  // @ts-expect-error: Need to find a better way to do union below
```
lukelooker, line 39: TS 4.2.x can now properly infer this union. 🎉

mdodgelooker: Oh neat!

mdodgelooker (approved):
👍

PR #2223, opened 2021-04-20
chore(deps): Loosen package.json resolutions

lukelooker:
Generally we use package.json resolutions to REQUIRE a newer version of a transient dependency to prevent the usage of transient dependencies with known vulnerabilities.

This PR loostens that requirement to require the fixed version or NEWER instead of explicitly locking to the fix version specifically

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,18 +102,18 @@
   },
   "resolutions": {
     "axios": ">= 0.21.0",
-    "eslint-plugin-react-hooks": "4.2.0",
-    "highlight.js": "10.4.1",
-    "immer": "8.0.1",
-    "is-svg": "4.2.2",
-    "node-fetch": "2.6.1",
-    "node-forge": "0.10.0",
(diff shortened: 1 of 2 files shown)


lukelooker:
@mdodgelooker - If you have a chance can you give this a look? Hoping to land this so I can get a few more dependency upgrade PRs rolling (I suspect `yarn.lock` changes will conflict so trying to keep number of concurrents low :)

mdodgelooker:
👀

mdodgelooker (approved):
LGTM

Each of those resolutions were due to an issue with older version of the package, right – never with a newer version?

lukelooker:
> Each of those resolutions were due to an issue with older version of the package, right – never with a newer version?

Yep, all are cases where a transient dependency had/has a vulnerability so we wanted to make sure that the package is always installed with the fix version [or better]. Before the specificity meant that even if the dependency specified a newer version our explicitly specified version would win.

PR #2227, opened 2021-04-20
feat(Aside): Standardize width sizing

laraharrow:
- [ ] 📚 Documentation updated
- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11


packages/components/src/Layout/Semantics/Aside.tsx
```diff
@@ -41,7 +43,7 @@ export interface AsideProps extends SemanticLayoutBase, SemanticBorderProps {
    * Specify width of aside
    * @default '12rem'
    */
-  width?: string
+  width?: ResponsiveValue<DialogSizeRamp | string>
```
lukelooker, line 46: `DialogSizeRamp` should be lifted out of `Dialog` and renamed to something like `LayoutWidthRamp` and placed in the `/Layout` folder.

Then `dialogWidth` should be updated to reference that central, more generic version.

packages/components/src/Layout/Semantics/Aside.tsx
```diff
@@ -56,6 +58,8 @@ export const Aside = styled.aside
   min-width: ${({ width }) => width};
   overflow: auto;
   width: 0;
-  ${borderHelper}
   ${({ scrollWithin }) => scrollWithin && 'height: fit-content;'}
+
+  ${borderHelper}
+  ${dialogWidth}
```
lukelooker, line 64: Seems really odd to have `dialogWidth` specifying the `aside`'s width. Perhaps create a `layoutWidth` helper that can be used in both places?

packages/components/src/Layout/Semantics/Semantics.test.tsx
```diff
@@ -152,4 +152,16 @@ describe('Semantics', () => {
       height: 'fit-content',
     })
   })
+
+  test('Aside can use t-shirt sized for its width.', () => {
+    renderWithTheme(
+      <Layout hasAside>
+        <Aside width="medium">Aside content</Aside>
+        <Section as="main">Main content</Section>
+      </Layout>
```
lukelooker, line 161: Example can be narrowed down to just the `Aside` itself.

```suggestion
      <Aside width="medium">Aside content</Aside>
```

packages/components/src/Layout/Semantics/Story/ExamplePage.story.tsx
```diff
@@ -345,3 +345,16 @@ export const ScrollAllAreas = () => (
...
+  <Highlighter>
+    <Layout hasAside>
+      <Aside p="xxlarge" width="xsmall">
+        Aside content
+      </Aside>
+      <Section p="xxlarge" as="main">
+        {longContent}
+      </Section>
+    </Layout>
+  </Highlighter>
```
lukelooker, line 359: Narrow to just the `Aside` and make this an `arg` that the Storybook user can play with. See Dialog story at https://looker-open-source.github.io/components/latest/storybook/?path=/docs/dialog--basic

https://github.com/looker-open-source/components/blob/main/packages/components/src/Dialog/stories/Dialog.story.tsx#L169-L186

image

packages/components/src/Layout/Semantics/Aside.tsx
```diff
@@ -56,6 +58,8 @@ export const Aside = styled.aside
   min-width: ${({ width }) => width};
   overflow: auto;
   width: 0;
-  ${borderHelper}
   ${({ scrollWithin }) => scrollWithin && 'height: fit-content;'}
+
```
lukelooker, line 62: `attrs` above should default `width = 'xsmall`` I guess?

Also `max-width` and `min-width` should probably be removed to not conflict with `dialogWidth` (to be renamed)

lukelooker (changes requested):
As you pointed out on chat we'll need to rationalize the sizes a bit to match-up with `Aside`'s existing default.

Since `Aside` (and all of the layout components for that matter) are marked as "experimental" I think we can change the default safely so I'd probably just go with `xsmall` for now and we can iterate from there.

laraharrow:
@lukelooker I was wondering if it make sense to lift `dialogWidth` just because maybe we will want to use other sizes then what is set for `Dialog` and `Drawer` for instance right now Aside width is set to `12rem` and the smaller size on that file is `xxsmall` and it is set as `16rem`

lukelooker:
@laraharrow - I definitely think we'll want to *add* a "LayoutWidths" - I'm admittedly not sure if we could use that same collection of sizes in Drawer & Dialog or not so perhaps that's a good reason to keep them seperate for now. I'd say you could just copy the existing values for DialogWidths and then @jhardy & I can audit usage within Helltool and come up with an ideal list.

packages/components/src/Layout/Semantics/Story/ExamplePage.story.tsx
laraharrow: Hey @lukelooker I added `width` to the argument list on Storybook. But changing the values doesn't change the `Aside`'s width - Same seemed true to me in Dialog.

lukelooker: Ah - weird, yeah it looks like that example isn't working. 

Drawer (https://looker-open-source.github.io/components/latest/storybook/?path=/story/drawer--basic) does appear to work so hopefully you can borrow from there and perhaps take on getting `Dialog` working properly as well?

laraharrow: sure thing!

lukelooker (changes requested):
See notes in Buganizer about next steps.

packages/components/src/Layout/Semantics/LayoutWidth.ts
```diff
@@ -0,0 +1,61 @@
...
+  | SizeMedium
+  | SizeLarge
+
+export type LayoutSizeRamp = Record<LayoutSizes, string>
+export type LayoutWidth = ResponsiveValue<LayoutSizeRamp | string>
+
+/* eslint-disable sort-keys-fix/sort-keys-fix */
+export const layoutSizes: LayoutSizeRamp = {
+  xxsmall: '16rem',
+  xsmall: '12rem',
```
lukelooker, line 49: Set back to `21rem`

mdodgelooker (approved):
LGTM!

packages/components/src/Layout/Semantics/LayoutWidth.ts
```diff
@@ -0,0 +1,68 @@
...
+  | SizeSmall
+  | SizeMedium
+  | SizeLarge
+
+export type LayoutSizeRamp = Record<LayoutSizes, string>
+
+export type LayoutWidth = ResponsiveValue<LayoutSizeRamp | string>
+
+/* eslint-disable sort-keys-fix/sort-keys-fix */
+export const layoutSizes: LayoutSizeRamp = {
```
jhardy, line 51: I have mixed feelings about combining t-shirt sizes with named values like `navigation` and `rail`.  One idea Luke and I had talked about was naming these items like `size="rail"` instead of width as the prop. 

I'm questioning if we even need the t-shirt sizes here? Should width just be `'rail' | 'navigation' | 'sidebar' | number`?

packages/components/src/Layout/Semantics/Story/ExamplePage.story.tsx
```diff
@@ -345,3 +351,33 @@ export const ScrollAllAreas = () => (
...
+AsideDefaultWidthSizeXsmall.args = {}
+
+export const AsideWidthSizeNavigation = WidthTemplate.bind({})
+AsideWidthSizeNavigation.args = {
+  width: 'navigation',
+}
+
+export default {
+  argTypes: {
+    width: {
```
jhardy, line 374: having these width controls be apart of the default makes it Look like they are available on every layout story. Is there a way to limit them just to the `Aside` stories?

packages/components/src/Layout/Semantics/LayoutWidth.ts
jhardy: the name here `layoutSizes` seems again like it would apply to all layout pieces, but `rail`, `navigation`, and `sidebar` apply to just `Aside`

PR #2233, opened 2021-04-21
feat(Tree): forceLabelPadding prop

ElliotTheHuman:
Adds the `forceLabelPadding` prop to `Tree`. Also refactors some of the indent-related util functions to:
A) Support the `forceLabelPadding` prop
B) Changed the parameter structure to take an object (since they now need 4+ fields to work out the logic)

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


lukelooker (approved):
Looks great and works as specified :)

`generateIndent` definitely got complicated here but I don't see a great alternative immediately and on the plus side it's an internal implementation detail so if a simpler mechanism occurs to someone at a later point it'd be a non-breaking change to refactor.

PR #2235, opened 2021-04-21
fix(Accordion): Correctly infer icon style from indicatorPosition="left"

lukelooker:
Also changes `Fieldset` icon size from `medium` to `small` to be consistent with `Accordion`

- [x] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari


packages/components/src/Accordion/accordionDefaults.tsx
```diff
@@ -44,7 +46,7 @@ export interface AccordionContextProps {
   onOpen?: () => void
 }
 
-export const accordionContextDefaults: AccordionContextProps = {
+export const accordionDefaults: AccordionDefaultProps = {
```
ElliotTheHuman, line 49: I think we're safe to scrap the control-related props from `accordionDefaults` (and from the interface).

These don't have a "default" per say, but some of them previously had default values for typing purposes.

ElliotTheHuman, line 49: I think we're safe to scrap the control-related props from accordionDefaults (and from the interface).

These don't have a "default" per say, but some of them previously had default values for typing purposes.

ElliotTheHuman (changes requested):
Just a minor comment about the `AccordionContext => accordionDefaults` conversion. Outside of that, looks shway!

packages/components/src/Accordion/accordionDefaults.tsx
lukelooker: Ah yes! Fixed! :)

ElliotTheHuman (approved):
Shway!

PR #2238, opened 2021-04-22
fix(useHovered): Invalid argument error

mdodgelooker:
`useHovered` was typecasting `as Node` on arguments that might be `null` in order to call `compareDocumentPosition` and detect if a child of the hover element was currently focused. This caused an `Invalid argument` error in IE11 (though no other breakage, as far as I could tell).

This PR adds a check before calling `compareDocumentPosition` and removes the typecasting.

I also added a test suite.

- [ ] 📚 Documentation updated
- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11

### DIFF
--- a/packages/components/src/Menu/Menu.story.tsx
+++ b/packages/components/src/Menu/Menu.story.tsx
@@ -250,7 +250,7 @@ export const Hover = () => {
   const close = () => setOpen(false)
 
   return (
-    <Card ref={hoverRef} p="large" raised height="auto">
+    <Card ref={hoverRef} p="large" raised height="auto" tabIndex={0}>
       <Space between>
         <Paragraph>
           Hovering in this card will show the button that triggers the menu.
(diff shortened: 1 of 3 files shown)


packages/components/src/Menu/Menu.story.tsx
```diff
@@ -250,7 +250,7 @@ export const Hover = () => {
   const close = () => setOpen(false)
 
   return (
-    <Card ref={hoverRef} p="large" raised height="auto">
+    <Card ref={hoverRef} p="large" raised height="auto" tabIndex={0}>
```
mdodgelooker, line 253: Making this focusable to test `useHovered`'s focusin/focusout behavior.

lukelooker (approved):
Nice!

PR #2239, opened 2021-04-22
chore(deps): Update testing-library & (most) webpack dependencies

lukelooker:


packages/components/src/Form/Inputs/InputSearch/InputSearch.test.tsx
```diff
@@ -151,7 +151,7 @@ describe('InputSearch', () => {
       // Close popover to silence act() warning
       fireEvent.click(document)
       // Value is cleared on list close by default
-      expect(input).toHaveDisplayValue('')
+      expect(input).toHaveDisplayValue(['F'])
```
mdodgelooker, line 154: Oops, this test was always wrong. The value shouldn't be cleared unless `clearOnClose` is true. Maybe something tricky about the difference b/t `toHaveValue` and `toHaveDisplayValue`... anyway, might as well delete this line as well as the comment above it.
```suggestion
```

mdodgelooker (changes requested):

mdodgelooker (approved):

PR #2246, opened 2021-04-23
chore: Jest updates to support Filter Packages

mdodgelooker:
- Deleted the `i18next` mock file. Long story short it's actually not needed (just the `react-i18next` mock file is) and it was interfering with tests in `@looker/filter-expressions`.
- Changed some jest setup & config stuff to get tests passing

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


mdodgelooker:
@dbchristopher and @lukelooker here are the jest changes I needed to make to get it passing after copying over the filter packages.

I included the bash script in this PR in case you want to run it and see it working on those filter packages – but I'll remove it before merging. You would run this command, though you might need to adjust permissions on the file (let me know if you run into that):
```
SOURCE_PATH=name-of-core-product-folder-goes-here ./bin/filter-packages.sh
```

If you do run the script, there is still one jest failure in `Filter.spec.tsx`, but that is being fixed in the current upgrade PR.

packages/components/src/Breakpoint/Breakpoint.test.tsx
```diff
@@ -38,6 +38,14 @@ import { Breakpoint } from './Breakpoint'
...
 
 describe('Breakpoint', () => {
+  Object.defineProperty(HTMLElement.prototype, 'clientWidth', {
+    configurable: true,
+    value: 800,
+  })
+  Object.defineProperty(HTMLElement.prototype, 'clientHeight', {
+    configurable: true,
+    value: 600,
+  })
```
dbchristopher, line 48: We already [define these in jest.setup.js](https://github.com/looker-open-source/components/blob/main/jest.setup.js#L42-L49). Is it just the `configurable: true` property that's needed? I imagine you could just update the original statement rather than redefining the property here.

dbchristopher: Otherwise if the change is only relevant to this file, then we'd probably need to tear these back down in an `afterEach` so as not to bleed into other tests

dbchristopher: Ok sorry, I see that was in fact removed from jest.setup altogether. I really need to read the whole PR before commenting! Is there any reason to move it here instead? These probably would still bleed into every test ran after `Breakpoint` without an afterEach reset

mdodgelooker (commented):
Ack forgot to submit my comments, sorry:

packages/components/src/Breakpoint/Breakpoint.test.tsx
mdodgelooker: So sorry, I had a comment about the move but neglected to submit. I’m still a little bit at a loss as to why the overrides in the RelativeTimeframes were getting clobbered, but it is probably better to have them restricted to the test where they’re needed. Great point about reset though - I’ll add that...

mdodgelooker:
Updated `Breakpoint.test.tsx` with `spyOn` and `afterEach` resets.

bin/filter-packages.sh
```diff
@@ -0,0 +1,42 @@
...
+  # copy the tsconfig
+  cp packages/components/tsconfig.json packages/filter-$name/tsconfig.json
+
+  # use version # for @looker/sdk in package.json
+  sed -i "s/@looker\/sdk\": \"\*/@looker\/sdk\": \"$sdkVersion/g" packages/filter-$name/package.json
+  # update main path, add module, files, types, etc
+  sed -i "s/\"main\": \"src\/index.ts\",/$packageAdd/g" packages/filter-$name/package.json
+  # remove @looker/cli
+  sed -i '/"@looker\/cli": "\*",/d' packages/filter-$name/package.json
+  # remove scripts block
```
lukelooker, line 34: Nit prefer not to expose `core product` package names to the world - I suppose this will resolve soon since it's likely going away from the filters package soon anyway?

jest.config.js
```diff
@@ -41,7 +41,8 @@ module.exports = {
   moduleDirectories: ['./node_modules', './packages'],
   moduleFileExtensions: ['js', 'jsx', 'ts', 'tsx', 'json', 'node'],
   moduleNameMapper: {
-    '@looker/([^/]+)$': '<rootDir>/packages/$1/src',
+    '@looker/components/lib/date': '<rootDir>/packages/components/src/date',
```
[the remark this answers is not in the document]
lukelooker: Probably an argument to just move these to `@looker/components-date` or similar

jest.setup.js
```diff
@@ -42,15 +42,6 @@ globalAny.IntersectionObserver = observeMock
...
 
-// Ensure jest viewport has real dimensions to support responsively rendering content
-Object.defineProperty(document.body, 'clientWidth', {
-  value: 800, // match storyshots default viewport
-})
-
-Object.defineProperty(document.body, 'clientHeight', {
-  value: 600, // match storyshots default viewport
-})
-
```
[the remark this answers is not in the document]
lukelooker: This also breaks some stuff in Storybook 6.2 (blocking upgrade currently) so I suspect we'll be looking for a new strategy around this stuff in general. (I'm leaning towards using Loki or another tool instead of Storyshots)

lukelooker (dismissed):

mdodgelooker:
Deleted the bash script since it's just for dev purposes.

lukelooker (approved):

PR #2247, opened 2021-04-23
chore: Fix typescript-eslint dependency version and related lint breakages

lukelooker:
16 errors remaining all are of these two types:

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,6 +103,7 @@
   "resolutions": {
     "axios": ">= 0.21.0",
     "eslint-plugin-react-hooks": ">= 4.2.0",
+    "@typescript-eslint/eslint-plugin": "4.22.0",
     "highlight.js": ">= 10.4.1",
     "immer": ">= 8.0.1",
     "is-svg": ">= 4.2.2",
--- a/packages/components/src/Form/Inputs/InputTimeSelect/InputTimeSelect.tsx
+++ b/packages/components/src/Form/Inputs/InputTimeSelect/InputTimeSelect.tsx
@@ -25,7 +25,7 @@
  */
 import { useTranslation } from 'react-i18next'
 import React, {
-  useCallback,
+  useMemo,
   forwardRef,
   KeyboardEvent,
   useState,
@@ -320,15 +320,15 @@ const InputTimeSelectLayout = forwardRef(
       }
     }
 
-    /* eslint-disable-next-line @typescript-eslint/unbound-method */
-    const throttledHandleChange = useCallback(
-      throttle(
-        (v: MaybeComboboxOptionObject) => {
-          handleChange(v)
-        },
-        50,
-        { trailing: false }
-      ),
+    const throttledHandleChange = useMemo(
+      () =>
+        throttle(
+          (v: MaybeComboboxOptionObject) => {
+            handleChange(v)
+          },
+          50,
+          { trailing: false }
+        ),
       [handleChange]
     )
 
--- a/packages/components/src/Form/Inputs/InputTimeSelect/InputTimeSelect.tsx
+++ b/packages/components/src/Form/Inputs/InputTimeSelect/InputTimeSelect.tsx
@@ -310,26 +310,22 @@ const InputTimeSelectLayout = forwardRef(
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [value])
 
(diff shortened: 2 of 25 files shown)


package.json
```diff
@@ -103,6 +103,7 @@
   "resolutions": {
     "axios": ">= 0.21.0",
     "eslint-plugin-react-hooks": ">= 4.2.0",
+    "@typescript-eslint/eslint-plugin": "4.22.0",
```
lukelooker, line 106: Ignore the terribly old version that Gatsby tries to force on us.

packages/components/src/Panel/Panel.test.tsx
```diff
@@ -218,11 +218,11 @@ describe('Panel', () => {
     expect(screen.getByText('My neat dialog')).toBeInTheDocument()
   })
 
-  test('triggers console.warn if no children is passed', () => {
-    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  test('triggers console.warn if no children are passed', () => {
+    /* eslint-disable-next-line @typescript-eslint/ban-ts-comment */
     // @ts-ignore
```
lukelooker, line 223: ESLint / TS would *really* prefer we not do this :)

packages/eslint-config/index.js
```diff
@@ -99,6 +100,7 @@ module.exports = {
       },
     ],
     'no-undef': 'off', // Not needed with TS around
+    'no-use-before-define': 'off',
```
dbchristopher, line 103: Is this mainly because of our pattern to define styled components at the bottom of the file?

lukelooker: Yep! Frankly, it always strikes me as a funny lint rule since JS has long supported hoisting and with TS in play we'll definitely know if something isn't defined at all.

packages/components/src/Menu/MenuHeading.hooks.ts
```diff
@@ -29,19 +29,19 @@ import { RefObject, useEffect, useState } from 'react'
...
+        ? null
+        : new IntersectionObserver(
+            ([entry]) => {
+              setIsVisible(entry.intersectionRatio > 0)
+            },
+            {
+              threshold: [0, 1],
+            }
+          )
+
```
dbchristopher, line 44: resolves warning:

>/usr/local/google/home/dbchristopher/components/packages/components/src/Menu/MenuHeading.hooks.ts
>  32:9  warning  The 'observer' conditional could make the dependencies of useEffect Hook (at line 54) change on every render. Move it inside the useEffect callback. Alternatively, wrap the initialization of 'observer' in its own useMemo() Hook  react-hooks/exhaustive-deps

packages/components/src/Form/Inputs/InputTimeSelect/InputTimeSelect.tsx
```diff
@@ -310,26 +310,22 @@ const InputTimeSelectLayout = forwardRef(
...
+              ? newSelectedOption.value
+              : undefined
+            if (isFunction(onChange) && isValidTime(newValue)) {
+              onChange(newValue)
+            }
+          },
+          50,
+          { trailing: false }
+        ),
+      [onChange]
```
dbchristopher, line 328: Apparently `useMemo` is preferred to `useCallback` for stuff like this:
https://github.com/facebook/react/issues/19240#issuecomment-652945246

Also merging `handleChange` and `throttledHandleChange` resolve issue where callback dependency could change on every render

packages/components/src/Panel/Panel.tsx
```diff
@@ -35,9 +35,7 @@ const isRenderProp = (
 export const Panel: FC<PanelProps> = ({ children, content, ...props }) => {
   const { domProps, panel } = usePanel({ content, ...props })
 
-  if (children === undefined) {
-    return <>{panel}</>
-  } else if (isValidElement(children)) {
```
dbchristopher, line 40: This actually _wasn't_ conforming to the behavior described in that test `'triggers console.warn if no children are passed'`.

With this logic branch deleted, the console.warn happens in the final `else` block. And the default return should still be fine in cases where children is undefined, right?

```js
return (
    <>
      {children}
      {panel}
    </>
  )
```

lukelooker: Yep! Right-o! Thanks for catching that.

mdodgelooker (approved):
LGTM!

PR #2249, opened 2021-04-23
chore: Fix text coverage calculation to ignore generated & TS-only files

lukelooker:
Minor changes to coverage calculation to (correctly) ignore: 

  - `packages/src/icons` (dynamically generated code) 
  - `index.ts` / `types.ts` files that are simple pass-throughs (used properly TS coverage would make us aware of anything gone wrong in thse.


mdodgelooker (approved):
Nice

PR #2263, opened 2021-04-27
feat(IconButton): Support for toggleColor

lukelooker:
IconButton can now support icon and background color when toggled

- [x] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari


packages/components/src/Button/IconButton.tsx
```diff
@@ -142,6 +60,7 @@ const IconButtonComponent = forwardRef(
       label,
       toggle,
       toggleBackground,
+      toggleColor = 'key',
```
ElliotTheHuman, line 63: It might be nice to create a constant for the places in `IconButton.tsx` and `iconButtonColor.ts` where "key" is used as a fallback / default value.

```
const DEFAULT_COLOR = 'key'
```

That being said, I can see argument for not coupling color and background-color related functions too tightly.

www/src/documentation/components/actions/icon-button.mdx
```diff
@@ -96,6 +96,27 @@ Icon buttons have the `toggleBackground` prop which can be used to show a backgr
 </Space>
 ```
 
+### ToggleColor
+
+Icon buttons have the `toggleColor` prop which can be used to specify a theme-derived background color to the Icon. The `toggle` prop has to be explictly set to either `true` or `false`.
```
ElliotTheHuman, line 101:
```suggestion
Icon buttons have the `toggleColor` prop which can be used to specify what the icon's color and background color are when `toggle = true`. It accepts the following values:
- 'key'
- 'calculation' 
- 'dimension' 
- 'measure'

**Note:** The `toggle` prop has to be explicitly set to either `true` or `false`.
```

Tried to add more specificity to the doc portion, but this is just a nit (i.e. feel free to disregard).

ElliotTheHuman (dismissed):
Some minor feedback, but they're more suggestions than hard blockers.

Feel free to implement suggestions or merge as is!

packages/components/src/Button/IconButton.tsx
lukelooker: Great point! Updated! :)

ElliotTheHuman (approved):
:shipit: !!

PR #2269, opened 2021-04-29
fix(RangeSlider): Value stays within bounds when min/max changes

mdodgelooker:
Reverts change in https://github.com/looker-open-source/components/commit/6b1ae1945f9c3951ceede4bb980ba778c95955cd to fix an issue where changing the min/max doesn't update the value to stay within bounds.
Repro here: https://codesandbox.io/s/determined-breeze-jwi25


lukelooker (approved):
LGTM

PR #2273, opened 2021-04-29
feat(Pagination): and PageSize support alwaysVisible - visible even when there are no pages to paginate

lukelooker:


ElliotTheHuman:
Commit 902e7fb
- Added logic to prevent value of `PageSize` from being greater than `total`
- Added a `<span>` wrapper to `of X` JSX for testing purposes (without this, jest fails to find that string because `getByText` matches based on exact text children not partial as far as I understand)

ElliotTheHuman:
Also added auto-disable to `PageSize`'s internal `Select` if all option values > total value.

ElliotTheHuman (approved):
🚀

PR #2277, opened 2021-04-30
feat: useFocusVisible hook

mdodgelooker:
### Changes
- **Initial problem: visible focus was turning off when clicking via spacebar**
- The behavior was duplicated across many components, so I consolidated into a `useFocusVisible` hook
- Fixed the spacebar issue simply not turning off `focusVisible` in the `onClick`. This follows browser behavior for the `:focus-visible` pseudo-selector
- Added a CSS utility for applying styles via `focusVisible` _OR_ `:focus-visible` – the latter is supported only in Chrome & Firefox, but it does capture one scenario that `useFocusVisible` misses (see comment in code).
- **Affected components**
  - Accordion (disclosure)
  - Button
  - IconButton
  - ButtonGroup/Toggle
  - DataTable (cell)
  - Tab

### Tests
- `useFocusVisible.test.tsx`
- Added "Focused" stories for snaphsot-testing purposes
- Added CSS in `manager-head.html` to hide these stories in the storybook sidebar, since they just look like duplicates

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/Tabs/Tab.tsx
```diff
@@ -55,7 +60,7 @@ const TabStyle = styled.button
   .withConfig({ shouldForwardProp })
   .attrs(({ type = 'button' }) => ({
     type,
-  }))<TabProps>`
+  }))<TabProps & FocusVisibleProps>`
```
mdodgelooker, line 63: Moved `focusVisible` down here, it should only be used internally.

packages/components/src/Tabs/Tab.tsx
```diff
@@ -107,46 +112,34 @@ const TabJSX = forwardRef((props: TabProps, ref: Ref<HTMLButtonElement>) => {
...
-  }
-
-  const onClick = () => {
+  const handleClick = useWrapEvent(() => {
     if (!disabled && onSelect) {
       onSelect()
     }
-    setFocusVisible(false)
-  }
+  }, onClick)
```
mdodgelooker, line 126: Realized `props.onClick`, if it exists, was not getting called.

packages/components/src/Tree/Tree.tsx
```diff
@@ -57,7 +57,6 @@ const TreeLayout: FC<TreeProps> = ({
   keyColor: propsKeyColor,
   label: propsLabel,
   labelBackgroundOnly: propsLabelBackgroundOnly,
-  onClick,
```
mdodgelooker, line 60: Was this left unused for a reason? By removing it, it goes onto `Accordion` along with `...restProps`.

packages/components/src/Tree/types.ts
```diff
@@ -37,7 +37,7 @@ export interface TreeProps
       | 'indicatorPosition'
       | 'indicatorSize'
     >,
-    ListItemProps {
+    Omit<ListItemProps, 'onClick'> {
```
mdodgelooker, line 40: Due to adding `useClickable` and `GenericClickProps` in `AccordionDisclosure`, the `onClick`s had a TS conflict. Now it's going onto `Accordion` (see comment above) so it should get the onClick type from there.

packages/components/src/Accordion/AccordionDisclosure.tsx
```diff
@@ -160,13 +161,11 @@ export const AccordionDisclosureStyle = styled.div
...
-    focusVisible &&
+  ${focusVisibleCSSWrapper(
+    ({ theme }) => css`
+      box-shadow: inset 0 0 0 2px ${theme.colors.keyFocus};
     `
-      &:focus {
-        box-shadow: inset 0 0 0 2px ${theme.colors.keyFocus};
-      }
-    `}
+  )}
```
mdodgelooker, line 168: This helper function inserts the returned style inside `:focus-visible` for browsers that support it (everyone except Safari and IE11) and toggles it on when `focusVisible` is true, for IE11 & Safari, and for the cases where the element receiving the styling is a parent of the focused element (see `ListItem`).

`:focus-visible` is useful because it's able to capture a situation that we can't with `useFocusVisible`: 
1. User hits the escape key on an open `Menu`
2. The menu closes, and focus moves back to the button
3. `useFocusVisible` relies on the keyup event, but in this situation, focus moves to the button after that, so `focusVisible` is not updated to true for the button.

packages/components/src/Accordion/Accordion.story.tsx
```diff
@@ -53,7 +55,25 @@ export const LegacyComposition = () => (
...
+Focused.args = {
+  ...Basic.args,
+}
+
+Focused.parameters = {
+  beforeScreenshot: async (page: Page) => {
+    const disclosure = await page.$('[role="button"]')
+    await disclosure?.type(' ')
+  },
+}
```
mdodgelooker, line 76: Added these stories for the various components to generate image snapshots of the focus visible styling. When you run storybook in the browser they all just look like duplicates, so I added CSS in `manager-head.html` to hide them from the sidebar (see below).

packages/components/src/Accordion/Accordion.story.tsx
```diff
@@ -53,7 +55,26 @@ export const LegacyComposition = () => (
...
+export const Basic = Template.bind({})
+Basic.args = {
+  children: 'Hello',
+  content: 'World',
+}
+
+export const Focused = Template.bind({})
+Focused.args = {
+  ...Basic.args,
+}
```
mdodgelooker, line 69: Added these "Focused" stories for each of the various components. When running storybook in the browser they just look like duplicates so I hid them from the sidebar using CSS in `manager-head.html` (see below).

packages/components/src/Accordion/Accordion.story.tsx
```diff
@@ -53,7 +55,26 @@ export const LegacyComposition = () => (
...
+
+Focused.parameters = {
+  beforeScreenshot: async (page: Page) => {
+    const disclosure = await page.$('[role="button"]')
+    await disclosure?.type(' ')
+    await page.waitForTimeout(50)
```
mdodgelooker, line 75: `beforeScreenshot` allows for interaction before the snapshot is taken. I was getting snapshot instability issues, which I think were due to the focus ring taking a tiny amount of time to "re-rasterize" – adding the `waitForTimeout` resolved it.

packages/components/src/Button/ButtonGroup.story.tsx
```diff
@@ -56,6 +57,20 @@ Basic.args = {
...
+Focused.args = {
+  ...Basic.args,
+  margin: 'small',
+}
+
+Focused.parameters = {
+  beforeScreenshot: async (page: Page) => {
+    const button = await page.$('button')
+    await button?.type(' ')
+    await page.waitForTimeout(200)
```
mdodgelooker, line 70: The background color animation takes 150ms, plus a 50ms buffer for the "re-rasterization".

packages/components/src/Button/Button.story.tsx
```diff
@@ -38,18 +39,53 @@ Basic.args = {
   children: 'Button Text',
 }
 
+export const Focused = Template.bind({})
+Focused.args = {
+  ...Basic.args,
+  margin: 'small',
```
mdodgelooker, line 45: Added margin on some of the Focus stories to be able to capture the whole focus ring.

packages/components/src/Accordion/Accordion.story.tsx
dbchristopher: nice!

dbchristopher (approved):
This is a really nice refactor! Looks great 👍

PR #2278, opened 2021-04-30
chore: Add `@looker/eslint-config/oss-compat` rule set for synchorniz…

lukelooker:


www/package.json
```diff
@@ -44,5 +44,8 @@
     "@styled-icons/material": "^10.28.0",
     "@styled-icons/material-outlined": "^10.28.0",
     "@styled-icons/material-rounded": "^10.28.0"
+  },
+  "eslintConfig": {
+    "extends": ["@looker/eslint-config/oss-compat"]
   }
```
lukelooker, line 50: Just an example use case for now (this package doesn't really need it but it's a POC of the rule itself) :)

packages/eslint-config/oss-compat.js
```diff
@@ -0,0 +1,34 @@
...
+ SOFTWARE.
+
+ */
+
+module.exports = {
+  extends: ['@looker/eslint-config', '@looker/eslint-config/license-header'],
+  rules: {
+    camelcase: true,
+    'import/order': 'off',
+    'sort-keys-fix/sort-keys-fix': 'off',
```
lukelooker, line 32: Choices here TBD. This should work as starting place I think?

dbchristopher: since it's `lint-js-fix`able I actually feel like it's fine to keep on. I have the vis packages running with sorted imports now.

packages/eslint-config/oss-compat.js
```diff
@@ -0,0 +1,34 @@
...
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+module.exports = {
+  extends: ['@looker/eslint-config', '@looker/eslint-config/license-header'],
+  rules: {
+    camelcase: true,
```
dbchristopher, line 30: camelcase we would want `false`. but maybe we can just turn it off for the entire monorepo? The strict convention doesn't seem all that particularly helpful to me in general

lukelooker:
@mdodgelooker This should be ready for a "final" review I think

PR #2281, opened 2021-05-01
feat(Accordion): Toggle on spacebar key

mdodgelooker:
_Preliminary work for focus visible refactor._

As I was pulling out the focus visible logic in here to replace with the upcoming `useFocusVisible` hook, I realized `AccordionDisclosure` is actually a great use case for `useClickable`. This hook is aimed at accessibility for non-buttons that are being used in a button-y way. It adds keyup handlers for enter and spacebar keys, and it has focus visible behavior built-in. The enter key was already triggering the accordion toggle, so the only change from `useClickable` is toggling on the spacebar key.

Doing this change in a separate PR helps make the focus visible refactor more sane. 

- [ ] 📚 Documentation updated
- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11

### DIFF
--- a/packages/components/src/Accordion/AccordionDisclosure.tsx
+++ b/packages/components/src/Accordion/AccordionDisclosure.tsx
@@ -24,19 +24,23 @@
 
  */
 
-import React, { FC, KeyboardEvent, Ref, forwardRef, useState } from 'react'
+import React, { FC, Ref, forwardRef } from 'react'
 import styled from 'styled-components'
 import {
   TypographyProps,
   typography,
-  CompatibleHTMLProps,
   padding,
   PaddingProps,
   shouldForwardProp,
   TextColorProps,
   color as colorStyleFn,
 } from '@looker/design-tokens'
-import { useWrapEvent } from '../utils'
+import {
+  FocusVisibleProps,
+  GenericClickProps,
+  useClickable,
+  useWrapEvent,
+} from '../utils'
 import { simpleLayoutCSS, SimpleLayoutProps } from '../Layout/utils/simple'
 import { AccordionDisclosureLayout } from './AccordionDisclosureLayout'
 import { AccordionControlProps, AccordionIndicatorProps } from './types'
@@ -45,12 +49,11 @@ import { accordionDefaults } from './accordionDefaults'
 export interface AccordionDisclosureProps
   extends TypographyProps,
     Omit<AccordionDisclosureStyleProps, 'focusVisible'>,
-    CompatibleHTMLProps<HTMLElement>,
+    GenericClickProps<HTMLElement>,
     SimpleLayoutProps,
     AccordionControlProps,
     AccordionIndicatorProps {
   className?: string
-  focusVisible?: boolean
   ref?: Ref<HTMLDivElement>
   /**
    * ID of the corresponding AccordionContent container
@@ -72,9 +75,9 @@ const AccordionDisclosureInternal: FC<AccordionDisclosureProps> = forwardRef(
       accordionDisclosureId,
       children,
       className,
+      disabled,
       onBlur,
-      onClick,
-      onKeyDown,
+      onClick: propsOnClick,
       onKeyUp,
       defaultOpen,
       isOpen,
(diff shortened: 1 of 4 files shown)


packages/components/src/Accordion/AccordionDisclosure.tsx
```diff
@@ -45,12 +49,11 @@ import { accordionDefaults } from './accordionDefaults'
 export interface AccordionDisclosureProps
   extends TypographyProps,
     Omit<AccordionDisclosureStyleProps, 'focusVisible'>,
-    CompatibleHTMLProps<HTMLElement>,
+    GenericClickProps<HTMLElement>,
     SimpleLayoutProps,
     AccordionControlProps,
     AccordionIndicatorProps {
   className?: string
-  focusVisible?: boolean
```
mdodgelooker, line 53: This prop should be internal only – I moved it into `AccordionDisclosureStyleProps` using `FocusVisibleProps`.

packages/components/src/Tree/Tree.tsx
```diff
@@ -57,7 +57,6 @@ const TreeLayout: FC<TreeProps> = ({
   color: propsColor,
   label: propsLabel,
   labelBackgroundOnly: propsLabelBackgroundOnly,
-  onClick,
```
mdodgelooker, line 60: Was this left unused for a reason? By not destructuring it here, it goes onto `Accordion` along with `...restProps` – would there be any issue with that?

packages/components/src/Tree/types.ts
```diff
@@ -37,7 +37,7 @@ export type TreeProps = Omit<
   | 'indicatorPosition'
   | 'indicatorSize'
 > &
-  ListItemProps & {
+  Omit<ListItemProps, 'onClick'> & {
```
mdodgelooker, line 40: Using `GenericClickProps` in `AccordionProps` caused a TS conflict with the `onClick` from `ListProps` (because of the whole `KeyboardEvent` thing). Now it's going onto `Accordion` (see comment above) so it should get the onClick type from there.

packages/components/src/Tree/Tree.tsx
ElliotTheHuman: No reason that I can think of. I think that was an accident. Thanks for catching that!

In that case, I _believe_ `onKeyUp` should be removed from the destructuring as well since it should be passed down to the `Accordion` like `onClick`.

packages/components/src/Accordion/AccordionDisclosure.tsx
```diff
@@ -89,50 +93,31 @@ const AccordionDisclosureInternal: FC<AccordionDisclosureProps> = forwardRef(
...
-    }, onClick)
-
-    const handleBlur = useWrapEvent(() => setFocusVisible(false), onBlur)
+    const clickableProps = useClickable({
+      disabled,
+      onBlur,
+      onClick,
+      onKeyUp,
+      role,
+    })
```
ElliotTheHuman, line 110: One consequence of using `useClickable` instead of the old `handleKeyUp` is that if you're focused on a `Tree` and press any key, the focus visible border appears (before it was just on tab and any other key press would remove the border).

That being said, if that behavior is standard for other components / we're okay with that, I think using `useClickable` is a huge improvement here!

ElliotTheHuman (changes requested):
`AccordionDisclosure` looks way better with `useClickable`. Thank you for implementing that!

I had one suggestion about `Tree` and one clarifying question about `AccordionDisclosure`'s `onKeyUp` behavior.

packages/components/src/Accordion/AccordionDisclosure.tsx
mdodgelooker: Great question – I think that's how _most_ of the other focus visible components are working now and that's how it'll be after my refactor. I'm going off of the not-yet-universally-supported `:focus-visible` pseudo-selector behavior that does go on for any key: https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible

packages/components/src/Tree/Tree.tsx
mdodgelooker: Ooh thanks for catching that. Removed [here](https://github.com/looker-open-source/components/pull/2281/commits/55b220c35d17b843dee78f3bd01910d0ca3b6aa6)

ElliotTheHuman (approved):
👍

PR #2286, opened 2021-05-03
feat(visualizations): Vis Component prerelease infrastructure

dbchristopher:
This PR adds is a perquisite for the eventual release of vis components into our open source library.

- package directories
- babel config files
- package.json files
- i18nIgnoredProp updates

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "@babel/preset-typescript": "^7.13.0",
     "@babel/runtime": "^7.14.0",
     "@babel/runtime-corejs3": "^7.13.17",
+    "@looker/sdk": "^21.6.0",
     "@testing-library/jest-dom": "^5.12.0",
     "@types/jest-image-snapshot": "^4.3.0",
     "@types/node": "^15.12.1",
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -46,7 +46,7 @@
     "@types/lodash": "^4.14.168",
     "@types/react": "^16.9.56",
     "@types/react-dom": "^16.9.9",
-    "@types/styled-components": "^5.1.9",
+    "@types/styled-components": "5.1.9",
     "@types/styled-system": "^5.1.11",
     "@types/uuid": "^8.3.0",
     "csstype": "^3.0.8",
--- a/packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
+++ b/packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
@@ -239,7 +239,7 @@ const ComboboxListInternal = forwardRef(
       }
 
       const timeoutValue = 50
-      let t: NodeJS.Timeout
+      let t: ReturnType<typeof setTimeout>
       const scrollListener = throttle(() => {
         if (contentContainer) {
           updateScrollState(contentContainer)
--- a/packages/eslint-config/index.js
+++ b/packages/eslint-config/index.js
@@ -70,6 +70,7 @@ module.exports = {
   rules: {
     '@typescript-eslint/explicit-function-return-type': 'off',
     '@typescript-eslint/explicit-module-boundary-types': 'off',
+    '@typescript-eslint/no-redeclare': ['error'],
     '@typescript-eslint/no-use-before-define': 'off',
     'import/default': 'off',
     'import/named': 'off',
@@ -77,6 +78,7 @@ module.exports = {
     'import/order': 'error',
     indentation: 'off',
     'no-console': 'warn',
+    'no-redeclare': 'off', // prefer TS version of `no-redeclare` to support function overload syntax
     'no-restricted-imports': [
       'error',
       {
@@ -110,6 +112,7 @@ module.exports = {
     'sort-keys-fix/sort-keys-fix': 'error',
     // Work to enable these soon-ish
     'testing-library/no-node-access': 'off',
+    'testing-library/prefer-screen-queries': 'off',
   },
 
   settings: {
--- a/packages/visualizations-nivo/package.json
+++ b/packages/visualizations-nivo/package.json
@@ -1,0 +1,53 @@
+{
+ "name": "@looker/visualizations-nivo",
+ "main": "src/index.ts",
+ "version": "0.1.0",
+ "publishConfig": {
+  "access": "restricted"
+ },
+ "license": "MIT",
+ "module": "lib/index.js",
+ "sideEffects": false,
+ "dependencies": {
+  "@looker/visualizations": "*",
+  "@nivo/axes": "0.67.0",
+  "@nivo/bar": "0.67.0",
+  "@nivo/core": "0.67.0",
+  "@nivo/line": "0.67.0",
+  "lodash": "4.17.21"
+ },
+ "devDependencies": {
+  "@looker/components-test-utils": "*",
+  "@looker/eslint-config": "*",
+  "@looker/prettier-config": "*",
+  "@looker/stylelint-config": "*",
+  "@testing-library/react": "*",
+  "@types/react": "*",
+  "@types/react-dom": "*",
+  "@types/styled-components": "*",
+  "@types/styled-system": "*",
+  "eslint-plugin-header": "*",
+  "eslint-plugin-i18next": "*",
+  "eslint-plugin-mdx": "*",
+  "eslint-plugin-node": "*",
+  "eslint-plugin-prettier": "*",
+  "eslint-plugin-react-hooks": "*",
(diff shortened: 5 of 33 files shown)


tsconfig.common.json
```diff
@@ -4,7 +4,7 @@
     "target": "esnext",
 
     "jsx": "react",
-    "lib": ["es6", "es2017.object", "dom"],
+    "lib": ["es6", "es2019.object", "es2019.array", "dom"],
```
dbchristopher, line 7: Resolves object and array method errors:

> packages/visualizations-nivo/src/Bar/barUtils.ts:33:29 - error TS2550: Property 'fromEntries' does not exist on type 'ObjectConstructor'. Do you need to change your target library? Try changing the `lib` compiler option to 'es2019' or later.

> packages/visualizations-nivo/src/Bar/Bar.tsx:65:14 - error TS2550: Property 'flatMap' does not exist on type 'BarDatum[]'. Do you need to change your target library? Try changing the `lib` compiler option to 'es2019' or later.

packages/eslint-config/index.js
```diff
@@ -71,6 +71,7 @@ module.exports = {
     '@typescript-eslint/explicit-function-return-type': 'off',
     '@typescript-eslint/explicit-module-boundary-types': 'off',
     '@typescript-eslint/no-use-before-define': 'off',
+    camelcase: 'off',
```
dbchristopher, line 74: I figured that nobody was that attached to this rule, so it's easier to just assume support of all the sdk response keys

packages/visualizations-nivo/package.json
```diff
@@ -0,0 +1,31 @@
...
+ "license": "UNLICENSED",
+ "dependencies": {
+  "@nivo/axes": "0.67.0",
+  "@nivo/bar": "0.67.0",
+  "@nivo/core": "0.67.0",
+  "@nivo/line": "0.67.0",
+  "lodash": "4.17.21"
+ },
+ "devDependencies": {
+  "@looker/components-test-utils": "1.1.5",
```
mdodgelooker, line 15: I left a comment about this over in the core product release script PR, but it looks like the `@looker/*` packages aren't getting the right version.

mdodgelooker (dismissed):

lukelooker:
@dbchristopher Is this PR still a going concern or are you planning to re-open later on? (just checking as it's on my review list but looks like it's parked) :)

dbchristopher:
@lukelooker I was planning to request review as part of the general release process. Not urgent now, but more updates will be forthcoming!

packages/eslint-config/index.js
dbchristopher: thanks @mdodgelooker! Just getting back to this side of the release process now. I don't think we'll need to disable `order/properties-alphabetical-order` -- that's easily autofixable and should not get out-of-sync anymore with the packages in helltool extending our eslint config.

I also think it's probably worth keeping the `no-explicit-any` turned on! I've resolved some instances of that in vis packages, and manually disabled the rule where `any` really is as specific as I can be for now

packages/components/src/Form/Inputs/Combobox/ComboboxList.tsx
```diff
@@ -240,7 +240,7 @@ const ComboboxListInternal = forwardRef(
       }
 
       const timeoutValue = 50
-      let t: NodeJS.Timeout
+      let t: ReturnType<typeof setTimeout>
```
dbchristopher, line 243: Running `yarn lint` I'd get this error:

```
Type 'number' is not assignable to type 'Timeout'.
```

But then if I set `let t: number`, when running `yarn build` I'd get this error:

```
Type 'Timeout' is not assignable to type 'number'.
```

So going with `ReturnType` was a compromise to keep all build environments happy relative to their own definitions.

dbchristopher:
Not sure why all the build stuff is failing suddenly 🤔  I'll have to take more of a look at this on monday

package.json
```diff
@@ -67,6 +67,12 @@
     "@babel/preset-typescript": "^7.13.0",
     "@babel/runtime": "^7.14.0",
     "@babel/runtime-corejs3": "^7.13.17",
+    "@looker/components": "*",
+    "@looker/design-tokens": "*",
+    "@looker/eslint-config": "*",
+    "@looker/prettier-config": "*",
+    "@looker/sdk": "^21.6.0",
+    "@looker/stylelint-config": "*",
```
lukelooker, line 75: I think this will break Lerna publishing. 

Is there a reason the central package.json should need these (they should get inferred by the `workspaces` above?

NOTE: `@looker/sdk` is okay here - just meant the local references.

packages/visualizations-nivo/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "eslint-plugin-i18next": "*",
+  "eslint-plugin-mdx": "*",
+  "eslint-plugin-node": "*",
+  "eslint-plugin-prettier": "*",
+  "eslint-plugin-react-hooks": "*",
+  "eslint-plugin-sort-keys-fix": "*",
+  "eslint-plugin-testing-library": "*",
+  "react": "^16.14.0",
+  "react-dom": "^16.14.0",
+  "typescript": "4.2.4"
```
lukelooker, line 39: Other packages within the monorepo don't specify `@typescript` so I'd probably forego it here.

packages/visualizations-nivo/package.json
```diff
@@ -0,0 +1,53 @@
...
+ },
+ "peerDependencies": {
+  "@looker/components": "1.x",
```
lukelooker, line 42: This needs to bump to 2.x shortly. Perhaps make it a range of 1.x or 2.x (or just make the bump to 2.x)

packages/visualizations-nivo/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "@looker/visualizations": "*",
+  "@nivo/axes": "0.67.0",
+  "@nivo/bar": "0.67.0",
+  "@nivo/core": "0.67.0",
+  "@nivo/line": "0.67.0",
+  "lodash": "4.17.21"
+ },
+ "devDependencies": {
+  "@looker/components-test-utils": "1.3.1",
+  "@looker/eslint-config": "*",
```
lukelooker, line 21: Since these are both local I'm thinking they should both be `*` or version specific?

Not _exactly_ sure if this is going to cause concerns publishing (some security scanning tools want an explicit range so they can calculate full dep tree). That said we could probably just strip all of the `devDeps` out during publishing step.

packages/visualizations-nivo/tsconfig.build.json
```diff
@@ -0,0 +1,8 @@
+{
```
lukelooker, line 1: NOTE TO SELF: TS4.2 or 4.3 add a feature (name I'm spacing) that _should_ make all of these sub-extends moot – investigate how so we can don't have to keep copying these around. :)

package.json
lukelooker: Ok - just checked this out and I'm pretty sure this leads to `yarn` install failures as well (at least it did on my instance)

lukelooker (changes requested):
Overall this PR looks good to me. Ran into a specific issues:

- The local packages references in the root `package.json` seem to break `yarn install` in some interesting ways. See note on that file for more detail but I think those local references need to be removed.
- Made notes about `visualizations-nivo/package.json` but same questions apply to all of the new package.json added here:
  - Local package refs in `devDeps` are all `*` except `"@looker/components-test-utils"` - seems like these should be consistent. All `*`?
  - Remove `typescript` ref (it's provided by the monorepo and isn't declared per package elsewhere in the monorepo except `eslint-config` (that's a special case)
- Please run `yarn deduplicate` on this PR so can keep the `yarn.lock` as narrow as possible :)

package.json
dbchristopher: [Lerna list](https://www.npmjs.com/package/@lerna/list) is a much better way to track local deps. I can safely remove the starred versions from here!

packages/components/package.json
```diff
@@ -46,7 +46,7 @@
     "@types/lodash": "^4.14.168",
     "@types/react": "^16.9.56",
     "@types/react-dom": "^16.9.9",
-    "@types/styled-components": "^5.1.9",
+    "@types/styled-components": "5.1.9",
```
dbchristopher, line 49: Pinned this to 5.1.9. I changed the `components-providers` to a starred version to ensure there would only ever be one version of `types/styled-components` in the monorepo

lukelooker (approved):
Ship all the things! :)

PR #2288, opened 2021-05-04
feat(Aside): adding prop collapse

laraharrow:
- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/Layout/Semantics/Aside.tsx
```diff
@@ -30,6 +30,11 @@ import { SemanticLayoutBase, semanticLayoutCSS } from './semanticStyledBase'
 import { borderHelper, SemanticBorderProps } from './semanticBorderHelper'
 
 export interface AsideProps extends SemanticLayoutBase, SemanticBorderProps {
+  /**
+   * To be used when desiring the Aside to be collapsible
```
lukelooker, line 34:
```suggestion
   * Prevent `Aside` from being rendered.
```

packages/components/src/Layout/Semantics/Aside.tsx
```diff
@@ -51,11 +56,12 @@ export const Aside = styled.aside
   }))<AsideProps>`
   ${semanticLayoutCSS}
 
+  ${({ collapse }) => collapse && 'display: none;'}
```
lukelooker, line 59: Rather than using CSS I _think_ this should actually prevent the JSX from being rendered at all - that way if there's any cost to the underlying rendering (complex `Tree` for instance) there will be no performance impact of rendering this hidden element.

I suspect to do this `Aside` will need to have an underling component like:

```

const AsideLayout: FC<AsideProps> = ({ collapse, ...props }) => 
  collapse ? null : <aside {...props} />

export const Aside = styled(AsideLayout)  ...

```

packages/components/src/Layout/Semantics/Semantics.test.tsx
```diff
@@ -152,4 +152,8 @@ describe('Semantics', () => {
       height: 'fit-content',
     })
   })
+  test('Aside collapse will display none.', () => {
+    renderWithTheme(<Aside collapse>Aside content</Aside>)
+    expect(screen.getByText('Aside content')).toHaveStyleRule('display: none;')
```
lukelooker, line 157: This will need to get refactored since `Aside content` will no longer be on the page.

packages/components/src/Layout/Semantics/Story/ExamplePage.story.tsx
```diff
@@ -345,3 +345,50 @@ export const ScrollAllAreas = () => (
...
+        <AsideAlt scrollWithin p="xxlarge" width="10rem">
+          I'm baby man braid cold-pressed seitan sartorial, tumblr ennui selfies
+          chia twee subway tile af 90's celiac. Gochujang distillery tumeric
+          flannel lumbersexual gastropub fashion axe viral neutra. Pickled
+          narwhal everyday carry activated charcoal succulents occupy YOLO
+          yuccie forage seitan kitsch. Celiac bespoke cloud bread asymmetrical
+          bicycle rights XOXO cold-pressed hashtag listicle hell of migas.
+          Chillwave brooklyn fam occupy microdosing leggings. Poke af cornhole
+          hot chicken. Portland tattooed +1 chia ennui.
+        </AsideAlt>
```
lukelooker, line 390: Can we use a simpler example here so it's easier to "see" what's being tested here:

```suggestion
        <Aside collapse>Aside</Aside>
        <Section as="main">Section</Section>
```

lukelooker (changes requested):

packages/components/src/Layout/Semantics/Aside.tsx
```diff
@@ -44,8 +50,13 @@ export interface AsideProps extends SemanticLayoutBase, SemanticBorderProps {
...
 }
 
-export const Aside = styled.aside
-  .withConfig({ shouldForwardProp })
+const AsideLayout: FC<AsideProps> = ({ collapse, ...props }) =>
+  collapse ? null : <aside {...props} />
+
+export const Aside = styled(AsideLayout)
+  .withConfig({
+    shouldForwardProp: (prop) => prop === 'collapse' || shouldForwardProp(prop),
```
lukelooker, line 58: @laraharrow - the problem was the `shouldForwardProp` was filtering out `collapse` and therefore not forwarding it on the the underlying `AsideLayout` component.

This change makes it work as it should

packages/components/src/Layout/Semantics/Story/ExamplePage.story.tsx
```diff
@@ -345,3 +347,24 @@ export const ScrollAllAreas = () => (
...
+  return (
+    <Layout hasAside>
+      <Aside p="small" collapse={!value}>
+        Aside
+      </Aside>
+      <Section p="small">
+        <FieldToggleSwitch label="Show Aside" onChange={toggle} on={value} />
+      </Section>
+    </Layout>
+  )
```
lukelooker, line 363: Made this interactive so it's easier to understand what it does.

packages/components/src/Layout/Semantics/Story/ExamplePage.story.tsx
```diff
@@ -345,3 +347,24 @@ export const ScrollAllAreas = () => (
...
+}
+
+AsideCollapse.parameters = {
+  storyshots: {
+    disable: true,
```
lukelooker, line 368: Screenshot is of pretty limited use since a simple Jest test of whether or not the `Aside` content is visible will give the same quality of coverage.

lukelooker (changes requested):
I went ahead and fixed this up (took me a bit of poking to understand what had gone awry). 

This needs test coverage and then it should be good-to-go.

packages/components/src/Layout/Semantics/Semantics.test.tsx
```diff
@@ -152,4 +152,8 @@ describe('Semantics', () => {
       height: 'fit-content',
     })
   })
+  test('Aside collapse will display none.', () => {
+    renderWithTheme(<Aside collapse>Aside content</Aside>)
+    expect(screen.queryByText('Aside content')).not.toBeInTheDocument()
+  })
```
lukelooker, line 158: This test file is getting pretty huge - can we go ahead and move this (as well as any other `Aside`-specific tests) into `Aside.test.tsx` to follow along with our other component file-layouts?

lukelooker (changes requested):
Let's move `Aside` tests into their own file so that it easier for reviewers to find related tests :)

packages/components/src/Layout/Semantics/Semantics.test.tsx
laraharrow: Should I do this here or wait until we have all the Aside/Layout work done and I can organize the test separated then?
I'm aware that other PR's that I'm working now also have test that could be subdivided

lukelooker: I suppose it's fine if it gets done in a follow-up PR. 👍

lukelooker (approved):

PR #2295, opened 2021-05-05
feat(Tree): Keyboard Navigation

ElliotTheHuman:
**TODO:** Add the following keyboard navigation tests:
1. Move right in treeitem, move down to next treeitem
2. Home button
3. End button
4. Preserves last focused element

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/Tree/types.ts
```diff
@@ -37,7 +37,17 @@ export type TreeProps = Omit<
...
+    ListItemProps,
+    | 'color'
+    | 'density'
+    | 'detail'
+    | 'disabled'
+    | 'icon'
+    | 'itemRole'
+    | 'keyColor'
+    | 'truncate'
+  > & {
```
ElliotTheHuman, line 50: Unrelated change: `Tree` accepted `ListItem` props, but funneled anything that wasn't one of these props to the nested `Accordion` (which is a no-go).

Getting `Tree` to support all `ListItem` props is possible but I'd like to keep those changes separate from all the keyboard nav business. For now, restricting props down to only the ones that are needed `Tree`'s nested `TreeItem` / `ListItem` seemed like a decent temp solution.

packages/components/src/List/ListItem.tsx
```diff
@@ -236,6 +232,20 @@ const ListItemInternal = forwardRef(
...
+      const focusableElements = wrapperRef?.current?.querySelectorAll(
+        'a, button, input'
+      )
+
+      if (focusableElements) {
+        focusableElements.forEach((activeElement) => {
+          activeElement.setAttribute('tabIndex', '-1')
+        })
+      }
+    })
```
ElliotTheHuman, line 247: In an ideal world, I'd put this `useEffect` call in `TreeItem`. But two points:
1. Putting this in `TreeItem` leads to the `querySelectorAll` call failing to find child focusable elements. I _think_ this has to do with the fact that `ListItem` (which is nested within `TreeItem`) is the one being rerendered, and so hovering / keyboard navigating rerenders the `ListItem` and calls the `useEffect` again with hover disclosure icons present. Placing in `TreeItem` means this `useEffect` call doesn't get re-called when hovering / keyboard navigating. Not sure on this 100%, but TLDR the child elements don't get their auto `tabIndex=-1` when this code lives in `TreeItem`.
2. It's might be nice for actual `List`s elements to include focusable child elements in their keyboard nav flow.

packages/components/src/List/ListItem.tsx
```diff
@@ -162,7 +165,17 @@ const ListItemInternal = forwardRef(
...
+    let hovered = internalHoveredState[0]
+    let setHovered = internalHoveredState[1]
+
+    if (propsSetHovered && propsHovered !== undefined) {
+      hovered = propsHovered
+      setHovered = propsSetHovered
+    } else {
+      hovered = internalHoveredState[0]
+      setHovered = internalHoveredState[1]
+    }
```
ElliotTheHuman, line 178: For the record, I don't love this. Buuuuut I needed a way for state to exist up in the `TreeItem` so that `TreeItem` could know when it's being hovered or not. It needs to know hovered state because it needs to know when to call its `useEffect` which adds `tabIndex=-1` to child button / a / input elements.

Alternatives to this:
- Move the `useEffect` into `ListItem.tsx` (but this couples `ListItem` and `TreeItem` because the `querySelectorAll` call references Tree-related classes)
- Move the `useEffect` call into `ListItemLabel` and `ListItemDetail`
  - I actually think this might be the call...

ElliotTheHuman: Update: Now that we've forced `TreeItemInner` (i.e. `Tree`'s internal item used to keep code DRY), we _should_ be free to apply `tabIndex=-1` to all links / buttons / inputs present within a `TreeItem` / `ListItem`. In which case, keeping the `useEffect` in `ListItem.tsx` creates the least amount of tech debt afaik.

Follow-up work: Update `List` keyboard navigation to support similar keyboard nav behavior as `Tree` (i.e. up / down arrow traverses items, left / right traverses buttons within the item / item detail container).

packages/components/src/Tree/stories/FieldPicker.story.tsx
```diff
@@ -179,7 +184,7 @@ const fields = (
         fontWeight="semiBold"
         pt="xsmall"
         pb="xxsmall"
-        pr="xxsmall"
+        px="xxsmall"
```
ElliotTheHuman, line 187: Minor visual adjustment to get field type headings to line up with field labels

packages/components/src/utils/useArrowKeyNav.ts
```diff
@@ -76,6 +82,7 @@ export interface UseArrowKeyNavProps<E extends HTMLElement> {
  */
```
mdodgelooker, line 82: Now that this hook has a different result if disabled, can you define the result type here, and add it as the return type on the function definition below?
```
export type UseArrowKeyNavResult<E extends HTMLElement> = Pick<
  UseArrowKeyNavProps<E>,
  'ref' | 'onBlur' | 'onFocus' | 'onKeyDown'
> & {
  tabIndex?: number
}
```

packages/components/src/utils/useArrowKeyNav.ts
```diff
@@ -183,4 +190,6 @@ export const useArrowKeyNav = <E extends HTMLElement = HTMLElement>({
     // landing back on the container when shift-tabbing from the first item
     tabIndex: focusInside ? undefined : 0,
   }
+
+  return disabled ? null : navProps
```
mdodgelooker, line 194:
```suggestion
  return disabled ? {} : navProps
```

packages/components/src/List/ListItem.tsx
```diff
@@ -284,6 +304,32 @@ const ListItemInternal = forwardRef(
...
+    const handleWrapperBlur = (event: React.FocusEvent<HTMLElement>) => {
+      const nextFocusTarget = getNextFocusTarget(event)
+
+      if (
+        nextFocusTarget &&
+        !event.currentTarget.contains(nextFocusTarget as Node)
+      ) {
+        setHovered(false)
+      }
+    }
```
mdodgelooker, line 321: This change is good but I think it's causing `AccordionDisclosure` on the field picker to look funky when focused:
image

packages/components/src/List/ListItemLabel.tsx
```diff
@@ -85,6 +85,10 @@ export const ListItemLabel = styled(ListItemLabelLayout).withConfig({
   ${({ height, itemRole }) => itemRole === 'none' && `height: ${height}px;`}
   ${listItemBackgroundColor}
 
+  &:focus {
+    box-shadow: inset 0 0 0 2px ${({ theme }) => theme.colors.keyFocus};
+  }
+
```
mdodgelooker, line 91: Is the focus ring supposed to always show on `ListItem` now, not just for focus via keyboard? If so, we should remove `useFocusVisible`.

packages/components/src/Tree/Tree.tsx
```diff
@@ -123,14 +127,19 @@ const TreeLayout: FC<TreeProps> = ({
...
+  const treeItemInnerRole = 'none'
   const label = (
     <TreeItemInner
+      renderAsDiv
       color={color}
       density={density}
       detail={detail}
       disabled={disabled}
       icon={icon}
+      itemRole={treeItemInnerRole}
```
mdodgelooker, line 139: I guess this is causing the snapshot diffs? Not sure if that's a real issue or not...

mdodgelooker (changes requested):
Looking pretty good! I have a few comments & questions here & there.

packages/components/src/utils/useArrowKeyNav.ts
mdodgelooker: Whoops, it looks like my requested change caused some TS errors. You can fix those by passing the element type like:
```
const navProps = useArrowKeyNav<HTMLTableElement>({
```

packages/components/src/Tree/TreeStyle.tsx
```diff
@@ -192,6 +192,27 @@ export const TreeStyle = styled(TreeStyleLayout)`
           indicatorSize: indicatorContainerSize,
           theme,
         })}
+
+      /**
+        Gets the box-shadow to sit above the ListItem background
+       */
+      &:focus::after {
```
mdodgelooker, line 199: Shouldn't the focus ring only be for keyboard nav? If so, it might work to update `Accordion` with something like
```
className={ `${clickableProps.focusVisible ? 'focusVisible ' : ''}${className}`}
```
and then here you could do `.focusVisible:after`.

packages/components/src/Tree/TreeStyle.tsx
```diff
@@ -192,6 +192,27 @@ export const TreeStyle = styled(TreeStyleLayout)`
...
+        top: 0;
+        z-index: 1;
+      }
+
+      /**
+        Disables AccordionDisclosure's inherent box-shadow
+       */
+      &:focus-visible {
+        box-shadow: none;
+      }
```
mdodgelooker, line 215: This won't work for Safari or IE11. But if you end up using my suggestion above, you could include this with that.

packages/components/src/Tree/utils/getNextTreeFocus.ts
```diff
@@ -0,0 +1,78 @@
...
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+export const getTabStops = (ref: HTMLElement): HTMLElement[] =>
+  Array.from(ref.querySelectorAll('[tabindex="-1"]:not(:disabled)'))
+
+// Returns a fallback element (called when the element with focus has been removed from the DOM)
+const getFallbackElement = (
```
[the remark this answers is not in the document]
mdodgelooker: Oh good point codeclimate. You could export the one in `utils/getNextFocus.ts` and then just use that here.

mdodgelooker (changes requested):
This is looking really slick! A few more suggestions.

packages/components/src/List/ListItemLabel.tsx
```diff
@@ -75,6 +75,7 @@ type ListItemLabelProps = CompatibleHTMLProps<HTMLElement> &
   ListItemStatefulProps &
   FlexibleColor & {
     disabled?: boolean
+    focusVisible: boolean
```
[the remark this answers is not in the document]
ElliotTheHuman: 👍  Good call.

And (iirc) I moved `focusVisible` from the wrapper to the label because the label container is the one that actually takes focus (i.e. has `tabIndex=-1`). The wrapper is just a layout container / li stand-in while the label is the button / a usually (in Explore's instance it's a div with an onClick / onKeyDown because buttons inside of buttons).

packages/components/src/Tree/TreeStyle.tsx
ElliotTheHuman: Good catch. With the previous styling you could click on the indicator and that would trigger the focus visible ring 🤦 .

Went ahead and used the `useFocusVisible` hook, using the returned `focusVisible` prop to conditionally pass the `focusVisible` class name to `TreeStyle`.

ElliotTheHuman:
GH Actions is stalling, but the code should be ready for another look!

packages/components/src/Tree/TreeStyle.tsx
mdodgelooker: I think you should be able to just add the `className` in `AccordionDisclosure` where `useFocusVisible` is already in place. Would this work? https://github.com/looker-open-source/components/pull/2372

mdodgelooker (approved):
Gorgeous. Ship it! 🚢

PR #2300, opened 2021-05-05
feat(Layout): Supports displaying shadow when content overflows container height

laraharrow:
- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


laraharrow:
@lukelooker any suggestions here?

laraharrow:
hey @lukelooker, as soon as I make `Layout` into a FC it breaks the display. When you have a sec would you look into this?

packages/components/src/Layout/Semantics/Layout.tsx
```diff
@@ -46,16 +47,102 @@ export interface LayoutProps
...
 const hasAsideCSS = css`
   flex-direction: row;
-
   & > ${Section} {
     width: 0;
   }
 `
+export const Layout: FC<LayoutProps> = ({ children }) => (
+  <InnerLayout>{children}</InnerLayout>
+)
```
lukelooker, line 56: The `InnerLayout` needs to "know" about `hasAside` & `fixed` to properly specify the CSS:

```suggestion
export const Layout: FC<LayoutProps> = ({ children, fixed, hasAside }) => (
  <InnerLayout hasAside={hasAside} fixed={fixed}>{children}</InnerLayout>
)
```

packages/components/src/Layout/Semantics/Layout.tsx
```diff
@@ -46,16 +48,143 @@ export interface LayoutProps
 
 const hasAsideCSS = css`
   flex-direction: row;
-
   & > ${Section} {
     width: 0;
   }
 `
+export const Layout: FC<LayoutProps> = ({ children, fixed, hasAside }) => {
```
mdodgelooker, line 55:
```suggestion
export const Layout: FC<LayoutProps> = ({ children, fixed, hasAside, ...props }) => {
```

packages/components/src/Layout/Semantics/Layout.tsx
```diff
@@ -46,16 +48,143 @@ export interface LayoutProps
...
+    if (container) {
+      setHasOverflow(container.offsetHeight < container.scrollHeight)
+    }
+  }, [height])
+  return (
+    <InnerLayout
+      fixed={fixed}
+      hasAside={hasAside}
+      hasOverflow={hasOverflow}
+      ref={internalRef}
```
mdodgelooker, line 79:
```suggestion
      ref={internalRef}
      {...props}
```

mdodgelooker (commented):
`Layout` needs to pass through the rest of `props`.

laraharrow:
@lukelooker  Here is my first attempt on making the hooks for overflow into `useOverflow` - its not quite working 🤦🏽‍♀️
I saw it working when I had the logic inside each component ( `Layout` `Aside` `Section` ) once I created `useOverflow` and tried adding it to the 3 components it's not working anymore. Also`shouldForwardProp` is causing a bug on `collapse`.

packages/components/src/Layout/Semantics/Aside.tsx
```diff
@@ -52,16 +53,28 @@ export interface AsideProps extends SemanticLayoutBase, SemanticBorderProps {
...
-const AsideLayout: FC<AsideProps> = ({ collapse, ...props }) =>
-  collapse ? null : <aside {...props} />
+export const Aside: FC<AsideProps> = ({ collapse, children, ...props }) => {
+  const internalRef = useRef<HTMLDivElement>(null)
+  const hasOverflow = useOverflow(internalRef)
 
-export const Aside = styled(AsideLayout)
-  .withConfig({
+  return collapse ? null : (
+    <InnerAside hasOverflow={hasOverflow} ref={internalRef} {...props}>
```
lukelooker, line 61: Create a single component that be used in all the components (see mention of `OverflowShadow` below:

```suggestion
    <OverflowShadow as="aside" hasOverflow={hasOverflow} ref={internalRef} {...props}>
```

packages/components/src/Layout/Semantics/Semantics.test.tsx
```diff
@@ -29,7 +29,31 @@ import React from 'react'
...
+      Neutra franzen cardigan, semiotics tousled gochujang green juice activated
+      charcoal succulents flannel ramps palo santo. Kale chips williamsburg
+      hexagon, etsy cray 8-bit cornhole tilde neutra DIY snackwave whatever food
+      truck marfa fashion axe. Polaroid master cleanse twee DIY tbh pop-up
+      biodiesel austin hell of scenester woke man bun 3 wolf moon. Venmo
+      coloring book beard adaptogen four loko locavore wolf street art +1
+      kombucha marfa 90's unicorn everyday carry.
+    </p>
+  </>
+)
```
lukelooker, line 56: Can you pull the text block here from a variable in `__mock__` file so the test itself is easy to read? (I believe the `Constitution.tsx` file has some big chunks of text)

```suggestion
```

packages/components/src/Layout/Semantics/Story/ExamplePage.story.tsx
```diff
@@ -188,15 +188,15 @@ const Highlighter = styled.div`
     background-color: lightcoral;
   }
 
-  ${Aside} {
+  aside {
```
lukelooker, line 191: Ideally we should never target HTML selectors and instead use Styled Component references to ensure we're targeting just our own components.

packages/components/src/Layout/Semantics/Story/ExamplePage.story.tsx
```diff
@@ -153,7 +153,7 @@ const Template2: Story<LayoutProps> = (args) => (
           {menuItems}
         </Aside>
         <Layout>
-          <Section p="xxlarge" as="main">
```
lukelooker, line 156: Why did you remove the `as="main"` throughout this file? (it does add semantic meaning and helps screen readers so there will probably be real-world use cases for it)

packages/components/src/Layout/Semantics/Story/ExamplePage.story.tsx
```diff
@@ -380,23 +389,78 @@ AsideWidthSizeRail.args = {
...
+        {longContent}
+        {longContent}
+        {longContent}
+      </Section>
+    </Layout>
+    <Footer height="3rem" px="large">
+      Page Footer
+    </Footer>
+  </Page>
+)
```
lukelooker, line 466: This file is getting too large to read. Can we look at breaking it up in this PR or doing a follow-up to do so?

packages/components/src/Layout/Semantics/Story/Semantics.story.tsx
```diff
@@ -98,7 +98,7 @@ export const CommonLayouts = () => (
 
 const HeaderLayoutAsideMain = () => (
   <Highlighter>
-    <Layout>
+    <Layout height="100%">
```
lukelooker, line 101: We should not need to explicitly style `Layout` like this (if it's need we probably have a bug that needs to be addressed)

lukelooker (changes requested):
Most of my comments are focused around consolidating some duplicate code across Aside, Layout & Section (I made comments on `Aside` but refactors applied there should get applied to the other `useOverflow` components as well)

packages/components/src/Layout/Semantics/Aside.tsx
```diff
@@ -52,16 +53,28 @@ export interface AsideProps extends SemanticLayoutBase, SemanticBorderProps {
   width?: ResponsiveValue<AsideSizeRamp | string>
 }
 
-const AsideLayout: FC<AsideProps> = ({ collapse, ...props }) =>
-  collapse ? null : <aside {...props} />
+export const Aside: FC<AsideProps> = ({ collapse, children, ...props }) => {
```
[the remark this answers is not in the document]
mdodgelooker: @lukelooker yeah, we probably should use `forwardRef` – I could see people wanting to measure these elements, get their scroll position, stuff like that.
@laraharrow let's chat tomorrow about how to modify `useOverflow` to support that change.

packages/components/src/Layout/Semantics/Story/ExamplePage.story.tsx
laraharrow: yes I'll add that to the PR that I'll separate the tests.

packages/components/src/Layout/Semantics/Section.tsx
```diff
@@ -39,15 +45,36 @@ export interface SectionProps extends SemanticLayoutBase {
...
 }
 
-export const sectionCSS = css`
-  ${semanticLayoutCSS}
-  flex: 1 0 auto;
+interface OverflowProps extends SectionProps, UseOverflowProps {}
+
+export const InnerSection = styled.div<OverflowProps>`
+  ${({ hasOverflow }) => hasOverflow && OverflowShadow}
 `
```
lukelooker, line 52: This repeats in all `Aside`, `Layout` & here (`Section`).

How about renaming it as `OverflowShadow` and placing it centrally (perhaps with the `useOverflow` hook) and then using it for all of the usages?

packages/components/src/Layout/Semantics/Story/Semantics.story.tsx
lukelooker: Still worried about this change. Why was it needed?

lukelooker (changes requested):

packages/components/src/Layout/Semantics/Story/Semantics.story.tsx
lukelooker, line 101: Still worried about this change. Why was it needed?

packages/components/src/Layout/Semantics/Section.tsx
laraharrow: The only things about that is that we use `aside` and `div` as the styled element. Also the specific component prop is part of the interface. I could just add the 3 props: `SectionProps` `AsideProps` `LayoutProps` what to do about the fact that `Aside` uses `aside` as its element on `styled` and the other two use `div`.

packages/components/src/Layout/Semantics/Story/Semantics.story.tsx
laraharrow: oops forgot to remove it, I fixed the issue before just forgot to remove it on story. fixed now

lukelooker (approved):
Awesome! `useOverflow` & `OverflowShadow` look great.

Ignore test / storybook structure since they're a bit of a mess but that's going to be addressed shortly.

PR #2304, opened 2021-05-06
chore: add colorful looker logo icon

hkbarton:
Add colorful Looker logo icon.

image


jhardy:
@hkbarton Thanks for the PR. I am going to close this for now as I don't think we should be including the branded Logo in our open source package. If you have additional context or questions, feel free to open a buganizer ticket against the components team and we can chat further there.

PR #2322, opened 2021-05-10
chore: Update dependencies and use resolutions to address transient dependencies with vulnerabilities

lukelooker:


packages/components/package.json
```diff
@@ -31,7 +31,7 @@
     "i18next": "19.9.1",
     "lodash": "^4.17.20",
     "react-hotkeys-hook": "2.3.1",
-    "react-i18next": "11.8.8",
+    "react-i18next": "11.8.15",
```
lukelooker, line 34: Newer versions appear to have removed usage of `html-parse-stringify2` and instead use a newer version of `html-parse-stringify`

mdodgelooker (approved):

mdodgelooker (changes requested):
Oh whoops, i18n isn't working in one of the snapshots. Looking into it...

mdodgelooker (approved):

PR #2337, opened 2021-05-12
fix: Correct List overflow & Tree color typing

lukelooker:


ElliotTheHuman (approved):

ElliotTheHuman (changes requested):
Seeing issues with windowing with the `overflow: auto` removed from `List`.

PR #2338, opened 2021-05-12
fix(NavList): Properly style Trees within NavList

lukelooker:
Also addresses issue where non-StyledIcon wouldn't receive proper gap within ListItem


packages/components/src/NavTree/NavTree.story.tsx
```diff
@@ -0,0 +1,115 @@
...
+  </NavList>
+)
+
+const ListItemNestedInTree = styled(ListItem)`
+  ${ListItemLabel} {
+    padding-left: calc(
+      ${({ theme }) => `${theme.space.medium} + ${theme.space.xsmall}`}
+    );
+  }
+`
```
lukelooker, line 111: This is the slightly unfortunate workaround to get `ListItem` to have correct indent within a `Tree`. 

Thoughts on how we can eventually incorporate this? (Perhaps `TreeItem` in a `TreeCollection`?)

packages/components/src/NavTree/NavTree.tsx
```diff
@@ -0,0 +1,54 @@
...
+ * This component is in active development and may see significant change in
+ * it's behavior, interface & presentation. It may also be deprecated without
+ * SemVer major version change. _It is not recommended_ to use this component
+ * at this time.
+ *
+ * KNOWN ISSUES
+ *
+ * `Tree` w/ selected does not get text-color properly specified because
+ * `[aria-selected='true'] is not specified properly
+ *
```
lukelooker, line 49: Untrue, fixing in follow-up commit

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
```diff
@@ -109,7 +109,11 @@ const OptionLayout = ({ option, ...rest }: OptionLayoutProps) => {
   const { hasIcons } = useContext(SelectOptionsContext)
   const { indicatorPropRef } = useContext(ComboboxContext)
   const iconPlaceholder = hasIcons ? (
-    <IconPlaceholder size="small" data-testid="option-icon-placeholder" />
+    <IconPlaceholder
+      mr="small"
```
lukelooker, line 113: Had to add an explicit specification of `mr` here as it was removed from `IconPlaceholder`

packages/components/src/Icon/IconPlaceholder.tsx
```diff
@@ -38,12 +38,9 @@ export interface IconPlaceholderProps
...
 
-export const IconPlaceholder = styled.div.attrs<IconPlaceholderProps>(
-  ({ mr = 'xsmall' }) => ({
-    'aria-hidden': true,
-    mr,
-  })
-)<IconPlaceholderProps>`
+export const IconPlaceholder = styled.div.attrs<IconPlaceholderProps>(() => ({
+  'aria-hidden': true,
+}))<IconPlaceholderProps>`
```
lukelooker, line 43: Remove built-in mr since it's not really evident that it'll be there from just using the component and in other cases (read: List) it's undesirable.

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
ElliotTheHuman: Should this be "xsmall"? `IconPlaceholder` previously had a default "xsmall" margin-right.

ElliotTheHuman: Screen Shot 2021-05-12 at 2 28 52 PM

packages/components/src/NavTree/NavTree.story.tsx
ElliotTheHuman: I can't think of anything radically better than this for now. Because `TreeStyle` (which is nested within `Tree`) handles all the indentation logic, we would need to do something like `TreeArtificial` if we wanted to indent them without a styled override.

ElliotTheHuman (changes requested):
Looks good overall. Had some minor style questions but nothing major.

packages/components/src/Form/Inputs/Select/SelectOptions.tsx
lukelooker: Yes, will correct! :)

packages/components/src/NavList/NavList.tsx
```diff
@@ -24,12 +24,29 @@
...
 import { ListItemDetail } from '../List/ListItemDetail'
 import { List, ListItemLabel } from '../List'
+import { ListItemIconPlacement } from '../List/ListItemLayout'
+
+export const navStyles = css`
+  ${AccordionDisclosureStyle}, ${ListItemLabel} {
+    border-bottom-right-radius: 5rem;
+    border-top-right-radius: 5rem;
+
+    &[aria-selected='true'] {
```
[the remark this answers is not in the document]
lukelooker: For now at least i'm going to say no - the developer can set both and I can imagine a case where they don't want `current` to be highlighted even tho it's the "current" page.

lukelooker (commented):
@ElliotTheHuman Ready for re-review. I did include two snapshot updates that were unrelated just-because (they new versions are more correct, the difference is below the check-threshold)

ElliotTheHuman (approved):
LGTM

PR #2341, opened 2021-05-12
fix(reset): Refine reset for Legend and provide css classNames for downstream extension as-needed

lukelooker:


mdodgelooker (approved):
I like it!

PR #2348, opened 2021-05-12
feat(Aside): standardized width size

laraharrow:
- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


lukelooker (changes requested):
Need to add `AsideSizes` to the sizes supported by `Drawer` so that folks can move between the two seamlessly then this is totally good to go!

packages/components/src/Drawer/DrawerSurface.tsx
```diff
@@ -75,6 +76,7 @@ export const DrawerSurface = styled(SurfaceBase).attrs<DrawerSurfaceProps>(
 
   ${placement}
   ${dialogWidth}
+  ${asideWidth}
```
lukelooker, line 79: This will output `width` twice in the CSS.

There should be a new `drawerWidth` function that combines the two Ramps and then allows a single use here.

lukelooker (changes requested):
Seeing some test failures and the `DrawerWidth` is going to need to be an extension of `DialogWidth` so there's only a single helper in the CSS block.

packages/components/src/Drawer/DrawerSurface.tsx
```diff
@@ -43,7 +44,7 @@ export interface DrawerSurfaceProps {
    * the specified width or the viewport width.
    * @default 'medium'
    */
-  width?: ResponsiveValue<DialogSizeRamp | string>
+  width?: ResponsiveValue<DialogSizeRamp | AsideSizeRamp | string>
```
mdodgelooker, line 47: If you make this a new type...
```
export type DrawerWidth = ResponsiveValue<DialogSizeRamp | AsideSizeRamp | string>
...
width?: DrawerWidth
```
You can then add it to `UseDialogPropsInternal` in `useDialog.tsx`:
```
export interface UseDialogPropsInternal
  extends Omit<UseDialogProps, 'placement' | 'width'> {
  placement?: DialogPlacements | DrawerPlacements
  width?: DialogWidth | DrawerWidth
}
```

packages/components/src/Dialog/DialogSurface.tsx
```diff
@@ -63,7 +64,7 @@ export interface DialogSurfaceProps {
    * Explicitly specifying a width will set the Surface to be the lesser of
    * the specified width or the viewport width.
    */
-  width?: DialogWidth
+  width?: DrawerWidth
```
lukelooker, line 67: ?

```suggestion
  width?: DialogWidth
```

packages/components/src/Drawer/DrawerSurface.tsx
```diff
@@ -43,7 +47,7 @@ export interface DrawerSurfaceProps {
    * the specified width or the viewport width.
    * @default 'medium'
    */
-  width?: ResponsiveValue<DialogSizeRamp | string>
+  width?: DrawerWidth
```
lukelooker, line 50: This needs to support `string` still as folks may want to use "off standard" sizes:

```suggestion
  width?: DrawerWidth | string
```

Also removing `ResponsiveValue` will act as a breaking change for anybody using that so we need to put it back so as not to break folks who might be using responsive drawer sizes (not sure if anyone is but it's hard to know for sure since extension developers use stuff in new ways).

packages/components/src/Dialog/DialogSurface.tsx
```diff
@@ -28,7 +28,8 @@ import styled, { css } from 'styled-components'
 import { height, ResponsiveValue } from 'styled-system'
 import { theme } from '@looker/design-tokens'
 import { SurfaceBase, surfaceTransition } from '../Dialog/SurfaceBase'
-import { dialogWidth, DialogWidth } from './dialogWidth'
+import { DrawerWidth } from '../Drawer/DrawerSurface'
```
lukelooker, line 31: I'm thinking the `DrawerWidth` shouldn't be pulled into `DialogSurface`?

lukelooker (changes requested):
Appears that sine `Drawer` stuff snuck in `Dialog` and I see one spot where a change to `width` on `Drawer` could/would be a breaking change for folks.

laraharrow:
Hey Luke, I'm not sure I follow your suggestions here. I was getting errors by using `DialogWidth` within `Drawer` and `Dialog ` so to fix it I updated that value to include the new `Aside` values. Nothings should have changed to `Dialog` as we still support all the value expected by it. I just added the values that `Aside` is adding. I was a bit skeptical of adding `DrawerWidth` inside `DialogSurface` but it seemed to be needed to avoid a bug since `Drawer` is actually using `Dialog` underneath.

packages/components/src/Dialog/DialogSurface.tsx
lukelooker: `DialogDrawerWidth`

lukelooker (approved):
Ship it! :)

PR #2354, opened 2021-05-13
fix(Tree): Added explicit role="none" to nested TreeItem in Tree

ElliotTheHuman:

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/Tree/Tree.tsx
```diff
@@ -136,6 +137,8 @@ const TreeLayout: FC<TreeProps> = ({
       disabled={disabled}
       icon={icon}
       truncate={truncate}
+      itemRole={treeItemInnerRole}
```
ElliotTheHuman, line 140: There really isn't a reason this inner `TreeItem` should ever be a `button`, so setting `itemRole="none"` made sense to give people room to put focusable elements like buttons / links inside the disclosure.

packages/components/src/Tree/TreeItem.test.tsx
```diff
@@ -82,6 +82,9 @@ describe('TreeItem', () => {
   })
 
   test('Triggers onClickWhitespace when indent padding is clicked, itemRole = "none", and labelBackgroundOnly = true', () => {
+    // eslint-disable-next-line no-console
+    console.warn = jest.fn()
```
ElliotTheHuman, line 86: Normally we shouldn't be using `labelBackgroundOnly` and an `onClick`, but I have both on this test to track whether whitespace clicks trigger the `onClick`.

lukelooker (approved):
LGTM

PR #2360, opened 2021-05-14
fix(Panel): PanelHeader icon should always be `ArrowBack`

lukelooker:
## Ticket
https://buganizer.corp.google.com/issues/188071366

## Screenshot
image

- [x] 📚 Documentation updated


www/src/documentation/components/layout/panel.mdx
```diff
@@ -4,7 +4,7 @@ storybook: true
 ---
 
 `<Panel />` component displays content with directional animation by using prop `direction` - `left`, `right` currently. The elements to be displayed as the compoennt must be
-passed as the prop `content`. `<Panels />` should be the parent component grouping all the elements involved on the `Panel. The `Panel Heading`can be defined as a string or a ReactNode and will be displayed above the main`content`.
+passed as the prop `content`. `<Panels />` should be the parent component grouping all the elements involved on the `Panel`. The `PanelHeader`can be defined as a string or a ReactNode and will be displayed above the main`content`.
```
jhardy, line 7: Minor nit, this description is worded a bit strange, plus a small typo.  Part of me wonders if the example is enough and we don't need to describe all the props. 
A rough pass at a few ideas

## More

Use the `<Panel />` component to display content with a directional animation by using  the `direction` prop.  The content to be displayed must be passed to the prop `content` prop. 

To ensure proper display of the the content, all `Panels` should be wrapped in the `<Panels />` component. The `PanelHeader` can be defined as a string or a ReactNode and will be displayed above the main content.

---

## Alternate

Use the `<Panel />` component to display content with a directional animation. All `<Panels />` need to be wrapped in a `<Panels />` component to ensure proper animation.

jhardy (approved):
Only one small nit on the documentation.

www/src/documentation/components/layout/panel.mdx
lukelooker: Agreed, shorted seems better here - the Storybook doc-props fill in the gaps for implementation as well.

jhardy (approved):
LGTM

PR #2373, opened 2021-05-18
feat(HoverDisclosure): Support strategy to allow developer to use display or visibility control

lukelooker:

- [x] 📚 Documentation updated
- [x] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/Tree/stories/FieldItem.tsx
```diff
@@ -0,0 +1,155 @@
...
+  selected = false,
+}) => {
+  const [overlay, setOverlay] = useState<string | undefined>(undefined)
+
+  const [isFilter, setIsFilter] = useState(filter)
+  const [isPivot, setIsPivot] = useState(pivot)
+  const [isSelected, setIsSelected] = useState(selected)
+
+  const toggleMenu = () =>
+    overlay === 'menu' ? setOverlay(undefined) : setOverlay('menu')
```
ElliotTheHuman, line 67: nit: Because we don't have an overlay for the `filter` icon button anymore, I think we can simplify this callback and the `overlay` state to just be boolean. So...

```
  const [isFieldMenuOpen, setIsFieldMenuOpen] = useState<boolean>(false)
  ...
  const toggleMenu = () =>
    isFieldMenuOpen ? setIsFieldMenuOpen(false) : setIsFieldMenuOpen(true)
```

ElliotTheHuman: Would also need to change `hoverDisclosure: !overlay` to `hoverDisclosure: !isFieldMenuOpen`

ElliotTheHuman (approved):
Minor nits. Otherwise, LGTM!

PR #2374, opened 2021-05-18
fix(eslint): Inline license header content

dbchristopher:
* Move eslint-header content out of `config/license-header.js` and embed it directly in `eslint-config/license-header.js` configuration. 
* Update license header formatting to match how eslint renders it.
* Fix a few eslint errors like no-extraneous-dependencies which had snuck into the codebase somehow.


packages/eslint-config/license-header.js
```diff
@@ -38,6 +38,37 @@ module.exports = {
   ],
   plugins: ['header'],
   rules: {
-    'header/header': [2, './config/license-header.js'],
```
dbchristopher, line 41: Unfortunately this filepath has to be relative to the monorepo root, rather than the package itself. So the only way I could actually ship the license header text with the `eslint-config` package was to inline the content directly here.

From the docs:

> Due to limitations in eslint plugins, the file is read relative to the working directory that eslint is executed in. If you run eslint from elsewhere in your tree then the header file will not be found.

packages/eslint-config/license-header.js
```diff
@@ -38,6 +38,37 @@ module.exports = {
...
+    'header/header': [
+      2,
+      'block',
```
dbchristopher, line 43: The way eslint-header renders a comment block does not include a space before the closing `*/`. So this PR includes like a thousand instances of this:

Screen Shot 2021-05-18 at 3 45 16 PM

storybook/package.json
```diff
@@ -24,6 +24,8 @@
     "@storybook/addon-a11y": "^6.1.15",
     "@storybook/addon-essentials": "^6.1.15",
     "@storybook/addon-storyshots": "^6.1.15",
-    "@storybook/addon-storyshots-puppeteer": "^6.1.15"
+    "@storybook/addon-storyshots-puppeteer": "^6.1.15",
+    "babel-loader-exclude-node-modules-except": "^1.1.2",
+    "tsconfig-paths-webpack-plugin": "^3.5.1"
```
dbchristopher, line 29: I got some precommit errors about extraneous dependencies. I added the deps explicitly to resolve that. 

I'm not sure why these errors all the sudden appeared _now_. My guess is that changing so many files flushed some eslint cache which caused some of these files to actually get checked for the first time in a while.

packages/icons/package.json
```diff
@@ -21,7 +21,8 @@
     "prebuild": "./bin/prebuild.js"
   },
   "dependencies": {
-    "@styled-icons/styled-icon": "^10.6.3"
+    "@styled-icons/styled-icon": "^10.6.3",
+    "h2x-types": "^1.1.0"
```
dbchristopher, line 25: extraneous dep

packages/icons/bin/prebuild.js
```diff
@@ -37,14 +37,33 @@ const fg = require('fast-glob')
...
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+*/
+
+`
```
dbchristopher: Somehow the `config/license-header.js` file itself was now failing the eslint-header check. Since the content was going to be duplicated anyways I just went ahead and inlined the text content.

packages/icons/package.json
lukelooker: Should be a `devDependency` so it's not pushed into downstream consumer's dependency graphs (`h2x` is just used in the build process)

lukelooker (commented):
Opened an alternative, narrower PR (yours squashed with a slightly tweaked eslint setup that adds the leading space to the closing comment and removes the 2 required newlines afterwards): https://github.com/looker-open-source/components/pull/2395

My main concern bumping the formatting around on all of the files as part of this PR is that it will make scanning git history a bit harder so I wanted to see if it was possible to make the narrowest change possible. I'd argue when we need to update the license language (either re-licencing or 2021 => 2022) would be a good time to also trim that closing white space out and perhaps make the following break required rather than optional (we'll have to have funny Git history then whether we like it or not)

Thoughts?

PR #2397, opened 2021-05-19
feat(Tree): Keyboard Navigation

lukelooker:
Re-merging this PR. Identical to version previously reviewed in #2295


ElliotTheHuman (approved):
My cloudtop is rebooting so can't test locally atm. That being said, it's identical to my previous PR so should be 👍

PR #2398, opened 2021-05-19
feat(HoverDisclosure): Support strategy to allow developer to use display or visibility control

lukelooker:
This is a re-do of #2373


ElliotTheHuman (approved):
:shipit:

PR #2406, opened 2021-05-20
fix(Tree): Properly disperse ListItem related props onto inner TreeItem

ElliotTheHuman:
- Allowed (most) props to be passed to `Tree`'s inner `TreeItem`
- Added tests to check `itemRole` prop behavior on `Tree`

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/Tree/Tree.tsx
```diff
@@ -156,12 +166,7 @@ const TreeLayout: FC<TreeProps> = ({
       aria-current={current}
       aria-selected={selected}
       content={
-        <List
-          density={density}
```
ElliotTheHuman, line 160: Unrelated change, but I'm fairly certain passing to `List`'s `density` prop is unnecessary. `List` will just pass this to a `ListItemContext`, but `Tree`'s and `TreeItem`'s get their density values from `TreeContext` which receives `density` below.

ElliotTheHuman: If snapshots pass, I feel good about removing this to avoid confusing folks as to why `density` is passed to both `List` and `TreeContext.Provider`.

packages/components/src/Tree/Tree.test.tsx
```diff
@@ -268,4 +268,30 @@ describe('Tree', () => {
...
+      expect(screen.getByText('Default Tree').closest('a')).toBe(null)
+    })
+
+    test('renders inner <TreeItem> as <a> and receives link-related props when itemRole="link"', () => {
+      renderWithTheme(
+        <Tree
+          href="https://google.com"
+          itemRole="link"
+          label="Link Tree"
+          target="_blank"
```
lukelooker, line 286: Should probably verify that `onClick` can "block" `href` so that patterns like React Router will work properly.

Basically `onClick` needs to get assigned at the same level within the DOM as `href`

lukelooker (approved):
LGTM! :)

lukelooker (approved):

PR #2408, opened 2021-05-21
fix(Panel): always stack on top

mdodgelooker:
### Issue
https://codesandbox.io/s/optimistic-snowflake-e43le

Any element in `Panels` that comes after `Panel` and has `position: relative` (or `absolute`) will stack on top of the panel surface. This is happening in the core product left nav due to a recent fix for a focus-ring display bug using `position: relative` on `AccordionDisclosure` inside of `Tree`.

### Fix
Render `Panel` in a `Portal` and use context to store the dimensions and position of `Panels` for use in `usePanel`. 

FWIW I did try another route of artificially moving the panel(s) down the bottom of `Panels` instead of using `Portal`, but gave up once it started getting ridiculously complex.

*Verified fix in core product.*

**Note:** This PR does remove `direction` from `PanelHeader` which was a TODO, but it's not a breaking change since `PanelHeader` was not exported from the `index.ts`.

- [ ] ♿️ Accessibility addressed – **I did not look at this at all. A future PR, perhaps?**
- [ ] 📚 Documentation updated
- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11


packages/components/src/Panel/PanelSurface.tsx
```diff
@@ -29,7 +29,7 @@ import { variant } from 'styled-system'
 import { PanelSurfaceProps } from './types'
 
 const surfaceTransition = () => css`
-  ${({ theme }) => `${theme.transitions.moderate}ms ${theme.easings.ease}`}
+  ${({ theme }) => `${theme.transitions.moderate}ms ${theme.easings.easeOut}`}
```
mdodgelooker, line 32: Feel free to object –`ease` just looked a little intense when the `Panel` is wide. (See http://localhost:3333/?path=/story/panel--center-placement)

packages/components/src/Panel/PanelSurface.tsx
```diff
@@ -45,6 +45,8 @@ export const PanelSurface = styled.div.attrs<PanelSurfaceProps>(
 )<PanelSurfaceProps>`
   background: ${({ theme }) => theme.colors.background};
   bottom: 0;
+  display: flex;
+  flex-direction: column;
```
mdodgelooker, line 49: Supports scrollability of content.

packages/components/src/Panel/PanelWindow.tsx
```diff
@@ -0,0 +1,44 @@
...
+    top,
+    width,
+  })
+)<PanelSurfaceProps>`
+  ${height}
+  ${width}
+  ${position}
+  overflow: hidden;
+  position: absolute;
+`
```
mdodgelooker, line 52: This component supports animating a `Panel` from a direction that's not attached to the edge of the screen. (`overflow: hidden` hides the bit you don't want to see as it animates.)

packages/components/src/Panel/Panels.tsx
```diff
@@ -24,9 +24,31 @@
...
+  (
+    props: CompatibleHTMLProps<HTMLDivElement>,
+    forwardedRef: Ref<HTMLDivElement>
+  ) => {
+    const [element, ref] = useCallbackRef(forwardedRef)
+    const [rect] = useMeasuredElement(element)
+    return (
+      <PanelsContext.Provider value={rect}>
+        <div ref={ref} {...props} />
+      </PanelsContext.Provider>
```
mdodgelooker, line 44: Captures the dimensions and position and saves them in context.

packages/components/src/Panel/usePanel.tsx
```diff
@@ -57,10 +62,12 @@ export const usePanel = ({
 
   const isOpen = isControlled ? controlledIsOpen || false : uncontrolledIsOpen
 
+  const firstRender = useRef(true)
   const { busy, className, renderDOM } = useAnimationState(
     isOpen,
-    defaultOpen ? 'none' : undefined
+    isOpen && firstRender.current ? 'none' : undefined
```
mdodgelooker, line 68: Using `defaultOpen` here resulted in the `Panel` never animating in on subsequent re-opening.

packages/components/src/Panel/usePanel.tsx
```diff
@@ -101,3 +117,8 @@ export const usePanel = ({
     setOpen,
   }
 }
+
+const PanelContent = styled.div`
+  flex: 1;
+  overflow: auto;
+`
```
mdodgelooker, line 124: Support scrolling of long content.

packages/components/src/Panel/PanelHeader.tsx
```diff
@@ -25,58 +25,36 @@
...
-   * callback to close Panel
-   */
-  handleClose?: () => void
-  title: string
-}
-
-export const PanelHeader: FC<PanelHeaderProps> = ({
-  className,
-  handleClose,
+export const PanelHeader: FC<PanelBaseProps> = ({
```
mdodgelooker, line 36: Did a little interface refactoring in `types.ts`.

packages/components/src/Panel/types.ts
```diff
@@ -30,27 +30,45 @@ export type PanelDirection = 'left' | 'right'
...
-export interface UsePanelResponse {
+export type UsePanelResponse = {
   isOpen: boolean
   setOpen: (open: boolean) => void
   panel: ReactNode
   domProps: UsePanelResponseDom
 }
 
-export interface UsePanelProps {
+export type PanelBaseProps = {
```
mdodgelooker, line 52: Extracts the props that are common to `PanelHeader` and `usePanel`.

packages/components/src/Portal/Portal.tsx
```diff
@@ -63,6 +63,13 @@ interface PortalPlacementProps {
    * @default 'center
    */
   vertical?: 'center' | 'top' | 'bottom'
+
+  /**
+   * Enables fixed positioning on the portal.
+   * Setting to false will allow content to scroll with the rest of the page.
+   * @default true
+   */
+  fixed?: boolean
```
mdodgelooker, line 72: Having to add this makes me wonder if `InvisiBox` could be its own component separate from `Portal` that `Dialog` renders (since I think that's what most of its styles are for) and `styleDefenderCSS` gets applied to `#modal-root`? Probably not a big deal though.

packages/components/src/Panel/Panel.test.tsx
```diff
@@ -206,7 +206,7 @@ describe('Panel', () => {
 
     expect(screen.getByText('Panel content')).toBeInTheDocument()
 
-    fireEvent.click(screen.getByText('Close Panel Hook'))
+    fireEvent.click(screen.getByText('CloseTitle Panel Hook'))
```
mdodgelooker, line 209: Due to i18n-izing this text.

www/src/documentation/components/layout/panel.mdx
```diff
@@ -43,8 +43,8 @@ Use the `Panel` component to display content with a directional animation. All `
 We provide a custom hook that returns the opener function and rendered panel.
 Note: `panel` needs to be placed before `children`. The order in which the `panel` will be render metters for the component to be rendered correctly.
 
-```jsx
-;() => {
+```jsx static
+const PanelListItem = () => {
```
mdodgelooker, line 47: I had to make this the `usePanel` example static code only because now if you call this hook from outside of `Panels`, it doesn't know the proper dimensions and the panel covers the entire page. And as far as I know there's no way to declare _then_ render a new component in a live example.

laraharrow (dismissed):
Thanks for all the feedback great to look into it and way more clear!
LGTM🚢

lukelooker (approved):
Works flawlessly. Bravo!

packages/components/src/Panel/PanelSurface.tsx
jhardy: no objections

packages/components-providers/src/I18n/resources.ts
```diff
@@ -118,6 +118,9 @@ export const i18nResources: Resource = {
       'Previous page of results': 'Previous page of results',
       of: 'of',
     },
+    PanelHeader: {
+      CloseTitle: 'Close {{title}}',
+    },
```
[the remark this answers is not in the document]
jhardy: I am wondering if we want to allow customization of this the `CloseTitle`. I can see in some cases wanting to say `Back  to {{previous panel title}}` in some cases instead.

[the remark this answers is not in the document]
mdodgelooker: Really good point @jhardy - I’d love to make that tweak and a couple other small things I belatedly thought of when I get back. If the fix can wait that long.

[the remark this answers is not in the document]
lukelooker: Fix can totally wait for that refinement (I'm 🤞  that we won't be doing any more releases until 2.0 so this will ship with that anyway)

www/src/documentation/components/layout/panel.mdx
lukelooker: I think it's nice to make clear it exists. I'd say leave it for now :)

lukelooker (approved):
LGTM :)

PR #2410, opened 2021-05-21
fix(design-tokens): text5 color is now identical to text (blend=100%)

lukelooker:
Make text5 = text

### DIFF
--- a/packages/design-tokens/src/color/blendPoints.ts
+++ b/packages/design-tokens/src/color/blendPoints.ts
@@ -24,8 +24,7 @@
 
  */
 
-export type TextColorLevels = 0 | 1 | 2 | 3 | 4 | 5
-export const textBlends = [45, 65, 78, 88, 95, 99]
+export const textBlends = [45, 65, 78, 88, 100]
 
 export type UIColorLevels = 0 | 1 | 2 | 3 | 4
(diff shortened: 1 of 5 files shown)


packages/design-tokens/src/color/blendPoints.ts
```diff
@@ -25,7 +25,7 @@
  */
 
 export type TextColorLevels = 0 | 1 | 2 | 3 | 4 | 5
-export const textBlends = [45, 65, 78, 88, 95, 99]
```
lukelooker, line 28: This is the _actual_ change.

lukelooker: Changes text5 from being `95%` to `100%` - upper bound here wasn't in use (mistake)

packages/design-tokens/src/color/utils/mixColors.ts
```diff
@@ -30,4 +30,5 @@ export const mixColors = (
   mixAmount: number,
   foreground: string,
   background: string
-) => mix(mixAmount / 100, foreground, background)
+) =>
+  mixAmount === 100 ? foreground : mix(mixAmount / 100, foreground, background)
```
lukelooker, line 34: No need to blend if color mix is 100%

jhardy (approved):
👍

PR #2411, opened 2021-05-21
refactor(ListItem): ARIA prop refactor

ElliotTheHuman:
**Changes:**
- Passed in aria-related props like `aria-current` or `aria-describedby` will now attach to the label container (i.e. the element that receives focus on `ListItem`) rather than the wrapper.
- `current` `ListItem`s no longer receives a darkened background color like `selected` items (i.e. having a `current` `ListItem` does not change the background color at all)

**NOTE:** In the case of `Tree`, the focusable element is the `AccordionDisclosure` div that wraps around the indicator and the nested `TreeItem / ListItem`. To clarify, would this work for DiscoNav or would these aria props also need to land on the nested item in this case? (IIRC, when `labelToggle` is introduced the disclosure and the indicator will be focusable, but _not_ the nested item).

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


ElliotTheHuman:
My only fear with the `current` background change is that `MenuItem`s that use current to get the background color might "break". In browsing HT, I'm seeing at least 2 spots where the current prop is used on a `MenuItem`.

lukelooker (changes requested):
After reading over your thoughts I think we should break this into two different PRs:

1. Move all `aria` props to spread onto the proper target within `ListItem`
2. Deprecate `current` (we can do this as part of the upcoming 2.0 release)


> NOTE: In the case of Tree, the focusable element is the AccordionDisclosure div that wraps around the indicator and the nested TreeItem / ListItem. To clarify, would this work for DiscoNav or would these aria props also need to land on the nested item in this case? (IIRC, when labelToggle is introduced the disclosure and the indicator will be focusable, but not the nested item).

When `labelToggle` is introduced the `aria-*` attributes should land on the nested item – I would assume the nested item would likely be an `<a>` and would therefore need to

ElliotTheHuman:
Think splitting these off into 2 separate PRs makes sense. I'll close this out and get new PRs up in that case.

PR #2413, opened 2021-05-21
feat(TabPanel): Added tabStop boolean prop to TabPanel

ElliotTheHuman:
In the new Field Picker experience, a keyboard user has to tab from the `Tab` to the `TabPanel` to the first `Tree` in order to focus on the `Tree` items and begin field keyboard traversal. Ideally, I think it'd be cleaner to just tab from the `Tab` directly to the `TreeCollection` since it is the only element in the `TabPanel`.

Adding `tabIndex` as a prop allows devs to pass a non-zero value to `TabPanel`'s `tabIndex` prop and thus remove it from the tab flow.

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Tabs/TabPanel.tsx
+++ b/packages/components/src/Tabs/TabPanel.tsx
@@ -30,6 +30,10 @@ import styled from 'styled-components'
 export interface TabPanelProps {
   className?: string
   selected?: boolean
+  /**
+   * Set tabIndex to 0 if you'd like the TabPanel to be reached via keyboard tabbing
+   * @default -1
+   */
   tabIndex?: number
 }
 
@@ -37,7 +41,7 @@ const TabPanelLayout: FC<TabPanelProps> = ({
   children,
   className,
   selected,
-  tabIndex = 0,
+  tabIndex = -1,
 }) =>
   selected ? (
     <div className={className} tabIndex={tabIndex}>
--- a/packages/components/src/Tabs/TabPanel.tsx
+++ b/packages/components/src/Tabs/TabPanel.tsx
@@ -42,10 +42,10 @@ const TabPanelLayout: FC<TabPanelProps> = ({
   children,
   className,
   selected,
-  tabIndex = -1,
+  tabStop = false,
 }) =>
   selected ? (
-    <div className={className} tabIndex={tabIndex}>
+    <div className={className} tabIndex={tabStop ? 0 : -1}>
       {children}
     </div>
   ) : null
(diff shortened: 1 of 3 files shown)


ElliotTheHuman:
There's an argument to add `CompatibleHTMLProps<HTMLDivElement>` to `Tab` but I wasn't sure if that was excluded here purposefully.

packages/components/src/Tabs/TabPanel.tsx
```diff
@@ -30,11 +30,21 @@ import styled from 'styled-components'
 export interface TabPanelProps {
   className?: string
   selected?: boolean
+  /**
+   * Set tabIndex to 0 if you'd like the TabPanel to be reached via keyboard tabbing
+   * @default -1
+   */
+  tabIndex?: number
```
lukelooker, line 37:
```suggestion
  /**
   * Set to `true` if you would like TabPanel to be reached via tab-key.
   * Generally this is _only_ the case when the TabPanel contains no tab-stopping items (a, button, etc.)
   * @default false
   */
  tabStop?: boolean
```

lukelooker (changes requested):
I'd prefer not to ask developers to understand `tabIndex` and instead offer a more semantic prop. See proposed change.

lukelooker (dismissed):
Assuming @jhardy's research supports this I'm happy with it :)

jhardy:
I dug in a bit more and this seems to align closely with some [updated guidance from the w3c aria best practices.](https://github.com/w3c/aria-practices/issues/1164) This is not published on their official docs just yet, but the new guidance is for the `tab` key is 

> - When focus moves into the tab list, places focus on the active tab element.
> - When the tab list contains the focus, moves focus to the next element in the page tab sequence outside the tablist, which is the tabpanel unless the first element containing meaningful content inside the tabpanel is focusable. 

I am wondering about the prop name `tabStop`, that feels a bit too generic as first glance. I don't dislike it enough to block the PR. I am wondering if a more descriptive prop name but could be something we reuse elsewhere if we need to give the option to programmatically add focus.

`<TabPanel makeFocusable>` 
`<TabPanel includeInTabSequence>`
`<TabPanel addAsTabStop>`
`<TabPanel canReceiveFocus>`

Not sure if those are any better or too verbose, but just a few early morning thoughts.

lukelooker:
Great point on naming... what about `isTabStop`? (fits with our other `isX` conventions)

jhardy:
that works for me!

lukelooker (approved):
LGTM!

PR #2426, opened 2021-05-24
fix(ListItem): Removed background transition

ElliotTheHuman:
The `HoverDisclosure` component can be used to conditionally hide and show elements in a `ListItem` on hover. However, child elements of `HoverDisclosure` do not have a transition when appearing or disappearing, which makes the background hover appear to "lag" behind. 

Removing the background transition gets both `HoverDisclosure` children and the background color change to happen at around the same time.

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Tree/stories/FieldItem.tsx
+++ b/packages/components/src/Tree/stories/FieldItem.tsx
@@ -110,7 +110,10 @@ export const FieldItem: FC<FieldItemProps> = ({
       onKeyDown={(event) => {
         if (event.key === 'Enter' && event.metaKey) {
           alert(`CMD + Enter'ed on ${children}!`)
-        } else if (event.key === 'Enter') {
+        } else if (
+          event.key === 'Enter' &&
+          event.currentTarget === event.target
+        ) {
           toggleField()
         }
       }}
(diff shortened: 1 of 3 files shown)


packages/components/src/Tree/stories/FieldItem.tsx
```diff
@@ -110,7 +110,10 @@ export const FieldItem: FC<FieldItemProps> = ({
       onKeyDown={(event) => {
         if (event.key === 'Enter' && event.metaKey) {
           alert(`CMD + Enter'ed on ${children}!`)
-        } else if (event.key === 'Enter') {
+        } else if (
+          event.key === 'Enter' &&
+          event.currentTarget === event.target
+        ) {
```
ElliotTheHuman, line 116: We _don't_ want icon key presses to trigger the parent item's selected state toggle handler. This `onKeyDown` should only fire when focused on the item's label container and pressing "Enter".

lukelooker (approved):
LGTM.

Once merged I re-open a PR from `release-1.4` branch to backport the squashed commit :)

PR #2431, opened 2021-05-25
feat(ListItem): Remove "current" prop

ElliotTheHuman:
- Removed the "current" prop from `ListItem`'s interface
- Removed all use of "current" from `ListItem`/`TreeItem`/`MenuItem` stories and tests

BREAKING CHANGE: removed the `current` prop from `ListItem`'s interface


packages/components/src/List/ListItemWrapper.tsx
```diff
@@ -86,11 +86,6 @@ export const ListItemWrapper = styled(ListItemWrapperInternal)
     }
   }
 
-  &[aria-current='true'] {
-    background: ${({ theme: { colors } }) => colors.ui2};
-    font-weight: ${({ theme: { fontWeights } }) => fontWeights.semiBold};
-  }
-
```
ElliotTheHuman, line 93: Pretty sure this isn't necessary since background color is handled by the `ListItemLabel` component. In addition, we _don't_ want `aria-current` to trigger a background color change, if I understand correctly.

lukelooker (approved):
Awesome-sauce!

Please make sure the following is included in the squash message:

```
Removed the "current" prop from ListItem's interface
Removed all use of "current" from ListItem/TreeItem/MenuItem stories and tests

BREAKING CHANGE: removed the current prop from ListItem's interface
```

The "BREAKING CHANGE" _should_ trigger a MAJOR release (see https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-description-and-breaking-change-footer)

NOTE: `feat!` doesn't seem to work as-expected with Lerna's CHANGELOG extraction (and in fact seems to completely hide them from CHANGELOG)

lukelooker (approved):
Sold!

PR #2433, opened 2021-05-25
fix(ListItem): Properly disperse ARIA props onto ListItem's label container

ElliotTheHuman:
Moved aria related props to be passed to label rather than wrapper

**TODO:** Follow-up PR that does the following:
- Extract "aria partitioning" logic into its own function and file (`Tree` can then use this same function to partition out its `aria`-related props and send them to the nested item when `labelToggle === true`)
- Move `ListItemProps` to a separate file (need to do this otherwise we'll have a circular dependency between the "aria partitioning" function file and `ListItem.tsx`)

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/List/ListItem.tsx
+++ b/packages/components/src/List/ListItem.tsx
@@ -266,6 +266,14 @@ const ListItemInternal = forwardRef(
       selected,
     }
 
+    const ariaProps = {}
+    const wrapperProps = {}
+    Object.entries(restProps).forEach(([propKey, propValue]) =>
+      propKey.startsWith('aria-')
+        ? (ariaProps[propKey] = propValue)
+        : (wrapperProps[propKey] = propValue)
+    )
+
     const LabelCreator: FC<{
       children: ReactNode
       className?: string
@@ -283,6 +291,7 @@ const ListItemInternal = forwardRef(
         role={role || 'listitem'}
         target={target}
         tabIndex={tabIndex}
+        {...ariaProps}
         {...focusVisibleHandlers}
         {...statefulProps}
       >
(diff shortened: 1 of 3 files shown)


packages/components/src/List/ListItem.tsx
```diff
@@ -266,6 +266,14 @@ const ListItemInternal = forwardRef(
       selected,
     }
 
+    const ariaProps = {}
+    const wrapperProps = {}
+    Object.entries(restProps).forEach(([propKey, propValue]) =>
+      propKey.startsWith('aria-')
+        ? (ariaProps[propKey] = propValue)
+        : (wrapperProps[propKey] = propValue)
+    )
```
lukelooker, line 275: I think it'd be nice to extract this into a helper function:

```
const [ ariaProps, wrapperProps ] = extractAriaProps(props: {}) => {
  const aria = {}
  const other = {}
    
  Object.entries(restProps).forEach(([key, value]) =>
    key.startsWith('aria-') ? (aria[key] = value) : (other[key] = value)
  )

    return [ aria, other ]
}
```

lukelooker (changes requested):

packages/components/src/List/ListItem.tsx
lukelooker: Then I read your PR comments more closely and realized you're planning this in a later PR. Feel free to defer to later if you want.

lukelooker (approved):

PR #2436, opened 2021-05-26
refactor(ListItem): Created getAriaProps helper function

ElliotTheHuman:
- Moved ARIA prop partitioning logic into its own `getAriaProps` helper function
- Moved `ListItemProps` into `./types` file to avoid circular dependency

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/List/utils/getAriaProps.ts
```diff
@@ -0,0 +1,42 @@
...
+    propKey.startsWith('aria-')
+      ? (ariaProps[propKey] = propValue)
+      : (restProps[propKey] = propValue)
+  )
+
+  return {
+    ariaProps,
+    restProps,
+  }
+}
```
ElliotTheHuman, line 42: This is the bulk of the PR, but the rest of the LOCs were moving `ListItemProps` into the types file and changing its references.

packages/components/src/List/utils/getAriaProps.ts
```diff
@@ -0,0 +1,42 @@
...
+  Object.entries(props).forEach(([propKey, propValue]) =>
+    propKey.startsWith('aria-')
+      ? (ariaProps[propKey] = propValue)
+      : (restProps[propKey] = propValue)
+  )
+
+  return {
+    ariaProps,
+    restProps,
+  }
```
lukelooker, line 41: Rather than returning an object return this as a tuple:

```suggestion
  return [
    ariaProps,
    restProps,
  ]
```

lukelooker (changes requested):

packages/components/src/List/ListItem.tsx
```diff
@@ -266,12 +205,8 @@ const ListItemInternal = forwardRef(
...
     }
 
-    const ariaProps = {}
-    const wrapperProps = {}
-    Object.entries(restProps).forEach(([propKey, propValue]) =>
-      propKey.startsWith('aria-')
-        ? (ariaProps[propKey] = propValue)
-        : (wrapperProps[propKey] = propValue)
+    const [ariaProps, wrapperProps] = getAriaProps(
+      restProps as Record<string, unknown>
```
ElliotTheHuman, line 209: I really don't want to have the `as` keyword here, but I keep getting an "Index signature is missing in type" error whenever I try to pass `restProps` directly.

Everything I've read online points to replacing `interface`s with `type`s to get around this but I'm already using the `type` keyword for `ListItemProps` which is where `restProps` gets its typing.

packages/components/src/List/utils/getAriaProps.ts
```diff
@@ -0,0 +1,44 @@
...
+  const ariaProps = {}
+  const restProps = {}
+  Object.entries(props).forEach(([propKey, propValue]) =>
+    propKey.startsWith('aria-')
+      ? (ariaProps[propKey] = propValue)
+      : (restProps[propKey] = propValue)
+  )
+
+  return [ariaProps, restProps]
+}
```
lukelooker, line 44:
```suggestion
/**
 * Partitions an object into 2 objects, the first containing all aria related prop keys and their respective values
 * and the second containing all other prop keys and their respective values
 *
 * @param {Record<string, unknown>} props
 * @returns {Array} A tuple where the first object contains all aria related props and the second object contains the remaining props
 */
export const getAriaProps = (props: Record<string, unknown>) => {
  const aria = {}
  const remainder = {}
  Object.entries(props).forEach(([key, value]) =>
    key.startsWith('aria-') ? (aria[key] = value) : (remainder[key] = value)
  )

  return [aria, remainder]
}
```

lukelooker (approved):
Ship it!

PR #2437, opened 2021-05-26
refactor(ListItem): Refactor getAriaProps

ElliotTheHuman:
- Changed name to `partitionAriaProps` (after lodash's partition method)
- Updated props parameter type to use generic type that extends `React.AriaAttributes` type

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/List/utils/getAriaProps.ts
+++ b/packages/components/src/List/utils/getAriaProps.ts
@@ -28,10 +28,10 @@
  * Partitions an object into 2 objects, the first containing all aria related prop keys and their respective values
  * and the second containing all other prop keys and their respective values
  *
- * @param {Record<string, unknown>} props
+ * @param {Record<string, any>} props
  * @returns {Array} A tuple where the first object contains all aria related props and the second object contains the remaining props
  */
-export const getAriaProps = (props: Record<string, unknown>) => {
+export const getAriaProps = (props: Record<string, any>) => {
   const aria = {}
   const remainder = {}
   Object.entries(props).forEach(([key, value]) =>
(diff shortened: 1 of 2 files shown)


packages/components/src/List/utils/getAriaProps.ts
```diff
@@ -28,10 +28,10 @@
  * Partitions an object into 2 objects, the first containing all aria related prop keys and their respective values
  * and the second containing all other prop keys and their respective values
  *
- * @param {Record<string, unknown>} props
+ * @param {Record<string, any>} props
  * @returns {Array} A tuple where the first object contains all aria related props and the second object contains the remaining props
  */
-export const getAriaProps = (props: Record<string, unknown>) => {
+export const getAriaProps = (props: Record<string, any>) => {
```
ElliotTheHuman, line 34: There's a choice here, "Should we use `Record<string, any>` or `object` as the parameter type?"

Using `Record<string, any>` doesn't require an eslint ignore, so I'm opting for that. In addition, the only place the `props` parameter is used is in the `Object.entries` call. `Object.entries`' parameter typing is similar (i.e. string key, any value) so using any in the `props` typing _shouldn't_ have any major pitfalls afaik.
Screen Shot 2021-05-26 at 3 58 40 PM

lukelooker: I can live the `any` in this case - it's a funny trade-off, will be interesting to see if TS eventually offers a better way to type this scenario. :)

lukelooker (approved):
LGTM

lukelooker (approved):
Sold! (yet again)

PR #2445, opened 2021-06-01
feat(palette): No longer export palette, mark all as deprecated

lukelooker:
Palette can be deep-imported during 2.x series but may be removed entirely in 3.x
```import { palette} from '@looker/design-tokens/lib/legacy'```

BREAKING CHANGE: Design-tokens `palette` is no longer exported from root, deprecated for 3.x


mdodgelooker (approved):
LGTM

PR #2448, opened 2021-06-01
feat(PopoverContent): new component to enhance Popover layout

laraharrow:
Work done:
- Created `ModalContent` a simplified version of `DialogContent` to be used on `PopoverContent` as well - DRY.
- Simplified the work done on `DialogContent` using now `Modal`.
- PopoverContent` was restructured to use `Modal`

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Modal/ModalContent/ModalContent.text.tsx
+++ b/packages/components/src/Modal/ModalContent/ModalContent.text.tsx
@@ -24,10 +24,12 @@
 
  */
 
+/* eslint-disable i18next/no-literal-string */
+
 import React from 'react'
 import { renderWithTheme } from '@looker/components-test-utils'
 import { screen } from '@testing-library/react'
-import { DialogContent } from './DialogContent'
+import { ModalContent } from './ModalContent'
 
 const originalScrollHeight = Object.getOwnPropertyDescriptor(
   HTMLElement.prototype,
@@ -57,20 +59,20 @@ afterAll(() => {
 
 describe('DialogContent', () => {
   test('basic', () => {
-    renderWithTheme(<DialogContent>Stuff</DialogContent>)
+    renderWithTheme(<ModalContent>Stuff</ModalContent>)
     expect(screen.getByText('Stuff')).toBeInTheDocument()
   })
 
   test('hasHeader & hasFooter', () => {
     renderWithTheme(
-      <DialogContent hasHeader hasFooter>
+      <ModalContent hasHeader hasFooter>
         Stuff
-      </DialogContent>
+      </ModalContent>
     )
     expect(screen.getByText('Stuff')).toBeInTheDocument()
   })
 
-  test('content does not have a box shadow when content does not overflow', () => {
+  test('has no box shadow when it does not overflow', () => {
     Object.defineProperty(HTMLElement.prototype, 'scrollHeight', {
       configurable: true,
       value: 0,
@@ -80,19 +82,17 @@ describe('DialogContent', () => {
       value: 500,
     })
     renderWithTheme(
-      <DialogContent hasHeader hasFooter>
+      <ModalContent hasHeader hasFooter>
         Stuff
-      </DialogContent>
+      </ModalContent>
     )
 
     expect(
-      getComputedStyle(screen.getByTestId('dialog-content')).getPropertyValue(
-        'box-shadow'
-      )
+      getComputedStyle(screen.getByText('Stuff')).getPropertyValue('box-shadow')
     ).toEqual('')
   })
 
-  test('content has a box shadow when content overflows', () => {
+  test('has a box shadow when it overflows', () => {
     Object.defineProperty(HTMLElement.prototype, 'scrollHeight', {
       configurable: true,
       value: 500,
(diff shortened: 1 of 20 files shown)


packages/components/src/Modal/ModalContent/ModalContent.tsx
```diff
@@ -81,34 +82,29 @@ export const DialogContent: FC<DialogContentProps> = ({
...
   }, [height])
-
   return (
-    <InnerDialogContent
+    <InnerModalContent
       hasOverflow={hasOverflow}
       ref={internalRef}
-      px={['medium', 'xlarge']}
       pb={hasOverflow || !!hasFooter ? 'large' : 'xxxsmall'}
       pt={hasOverflow || !!hasHeader ? 'large' : 'xxxsmall'}
```
laraharrow, line 90: I know the ticket asked for ModalContent to not have padding. But because of the expected behavior in Dialog and in Popover I'm keeping this logic here since hasOverflow and hasFooter and hasHeader are Modal props. Open to suggestions if ModalContent really shouldn't have any padding.

lukelooker: I think this case for keeping the top & bottom padding here makes sense. 

It would be great to have specific tests for `hasFooter` & `hasHeader` that validate that padding gets assigned since this behavior is pretty nuanced. Noted on the test file as well.

packages/components/src/Modal/ModalContent/ModalContent.text.tsx
```diff
@@ -57,20 +59,20 @@ afterAll(() => {
...
 
   test('hasHeader & hasFooter', () => {
     renderWithTheme(
-      <DialogContent hasHeader hasFooter>
+      <ModalContent hasHeader hasFooter>
         Stuff
-      </DialogContent>
+      </ModalContent>
     )
     expect(screen.getByText('Stuff')).toBeInTheDocument()
```
lukelooker, line 72: I think it'd be great to add test coverage here to verify that top & bottom padding values get set properly and to also check the negative state in the "basic" test above.

packages/components/src/Popover/Layout/PopoverContent/PopoverContent.tsx
```diff
@@ -0,0 +1,42 @@
...
+import React, { FC } from 'react'
+import styled from 'styled-components'
+import { ModalContent, ModalContentProps } from '../../../Modal/ModalContent'
+
+const PopoverContentLayout: FC<ModalContentProps> = ({
+  children,
+  ...props
+}) => {
+  return (
+    <ModalContent pb="medium" pt="xsmall" px="large" {...props}>
```
lukelooker, line 36: Won't the `pb` & `pt` values get over-written by the values specified on `InnerModalContent` if it has overflow?

lukelooker (changes requested):
Looks pretty good overall but had some specific questions on implementation and a few test coverage additions that I think would be helpful.

Finally, I'd love to see `useOverflow` get leveraged in `DialogContent` to remove the complexity from that component since we have the hook available now.

packages/components/src/Modal/ModalContent/ModalContent.tsx
laraharrow: on a convo we had in some other PR I had done tests specifically for `paddings` I got a review that having that on storybook was enough. But I'm happy to add tests for sure

lukelooker: In that case the padding was a static value - in this case since it's controlled by the `hasHeader` / `hasFooter` props I'd like to have "proof" that they're working (mostly so we know if they regress) :)

packages/components/src/Modal/ModalContent/ModalContent.text.tsx
laraharrow: I was assuming the style testings were better done on storybook. I thought we had a convo about that before where I remove the tests that were related with padding since that gets exemplified in the stories.

packages/components/src/Popover/Layout/PopoverContent/PopoverContent.tsx
laraharrow: Yes I assume it will, but I was also assuming that it was the expected behavior.

lukelooker (commented):
Lots of snapshot breakage to address.

packages/components/src/Popover/Layout/PopoverContent/PopoverContent.tsx
lukelooker: I do think the PopoverContent padding settings need to be able to override the default settings in ModalContent so that PopoverContent fits the specs in Figma.

laraharrow:
I made ModalContent from a `FC` into a `forwardRef` that is exported as a `Styled Component` so I could use the logic on `useOverflow`. Some how the style on `ModalContent` is not rendering.
I added test for checking the padding of `modalContent` if ether of the specified props on the component are true.
I also added some logic to make sure `padding` is correct when `hasOverflow` and used on `PopoverContent`

lukelooker (approved):
LGTM!

PR #2449, opened 2021-06-01
fix(Tree): Updated Tree icon / indicator alignment

ElliotTheHuman:
**Changes:**
- Removed iconGapAdjuster
- Set consistent iconGap across densities (xxsmall)
- Scaled indicator icon according to density
- Added `iconSpacing` prop to align labels of `Tree`s with icons and labels of child `Tree`s without icons

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/List/utils/index.ts
```diff
@@ -25,7 +25,7 @@
  */
 
 export * from './createSafeRel'
-export * from './getAriaProps'
+export * from './partitionAriaProps'
```
ElliotTheHuman, line 28: Unrelated: Correcting the file name on this to reflect the actual function name of `partitionAriaProps`

packages/components/src/Tree/TreeArtificial.tsx
```diff
@@ -40,7 +40,13 @@ export const TreeArtificial: FC<{
...
   const { iconGap, iconSize } = listItemDimensions(density)
   return (
-    <TreeStyle depth={-1} iconGap={iconGap} iconSize={iconSize} dividers>
+    <TreeStyle
+      depth={-1}
+      iconGap={iconGap}
+      indicatorGap={iconGap}
+      indicatorSize={iconSize}
+      dividers
+    >
```
ElliotTheHuman, line 49: I plan on deprecating `TreeArtificial` as soon as this update gets merged.

For now, I needed to make some minor changes to get TS from complaining.

packages/components/src/Tree/Tree.tsx
```diff
@@ -162,6 +163,8 @@ const TreeLayout: FC<TreeProps> = ({
   const indicatorColor = disabled ? 'text1' : 'text5'
   const innerAccordion = (
     <Accordion
+      {...indicatorDefaults}
+      {...restProps}
```
lukelooker, line 167: Any concerns about breaking previous behavior that might've depended on overriding values "late"?

packages/components/src/Tree/Tree.tsx
```diff
@@ -170,20 +173,22 @@ const TreeLayout: FC<TreeProps> = ({
...
       tabIndex={-1}
-      {...indicatorDefaults}
-      {...restProps}
     >
       {label}
     </Accordion>
   )
 
+  const treeIndicatorGap = 'xxsmall'
+
```
lukelooker, line 191: Since this is used exactly once is there a reason to pull it out into a separate variable?

packages/components/src/Tree/types.ts
```diff
@@ -74,8 +74,16 @@ export type TreeProps = Omit<
      * If true, child `TreeItem` elements will receive additional indent padding to align parent `Tree` (with an optional `icon` prop) and child `TreeItem` labels.
      * If "no-icon", child `TreeItem` elements will receive additional indent padding to align parent `Tree` (without an optional `icon` prop) and child `TreeItem` labels.
      * @default false
+     * @todo - Remove in 2.x release
+     * @deprecated Use `iconSpacing` prop instead
```
lukelooker, line 78: Are we sure it makes sense to drop this in 2.x or perhaps wait for 3.x?

Asked another way, what's the overhead if we don't remove this immediately? Generally we try to indicate deprecations for at least a few releases before removing.

lukelooker: Alternatively, could we be "cute" and make this a proxy for `iconSpacing` in 2.x?

lukelooker (changes requested):
Some questions inline regarding deprecation strategy / plan.

The big question I have is about indicator alignment: 

image


It _appears_ like the indicator is no longer centered in the space it occupies? I think that's incorrect?

packages/components/src/Tree/Tree.tsx
ElliotTheHuman: I don't have any concerns, but this might not be the best change in retrospect. The reason I moved this up was because I wanted my explicit `indicatorGap` to be applied  (which conditionally applies the density-respective `iconGap` or the default `indicatorGap` based on the `iconSpacing` prop).

In c3fd26e, I moved the `restProps` destructure back to the bottom of the props, and destructured `indicatorDefaults` before the `Accordion` JSX (i.e. manually passing in the properties of `indicatorDefaults`).

This adds some lines, but I think it's a little easier to follow since I need to access that `indicatorDefault.indicatorGap` value anyhow 🤷

packages/components/src/Tree/Tree.tsx
ElliotTheHuman: It wouldn't ideally! But I'm getting a lint error when passing a literal string value:
Screen Shot 2021-06-02 at 10 46 51 AM

That being said, considering the comment above, it might make sense to pass the  `indicatorDefaults.indicatorGap (gets renamed to defaultGap)`  variable since I'm destructuring `indicatorDefaults` outside of the `Accordion` JSX.

packages/components/src/Tree/types.ts
ElliotTheHuman: We can probably do either!

1) The engineering overhead for keeping `forceLabelPadding` doesn't seem costly (it's just another variable that gets received by `Tree` and funneled into `TreeStyle` and `generateIndent`). However, supporting it and making sure it plays nice with any future alignment changes could be a nuisance.

2) It shouldn't be hard to make this a proxy for `iconSpacing` in 2.x. The only change we'd need to make would be within `generateIndent`. Happy to change the JSDoc to reflect this upcoming change if you'd like.

ElliotTheHuman:
> Some questions inline regarding deprecation strategy / plan.
> 
> The big question I have is about indicator alignment:
> 
> image
> 
> It _appears_ like the indicator is no longer centered in the space it occupies? I think that's incorrect?

The reason the indicator is no longer centered is because it has a "4px" gutter (in the form of `margin-right`). I _think_ that's the right way to apply that gutter (note: the optional `icon` on `ListItem`s also has a gutter, but this gutter is "4px + x" where x is some amount of minor label faux-padding Jared outlined in this spec:
https://www.figma.com/file/ADmrXwJUbKG0eVAhco8X10/LGM2-Transition?node-id=1495%3A222438

packages/components/src/Tree/types.ts
lukelooker: Ok, yes, I'd say go ahead and make it a "proxy" and mark it as `@deprecated` we can go ahead and remove it in the 3.x series.

jhardy:
> Some questions inline regarding deprecation strategy / plan.
> 
> The big question I have is about indicator alignment:
> 
> image
> 
> It _appears_ like the indicator is no longer centered in the space it occupies? I think that's incorrect?

I think this ok, as Elliot mentioned above its the artifact of the gutter, we could use `padding-left` on the label instead, but not sure if there is a down side to that. 

Either way I think the alignment is working pretty nicely

lukelooker:
> I think this ok, as Elliot mentioned above its the artifact of the gutter, we could use `padding-left` on the label instead, but not sure if there is a down side to that.
> 
> Either way I think the alignment is working pretty nicely

Comparing screenshots makes it look like a really big change but if the design-guy says it's okay I'm cool with it. 👍 

Reading remainder of feedback now. :)

packages/components/src/Tree/types.ts
ElliotTheHuman: Done in 38c0325

jhardy:
> > I think this ok, as Elliot mentioned above its the artifact of the gutter, we could use `padding-left` on the label instead, but not sure if there is a down side to that.
> > Either way I think the alignment is working pretty nicely
> 
> Comparing screenshots makes it look like a really big change but if the design-guy says it's okay I'm cool with it. 👍
> 
> Reading remainder of feedback now. :)

Just to compare side

> > I think this ok, as Elliot mentioned above its the artifact of the gutter, we could use `padding-left` on the label instead, but not sure if there is a down side to that.
> > Either way I think the alignment is working pretty nicely
> 
> Comparing screenshots makes it look like a really big change but if the design-guy says it's okay I'm cool with it. 👍
> 
> Reading remainder of feedback now. :)

Just to close the loop here, previously nested trees did not align with the parent icon, here is a visual side-by-side
image

In new spec we also slightly changed the spacing between the label and the icon, previously we were subtracting two pixels to save for some optical space, but I felt that was over complicated it, so now we stick with the default `ListItem` spacing, hopefully, to keep it a bit simpler. (@ElliotTheHuman can keep me honest here)

ElliotTheHuman:
> Just to close the loop here, previously nested trees did not align with the parent icon, here is a visual side-by-side
> image
> 
> In new spec we also slightly changed the spacing between the label and the icon, previously we were subtracting two pixels to save for some optical space, but I felt that was over complicated it, so now we stick with the default `ListItem` spacing, hopefully, to keep it a bit simpler. (@ElliotTheHuman can keep me honest here)

Correctamundo! This PR pulls out the `iconGapAdjuster` (i.e. the -2px to save for optical space). Now we use the exact   `iconGap` values from the density scale.

lukelooker (dismissed):
Awesome-sauce. 🚢

lukelooker (approved):
👍

PR #2459, opened 2021-06-03
fix(ListItem): Moved cursor styling up to ListItemWrapper

ElliotTheHuman:
This allows `TreeItem`s that are in `labelBackgroundOnly` mode have `cursor: pointer` when users hover over the indent whitespace (previously the cursor would switch to the default cursor when hovering over the indent whitespace).

**QUESTION:** Not sure if we should only apply `cursor: pointer` if an `onClick` or `onWhitespaceClick` is provided. If we're assuming that `ListItem`s should always be clickable elements like a button / link, I think it makes sense to always apply this style to the wrapper container (rather than conditionally when one of X props is provided).

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/List/ListItem.tsx
+++ b/packages/components/src/List/ListItem.tsx
@@ -190,6 +190,7 @@ const ListItemInternal = forwardRef(
     const renderedDetail = detail && (
       <HoverDisclosure visible={!hoverDisclosure}>
         <ListItemDetail
+          cursor={accessory ? 'default' : 'pointer'}
           pl={padding ? 'xsmall' : '0'}
           pr={accessory && padding ? itemDimensions.px : '0'}
         >
--- a/packages/components/src/List/ListItem.tsx
+++ b/packages/components/src/List/ListItem.tsx
@@ -190,7 +190,7 @@ const ListItemInternal = forwardRef(
     const renderedDetail = detail && (
       <HoverDisclosure visible={!hoverDisclosure}>
         <ListItemDetail
-          cursor={accessory ? 'default' : 'pointer'}
+          cursorPointer={!accessory}
           pl={padding ? 'xsmall' : '0'}
           pr={accessory && padding ? itemDimensions.px : '0'}
         >
@@ -216,6 +216,7 @@ const ListItemInternal = forwardRef(
         itemRole={itemRole}
         aria-selected={selected}
         className={className}
+        cursorPointer={!!(href || onClick)}
         focusVisible={focusVisible}
         height={itemDimensions.height}
         href={href}
@@ -287,6 +288,7 @@ const ListItemInternal = forwardRef(
         <ListItemWrapper
           className={className}
           color={listItemLabelColor(color, disabled)}
+          cursorPointer={!!onClickWhitespace}
           description={description}
           disabled={disabled}
           onBlur={handleWrapperBlur}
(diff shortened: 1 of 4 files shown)


packages/components/src/List/ListItem.tsx
```diff
@@ -190,6 +190,7 @@ const ListItemInternal = forwardRef(
     const renderedDetail = detail && (
       <HoverDisclosure visible={!hoverDisclosure}>
         <ListItemDetail
+          cursor={accessory ? 'default' : 'pointer'}
```
lukelooker, line 193: I _think_ it'd be more ideal to do _nothing_ if the `ListItemDetail` is in accessory mode?

```suggestion
          cursor={accessory ? undefined : 'pointer'}
```

packages/components/src/List/ListItemDetail.tsx
```diff
@@ -27,13 +27,21 @@
 import { padding, PaddingProps } from '@looker/design-tokens'
 import styled from 'styled-components'
 
-export const ListItemDetail = styled.div.attrs<PaddingProps>((props) => ({
-  ...props,
-  pl: props.pl || 'xsmall',
-}))<PaddingProps>`
+export type ListItemDetailProps = PaddingProps & {
+  cursor?: string
```
lukelooker, line 31: Could this be narrowed to `boolean` so folks could just say `<ListItemDetail cursorPointer />`

There are a lot of potential values for cursor so i'd prefer to keep the scope to smallest possible set.

packages/components/src/List/ListItemDetail.tsx
```diff
@@ -27,13 +27,21 @@
...
+}
+
+export const ListItemDetail = styled.div.attrs<ListItemDetailProps>(
+  (props) => ({
+    ...props,
+    cursor: props.cursor || 'default',
```
lukelooker, line 37: Ideally I'd like to see this narrowed as discussed above but if not I'd probably "default" to "undefined" or if it needs to be an explicit value `initial`

packages/components/src/List/ListItemWrapper.tsx
```diff
@@ -68,6 +68,7 @@ export const ListItemWrapper = styled(ListItemWrapperInternal)
   })
   .attrs(({ color = 'text5' }) => ({ color }))`
   align-items: center;
+  cursor: pointer;
```
lukelooker, line 71: Should it be a pointer if the item isn't actually clickable?

lukelooker (changes requested):
Had a bunch of questions here... let me know if you want to chat directly instead of back-and-forth via PR :)

packages/components/src/List/ListItem.tsx
```diff
@@ -215,6 +216,7 @@ const ListItemInternal = forwardRef(
         itemRole={itemRole}
         aria-selected={selected}
         className={className}
+        cursorPointer={!!(href || onClick)}
```
ElliotTheHuman, line 219: I don't love adding explicit `cursorPointer` props to `ListItemWrapper` and `ListItemLabel`, but because both their `onClick`s are technically always truthy (i.e. they both receive handlers that wrap the passed in `onClick` / `onClickWhitespace`) it seemed to be the most readable solution.

I could technically rewire things in such a way that the `handleOnClick` and `handleOnClickWhitespace` handlers are undefined if their respective props aren't passed in, but that felt less readable for people seeing the `ListItem` code for the first time. I could definitely be wrong about that though.

lukelooker (approved):
👍

PR #2472, opened 2021-06-04
feat(PopoverLayout ModalLayout): new component, also update DialogLayout

laraharrow:
- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


laraharrow:
Ok I think this is ready for review!
A few notes here:
1. I made a few changes to the snapshots based on the new code that was added. @lukelooker maybe we can chat about what is the best way to have every option well represented but with no repetition?
2. I don't think that the `DialogFooter` storybook examples are exactly relatable to the use cases we have in helltool, not sure if we want to update that.

lukelooker (changes requested):
Needs rebase :)

packages/components/src/Dialog/Layout/DialogHeader/DialogHeader.story.tsx
```diff
@@ -36,9 +36,7 @@ export const Basic = () => <DialogHeader>Heading</DialogHeader>
 
 export const HideClose = () => <DialogHeader hideClose>Heading</DialogHeader>
 
-export const Detail = () => (
-  <DialogHeader detail="Detail text">Heading</DialogHeader>
-)
```
lukelooker, line 41: Can we revert this change?

packages/components/src/Dialog/Layout/DialogHeader/DialogHeader.test.tsx
```diff
@@ -42,11 +42,6 @@ describe('DialogHeader', () => {
     expect(screen.getByText('Close')).toBeInTheDocument()
   })
 
-  test('Detail text visible', () => {
-    renderWithTheme(<Detail />)
-    expect(screen.getByText('Detail text')).toBeInTheDocument()
-  })
-
```
lukelooker, line 49: Let's revert this change and the one below (effectively just moves around the test and I'd prefer to limit the amount of change since we've had "jitter" on these components recently)

packages/components/src/__mocks__/DialogExampleLayout.tsx
```diff
@@ -36,13 +36,11 @@ export const DialogExampleLayout: FC<{ header: string }> = ({
   return (
     <DialogLayout
       header={header}
-      footer={
```
laraharrow, line 39: @lukelooker this is what I was mentioning about the stories

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
```diff
@@ -25,65 +25,64 @@
...
-export interface DialogLayoutExampleProps {
-  header?: string
-  footer?: string
-  length?: 'short' | 'long'
-}
+const extraButton = (
+  <ButtonTransparent color="neutral" size="medium">
+    Cancel
+  </ButtonTransparent>
+)
```
lukelooker, line 37: Can this set of stories be restore to how they were used before so that we can verify that no of the associated image snapshots were altered unexpectedly?

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
```diff
@@ -25,65 +25,64 @@
...
-  ...Basic.args,
-  footer: 'Footer text',
-}
+export const HeaderCloseButton = () => (
+  <Box bg="ui1">
+    <DialogLayout header="Header" headerCloseButton={true} footer="Footer">
+      <ConstitutionShort />
+    </DialogLayout>
+  </Box>
+)
```
lukelooker, line 61: This is a net-new test and can be kept.

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
```diff
@@ -25,65 +25,64 @@
...
+      footer={
+        <DialogFooter secondary={extraButton}>
+          <Button>Primary</Button>
+        </DialogFooter>
+      }
+    >
+      <ConstitutionShort />
+    </DialogLayout>
+  </Box>
+)
```
lukelooker, line 76: This is a net-new test and can be kept.

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
```diff
@@ -25,65 +25,64 @@
...
 
-const NoPaddingTemplate: Story<DialogLayoutExampleProps> = () => (
+export const NoPadding = () => (
   <Box bg="ui1">
     <DialogContent hasFooter={false} hasHeader={false}>
       <ConstitutionShort />
     </DialogContent>
   </Box>
 )
```
lukelooker, line 84: I'm okay with this change.

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
lukelooker: Specifically `DialogLayoutExampleProps`, `Template`, `Basic`, `Header` & `Footer`

packages/components/src/Modal/ModalFooter/ModalFooter.tsx
```diff
@@ -47,12 +48,16 @@ const ModalFooterLayout: FC<ModalFooterProps> = ({
...
   ...props
 }) => (
-  <Space as="footer" reverse between {...props}>
-    <Space reverse>{children}</Space>
-    {secondary && <Space>{secondary}</Space>}
-  </Space>
+  <Flex as="footer" justifyContent="flex-end" {...props}>
+    {secondary && <FooterLayout>{secondary}</FooterLayout>}
+    {children}
+  </Flex>
```
lukelooker, line 54: This will break the existing `DialogFooter` layout for "secondary" I think – "secondary" is supposed to go on the left-hand side of the `DialogFooter`

```
[secondary]                                                   [children]
```

e.g.:

```
[Add Another]                                                       [ Cancel,  Save Changes ]
```

packages/components/src/Modal/ModalHeader/ModalHeader.tsx
```diff
@@ -85,4 +85,5 @@ export const ModalHeader = styled(ModalHeaderLayout)<ModalHeaderProps>`
   align-items: center;
   display: flex;
   flex-shrink: 0;
+  width: 100%;
```
lukelooker, line 88: I'm not remember why we needed to add this (I think it's fine but always great to see a note on the PR noting why a change to a component was needed

packages/components/src/Popover/Layout/PopoverFooter/PopoverFooter.story.tsx
```diff
@@ -25,33 +25,26 @@
...
   title: 'PopoverFooter',
 }
 
-export const Basic = () => <PopoverFooter>Footer Text</PopoverFooter>
-Basic.parameters = {
-  storyshots: {
-    disable: true,
-  },
-}
+export const Basic = () => <PopoverFooter />
```
lukelooker, line 36: Pretty sure this still needs to have a child `Footer Text` ?

packages/components/src/__mocks__/DialogExampleLayout.tsx
```diff
@@ -36,13 +36,11 @@ export const DialogExampleLayout: FC<{ header: string }> = ({
...
-          <Button onClick={closeModal}>Done Reading</Button>
-          <ButtonTransparent color="neutral" onClick={closeModal}>
-            Finish Later
-          </ButtonTransparent>
-        </>
+      footer={<Button onClick={closeModal}>Done Reading</Button>}
+      footerSecondary={
+        <ButtonTransparent color="neutral" onClick={closeModal}>
+          Finish Later
+        </ButtonTransparent>
```
lukelooker, line 43: This change needs to be reversed as it alters the result of a lot of image snapshots. The previous version was the expected design (no use of `footerSecondary` in these)

lukelooker (changes requested):
Haven't dug into the `Popover` specific changes as of yet - I'd like to start with just having `Dialog*` snapshots stay identical (no changes). Additional snapshots is fine but I'd expect to see no changes to the existing "known good" snapshots.

packages/components/src/Dialog/Layout/DialogHeader/DialogHeader.story.tsx
laraharrow: I updated this change but as of right now its set as `storyshots: { disable: true },` so it shouldn't affect snapshots.

packages/components/src/Modal/ModalFooter/ModalFooter.tsx
laraharrow: That's one of the problems with my understanding of the `figma` the was that the component is displayed there there is no example with `secondary` so I assumed the `cancel` word there was the `secondary` prop. 
https://www.figma.com/file/zd6MjUXIIstvG7gTJWaX2Y/Looker-Components?node-id=3%3A569
So I actually change the code to accommodate that🤦🏽‍♀️.

packages/components/src/Modal/ModalHeader/ModalHeader.tsx
laraharrow: My bed, when we chat last there was a bug I was trying to fix. You had suggested adding that would maybe solve the problem. I added thinking I would go back and check and after solving the bug I forgot it that was there.

packages/components/src/Popover/Layout/PopoverFooter/PopoverFooter.story.tsx
laraharrow: Because when nothing is passed we would still have the `Done` available there is a chance that nothing needs to be passed. Am I missing something here?

lukelooker: Ah! My bad - that makes sense.

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
```diff
@@ -74,16 +68,51 @@ Full.args = {
...
-const NoPaddingTemplate: Story<DialogLayoutExampleProps> = () => (
+const extraButton = (
+  <ButtonTransparent color="neutral" size="medium">
+    Cancel
+  </ButtonTransparent>
+)
+
+export const HeaderDetail = () => (
+  <Box bg="ui1">
+    <DialogLayout header="Header" headerDetail={extraButton} footer="Footer">
```
lukelooker, line 79:
```suggestion
    <DialogLayout header="Header" headerDetail="Detail text" footer="Footer">
```

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
```diff
@@ -74,16 +68,51 @@ Full.args = {
   ...Header.args,
 }
 
-const NoPaddingTemplate: Story<DialogLayoutExampleProps> = () => (
+const extraButton = (
+  <ButtonTransparent color="neutral" size="medium">
+    Cancel
+  </ButtonTransparent>
+)
```
lukelooker, line 75:
```suggestion
```

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
```diff
@@ -74,16 +68,51 @@ Full.args = {
...
+
+export const HeaderDetail = () => (
+  <Box bg="ui1">
```
lukelooker, line 78: I worry that folks previewing these components might come to the false assumption that they have a `ui1` background if they don't look at the source.

So i'd prefer not to wrap see these in a box with background color since it's not something we expect to see in real implementations.

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
```diff
@@ -74,16 +68,51 @@ Full.args = {
...
+
+export const FooterSecondary = () => (
+  <Box bg="ui1">
+    <DialogLayout
+      header="Header"
+      footer={
+        <DialogFooter secondary={extraButton}>
+          <Button>Primary</Button>
+        </DialogFooter>
+      }
```
lukelooker, line 101: `DialogLayout` already wraps content output in `footer` within another `DialogFooter` so it's not needed here. Should instead be:

```suggestion
      footer={<Button>Primary</Button>
      footerSecondary="Secondary text"
```

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
```diff
@@ -74,16 +68,51 @@ Full.args = {
...
+  <Box bg="ui1">
+    <DialogLayout header="Header" headerDetail={extraButton} footer="Footer">
+      <ConstitutionShort />
+    </DialogLayout>
+  </Box>
+)
+
+export const HeaderCloseButton = () => (
+  <Box bg="ui1">
+    <DialogLayout header="Header" headerCloseButton={true} footer="Footer">
```
lukelooker, line 87:
```suggestion
    <DialogLayout header="Header" headerCloseButton footer="Footer">
```

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
lukelooker: NOTE: Please remove from all of these stories for consistency unless you see a specific benefit to keeping the background color (let me know if so - happy to discuss)

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
```diff
@@ -26,45 +26,39 @@
...
     </DialogLayout>
   </Box>
 )
 
-export const Basic = Template.bind({})
-Basic.args = {
-  length: 'short',
-}
+export const Content = Template.bind({})
+Content.args = {}
```
lukelooker, line 51: Any specific reason to rename this story? Most of our `.story` files start with a story named "Basic" so it'd be nice to keep with that convention unless there's a specific reason not to do so.

Also the `length` option made it easier to compose the stories below - wondering why you choose to remove that functionality and hard-code it instead?

```suggestion
export const Basic = Template.bind({})
Basic.args = {
  length: 'short'
}
```

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
```diff
@@ -74,16 +68,51 @@ Full.args = {
...
 }
 
-const NoPaddingTemplate: Story<DialogLayoutExampleProps> = () => (
+const extraButton = (
+  <ButtonTransparent color="neutral" size="medium">
+    Cancel
+  </ButtonTransparent>
+)
+
+export const HeaderDetail = () => (
```
lukelooker, line 77: This story could leverage the standard `Template` above by being rewritten as:

```suggestion
export const HeaderDetail = Template.bind({})
HeaderDetail.args = {
  ...Content.args,
  header: 'Header',
  headerDetail: 'Header detail text',
}
```

If you need to leverage the complete story in a Jest test you can do so by following this article: https://storybook.js.org/docs/react/workflows/unit-testing

TLDR: 

```
it('renders the button in the primary state', () => {
  render(<HeaderDetail {...HeaderDetail.args} />);
  expect(screen.getByText('Header detail text');
});
```

Note the `{...HeaderDetail.args} ` that spreads the args onto the component is a key piece here.

packages/components/src/__mocks__/DialogExampleLayout.tsx
lukelooker: This story should be restored to the way it was before - the intention (in this example) is to have the two buttons both on the left-side and not to have anything in the  `secondary`  (left) position.

lukelooker (changes requested):
Hey Lara - 

With the work to prepare 2.0 to launch we found quite a few spots where we had regressions in Dialog layouts (as well as the existing `PopoverContent` behavior so I'd like to be a bit more cautious moving forward so we don't accidentally cause any new issues. 

With that in mind I'd like to break this PR into two seperate PRs:

1. refactor: Extract new `ModalLayout` from `DialogLayout`
2. feat: Add new `PopoverLayout` component

Ideally the initial PR would be able to leave all existing stories and tests for `DialogLayout` intact (adding new ones is totally fine of course). 

I've made a bunch of additional comments as well as unresolving some unaddressed comments from my previous review that should help get the first PR ready to go (the "refactor" portion). 

If you need help with doing the git-magic to pull this apart into a couple of separate PRs I'd be happy to jump on a call with you and walk you through doing so.

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
laraharrow: the length option is not used anywhere as `long` so I didn't see the point of having it.

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
laraharrow: yeah I totally see that I was going off a different example that was already there that had that in place. I'm happy to remove it.

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
laraharrow: oops yep, that totally make sense, my bad!

packages/components/src/Dialog/Layout/DialogHeader/DialogHeader.test.tsx
laraharrow: I think this is done!

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
laraharrow: Why should we do this change?

PR #2481, opened 2021-06-06
feat(Tree): Add virtualization support

mdodgelooker:
1. Add `WindowedTreeCollection` – so much needs to be different in order to support virtualization (`trees` prop vs children, no internal state allowed, including the toggle state, and density needs to be defined at the top level) that I think a separate component will be less confusing.
2. Add internal-only `useWindowedTree` and `WindowedTreeNode` to make windowing work
3. ~~Refactor `useWindow` to support `useWindowedTree` and remove support for variable windowing, since it's not being used and it would've complicated the refactor.~~ Moving this to a separate PR: https://github.com/looker-open-source/components/pull/2485

**NOTE:** The Windowing story has some storybook-docs-related bottleneck (I tried disabling docs but it didn't seem to help) so it's pretty slow. Check it out with `yarn playground` for a more realistic test.

- [x] 📚 Documentation updated
- [x] 🏁 Performance impacts addressed
- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11

### DIFF
--- a/config/i18nIgnoredProps.js
+++ b/config/i18nIgnoredProps.js
@@ -34,7 +34,6 @@ const ours = [
   'iconVerticalAlign',
   'iconViewBox',
   'textDecoration',
-  'windowing',
   'layout',
   'indexBy',
   'indicatorPosition',
--- a/packages/components/src/List/List.tsx
+++ b/packages/components/src/List/List.tsx
@@ -44,20 +44,14 @@ import { HeightProps, fontFamily } from 'styled-system'
 import styled from 'styled-components'
 import { useArrowKeyNav, useWindow } from '../utils'
 import { ListItemContext } from './ListItemContext'
-import { DensityRamp, ListColorProps } from './types'
+import { DensityProps, ListColorProps } from './types'
 import { listItemDimensions } from './utils'
 
 export type ListProps = ListColorProps &
   HeightProps &
   WidthProps &
-  Omit<CompatibleHTMLProps<HTMLUListElement>, 'label'> & {
-    /**
-     * Determines how dense a list should be by affecting child ListItem
-     * size and spacing.
-     * @default 0
-     */
-    density?: DensityRamp
-
+  Omit<CompatibleHTMLProps<HTMLUListElement>, 'label'> &
+  DensityProps & {
     /**
      * Disables the nested List's keyboard nav capabilities
      * @private
--- a/packages/components/src/Tree/TreeCollection.tsx
+++ b/packages/components/src/Tree/TreeCollection.tsx
@@ -24,49 +24,43 @@
 
  */
 
-import React, { FC, KeyboardEvent, ReactNode, useRef } from 'react'
+import { CompatibleHTMLProps } from '@looker/design-tokens'
+import React, { forwardRef, KeyboardEvent, Ref, useRef } from 'react'
 import styled from 'styled-components'
-import { useArrowKeyNav } from '../utils'
+import { useArrowKeyNav, useForkedRef } from '../utils'
 import { getNextTreeFocus, getTreeItems } from './utils'
 
-export type TreeCollectionProps = {
-  children?: ReactNode
-  className?: string
-}
+export type TreeCollectionProps = CompatibleHTMLProps<HTMLUListElement>
 
-const TreeCollectionLayout: FC<TreeCollectionProps> = ({
-  children,
-  className,
-}) => {
-  const ref = useRef<HTMLUListElement>(null)
+const TreeCollectionInternal = forwardRef(
+  (props: TreeCollectionProps, forwardedRef: Ref<HTMLUListElement>) => {
+    const internalRef = useRef<HTMLUListElement>(null)
 
-  // Additional key shortcuts
-  const handleKeyDown = (event: KeyboardEvent<HTMLElement>) => {
-    const treeItems = getTreeItems(ref.current as HTMLElement)
-    if (event.key === 'Home') {
-      const firstItem = treeItems[0]
-      firstItem && firstItem.focus()
-    } else if (event.key === 'End') {
-      const lastItem = treeItems[treeItems.length - 1]
-      lastItem && lastItem.focus()
+    // Additional key shortcuts
+    const handleKeyDown = (event: KeyboardEvent<HTMLElement>) => {
+      const treeItems = getTreeItems(internalRef.current as HTMLElement)
+      if (event.key === 'Home') {
+        const firstItem = treeItems[0]
+        firstItem && firstItem.focus()
+      } else if (event.key === 'End') {
+        const lastItem = treeItems[treeItems.length - 1]
+        lastItem && lastItem.focus()
+      }
     }
+
+    const navProps = useArrowKeyNav<HTMLUListElement>({
+      axis: 'both',
+      getNextFocus: getNextTreeFocus,
+      onKeyDown: handleKeyDown,
+      ref: useForkedRef(internalRef, forwardedRef),
+    })
+
+    return <ul role="tree" {...props} {...navProps} />
   }
+)
 
-  const navProps = useArrowKeyNav<HTMLUListElement>({
-    axis: 'both',
-    getNextFocus: getNextTreeFocus,
-    onKeyDown: handleKeyDown,
-    ref,
-  })
-
-  return (
-    <ul className={className} role="tree" {...navProps}>
-      {children}
-    </ul>
-  )
-}
-
-export const TreeCollection = styled(TreeCollectionLayout)`
+export const TreeCollection = styled(TreeCollectionInternal)`
+  list-style-type: none;
   margin: 0;
   padding: 0;
 `
--- a/packages/components/src/Tree/TreeContext.tsx
+++ b/packages/components/src/Tree/TreeContext.tsx
@@ -25,12 +25,11 @@
  */
 
 import { createContext } from 'react'
-import { DensityRamp, ListColor } from '../List/types'
+import { DensityProps, ListColor } from '../List/types'
 
-export interface TreeContextProps {
+export type TreeContextProps = DensityProps & {
   border?: boolean
   color?: ListColor
-  density: DensityRamp
   depth?: number
   labelBackgroundOnly?: boolean
 }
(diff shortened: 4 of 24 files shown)


packages/components/src/List/List.tsx
```diff
@@ -44,20 +44,14 @@ import { HeightProps, fontFamily } from 'styled-system'
...
 export type ListProps = ListColorProps &
   HeightProps &
   WidthProps &
-  Omit<CompatibleHTMLProps<HTMLUListElement>, 'label'> & {
-    /**
-     * Determines how dense a list should be by affecting child ListItem
-     * size and spacing.
-     * @default 0
-     */
-    density?: DensityRamp
```
mdodgelooker, line 59: Extracting this into a utility type for use in various places.

packages/components/src/Tree/TreeCollection.tsx
```diff
@@ -24,49 +24,45 @@
...
-  className?: string
-}
+export type TreeCollectionProps = CompatibleHTMLProps<HTMLUListElement>
 
-const TreeCollectionLayout: FC<TreeCollectionProps> = ({
-  children,
-  className,
-}) => {
-  const ref = useRef<HTMLUListElement>(null)
+const TreeCollectionInternal = forwardRef(
```
mdodgelooker, line 35: Added `forwardRef` – this diff looks better with "Hide whitespace changes" on.

packages/components/src/Tree/TreeCollection.tsx
```diff
@@ -24,49 +24,45 @@
...
-    </ul>
-  )
-}
+    return <ul role="tree" {...props} {...navProps} />
+  }
+)
 
-export const TreeCollection = styled(TreeCollectionLayout)`
+export const TreeCollection = styled(TreeCollectionInternal)`
+  list-style-type: none;
```
mdodgelooker, line 65: The virtualization before/after spacer `li`s had bullets.

packages/components/src/Tree/TreeCollectionVirtual.tsx
```diff
@@ -0,0 +1,79 @@
...
+      </TreeCollection>
+    </TreeCollectionContext.Provider>
+  )
+}
+
+export const TreeCollectionVirtual = styled(
+  TreeCollectionVirtualInternal
+)<TreeCollectionProps>`
+  ${height}
+  ${width}
```
mdodgelooker, line 77: Setting height and width may be necessary depending on what it's rendered inside of.

packages/components/src/utils/useWindow.tsx
```diff
@@ -24,190 +24,56 @@
...
-export const useWindow = <E extends HTMLElement = HTMLElement>({
-  children,
+export type UseWindowProps<E extends HTMLElement> = Omit<
+  UseWindowBaseProps<E>,
+  'length'
+> & {
+  children?: JSX.Element | JSX.Element[]
+}
+
+export const useWindowBase = <E extends HTMLElement = HTMLElement>({
```
mdodgelooker, line 56: Added this base version of the hook that looks at `length` instead of `children` and returns `start`/`end` instead of `content` – with trees, it's much more complicated than a list to calculate which content to render, so that's handled in `useVirtualTree`.

packages/design-tokens/src/system/index.ts
```diff
@@ -50,6 +50,7 @@ export type {
   BoxShadowProps,
   DisplayProps,
   FlexboxProps,
+  HeightProps,
```
mdodgelooker, line 53: Any issue exporting this from here? `WidthProps` already is.

mdodgelooker:
Keeping this in draft mode for the moment, since I still have a few todo's, but I wanted to get some initial feedback before wrapping up.

packages/components/src/Tree/WindowedTree.tsx
```diff
@@ -0,0 +1,91 @@
+/*
```
mdodgelooker, line 1: This component is a helper that:
1. Adds some context to support windowing state management
2. Inserts `items` as `children`, as necessitated by the use of the `trees` prop as opposed to composition

packages/components/src/utils/useWindow.test.tsx
```diff
@@ -26,7 +26,7 @@
 
 import { fireEvent, render, screen } from '@testing-library/react'
 import React from 'react'
-import { ChildHeightFunction, useWindow, useToggle } from './'
```
mdodgelooker, line 29: I refactored `useWindow` so it could also support `Tree` windowing, removing unused "variable" windowing to make my life easier. 😄

config/i18nIgnoredProps.js
```diff
@@ -34,7 +34,6 @@ const ours = [
   'iconVerticalAlign',
   'iconViewBox',
   'textDecoration',
-  'windowing',
```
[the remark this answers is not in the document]
mdodgelooker: Oops, this one should've gone into the `useWindow` refactor. Sorry!

packages/components/src/Tree/WindowedTreeNode.tsx
```diff
@@ -0,0 +1,88 @@
+/*
```
mdodgelooker, line 1: This component is a helper that:
1. adds context for managing state related to virtualization
2. inserts `items` as `children` (necessitated by using data structure instead of composition).

packages/components/src/Tree/utils/useWindowedTree.tsx
```diff
@@ -0,0 +1,107 @@
...
+  // Keep track of the sub-trees opened / closed state by ID and
+  // the total number of SHOWN items(i.e.no ancestor is closed)
+  const [{ map, shownIDs, treesWithIDs }, dispatch] = useReducer(
+    windowedTreeReducer,
+    {
+      map: {},
+      shownIDs: [],
+      treesWithIDs: [],
+    }
+  )
```
mdodgelooker, line 54: I opted for `useReducer` here so I could store a map of the trees that makes re-calculating the # of shown items more efficient than having to loop through the entire `trees` data structure again.

mdodgelooker:
I set up this codesandbox for comparison: https://codesandbox.io/s/eager-curie-jn05f

It's super slow and only has 15 top-level trees, unlike the story, which has 100.

packages/components/src/Tree/stories/Windowing.story.tsx
```diff
@@ -0,0 +1,151 @@
...
+  return preamble.substr(startIndex, length)
+}
+
+const getItems = (
+  prefix: string,
+  labelLength: number,
+  canNest = false
+): WindowedTreeNodeProps[] => {
+  // const num = getRandomInteger(8)
+  // const itemsLength = Math.pow(num, 2)
```
ElliotTheHuman, line 62: These can probably be deleted, I think.

packages/components/src/Tree/utils/getWindowedTreeNodeFilterer.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+  return 'before'
+}
+
+export const getWindowedTreeNodeFilterer =
+  (
+    filteredList: WindowedTreeNodeIDProps[],
+    firstIDinWindow: number,
+    lastIDinWindow: number
+  ) =>
```
ElliotTheHuman, line 54: Just to make sure I understand what's going on, this function is used to figure out if a given node should be rendered in the window (or not). It calls `getTreeWindowIntersection` to check if a given node's ID is < the window's last ID and > the window's first ID (OR if the node is the open tree of the window's first element).

Hoping that's right 🤞

ElliotTheHuman (commented):
This is awesome!! 🎉 🎉 🎉  Had a few notes, and left a comment regarding `getWindowedTreeNodeFilterer.ts` just to make sure I'm understanding it correctly.

Would like to get @lukelooker to take a look before approving, but based on the `Windowing` story this works great.

packages/components/src/Tree/TreeContext.tsx
```diff
@@ -25,17 +25,16 @@
...
-import { DensityRamp, ListColor } from '../List/types'
+import { DensityProps, ListColor } from '../List/types'
 
-export interface TreeContextProps {
+export type TreeContextProps = DensityProps & {
   border?: boolean
   color?: ListColor
-  density: DensityRamp
   depth?: number
+  id?: string
```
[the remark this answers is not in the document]
mdodgelooker: An artifact that got refactored out. 😉

packages/components/src/Tree/WindowedTreeCollection.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+      }
+    })
+  })
+
+  afterEach(() => {
+    jest.runOnlyPendingTimers()
+    jest.useRealTimers()
+    jest.resetAllMocks()
+    /* eslint-disable-next-line @typescript-eslint/unbound-method */
+    Element.prototype.getBoundingClientRect = globalGetBoundingClientRect
```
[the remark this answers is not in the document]
mdodgelooker: That is an excellent question – it's for testing anything that happens after a delay. Not sure why I thought I needed it here though... 🤔 Taking it out!

packages/components/src/Tree/TreeCollectionVirtual.tsx
mdodgelooker: I needed height & width for the story to work and in my mind that's a good indication that they should be included. Though of course the typical usage will probably get dimensions based on the layout.

packages/components/src/Tree/utils/getWindowedTreeNodeFilterer.ts
mdodgelooker: Correct!

lukelooker:
Tried this out again with the `HoverDisclosure` revert and it feels much better than the composed version - still seeing occasional `[handler]` violations but a lot fewer than we were seeing earlier today. 

Given that this is an "opt-in" change I'd say we plan to ship it in `2.1` and see what benefits it offers to the target audience when integrated.

lukelooker (approved):
Tried this out again with the `HoverDisclosure` revert and it feels much better than the composed version - still seeing occasional `[handler]` violations but a lot fewer than we were seeing earlier today. 

Given that this is an "opt-in" change I'd say we plan to ship it in `2.1` and see what benefits it offers to the target audience when integrated.

Please wait to merge until 2.0 is cut.

PR #2483, opened 2021-06-07
chore: Format jsx in jsDoc with backticks

mdodgelooker:
Fixes Storybook docs tab throwing a React error when it sees jsx inside a jsDoc comment:
image


lukelooker (approved):
Nice catch. Thank you!

PR #2485, opened 2021-06-10
fix(useWindow): Simplify windowing for long lists

mdodgelooker:
Refactors `useWindow` to support [`Tree` virtualization](https://github.com/looker-open-source/components/pull/2481).

### DIFF
--- a/packages/components/src/Form/Inputs/Combobox/ComboboxContext.ts
+++ b/packages/components/src/Form/Inputs/Combobox/ComboboxContext.ts
@@ -65,7 +65,7 @@ export interface ComboboxContextProps<
   persistSelectionPropRef?: MutableRefObject<boolean>
   closeOnSelectPropRef?: MutableRefObject<boolean>
   inputReadOnlyPropRef?: MutableRefObject<boolean>
-  windowedOptionsPropRef?: MutableRefObject<boolean>
+  windowingPropRef?: MutableRefObject<boolean>
   freeInputPropRef?: MutableRefObject<boolean>
   isScrollingRef?: MutableRefObject<boolean>
   indicatorPropRef?: MutableRefObject<ComboboxOptionIndicatorProps['indicator']>
--- a/packages/components/src/List/List.tsx
+++ b/packages/components/src/List/List.tsx
@@ -79,10 +79,9 @@ export type ListProps = ListColorProps &
 
     /**
      * Use windowing for long lists (strongly recommended to also define a width on List or its container)
-     * 'none' - default with children are <= 100.
-     * 'fixed' - better performance, default when first child is a ListItem (height will default to 100%)
+     * Defaults to false with children <= 100 and true for > 100
      */
-    windowing?: 'fixed' | 'none'
+    windowing?: boolean
   }
 
 const getListItemHeight = (child: ReactChild, height: number) => {
@@ -118,14 +117,10 @@ export const ListInternal = forwardRef(
     )
 
     if (windowing === undefined) {
-      if (childArray.length > 100) {
-        windowing = 'fixed'
-      } else {
-        windowing = 'none'
-      }
+      windowing = childArray.length > 100
     }
 
-    if (height === undefined && windowing !== 'none') {
+    if (height === undefined && windowing) {
       // Need a height for windowing to work
       height = '100%'
     }
@@ -135,7 +130,7 @@ export const ListInternal = forwardRef(
         ? getListItemHeight(childArray[0] as ReactChild, itemDimensions.height)
         : 0,
       children: children as JSX.Element | JSX.Element[],
-      enabled: windowing !== 'none',
+      enabled: windowing,
       ref: forwardedRef,
       spacerTag: 'li',
     })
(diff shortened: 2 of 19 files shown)


packages/components/src/Form/Inputs/Combobox/ComboboxContext.ts
```diff
@@ -65,7 +65,7 @@ export interface ComboboxContextProps<
   persistSelectionPropRef?: MutableRefObject<boolean>
   closeOnSelectPropRef?: MutableRefObject<boolean>
   inputReadOnlyPropRef?: MutableRefObject<boolean>
-  windowedOptionsPropRef?: MutableRefObject<boolean>
+  windowingPropRef?: MutableRefObject<boolean>
```
mdodgelooker, line 68: Renaming `windowedOptions` to `windowing` for consistency with `List` now that they're both `boolean`.

packages/components/src/utils/useWindow.tsx
```diff
@@ -24,190 +24,55 @@
...
-  useMemo,
-  useReducer,
-} from 'react'
+import React, { Children, Ref, useMemo } from 'react'
 import { getWindowedListBoundaries } from './getWindowedListBoundaries'
 import { useCallbackRef } from './useCallbackRef'
 import { useMeasuredElement } from './useMeasuredElement'
 import { useScrollPosition } from './useScrollPosition'
 
-interface WindowHeightState {
```
mdodgelooker, line 41: Removing support for `'variable'` windowing since we're not using it, it's not the most performant, and it was an obstacle to the refactor I needed to do for `Tree` virtualization.

packages/components/src/utils/useWindow.tsx
```diff
@@ -24,190 +24,55 @@
...
-    start,
-  }
-}
-
-export type ChildHeightFunction = (child: ReactChild, index: number) => number
-
 export type WindowSpacerTag = 'div' | 'li' | 'tr'
 
-export interface UseWindowProps<E extends HTMLElement> {
+export type UseWindowBaseProps<E extends HTMLElement> = {
```
mdodgelooker, line 35: `Tree` virtualization is totally different from `List` in the way that the items get truncated to the visible window.

To enable that, `useWindowBase` takes `length` instead of `children` and returns the `after` & `before` spacers instead of `content` (`before`+truncated children+`after`).

packages/components/src/List/List.tsx
```diff
@@ -79,10 +79,9 @@ export type ListProps = ListColorProps &
 
     /**
      * Use windowing for long lists (strongly recommended to also define a width on List or its container)
-     * 'none' - default with children are <= 100.
-     * 'fixed' - better performance, default when first child is a ListItem (height will default to 100%)
+     * Defaults to false with children <= 100 and true for > 100
      */
-    windowing?: 'fixed' | 'none'
```
mdodgelooker, line 85: Making this a `boolean` – it used to support `variable` but that was removed awhile ago and now that `useWindow` no longer supports variable, it's time to simplify this prop.

packages/components/src/Menu/MenuList.test.tsx
```diff
@@ -39,7 +39,6 @@ const globalGetBoundingClientRect = Element.prototype.getBoundingClientRect
 describe('MenuList', () => {
   describe('windowing', () => {
     beforeEach(() => {
-      jest.useFakeTimers()
```
mdodgelooker, line 42: Turns out this was unnecessary.

packages/components/src/utils/useWindow.test.tsx
```diff
@@ -26,7 +26,7 @@
 
 import { fireEvent, render, screen } from '@testing-library/react'
 import React from 'react'
-import { ChildHeightFunction, useWindow, useToggle } from './'
+import { useWindow, useToggle } from './'
```
mdodgelooker, line 29: This diff benefits greatly from "Hide whitespace changes".

packages/components/src/utils/useWindow.tsx
mdodgelooker: Also, let me know if you prefer different naming, like
1. `useWindowLength` & `useWindow` or 
2. `useWindow` & `useWindowChildren`

packages/components/src/utils/useWindow.tsx
```diff
@@ -24,190 +24,55 @@
...
   enabled?: boolean
-  children?: JSX.Element | JSX.Element[]
-  /** Derive the height of each child using props, type, etc. */
-  childHeight: number | ChildHeightFunction
-  /** Tagname to use for the spacers above and below the window */
+  length: number
```
lukelooker, line 37: Can you add a JSDoc to clarify this the the count of items to be windowed.

Perhaps `itemCount` instead of length to clarify?

packages/components/src/utils/useWindow.tsx
lukelooker: I prefer #2 because it clarifies the inheritance better (`useWindow` is less "clever" and just requires an explicitly specified "length" prop whereas `useWindowChildren` makes the former a bit more clever and counts the `children` for you)

lukelooker (changes requested):
Minor naming changes but fundamentally this works for me.

packages/components/src/utils/useWindow.tsx
mdodgelooker: Sounds good!

lukelooker (approved):
Looks great!

PR #2486, opened 2021-06-10
fix(ReplaceText): Performance improvement

mdodgelooker:
Adds `useMemo` to `ReplaceText` to slightly improve performance. I'm using it in the example for `Tree` virtualization in https://github.com/looker-open-source/components/pull/2481 and wanted to get an accurate picture of the performance improvement from virtualization.


lukelooker (approved):
LGTM

PR #2489, opened 2021-06-11
fix(InputColor): Call onChange if input value is cleared

mdodgelooker:
This PR updates `InputColor` to call `onChange` if the input's new value is an empty string. Before it was only calling it if the value was a valid color, which an empty string is not.

### DIFF
--- a/packages/components/src/Form/Inputs/InputColor/InputColor.tsx
+++ b/packages/components/src/Form/Inputs/InputColor/InputColor.tsx
@@ -114,9 +114,8 @@ export const InputColorComponent = forwardRef(
       }
     }, [isFocused, value, inputTextValue])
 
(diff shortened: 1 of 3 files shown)


packages/components/src/Form/Inputs/InputColor/InputColor.tsx
```diff
@@ -77,7 +77,7 @@ function getColorFromText(text?: string) {
   return text && isValidColor(text) ? stringToSimpleHsv(text) : undefined
 }
 
-export const InputColorComponent = forwardRef(
+export const InputColorInternal = forwardRef(
```
lukelooker, line 80: 🎉

lukelooker (approved):
Looks great! Thank you :)

PR #2491, opened 2021-06-14
chore: Rationalize package versions in non-published packages

lukelooker:


packages/components-theme-editor/package.json
```diff
@@ -16,18 +16,16 @@
     "directory": "packages/theme-editor"
   },
   "dependencies": {
-    "@looker/components": "^1.4.0",
-    "@looker/design-tokens": "^1.3.1",
+    "@looker/components": "*",
+    "@looker/design-tokens": "*",
```
lukelooker, line 20: With the `--no-private` flag used for publishing these unpublished packages no longer get automatically moved forward with each release (good!) however - this means we need to loosen these to just pull "latest" from the monorepo.

packages/components-theme-editor/package.json
```diff
@@ -16,18 +16,16 @@
...
     "@styled-icons/material": "^10.34.0",
     "@styled-icons/material-outlined": "^10.28.0",
-    "lodash": "^4.17.20"
+    "lodash": "*"
```
lukelooker, line 23: No need to have a static version here since this isn't being published currently.

packages/components-theme-editor/package.json
```diff
@@ -16,18 +16,16 @@
...
   },
   "devDependencies": {
-    "@types/react": "^16.9.56",
-    "@types/styled-components": "^5.1.9",
     "react": "^16.14.0",
     "react-is": "^16.13.1",
-    "styled-components": "^5.3.0"
+    "styled-components": ">=5.3"
```
lukelooker, line 28: Might as well loosen here too (needs to be specified since this package will eventually get published)

storybook/package.json
```diff
@@ -18,23 +17,24 @@
...
   },
   "dependencies": {
-    "@looker/components": "^1.4.0",
-    "@looker/components-providers": "^1.3.1",
-    "@looker/components-theme-editor": "^1.4.0",
-    "@looker/design-tokens": "^1.3.1",
-    "@storybook/react": "^6.2.9",
-    "lodash": "^4.17.20",
+    "@looker/components": "*",
+    "@looker/components-date": "^2.0.0-alpha",
```
lukelooker, line 21: Annoyingly `yarn` gets upset about `*` because this package has not yet _ever_ been released as a publicly-available package.

Once `components-date` has been shipped (non-prelease) once this can be switched to `*`

mdodgelooker:
I'm getting a diff on `yarn.lock` after running `yarn`.

lukelooker:
Ooops, updated `yarn.lock`

mdodgelooker (approved):

PR #2495, opened 2021-06-15
fix: Include @types/styled-system in devDependencies to simplify installation

lukelooker:
Adds @types/styled-system to `devDependencies` to remove the need for downstream consumers to also install that dependency within their own projects.

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -47,6 +47,7 @@
     "@types/react": "^16.9.56",
     "@types/react-dom": "^16.9.9",
     "@types/styled-components": "^5.1.9",
+    "@types/styled-system": "^5.1.11",
     "@types/uuid": "^8.3.0",
     "csstype": "^3.0.8",
     "jest-styled-components": "^7.0.4",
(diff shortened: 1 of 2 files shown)


jhardy (commented):
Looks good to me, but one questions, do we need to update https://looker-open-source.github.io/components/latest/develop/ to reflect this changes? I see we recommended adding `@types/styled-components`  in the using typescript section. Do we either need to include `@types/styled-systems` there?

lukelooker:
> Looks good to me, but one questions, do we need to update https://looker-open-source.github.io/components/latest/develop/ to reflect this changes? I see we recommended using yarn to add `@types/styled-components` in the using typescript section but that is also included here under dev dependencies. Probably not worth blocking this improvement, but something that I saw.

Great point!

The _hope_ is that this will forego the need for folks to install that dependency but I went ahead and added this to the list because if this fix attempt doesn't work at least the README will help folks :)

jhardy (approved):
LGTM

PR #2497, opened 2021-06-15
fix(DialogHeader): Restore previous behavior for DialogHeader detail property

lukelooker:
During the recent refactor of the `DialogHeader` the behavior was changed so that the content `detail` property is not displayed when specified.

This PR restores the previous behavior and adds test coverage to validate that it is functional again.


jhardy (approved):
LGTM

PR #2503, opened 2021-06-15
fix(PopoverContent): Restore previous defaults and remove duplicate compnent

lukelooker:
There were two copies of `PopoverContent` in the repo -

- ⛔️ `/packages/components/src/Popover/PopoverContent.tsx` (used for image snapshot test, not exported)
- ✅ `/packages/components/src/Popover/Layout/PopoverContent` (exported, not used for tests)

The new version in `Layout/Popover` content has overflow support that we want to keep intact but also changes the padding from the version shipped previously. 

This PR reverts newer padding values specified for the upcoming `PopoverLayout` component but I believe this is the most pragmatic approach to address the short-term change issues we've seen in consumers of this component.


packages/components/src/Popover/Layout/PopoverContent/PopoverContent.tsx
```diff
@@ -36,7 +36,7 @@ const PopoverContentLayout: FC<PopoverContentProps> = ({
   ...props
 }) => {
   return (
-    <ModalContent pb="medium" pt="xsmall" px="large" {...props}>
+    <ModalContent p="small" {...props}>
```
lukelooker, line 39: `p="small"` was true of the "old" version of `PopoverContent`. To keep things stable this needs to stay as-is.

dbchristopher (approved):

PR #2506, opened 2021-06-15
fix(Panel): Render properly in a Drawer or Dialog

mdodgelooker:
`Panels` uses `useMeasuredElement` (which relies on `ResizeObserver`) to determine dimensions and position. Since `ResizeObserver` does not detect changes to position, when rendering inside a `Drawer`, the position gets stuck off the page, where the `Drawer` animation begins.

I added a `useEffect` to `Panels` that manually updates the dom rect after a delay, if `DialogContext` is present.

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Panel/Panels.tsx
+++ b/packages/components/src/Panel/Panels.tsx
@@ -24,9 +24,16 @@
 
  */
 
-import { CompatibleHTMLProps } from '@looker/design-tokens'
-import React, { createContext, forwardRef, Ref } from 'react'
+import { CompatibleHTMLProps, transitions } from '@looker/design-tokens'
+import React, {
+  createContext,
+  forwardRef,
+  Ref,
+  useContext,
+  useEffect,
+} from 'react'
 import styled from 'styled-components'
+import { DialogContext } from '../Dialog'
 import { useCallbackRef, useMeasuredElement } from '../utils'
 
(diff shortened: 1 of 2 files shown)


packages/components/src/Panel/Panels.tsx
```diff
@@ -37,7 +44,28 @@ const PanelsLayout = forwardRef(
...
+    // Changes to position aren't captured by useMeasuredElement(ResizeObserver)
+    // so if Panels is rendered in a Drawer or Dialog, we need to update the dom rect
+    // once the animation is done
+    useEffect(() => {
+      let t: NodeJS.Timeout
+      if (hasAnimation) {
+        t = setTimeout(() => {
+          refreshDomRect()
+          // moderate is the transition duration for Drawer/Dialog
+        }, transitions.moderate + 20)
```
mdodgelooker, line 61: This worries me. I tried + 1 and + 10 – the former never captured a change and the latter was inconsistent. I think it may just be a factor of however long it takes to insert the `Drawer` and its contents into the dom, which could easily be higher than 20. Should I go up to 100 instead just to be safe?

PR #2508, opened 2021-06-15
fix(Panel): Render properly in Drawer or Dialog

mdodgelooker:
Get rid of `Portal` in `Panel` because the positioning is buggy. Use `zIndexFloor` to resolve the stacking issue with `Tree`.


lukelooker (approved):
Works really nicely.  Thanks for chasing this down :)

PR #2510, opened 2021-06-17
fix(HoverDisclosure): Remove strategy prop, add placeholderWidth

lukelooker:


packages/components/src/Tree/stories/FieldItem.tsx
```diff
@@ -141,10 +142,7 @@ export const FieldItem: FC<FieldItemProps> = ({
             tooltipPlacement="top"
           />
         </HoverDisclosure>
-        <HoverDisclosure
-          visible={isFilter}
-          strategy={isPivot ? 'visibility' : 'display'}
-        >
+        <HoverDisclosure visible={isFilter} width={isPivot ? 24 : undefined}>
```
ElliotTheHuman, line 145: This looks good, but for some reason, the outermost flex container  above is shrinking this below 24px:
Screen Shot 2021-06-16 at 5 58 39 PM


However, looks like we can fix this by adding a div with `flex-basis: 24px` and `flex-shrink: 0`:
```
<div style={{ flexBasis: 24, flexShrink: 0 }}>
          <HoverDisclosure visible={isFilter} width={isPivot ? 24 : undefined}>
            <IconButton
              onClick={() => setIsFilter(!isFilter)}
              toggle={isFilter}
              toggleBackground
              toggleColor={color}
              icon={<FilterList />}
              label="Filter"
              tooltipPlacement="top"
            />
          </HoverDisclosure>
        </div>
```

ElliotTheHuman (approved):
Left a quick note about a `FieldItem` issue, but otherwise, I think we're good 👍

packages/components/src/Tree/stories/FieldItem.tsx
lukelooker: Eww... Went ahead and built those props into the HoverDisclosure we'd need the `flexBasis` to move with the width prop as specified here.

lukelooker: Clarification... my "eww..." was in reference to the movement visible without the flexShrink in place

ElliotTheHuman (approved):
LGTM!

PR #2512, opened 2021-06-17
fix(HoverDisclosure): Remove strategy prop, add placeholderWidth

lukelooker:
Make sure HoverDisclosure behaves within flex containers

BREAKING CHANGE: strategy prop is no longer supported. `width` may be used to workaround.


ElliotTheHuman (approved):
:shipit:

PR #2513, opened 2021-06-17
chore: Update resolutions to update vulnerable transient dependencies

lukelooker:


package.json
```diff
@@ -93,22 +93,24 @@
...
     "highlight.js": ">= 10.4.1",
     "immer": ">= 8.0.1",
     "is-svg": ">= 4.2.2",
     "node-fetch": ">= 2.6.1",
     "node-forge": ">= 0.10.0",
+    "normalize-url": ">=4.5.1",
     "prismjs": ">= 1.23.0",
-    "react": "16.14.0",
-    "react-dom": "16.14.0",
+    "postcss": "7.0.36",
```
lukelooker, line 107: This technically pulls the version leveraged by Gatsby "down" (from PostCSS 8.x) but it seems to build fine in development and production environments so I don't think it's a problem to downgrade for Gatsby.

Upgrading to all the way breaks `stylelint` currently - their next major update will support `8.x` - https://github.com/stylelint/stylelint/issues/5205

packages/visualizations-nivo/package.json
```diff
@@ -36,7 +36,7 @@
   "react-dom": "^16.14.0"
  },
  "peerDependencies": {
-  "@looker/components": "2.x",
+  "@looker/components": "*",
```
lukelooker, line 39: These were throwing errors in Yarn log since `2.x` is invalid and `2` isn't yet available as a released (non-alpha) version. Just starred them for now since this packages are all marked private.

packages/visualizations/package.json
```diff
@@ -34,14 +34,14 @@
...
+  "react-dom": "^16.14.0",
+  "styled-system": "5.1.5"
  },
  "peerDependencies": {
-  "@looker/components": "2.x",
+  "@looker/components": "*",
   "react": ">= 16.11.0",
   "react-dom": ">= 16.11.0",
-  "styled-components": "^5.2.1",
-  "styled-system": "5.1.5"
```
lukelooker, line 44: Shouldn't be required as a peerDep (not related to this PR, feel free to tell me to toss this out)

package.json
```diff
@@ -93,22 +93,24 @@
...
+    "glob-parent": ">= 5.1.2",
     "highlight.js": ">= 10.4.1",
     "immer": ">= 8.0.1",
     "is-svg": ">= 4.2.2",
     "node-fetch": ">= 2.6.1",
     "node-forge": ">= 0.10.0",
+    "normalize-url": ">=4.5.1",
     "prismjs": ">= 1.23.0",
-    "react": "16.14.0",
-    "react-dom": "16.14.0",
```
lukelooker, line 106: Don't want these here since I'd really prefer to keep `resolutions` limited to cases it's intended (pushing versions of transient dependencies)

ElliotTheHuman: Out of curiosity, why did we have `react` and `react-dom` in resolutions if we already have them in our dev dependencies?

packages/visualizations-table/package.json
```diff
@@ -33,7 +33,7 @@
   "react-dom": "^16.14.0"
  },
  "peerDependencies": {
-  "@looker/components": "2.x",
+  "@looker/components": "*",
```
ElliotTheHuman, line 36: I know `*` is just "matches any version" in `semver`, but in practice, does this just lead to the latest version of `@looker/components` being installed when installing the `visualizations-table` package?

ElliotTheHuman (approved):
Added some questions for my own edification, but seems good to me.

package.json
lukelooker: I think it was an attempt to have universal versions across all packages – it does work that way but also gives a bit of a "false signal" if package versions fall out of sync.

packages/visualizations-table/package.json
lukelooker: It does! This should get updated to `>= 2` eventually but since we're in the middle of the `alpha` series currently it acts odd. The `visualization*` packages are currently all marked as "private" so it's not _really_ a problem until we go ahead and publish them.

PR #2515, opened 2021-06-17
refactor(DialogLayout): extracted ModalLayout

laraharrow:
- [ ] ♿️ Accessibility addressed
- [x] 🖼 Image Snapshot coverage
- [ ] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11

PR #2516, opened 2021-06-17
feat(PopoverLayout): new component for laying out for popover

laraharrow:
- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11


packages/components/src/Popover/Layout/PopoverFooter/PopoverFooter.story.tsx
```diff
@@ -25,33 +25,26 @@
...
-  },
-}
+export const Basic = () => <PopoverFooter />
 
-export const CloseStringValue = () => (
-  <PopoverFooter close="Footer">Footer Text</PopoverFooter>
+export const FooterWithChildren = () => (
+  <PopoverFooter>
+    <ButtonTransparent color="neutral" size="small">
+      cancel
```
lukelooker, line 41: Would be nice to capitalize this so it matches `done`

```suggestion
      Cancel
```

packages/components/src/Popover/Layout/PopoverFooter/PopoverFooter.test.tsx
```diff
@@ -27,29 +27,22 @@
...
 
-  test('close prop with specified element ', () => {
-    renderWithTheme(<CloseElementValue />)
-    expect(screen.getByText('Footer Text')).toBeInTheDocument()
-    expect(screen.getByText('close')).toBeInTheDocument()
-    expect(screen.getByRole('button')).toBeInTheDocument()
+  test('with children', () => {
+    renderWithTheme(<FooterWithChildren />)
+    expect(screen.getByText('Done')).toBeInTheDocument()
+    expect(screen.getByText('cancel')).toBeInTheDocument()
```
lukelooker, line 46:
```suggestion
    expect(screen.getByText('Cancel')).toBeInTheDocument()
```

packages/components/src/Popover/Layout/PopoverFooter/PopoverFooter.tsx
```diff
@@ -53,15 +54,17 @@ const PopoverFooterLayout: FC<PopoverFooterProps> = (props) => {
...
       )}
-      <>{children}</>
+      {children}
     </ModalFooter>
   )
 }
 
-export const PopoverFooter = styled(PopoverFooterLayout)<PopoverFooterProps>``
+export const PopoverFooter = styled(PopoverFooterLayout)<PopoverFooterProps>`
+  align-items: center;
```
lukelooker, line 69: Why is this needed?

packages/components/src/Popover/Layout/PopoverLayout.story.tsx
```diff
@@ -0,0 +1,74 @@
...
+    Cancel
+  </ButtonTransparent>
+)
+
+export const Basic = () => (
+  <Box bg="ui1">
+    <PopoverLayout header="Header Text" footer>
+      <ConstitutionShort />
+    </PopoverLayout>
+  </Box>
```
lukelooker, line 44: Since these are brand new can we not include the `Box` here?
```suggestion
  <PopoverLayout header="Header Text" footer>
    <ConstitutionShort />
  </PopoverLayout>
```

lukelooker: Oh , actually these should all be converted to CSF like `DialogLayout`

https://storybook.js.org/docs/react/api/csf

packages/components/src/Popover/Layout/PopoverLayout.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { PopoverFooter } from './PopoverFooter'
+import { PopoverHeader } from './PopoverHeader'
+
+type WithHeader = {
+  header?: ModalLayoutProps['header']
+  hideHeading?: never
+}
+
+type WithoutHeader = {
+  header?: never
```
lukelooker, line 39: Currently this excludes `header` entirely but we want to only support `string` if `hideHeading=true`

>  - `heading` must be a `string` if `hideHeading=true` (use type-union to enforce)

packages/components/src/Popover/Layout/PopoverLayout.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { PopoverContent } from './PopoverContent'
+import { PopoverFooter } from './PopoverFooter'
+import { PopoverHeader } from './PopoverHeader'
+
+type WithHeader = {
+  header?: ModalLayoutProps['header']
+  hideHeading?: never
+}
+
+type WithoutHeader = {
```
lukelooker, line 38:
```suggestion
type WithHiddenHeader = {
```

lukelooker (changes requested):
Some small nits as well as I'd like the `PopoverLayout` stories to be written as CSF (Common Storybook Format) to make them consistent with `DialogLayout`

packages/components/src/Popover/Layout/PopoverLayout.story.tsx
```diff
@@ -0,0 +1,87 @@
...
+  footer: true,
+  header: 'Header text',
+  hideHeading: true,
+}
+
+export const HeaderCloseButton = Template.bind({})
+HeaderCloseButton.args = {
+  footer: true,
+  header: 'Header text',
+  headerCloseButton: true,
```
lukelooker, line 75: This composition shouldn't be possible. 

We don't want a case where users see both the "x" at the top and the "Done" at the bottom. I think just removing `headerCloseButton` support should prevent that possibility as specifying any `footer` should automatically force the `PopoverHeader` to remove the `x` icon

packages/components/src/Popover/Layout/PopoverLayout.story.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+export const Footer = Template.bind({})
+Footer.args = {
+  footer: true,
+}
+
+export const Full = Template.bind({})
+Full.args = {
+  footer: true,
+  header: 'Header text',
```
lukelooker, line 61:
```suggestion
  ...Header.args,
  footer: true,
```

packages/components/src/Popover/Layout/PopoverLayout.story.tsx
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export const HeaderHideHeading = Template.bind({})
+HeaderHideHeading.args = {
+  footer: true,
+  header: 'Header text',
+  hideHeading: true,
```
lukelooker, line 68:
```suggestion
  ...Full.args,
  hideHeading: true,
```

packages/components/src/Popover/Layout/PopoverLayout.story.tsx
```diff
@@ -0,0 +1,87 @@
...
+HeaderCloseButton.args = {
+  footer: true,
+  header: 'Header text',
+  headerCloseButton: true,
+}
+
+export const FooterExtraValue = Template.bind({})
+FooterExtraValue.args = {
+  footer: extraButton,
+  header: 'Header text',
```
lukelooker, line 81:
```suggestion
  ...header.Args,
  footer: extraButton,
```

packages/components/src/Popover/Layout/PopoverLayout.story.tsx
```diff
@@ -0,0 +1,87 @@
...
+)
+
+const Template: Story<Partial<PopoverLayoutProps>> = (args) => (
+  <PopoverLayout {...args}>
+    <ConstitutionShort />
+  </PopoverLayout>
+)
+
+export const Basic = Template.bind({})
+Basic.args = {}
```
lukelooker, line 46: The screenshot for this test indicates that the component defaults are off - there needs to be some way for the user to close the `Popover` - I think with this composition the `footer` should be displayed with a "Done" button to allow the user to close the `Popover`?

packages/components/src/Popover/Layout/PopoverLayout.story.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+export const Header = Template.bind({})
+Header.args = {
+  header: 'Header text',
+}
```
lukelooker, line 51: I'm thinking that in this case the PopoverLayout should display the "x" in the header and no footer? 

image

packages/components/src/Popover/Layout/PopoverLayout.story.tsx
lukelooker: image

lukelooker (changes requested):
Few nits on code-style things (sorry, missed these on first pass) as well as a larger question about defaults in how things are presented.

Also, I'd like to see #2515 merged so I can see if there are any unexpected visual snapshot test changes.

packages/components/src/Popover/Layout/PopoverLayout.story.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+export const Footer = Template.bind({})
+Footer.args = {
+  footer: true,
```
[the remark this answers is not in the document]
laraharrow: I fix the logic to always have `PopoverFooter unless its specified `footer = false`

packages/components/src/Popover/Layout/PopoverLayout.story.tsx
laraharrow: I kept `headerCloseButton` but made a type-union for `footer` and `headerCloseButton`.  I also added some logic into `PopoverHeader` to avoid adding the close button if `footer = true`.
That seems to have salved the issue.

packages/components/src/Popover/Layout/PopoverLayout.story.tsx
laraharrow: on the `HeaderHideHeading` example if I don't add the header value explicitly: `header: 'Header text',` I get an error that says: 
```
Types of property 'header' are incompatible.
Type 'ReactNode' is not assignable to type '(string & (boolean | ReactChild | ReactFragment | ReactPortal | null)) | undefined'.
Type 'null' is not assignable to type '(string & (boolean | ReactChild | ReactFragment | ReactPortal | null)) | undefined'.
```

packages/components/src/Popover/Layout/PopoverLayout.story.tsx
laraharrow: I don't really understand what the exactly you need done here.

packages/components/src/Popover/Layout/PopoverLayout.story.tsx
laraharrow: Types of property 'header' are incompatible.
Type 'ReactNode' is not assignable to type '(string & (boolean | ReactChild | ReactFragment | ReactPortal | null)) | undefined'.
Type 'null' is not assignable to type '(string & (boolean | ReactChild | ReactFragment | ReactPortal | null)) | undefined'.

packages/components/src/Popover/Layout/PopoverLayout.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+    renderWithTheme(<Basic {...Basic.args} />)
+    expect(
+      screen.getByText(/We the People of the United States/)
+    ).toBeInTheDocument()
+    expect(screen.getByText('Done')).toBeInTheDocument()
+  })
+
+  test('hideHeading prop hides Header', () => {
+    renderWithTheme(<HeaderHideHeading {...HeaderHideHeading.args} />)
+    expect(screen.queryByText('Header Text')).not.toBeInTheDocument()
```
lukelooker, line 43: The text should still be in the document (for screen-readers) but not visible - the `VisuallyHidden` should be doing this for us.

```suggestion
    expect(screen.queryByText('Header Text')).not.toBeVisible()
    expect(screen.queryByText('Header Text')).toBeIntheDocument()
```

packages/components/src/Popover/Layout/PopoverLayout.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+    expect(screen.getByText('Done')).toBeInTheDocument()
+  })
+
+  test('Header and no Footer ', () => {
+    renderWithTheme(<Header {...Header.args} />)
+    expect(screen.getByText('Close')).toBeInTheDocument()
+    expect(screen.queryByText('Done')).not.toBeInTheDocument()
+  })
+
+  test('Footer and no Header has no closing button', () => {
```
lukelooker, line 56:
```suggestion
  test('With header & footer display only "Done" button', () => {
```

laraharrow:
I looked into the DialogLayout snapshots that got changed on the PopoverLayout PR.
I'm a bit confused about what happened there:
here is what I see that got pushed (and merged) from the DialogLayout PR:
export const Header = Template.bind({})
Header.args = {
  header: 'Header text',
}

export const Footer = Template.bind({})
Footer.args = {
  footer: 'Footer text',
}

export const Full = Template.bind({})
Full.args = {
  ...Footer.args,
  ...Header.args,
}

export const HeaderDetail = Template.bind({})
HeaderDetail.args = {
  ...Full.args,
  headerDetail: 'Cancel',
}

export const HeaderCloseButton = Template.bind({})
HeaderCloseButton.args = {
  ...Full.args,
  headerCloseButton: true,
}
My understanding is that all 5 stories would have the same header and footer values and those are: header: 'Header text' and footer: 'Footer text'.

...BUT...

Also on the `DialogLayout PR:

the snapshots for:
1. DialogLayout/Header Close Button the values are: header: 'Header' and footer: 'Footer.
2.DialogLayout/Header Detail the values are: header: 'Header' and footer: 'Footer.

on the other hand on the same PR the story referent to this code:
export const FooterSecondary = Template.bind({})
FooterSecondary.args = {
  ...Full.args,
  footerSecondary: 'Cancel',
}
exactly the same as the 2 above returns the snapshot of: header: 'Header' and footer: 'Footer

That is what I would expect for all of them. 
As a secondary issue, the same 2 problem snapshots: Header Close Button and Header Detail are yet again different then the expected after I updated it, now presenting as I would expect
1. DialogLayout/Header Close Button the values are: header: 'Header text' and footer: 'Footer text.
2.DialogLayout/Header Detail the values are: header: 'Header text' and footer: 'Footer text.

Sorry this is a long message. Not sure if it's clear.
What I did was update the snapshots on the PopoverLayout pr and the way that it is presented seems to make sense to me. I know its not idea to mix the 2 separated component work. So let me know what you think the best approach is for that

lukelooker (approved):
Ok - I'm not exactly sure how the `DialogLayout` examples passed previously but it appears that they match what's `main` should have (the copies in `main` appear to be slightly out-of-date. 

We do still need to connect with @jhardy about figuring out how to align the horizontal padding values for `PopoverContent` with the header and footer versions but I'm ok with doing that in a follow-up PR (I've created a ticket to track that fix)

PR #2517, opened 2021-06-17
fix(Tooltip): Performance improvement

mdodgelooker:
Eliminate one `useCallbackRef`(for the trigger element) since it causes a "double render".

The `useCallbackRef` for the surface element "shouldn't" be causing much of a performance issue (on huge FieldPicker's, for example) since the state doesn't get updated till the tooltip renders.

Unfortunately, I tested this locally in a huge FieldPicker and didn't see a significant improvement – _maybe_ about 20%.

### DIFF
--- a/packages/components/src/Tooltip/useTooltip.tsx
+++ b/packages/components/src/Tooltip/useTooltip.tsx
@@ -24,7 +24,7 @@
 
  */
 
-import React, { MouseEvent, useMemo, useState } from 'react'
+import React, { MouseEvent, useCallback, useMemo, useState } from 'react'
 import {
   useAnimationState,
   useCallbackRef,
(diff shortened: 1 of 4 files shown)


packages/components/src/Tooltip/useTooltip.tsx
```diff
@@ -60,41 +60,54 @@ export function useTooltip({
...
 
-  const handleOpen = () => {
-    if (!disabled && (!element || !element.dataset.notooltip)) {
-      setIsOpen(true)
-    }
-  }
-  const handleClose = () => {
+  const handleOpen = useCallback(
+    (e: { currentTarget: HTMLElement }) => {
+      setTriggerElement(e.currentTarget)
```
mdodgelooker, line 72: We can get the trigger element from the handler (focus / mouseover) instead of ref.

packages/components/src/Tooltip/useTooltip.tsx
```diff
@@ -24,7 +24,7 @@
 
  */
 
-import React, { MouseEvent, useMemo, useState } from 'react'
+import React, { MouseEvent, useCallback, useMemo, useState } from 'react'
```
mdodgelooker, line 27: Added `useCallback` around all the handlers, which could _potentially_ reduce re-renders, just in case that's also contributing to poor performance.

PR #2527, opened 2021-06-23
feat(Accordion2): New improved & simpler accordion component

lukelooker:
- Adds `Accordion2` & `useAccordion2`
- Refactor `Fieldset` to use new accordion
- Refactor `Tree` to use new accordion
- Refactor `NavList` to work with `Tree` update above
- Refactor `Accordion` on top of `useAccordion2`

## Differences, or why not just change `Accordion`?

Differences between `Accordion` & `Accordion2` represent breaking changes to the existing `Accordion` interface.

- `Accordion2` does not support `padding` or `typography` props (put them on the children
- `Accordion2` has no built-in padding
- `Accordion2` does not take a `content` prop. Use `children` instead
- `Accordion2` requires a `label` prop (former `children` prop)

## Known Issues

- `Tree` `aria` props are spread to the wrong place now (two tests are disabled until `Tree` refactor using `useAccordion2` corrects composition.

## Review-structured commits

Commits tell a story in this PR. You may likely benefit from reading through each PR incrementally to get a understanding of the changes here.

image

## Developer Checklist [ℹ️](https://github.com/looker-open-source/components/blob/main//CONTRIBUTING.md#developer-checklist)

- [x] ♿️ Accessibility addressed
- [ ] 🌏 Localization & Internationalization addressed
- [x] 🖼 Image Snapshot coverage
- [x] 📚 Documentation updated
- [x] 🧳 Bundle size impact addressed
- [ ] 🏁 Performance impacts addressed
- [ ] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [ ] Safari
  - [ ] IE11

PR #2528, opened 2021-06-23
fix(DialogLayout): `footerSecondary` is only valid if `footer` is also specified

laraharrow:
- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Dialog/Layout/DialogLayout.story.tsx
+++ b/packages/components/src/Dialog/Layout/DialogLayout.story.tsx
@@ -29,7 +29,7 @@ import { ConstitutionShort } from '../../__mocks__/Constitution'
 import { Box } from '../../Layout'
 import { DialogContent, DialogLayout, DialogLayoutProps } from '.'
 
-const Template: Story<Partial<DialogLayoutProps>> = (args) => (
+const Template: Story<DialogLayoutProps> = (args) => (
   <Box bg="ui1">
     <DialogLayout {...args}>
       <ConstitutionShort />
--- a/packages/components/src/Dialog/Layout/DialogLayout.test.tsx
+++ b/packages/components/src/Dialog/Layout/DialogLayout.test.tsx
@@ -27,35 +27,49 @@
 import React from 'react'
 import { renderWithTheme } from '@looker/components-test-utils'
 import { render, screen } from '@testing-library/react'
-import {
-  Basic,
-  HeaderDetail,
-  HeaderCloseButton,
-  FooterSecondary,
-} from './DialogLayout.story'
+import { ConstitutionShort } from '../../__mocks__/Constitution'
+import { Basic, HeaderDetail, HeaderCloseButton } from './DialogLayout.story'
+import { DialogLayout, DialogLayoutProps } from './DialogLayout'
+
+const removeFooterSecondary = (
+  props?: Partial<DialogLayoutProps>
+): DialogLayoutProps => ({ ...props, children: '', footerSecondary: undefined })
 
 describe('DialogLayout', () => {
   test('Basic ', () => {
-    render(<Basic {...Basic.args} />)
+    render(<Basic {...removeFooterSecondary(Basic.args)} />)
     expect(
       screen.getByText(/We the People of the United States/)
     ).toBeInTheDocument()
   })
 
   test('Replaces the built-in `IconButton` with an arbitrary ReactNode', () => {
-    renderWithTheme(<HeaderDetail {...HeaderDetail.args} />)
+    renderWithTheme(
+      <HeaderDetail {...removeFooterSecondary(HeaderDetail.args)} />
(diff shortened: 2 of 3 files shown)


packages/components/src/Dialog/Layout/DialogLayout.test.tsx
```diff
@@ -58,4 +59,12 @@ describe('DialogLayout', () => {
...
     expect(screen.getByText('Cancel')).toBeInTheDocument()
   })
+  test('FooterSecondary without footer does not display', () => {
+    renderWithTheme(
+      <FooterSecondaryWithoutFooter {...FooterSecondaryWithoutFooter.args} />
+    )
+    expect(screen.getByText('Close')).toBeInTheDocument()
+    expect(screen.queryByText('Footer text')).not.toBeInTheDocument()
+    expect(screen.queryByText('Cancel')).not.toBeInTheDocument()
+  })
```
lukelooker, line 69: Include a test that "proves" that the type-union is preventing the developer from specifying `footerSecondary` without `footer`

See example at: https://github.com/looker-open-source/components/pull/2527/files#diff-70b3f612b78b582cb7e5380b31ecf51102774084bbfc3a482bca4dd532565d93R125-R132

packages/components/src/Dialog/Layout/DialogLayout.story.tsx
```diff
@@ -74,6 +74,15 @@ FooterSecondary.args = {
...
 }
 
+export const FooterSecondaryWithoutFooter = Template.bind({})
+FooterSecondaryWithoutFooter.args = {
+  ...Header.args,
+  footerSecondary: 'Cancel',
+}
+FooterSecondaryWithoutFooter.parameters = {
+  storyshots: { disable: true },
+}
```
lukelooker, line 84: I think this should be causing a TS error?

packages/components/src/Dialog/Layout/DialogLayout.tsx
```diff
@@ -35,31 +35,42 @@ import { DialogContent } from './DialogContent'
 import { DialogFooter } from './DialogFooter'
 import { DialogHeader, DialogHeaderProps } from './DialogHeader'
 
-export interface DialogLayoutProps extends ModalLayoutProps {
+type WithFooter = {
+  footer?: ModalLayoutProps['footer']
```
lukelooker, line 39:
```suggestion
  footer: ModalLayoutProps['footer']
```

packages/components/src/Dialog/Layout/DialogLayout.tsx
```diff
@@ -88,9 +107,7 @@ export const DialogLayout: FC<DialogLayoutProps> = ({
   headerDetail,
   isLoading,
 }) => {
-  const dialogFooter = (footer || footerSecondary) && (
-    <DialogFooter secondary={footerSecondary}>{footer}</DialogFooter>
-  )
+  const dialogFooter = constructDialogFooter(footer, footerSecondary)
```
lukelooker, line 110: Given the simplicity of this logic I'd probably just inline it here:

```suggestion
  const dialogFooter = footer ? (
    <DialogFooter secondary={footerSecondary}>{footer}</DialogFooter>
  ) : null
```

packages/components/src/Dialog/Layout/DialogLayout.tsx
lukelooker: `footer` should also have a JSDoc explaining what it does: displays content on the right-side of the `DialogFooter`, uses `Space` to add whitespace between each element.

lukelooker (changes requested):
Based on a read-through of the code it appears like the discriminated union isn't quite doing it's job yet. 

Added comments inline.

packages/components/src/Dialog/Layout/DialogLayout.tsx
laraharrow: I don't think we can remove the `?` from footer's type.
That would indicate that every `DialogLayout` would need a `footer`

lukelooker: With the optional undefined `?:` this type basically says that you can specify all the following without TS errors:

```
const testNone: FooterOptions = {}
const testFooter: FooterOptions = { footer: 'hello' }
const testFooterSecondary: FooterOptions = { footerSecondary: 'world' }
const testBoth: FooterOptions = { footer: 'hello', footerSecondary: 'world' }
```

But `testFooterSecondary` should cause a Typescript compile error.

The second `WithoutFooter` type allows for `footer` to be undefined entirely as long as `footerSecondary` is ALSO undefined.

lukelooker (approved):
Assuming tests pass this LGTM! :)

PR #2530, opened 2021-06-24
fix(PopoverContent): changed default horizontal padding from `p="small"` to `pb="medium" pt="xsmall" px="large"`

laraharrow:
- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11


packages/components/src/Popover/Layout/PopoverContent/PopoverContent.test.tsx
```diff
@@ -36,20 +36,20 @@ describe('PopoverContent', () => {
...
+      '1rem'
     )
   })
 
   test('Custom padding', () => {
     renderWithTheme(
-      <PopoverContent pl="none" pt="xlarge" pb="small">
+      <PopoverContent pb="small" pt="xlarge" px="small">
         Hello world
       </PopoverContent>
```
lukelooker, line 47: Can you add a Storybook / image snapshot example of this as well. I'd like to have that to verify that the browser is acting as-intended in case some future refactor makes the CSS output different in a way the assertions below might miss.

lukelooker (changes requested):
Can you add a Storybook / image snapshot for the "Custom padding" case. 

I'd like to have that to verify that the browser is acting as-intended in case some future refactor makes the CSS output different in a way the assertions below might miss.

packages/components/src/Popover/Layout/PopoverContent/PopoverContent.test.tsx
laraharrow: oh sorry about that I added the story but with the `snapshot { disable: true }` fixing it now.

lukelooker (approved):
Looks great! :)

PR #2543, opened 2021-06-28
fix(visualizations): Update Vis Components directory structure and package files

dbchristopher:
- Add file structure and build files for new `@looker/visualizations-adapters` package
- Update all visualization package.json files to match dependency changes in our internal repo

### DIFF
--- a/packages/visualizations-nivo/package.json
+++ b/packages/visualizations-nivo/package.json
@@ -9,7 +9,7 @@
  "module": "lib/index.js",
  "sideEffects": false,
  "dependencies": {
-  "@looker/visualizations": "0.1.0",
+  "@looker/visualizations-adapters": "0.1.0",
   "@nivo/axes": "0.67.0",
   "@nivo/bar": "0.67.0",
   "@nivo/core": "0.67.0",
@@ -17,7 +17,7 @@
   "lodash": "4.17.21"
  },
  "devDependencies": {
-  "@looker/components-test-utils": "^1.4.0",
+  "@looker/components-test-utils": "^1.3.1",
   "@looker/eslint-config": "^1.5.0",
   "@looker/prettier-config": "^1.4.0",
   "@looker/stylelint-config": "^1.4.0",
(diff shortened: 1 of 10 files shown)


dbchristopher:
@lukelooker here is the updated dependency graph that matches the changes in our internal repo

dbchristopher:
@lukelooker do you think this is worth merging for now just to resolve those build errors?

packages/visualizations-nivo/package.json
```diff
@@ -2,23 +2,25 @@
  "name": "@looker/visualizations-nivo",
  "main": "src/index.ts",
  "version": "0.1.0",
- "private": true,
+ "publishConfig": {
+  "access": "restricted"
+ },
```
mdodgelooker, line 7: I'm not familiar with how this is treated differently from `"private": true,` but trusting it's what we need for these packages. 🙂

mdodgelooker: Side note, should this config be added to the "Facilitating Pre-Launch Packages" doc?

mdodgelooker (approved):
LGTM!

packages/visualizations-nivo/package.json
lukelooker: This is going to cause trouble with publishing all the other packages. Within the OSS repo we need to keep "private": true in place so we can not publish these.

lukelooker (changes requested):

packages/visualizations-nivo/package.json
mdodgelooker: Oh good to know, thanks @lukelooker !

lukelooker:
I'm wondering if we should just pull all of these packages out of the OSS repo for right now? It seems like keeping all of that within a `launch-vis` branch in the prerelease repo would leave this more "self-serve" for now and reduce the number of "hops" needed to make an update.

packages/visualizations-nivo/package.json
dbchristopher: oh thanks for flagging that! I think I misunderstood some feedback from a previous PR

dbchristopher:
Closing in favor of #2570

PR #2544, opened 2021-06-28
feat(Popover): updated aria-labelledby and role for A11Y

laraharrow:
- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Dialog/useDialog.tsx
+++ b/packages/components/src/Dialog/useDialog.tsx
@@ -189,7 +189,6 @@ export const useDialog = ({
         <Backdrop className={className} onClick={handleClose} />
         <RenderSurface
           id={dialogId}
-          role="dialog"
           aria-labelledby={`${dialogId}-heading`}
           aria-busy={busy ? true : undefined}
           className={className}
--- a/packages/components/src/Dialog/useDialog.tsx
+++ b/packages/components/src/Dialog/useDialog.tsx
@@ -189,6 +189,7 @@ export const useDialog = ({
         <Backdrop className={className} onClick={handleClose} />
         <RenderSurface
           id={dialogId}
+          role="dialog"
           aria-labelledby={`${dialogId}-heading`}
           aria-busy={busy ? true : undefined}
           className={className}
--- a/packages/components/src/Popover/Layout/PopoverHeader/PopoverHeader.tsx
+++ b/packages/components/src/Popover/Layout/PopoverHeader/PopoverHeader.tsx
@@ -52,20 +52,24 @@ type DetailOptions = WithDetail | WithHideClose
 
 export type PopoverHeaderProps = DetailOptions & {
   children: ReactNode
-  headerVisuallyHidden?: boolean
+  /**
+   * Make PopoverHeader visually hidden.
+   * @default false
+   */
+  hidden?: boolean
 }
 
 const PopoverHeaderLayout: FC<PopoverHeaderProps> = ({
   children,
   hideClose = false,
   detail,
-  headerVisuallyHidden = false,
+  hidden = false,
   ...props
 }) => {
-  const { id: dialogId } = useContext(DialogContext)
-  const headingId = dialogId || undefined
+  const { id } = useContext(DialogContext)
+  const headingId = id ? `${id}-heading` : undefined
 
-  return headerVisuallyHidden ? (
+  return hidden ? (
     <VisuallyHidden id={headingId}>{children}</VisuallyHidden>
   ) : (
     <ModalHeader
--- a/packages/components/src/Popover/Layout/PopoverHeader/PopoverHeader.tsx
+++ b/packages/components/src/Popover/Layout/PopoverHeader/PopoverHeader.tsx
@@ -24,8 +24,10 @@
 
  */
 
-import React, { FC, ReactNode } from 'react'
+import React, { FC, ReactNode, useContext } from 'react'
 import styled from 'styled-components'
+import { DialogContext } from '../../../Dialog/DialogContext'
+import { VisuallyHidden } from '../../../VisuallyHidden'
 import {
   ModalHeaderCloseButton,
   ModalHeader,
@@ -50,19 +52,27 @@ type DetailOptions = WithDetail | WithHideClose
 
 export type PopoverHeaderProps = DetailOptions & {
   children: ReactNode
+  headerVisuallyHidden?: boolean
 }
 
 const PopoverHeaderLayout: FC<PopoverHeaderProps> = ({
   children,
   hideClose = false,
   detail,
+  headerVisuallyHidden = false,
   ...props
 }) => {
-  return (
+  const { id: dialogId } = useContext(DialogContext)
+  const headingId = dialogId || undefined
+
+  return headerVisuallyHidden ? (
+    <VisuallyHidden id={headingId}>{children}</VisuallyHidden>
+  ) : (
     <ModalHeader
       detail={detail || (!hideClose && <ModalHeaderCloseButton size="small" />)}
       fontSize="small"
       fontWeight="medium"
+      id={headingId}
       pl="large"
       pr="medium"
       py="small"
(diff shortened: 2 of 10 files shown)


packages/components/src/Dialog/useDialog.tsx
```diff
@@ -189,6 +189,7 @@ export const useDialog = ({
         <Backdrop className={className} onClick={handleClose} />
         <RenderSurface
           id={dialogId}
+          role="dialog"
```
lukelooker, line 192: This is not needed. It's pushed in by `...surfaceProps` below.

packages/components/src/Popover/Layout/PopoverHeader/PopoverHeader.tsx
```diff
@@ -50,19 +52,27 @@ type DetailOptions = WithDetail | WithHideClose
 
 export type PopoverHeaderProps = DetailOptions & {
   children: ReactNode
+  headerVisuallyHidden?: boolean
```
lukelooker, line 55: Given that this is in the context of `PopoverHeader` already this could be shortened a good bit.

```suggestion
  hidden?: boolean
```

Make sure to add a JSDoc explaining the use-case.

packages/components/src/Popover/Layout/PopoverHeader/PopoverHeader.tsx
```diff
@@ -50,19 +52,27 @@ type DetailOptions = WithDetail | WithHideClose
...
+  headerVisuallyHidden = false,
   ...props
 }) => {
-  return (
+  const { id: dialogId } = useContext(DialogContext)
+  const headingId = dialogId || undefined
+
+  return headerVisuallyHidden ? (
+    <VisuallyHidden id={headingId}>{children}</VisuallyHidden>
+  ) : (
```
lukelooker, line 70:
```suggestion
  return hidden ? (
    <VisuallyHidden id={headingId}>{children}</VisuallyHidden>
  ) : (
```

packages/components/src/Popover/Layout/PopoverHeader/PopoverHeader.tsx
```diff
@@ -50,19 +52,27 @@ type DetailOptions = WithDetail | WithHideClose
...
 const PopoverHeaderLayout: FC<PopoverHeaderProps> = ({
   children,
   hideClose = false,
   detail,
+  headerVisuallyHidden = false,
   ...props
 }) => {
-  return (
+  const { id: dialogId } = useContext(DialogContext)
+  const headingId = dialogId || undefined
```
lukelooker, line 66: The `heading` needs to have `-heading` prepended to it so that the `labelledby` on the "surface" can target it.

```suggestion
    const { id } = useContext(DialogContext)
    const headingId = id ? `${id}-heading` : undefined

```

packages/components/src/Popover/Layout/PopoverLayout.tsx
```diff
@@ -61,10 +46,15 @@ export const PopoverLayout: FC<PopoverLayoutProps> = ({
   isLoading,
 }) => {
   const internalFooter = typeof footer === 'boolean' ? null : footer
-  const popoverHeader = constructPopoverHeader(header, hideHeader, !!footer)
+  const popoverHeader = hideHeader ? (
+    <PopoverHeader headerVisuallyHidden>{header}</PopoverHeader>
+  ) : (
+    <PopoverHeader hideClose={!!footer}>{header}</PopoverHeader>
+  )
+
```
lukelooker, line 54:
```suggestion

```

packages/components/src/Popover/Layout/PopoverLayout.tsx
```diff
@@ -61,10 +46,15 @@ export const PopoverLayout: FC<PopoverLayoutProps> = ({
...
   return (
     <>
-      {popoverHeader}
+      {header && popoverHeader}
```
lukelooker, line 57:
```suggestion
      {header && (
        <PopoverHeader hidden={hideHeader} hideClose={!!footer}>
          {header}
        </PopoverHeader>
      )}
```

packages/components/src/Popover/usePopover.tsx
```diff
@@ -242,7 +242,14 @@ export const usePopover = ({
       }}
     >
       <Portal ref={scrollRef}>
-        <SurfaceComponent placement={placement} ref={ref} style={style}>
+        <SurfaceComponent
+          aria-labelledby={dialogId}
```
lukelooker, line 246:
```suggestion
          aria-labelledby={`${dialogId}-heading`}
```

lukelooker (changes requested):
Added inline comments about the missing pieces here.

- [x] Needs Storybook example that shows the full composition of `Popover` + `PopoverLayout`
- [x] Needs test coverage: Visible Header
- [x] Needs test coverage: Hidden Header

### Rough test steps:

1. Open Popover w/ PopoverLayout
2. getByRole('dialog') 
3. get `aria-labelledby` attribute
3. Verify that element with `id` = `aria-labelledby` exists and matches expected text

packages/components/src/Popover/Popover.test.tsx
```diff
@@ -44,6 +45,31 @@ describe('Popover', () => {
...
+
+    const openPopover = screen.getByText('Open')
+    fireEvent.click(openPopover)
+
+    expect(screen.queryByText('Header text')).toBeInTheDocument()
+    expect(screen.getByRole('dialog')).toBeInTheDocument()
+    expect(screen.getByRole('dialog')).toHaveAttribute(
+      'aria-labelledby',
+      'a11y-heading'
+    )
```
lukelooker, line 70: We want to verify that the id specified in `aria-labelledby` exists.

```suggestion
    )
    expect(screen.getByText('Header text')).toHaveAttribute('id', 'a11y-heading')
```

packages/components/src/Popover/usePopover.tsx
```diff
@@ -244,13 +244,16 @@ export const usePopover = ({
       <Portal ref={scrollRef}>
         <SurfaceComponent placement={placement} ref={ref} style={style}>
           <Flex
-            flexDirection="column"
             alignItems="flex-start"
+            aria-labelledby={`${dialogId}-heading`}
```
lukelooker, line 248: The `dialog` usage here is making me :eye-twitch:

Can you change the spot where we instanient this to:

```
  id = useID(id)
```

And then the various spots using it can go from ~`dialogId`~ to just `id` (I'm guessing it was my bad choice when copying things around from `dialog` but it'd be nice to get this rationalized since it's getting tweaked here anyway.

packages/components/src/Popover/usePopover.tsx
```diff
@@ -244,13 +244,16 @@ export const usePopover = ({
...
+            borderRadius="inherit"
+            data-testid="content-container"
```
lukelooker, line 250: This is used in exactly one spot `Select.test.tsx` - can you remove it and then update the test there to use `selectByRole('dialog')` instead of ~`selectByTestId('content-container')`~

```suggestion
```

lukelooker (approved):
A few small nits on test and ancillary clean-ups than I think this should be good-to-go.

lukelooker (changes requested):
A few small nits on test and ancillary clean-ups than I think this should be good-to-go.

packages/components/src/Popover/usePopover.tsx
laraharrow: hey @lukelooker I agree with the change of dialogId into something else. But it can't be id it will have a conflict with the prop id
how about something like a11yId?
not sure I'm really bad at naming things

packages/components/src/Popover/usePopover.tsx
```diff
@@ -232,25 +232,27 @@ export const usePopover = ({
 
   const SurfaceComponent = surface || OverlaySurface
 
-  const dialogId = useID(id)
+  const popoverId = useID(id)
```
lukelooker, line 235:
```suggestion
 id = useID(id)
```

lukelooker (approved):
Awesome-sauce! This looks great! :)

PR #2545, opened 2021-06-29
refactor(ListItem): createListItemPartitions

lukelooker:
`createListItemPartitions` accepts all `ListItemProps` and outputs the core pieces that make up `ListItem` (i.e. the item, the label, the detail). It outputs these pieces in a 2-tuple, the first element being a fragment that contains all elements to go "inside" the main container, and a second element being all elements that go "outside" the main container (which is just the detail when accessory mode is enabled).

The PR does the following:
- Adds `createListItemPartitions`
- Refactors `ListItem` pieces to utilize `createListItemPartitions`
- Deprecates `padding` on `ListItem` `detail`, now is always `false`

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/List/utils/listItemPadding.ts
```diff
@@ -24,27 +24,24 @@
 
  */
 
-import { SpaceRamp } from '@looker/design-tokens'
+import { SpaceRamp } from 'packages/design-tokens/src/system'
```
lukelooker, line 27: Oops. Bad path here.

packages/components/src/List/ListItemDetail.tsx
```diff
@@ -29,11 +29,20 @@ import React from 'react'
...
-  <HoverDisclosure width={width} visible={!hoverDisclosure}>
-    <div {...props} />
-  </HoverDisclosure>
-))<DetailOptions>`
+export type ListItemDetailProps = Omit<
+  DetailOptions,
+  'accessory' | 'padding'
+> & {
+  className?: string
+}
```
lukelooker: Was `className` actually needed here? It seems like that will still pass through even though Typescript isn't aware of it.

packages/components/src/List/ListItemContent.tsx
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+const ListItemContentInternal: FC<ListItemContentProps> = ({
+  children,
+  itemRole = 'button',
+  ...props
+}) => {
+  let Component = (p: CompatibleHTMLProps<HTMLElement>) => (
+    <Button {...p} type="button" />
+  )
```
ElliotTheHuman, line 78: Something about this line makes `TreeCollection` and some of the `Menu` tests really unhappy.

packages/components/src/Tree/Tree.tsx
```diff
@@ -197,29 +199,40 @@ const TreeLayout = ({
...
+    ...accordionProps,
+  })
+
+  const innerAccordion = (
+    <TreeAccordion className="tree-accordion-disclosure">
+      {!partialRender && (
+        <Accordion2Disclosure {...domProps} {...disclosureProps} />
+      )}
+      {treeContent}
+    </TreeAccordion>
```
ElliotTheHuman, line 235: It's incredible how easy it was to compose everything myself. `useAccordion2` already seems like a big win.

lukelooker (commented):
Found 1 remaining oddity in manual testing. 

 1. Go to "Tree: basic" story
 2. Click the label text of a `Tree` (not a `TreeItem`)
 3. Focus doesn't land on the `Accordion` so `<enter>`, `up-arrow`, etc.. (keyboard navigation) doesn't function properly

`document.activeElement` shows that focus is landing on the `ListItemContent` instead of an accordion piece as would be expected (I think)

Anyway, happy to help poke at it more but wanted to write it down before I lost track.

ElliotTheHuman (approved):
LGTM!

PR #2565, opened 2021-07-02
feat(Accordion2): Density support

lukelooker:
Refactors `Accordion2` (and `useAccordion2`) to support `density` directly. 

Also refactors the `generateIndent` function to use density and to directly ingest scope from Styled Components template literal rather than requiring a variety of parameters.


.github/workflows/ci.yml
```diff
@@ -80,7 +80,9 @@ jobs:
         if: ${{ always() && (steps.snapshots.outcome == 'failure') }}
         # This command assumes Storybook has already been built
         # It's built above in the `yarn test:image-snapshots` so that's a safe assumption
-        run: yarn workspace storybook snapshots-update
+        run: |
+          yarn workspace storybook storyshots-preview-restore \
+          && yarn workspace storybook snapshots-update
```
lukelooker, line 85: Rewrote CI setup here a bit to allow for `snapshots-update` to work predictably in local dev environments.

packages/components/src/Accordion/Accordion.tsx
```diff
@@ -133,6 +134,8 @@ const AccordionInternal: FC<AccordionProps> = ({
  * @deprecated Use `Accordion2` instead
  */
 export const Accordion = styled(AccordionInternal)`
+  font-size: ${({ theme, density = 0 }) =>
+    theme.fontSizes[accordionDimensions(density).fontSize]};
```
lukelooker, line 138: Moved font-size to the `Accordion` so that content will also follow the font-size default with density. The developer can always specific a more specific font-size on the content within the accordion if they so choose.

packages/components/src/Accordion2/Accordion2.story.tsx
```diff
@@ -75,12 +76,34 @@ DefaultOpen.parameters = {
...
+    </Grid>
+  </SpaceVertical>
+)
+
 export const CustomizedIndicator = Template.bind({})
 CustomizedIndicator.args = {
   ...Basic.args,
   indicatorIcons: { close: <ChevronLeft />, open: <ExpandMore /> },
-  indicatorPosition: 'left',
-  indicatorSize: 'large',
```
lukelooker, line 83: `indicatorSize` is no longer supported.

packages/components/src/Accordion2/Accordion2Disclosure.tsx
```diff
@@ -26,28 +26,31 @@
...
-    return (
-      <div ref={ref} {...props}>
-        {indicatorPosition === 'left' && indicator}
-        <AccordionLabel>{children}</AccordionLabel>
-        {indicatorPosition !== 'left' && indicator}
-      </div>
-    )
-  }
+  ({ children, indicator, indicatorPosition, ...props }, ref) => (
+    <div ref={ref} {...props}>
```
lukelooker, line 48: Just removed an extra return here.

packages/components/src/Accordion2/accordionDimensions.ts
```diff
@@ -0,0 +1,77 @@
...
+  indicatorSize: 'small',
+}
+
+// Negative density values
+export const densityNegative1: AccordionDensityDimensions = {
+  ...density0,
+}
+export const densityNegative2: AccordionDensityDimensions = {
+  ...densityNegative1,
+}
```
ElliotTheHuman, line 55: nit: I can imagine someone asking us why `density = -1` or `density = -2` doesn't change anything compared to `density = 0`. Not a big deal, but do you think we should have these actually change a knob or two? @jhardy

ElliotTheHuman (commented):
This looks great. Very amazed that the `Tree` snapshots stayed in tact.

Couple nits and questions, but I'm pretty much ready to approve.

packages/components/src/Accordion2/Accordion2Disclosure.tsx
```diff
@@ -61,8 +64,7 @@ Accordion2DisclosureInternal.displayName = 'Accordion2DisclosureInternal'
  *
  * @TODO 3.x / Accordion removal
  */
-const labelTypographyDefaults = css`
-  font-size: ${({ theme }) => theme.fontSizes.small};
+const labelTypographyDefaults = css<Accordion2DisclosureProps>`
```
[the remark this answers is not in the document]
lukelooker: Yep! It sure can! ✅

packages/components/src/Accordion2/accordionDimensions.ts
lukelooker: I think that's a fair point. I'd say we can safely tighten this down progressively if/when Jared has specific thoughts since we've left "slots" for those changes to fit within.

ElliotTheHuman (approved):
MTGL

PR #2566, opened 2021-07-02
refactor: TreeItem2-ish

lukelooker:
Work in progress... please look away for now.


packages/components/src/ListItem/ListItemDetail.tsx
```diff
@@ -48,5 +48,6 @@ export const ListItemDetail = styled(
   color: ${({ theme }) => theme.colors.text2};
   display: flex;
   font-size: ${({ theme }) => theme.fontSizes.xsmall};
+  height: 100%;
```
lukelooker, line 51: Why'd we need this?

ElliotTheHuman: I believe it's to help center the detail elements in the center. Without `height: 100%` , the `ListItemDetail` is only as tall as its content, which might be less than the rest of the item.

PR #2580, opened 2021-07-04
fix: move lodash to peer dependencies

AprilArcus:
I observed that components brings in its own version of lodash whenever it's imported to helltool. Although `packages/components/README.md` [states that lodash should be supplied as a peer dependency](https://github.com/looker-open-source/components/blame/12658bddd81bf4dc1625865de424cc674ea07e1c/packages/components/README.md#L15), this does not actually appear to be the case. This PR moves lodash to `peerDependencies` and `devDependencies` wherever it is found in the `components` tree.

### DIFF
--- a/packages/components-providers/package.json
+++ b/packages/components-providers/package.json
@@ -20,7 +20,6 @@
   "dependencies": {
     "@looker/design-tokens": "^2.0.0",
     "i18next": "20.3.1",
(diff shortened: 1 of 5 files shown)


lukelooker (approved):
LGTM (pushed a small change to update "getting started" page as well)

PR #2582, opened 2021-07-07
fix(PopoverHeader):  update width for use with Popover

laraharrow:
- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Popover/Layout/PopoverHeader/PopoverHeader.tsx
+++ b/packages/components/src/Popover/Layout/PopoverHeader/PopoverHeader.tsx
@@ -87,4 +87,6 @@ const PopoverHeaderLayout: FC<PopoverHeaderProps> = ({
   )
 }
 
-export const PopoverHeader = styled(PopoverHeaderLayout)<ModalHeaderProps>``
+export const PopoverHeader = styled(PopoverHeaderLayout)<ModalHeaderProps>`
+  width: 100%;
+`
--- a/packages/components/src/Popover/stories/Popover.story.tsx
+++ b/packages/components/src/Popover/stories/Popover.story.tsx
@@ -70,6 +70,23 @@ Basic.parameters = {
   storyshots: { disable: true },
 }
 
+export const PopoverPopoverLayoutWithCloseButton = () => (
+  <Popover
+    content={
+      <PopoverLayout header="Header text" footer={false}>
+        <ConstitutionShort fontSize="small" />
+      </PopoverLayout>
+    }
+  >
+    <Button>Open Popover</Button>
+  </Popover>
+)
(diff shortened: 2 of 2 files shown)


mdodgelooker (approved):
LGTM!

packages/components/src/Popover/Layout/PopoverHeader/PopoverHeader.tsx
```diff
@@ -87,4 +87,6 @@ const PopoverHeaderLayout: FC<PopoverHeaderProps> = ({
   )
 }
 
-export const PopoverHeader = styled(PopoverHeaderLayout)<ModalHeaderProps>``
+export const PopoverHeader = styled(PopoverHeaderLayout)<ModalHeaderProps>`
+  width: 100%;
```
lukelooker, line 91: I wonder if it would make sense to put this on `ModalHeader` so that `DialogHeader` gets it as well?

jhardy: I think it would make sense to add on `ModalHeader` as well, do we need to make sure `Footer` is also 100%?

laraharrow: I checked and this behavior does not appear to be a problem for `Dialog`

Screen Shot 2021-07-08 at 7 55 12 AM

lukelooker: I think it'd still be good to move this behavior up to `ModalHeader` to make it consistent - it does seem really odd that only this component would need this fix if `Footer` does not also need it. 

I would like to better understand why this one is an outlier.

mdodgelooker:
A little bit of context...

`PopoverHeader` and `PopoverFooter` are direct children of this `Flex` which has `alignItems="flex-start"`: https://github.com/looker-open-source/components/blob/main/packages/components/src/Popover/usePopover.tsx#L246

`PopoverFooter`/`ModalFooter` already has `width: 100%` because it renders a `Space`, which has a default width of 100%: https://github.com/looker-open-source/components/blob/main/packages/components/src/Layout/Space/Space.tsx#L172

`PopoverHeader`/`ModalHeader` does not render a `Space`: https://github.com/looker-open-source/components/blob/main/packages/components/src/Modal/ModalHeader/ModalHeader.tsx#L60

lukelooker:
Thank you @mdodgelooker, that makes sense. I put a PR that _attempts_ to reconcile the behavior a bit so hopefully I'll be less clueless next time we look at this code :)

https://github.com/looker-open-source/components/pull/2586

PR #2584, opened 2021-07-07
fix(usePopover): aria-haspopup updated to

laraharrow:
- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Popover/stories/Popover.story.tsx
+++ b/packages/components/src/Popover/stories/Popover.story.tsx
@@ -70,6 +70,19 @@ Basic.parameters = {
   storyshots: { disable: true },
 }
 
+export const PopoverAriaHasPopup = () => (
+  <Popover
+    ariaHaspopup="dialog"
+    content={<PopoverContent>Some content</PopoverContent>}
+  >
+    <Button>Open</Button>
+  </Popover>
+)
+
+PopoverAriaHasPopup.parameters = {
+  storyshots: { disable: true },
+}
+
 export const PopoverPopoverLayout = () => (
   <Popover
     content={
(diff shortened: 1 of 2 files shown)


laraharrow:
To see the `aria-haspopup` at work check the `MovingTargetInner` example on `Popover.story.tsx`
if `menu` or `true` is passed they both will show as `true`.

packages/components/src/Popover/stories/Popover.story.tsx
```diff
@@ -372,18 +372,19 @@ Placement.parameters = {
...
+      content: (
+        <PopoverContent p="large" width="360px">
+          <Paragraph>
+            The anchor is moving around so the Popover position must be updated
+            via popperInstanceRef.current.update.
+          </Paragraph>
+        </PopoverContent>
+      ),
+      placement: 'right-end',
+    })
```
mdodgelooker, line 387: You don't need to make a story with `usePopover` specifically – `Popover` supports all the same props. A majority of use cases will choose `Popover` and `usePopover` should only be used for very specific situations where the developer needs to control the behavior more closely.

I think it would be best to undo the changes to this story and just create a new story right under `Basic` that duplicates basic but adds `aria-haspopup="grid"` or something.

packages/components/src/Popover/usePopover.tsx
```diff
@@ -55,6 +55,13 @@ export interface UsePopoverProps extends UsePopoverToggleProps {
    */
   content: ReactNode
 
+  /**
+   * Use this prop to specify aria-haspopup value
+   * menu - listbox - tree - grid are used to identify the component popover is.
+
+   */
+  ariaHasPopup?: boolean | 'grid' | 'dialog' | 'menu' | 'listbox' | 'tree'
+
```
mdodgelooker, line 64: You could reduce the duplication if you remove this section as well as line 142 in `UsePopoverResponseDom` and add the following type at the top of the file, and extend both props from it (`AriaAttributes` can be imported from React):
```
type AriaHaspopupProps = Pick<AriaAttributes, 'aria-haspopup'>
```

One minor caveat is that you'd also have to change line 150 to the following, since there's a hyphen in the prop name:
```
'aria-haspopup': ariaHasPopup
```

PR #2585, opened 2021-07-07
fix(RangeSlider): Value reset due to intermediate rerender

mdodgelooker:
If `RangeSlider` is rendered inside a component with a "double re-render", e.g. one state variable is set due to the `onChange` being called, then another state variable is set in a `useEffect` as a result of the first state change, or certain implementations of redux can cause this, then it exposes a bug where the value is continually reverted: https://looker-open-source.github.io/components/canary/storybook/?path=/story/rangeslider--rerender-repro&globals=measureEnabled:false

My first attempt to fix this https://github.com/looker-open-source/components/pull/2174 introduced a different bug where the value would not respond to min/max changes.

This time around, my approach was to remove all the `// eslint-disable-next-line react-hooks/exhaustive-deps`, since breaking this rule does lead to confusing behavior (and the min/max bug I introduced), and add `useMemo` / `useCallback` as necessary.

To verify that this fix wouldn't introduce regressions in the core product, I created a test-only PR in that repo (see ticket comment).

### DIFF
--- a/packages/components/src/Form/Inputs/RangeSlider/RangeSlider.tsx
+++ b/packages/components/src/Form/Inputs/RangeSlider/RangeSlider.tsx
@@ -30,10 +30,10 @@ import React, {
   Ref,
   useState,
   KeyboardEvent,
-  useRef,
-  useEffect,
   useCallback,
+  useEffect,
   useMemo,
+  useRef,
 } from 'react'
 import {
   SpaceProps,
@@ -41,9 +41,8 @@ import {
   space,
   typography,
   TypographyProps,
-  width,
-  WidthProps,
 } from '@looker/design-tokens'
+import { WidthProps } from 'styled-system'
 import styled from 'styled-components'
 import sortBy from 'lodash/sortBy'
 import indexOf from 'lodash/indexOf'
@@ -252,7 +251,11 @@ export const InternalRangeSlider = forwardRef(
     const maxPos = ((maxValue - min) / (max - min)) * containerRect.width
     const fillWidth = maxPos - minPos
 
-    // Behavioral callbacks
+    /*
+     * Behavioral callbacks
+     * ------------------------------------------------------
+     */
+
     const roundSliderValue = partial(roundToStep, min, max, step)
 
     const focusChangedPoint = useCallback(
@@ -356,9 +359,11 @@ export const InternalRangeSlider = forwardRef(
       [handleMouseEvent]
     )
 
-    // Mouse down event (and re-measure the client rectangle values before calculating value).
-    // This ensures accurate calculation when slider has moved to new location on page due
-    // to asynchronous dom changes.
+    /*
+     * Mouse down event (and re-measure the client rectangle values before calculating value).
+     * This ensures accurate calculation when slider has moved to new location on page due
+     * to asynchronous dom changes.
+     */
     useEffect(() => {
       if (isMouseDown) {
         refreshDomRect() // re-measure rectangle when isMouseDown changes from false to true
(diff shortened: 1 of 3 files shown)


packages/components/src/Form/Inputs/RangeSlider/RangeSlider.tsx
```diff
@@ -187,6 +190,18 @@ export const InternalRangeSlider = forwardRef(
...
+    // (an array may be "new" even if the numbers haven't changed)
+    const valuePropMin = valueFromProps?.[0]
+    const valuePropMax = valueFromProps?.[1]
+    const valueProp = useMemo(() => {
+      if (valuePropMin === undefined || valuePropMax === undefined) {
+        return undefined
+      }
+      return [valuePropMin, valuePropMax]
+    }, [valuePropMin, valuePropMax])
+
```
mdodgelooker, line 203: With a value prop that's an array, the developer may not have gone to the trouble to have memoized it, so this component would see it as "new" every render, even if the numbers haven't changed. This bit ensures the memoization.

packages/components/src/Form/Inputs/RangeSlider/RangeSlider.tsx
```diff
@@ -197,14 +211,29 @@ export const InternalRangeSlider = forwardRef(
...
-    const [value, setValue] = useState(sort(boundedValue))
+    const [valueState, setValue] = useState(() => {
+      // make sure value array doesn't extend past min/max bounds
+      const boundedValue = boundValueProp(
+        min,
+        max,
+        valueProp || defaultValueProp
+      )
+      return sort(boundedValue)
+    })
```
mdodgelooker, line 227: Since the `boundedValueProp` and `sort` are only needed to get the initial value, the `useState` callback avoids calling them on subsequent renders.

packages/components/src/Form/Inputs/RangeSlider/RangeSlider.tsx
```diff
@@ -197,14 +211,29 @@ export const InternalRangeSlider = forwardRef(
...
+
+    // Create a more stable value state
+    // (an array may be "new" even if the numbers haven't changed)
+    const valueMin = valueState[0]
+    const valueMax = valueState[1]
+    const value = useMemo(() => {
+      return [valueMin, valueMax]
+    }, [valueMin, valueMax])
```
mdodgelooker, line 235: Also memoizing the `value` state since it's subject to the same "new without the numbers changing" issue as `valueProp`.

packages/components/src/Form/Inputs/RangeSlider/RangeSlider.tsx
```diff
@@ -312,53 +368,47 @@ export const InternalRangeSlider = forwardRef(
...
-      const boundedValue = boundValueProp(min, max, valueProp)
-      if (!isEqual(value, boundedValue)) {
-        setValue(sort(boundedValue))
+      const boundedValueProp = boundValueProp(min, max, valueProp)
+      if (
+        valueProp &&
+        !isEqual(value, boundedValueProp) &&
+        !isEqual(valueProp, previousValueProp)
+      ) {
+        setValue(sort(boundedValueProp))
```
mdodgelooker, line 399: Narrowing the condition here a bit to also check that the component is indeed controlled (`valueProp` isn't undefined) and `valueProp` has changed this render (with everything else now in the deps array, this will get checked more often).

packages/components/src/Form/Inputs/RangeSlider/RangeSlider.tsx
```diff
@@ -260,6 +291,7 @@ export const InternalRangeSlider = forwardRef(
           const newValue = sort([newPoint, value[unfocusedThumb]])
           focusChangedPoint(newValue, newPoint)
           setValue(newValue)
+          onChange?.(newValue)
```
mdodgelooker, line 294: Calling `onChange` explicitly now here and in `handleMouseEvent` since I narrowed the last `useEffect` a bit (see below) that used to be calling it any time the `value` state changed.

lukelooker (approved):
Played with the old vs. new and can't see the issue here or on the product-integration branch. LGTM!

dbchristopher (approved):
Wow, really phenomenal job debugging and tightening up all this logic! It all works great and seems absolutely bulletproof. You have a real mastery of all the intricacies of the react lifecycle hooks. 💯

mdodgelooker:
Aw thank you both!!

PR #2587, opened 2021-07-08
test(Tree): Correct `assumeIconAlignment` and `truncate` breakage

lukelooker:


packages/components/src/Tree/TreeStyle.tsx
```diff
@@ -92,13 +94,14 @@ const dividersCSS = css`
 
 interface TreeItemIndentProps extends GenerateIndentProps {
   labelBackgroundOnly: boolean
-  icon?: IconType
+  icon: boolean
```
lukelooker, line 97: Doesn't this come in via `GenerateIndentProps` now? Can we just remove it altogther?

ElliotTheHuman: It totally does. Good call.

ElliotTheHuman (approved):

PR #2591, opened 2021-07-08
docs: Add Color documentation to our foundation section

jhardy:
This PR adds a documentation pages about our color palette and the various groups of color we have.

image


www/src/documentation/foundations/demos/Color.tsx
```diff
@@ -0,0 +1,296 @@
...
+    specializedText: []
+  }
+
+  const colorBands = {
+    core: coreColors.length,
+    intent: coreColors.length + intentColorKeys.length,
+    derivative: coreColors.length + intentColorKeys.length + derivativeColorKeys.length
+  }
+
+  for(const [key, value] of Object.entries(theme.colors)) {
```
jhardy, line 197: there might be a better way to do this, but the idea here is to break out the theme into their logical grouping, this way we can document each group in a bit more detail

www/src/documentation/foundations/demos/Color.tsx
```diff
@@ -0,0 +1,296 @@
...
+const derivativeColorKeys: Array<keyof DerivativeColors> = [
+  'field',
+  'inverse',
+  'inverseOn',
+  'neutral',
+  'linkInteractive',
+  'informAccent',
+  'positiveAccent',
+  'warnAccent',
+]
```
jhardy, line 56: Declaring color keys here and below feels a bit brittle. I was curious if we should export these from `design-tokens` package. Interested in  thoughts here.

www/src/documentation/foundations/demos/Color.tsx
```diff
@@ -0,0 +1,296 @@
...
+    }
+    else if(specializedTextColorKeys.includes(key)) {
+      colorList.specializedText.push([key, value])
+    }
+    else {
+      colorList.stateful.push([key, value])
+    }
+  }
+
+  colorList.stateful = chunk(colorList.stateful, 7)
```
jhardy, line 221: This chunks the stateful colors into logical groupings that made it easer to lay out for the docs

lukelooker (approved):
@jhardy - I went ahead and hacked on top of your PR to add exports for all of the color groupings this wanted to leverage as well as moving the function to break-up the theme's `color` object into it's requisite parts.

I also moved most of the demo text into `MDX` to help with search discovery and minimize the mixture of code and documentation. 

Assuming tests pass I think this is fine to merge as it doesn't have any "functional" effects on the library outside of improving the documentation around color.

lukelooker (dismissed):
Had to update a single snapshot from theme-editor demo.

jhardy:
@lukelooker thanks for your improvements, much better. I refactored the some of the grids for better responsive handling on the color page. 

I also went ahead and added some docs for the `radii` theme values after I saw you use `theme.radii[3]` and I was confused why it wasn't working, but it needed to be `theme.radii.medium`.

www/src/documentation/foundations/demos/ColorList.tsx
```diff
@@ -41,22 +41,28 @@ const ColorSwatch = styled(({ name, color, ...props }: ColorSwatchProps) => (
   &::before {
     background: ${({ color }) => color};
     border: 1px solid ${({ theme }) => theme.colors.ui3};
-    border-radius: ${({ theme }) => theme.radii[3]};
-    content: ' ';
+    border-radius: ${({ theme }) => theme.radii.medium};
```
lukelooker, line 44: 🤦

www/src/documentation/foundations/demos/ColorList.tsx
```diff
@@ -41,22 +41,28 @@ const ColorSwatch = styled(({ name, color, ...props }: ColorSwatchProps) => (
...
     display: block;
     height: 4rem;
-    width: 7.5rem;
+    margin-bottom: ${({ theme }) => theme.space.xsmall};
+    width: 100%;
   }
 `
 
+const ColorListGrid = styled(Grid)`
+ grid-template-columns: repeat(auto-fit, minmax(100px, 1fr));
```
lukelooker: I wonder if we should add `autofit` to `Grid` as a prop that takes the minimum value as a string and assumes the max is always `1fr` - seems like a pattern that could/would be useful elsewhere.

lukelooker (approved):
Pushed a commit with a couple of prettier clean-ups (lint doesn't require them in MDX but it was distracting me).

The custom grids you created expose a nice enhancement we could probably do to the grid component at some point (see inline comment).

PR #2594, opened 2021-07-09
fix(InputSearch): Return focus to input after clearing

mdodgelooker:
Adds `focus()` to the clear button handler in `ComboboxInput`. (`InputSearch` is the most obvious use case but this also affects `Select` and `SelectMulti` if they have `isClearable`).

- [ ] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11


lukelooker (approved):
LGTM

PR #2595, opened 2021-07-09
fix(InputColor): Incorrect coloring and positioning when hueless or saturationless color is provided

ElliotTheHuman:
- Added fallback values to `simpleStringToHsv` function to prevent NaN values from appearing in color state's h and s values
- Added tests to assure that we replace NaN values with default values (i.e. 0)

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Form/Inputs/InputColor/utils/stringToSimpleHsv.ts
+++ b/packages/components/src/Form/Inputs/InputColor/utils/stringToSimpleHsv.ts
@@ -29,17 +29,17 @@ import { SimpleHSV } from '../types'
 
 export const stringToSimpleHsv = (color: string): SimpleHSV => {
   const hsvColor = hsv(color)
-
+  let { h, s, v } = hsv(color)
   /**
-   * The hsv helper function returns NaN for hueless and saturationless values
+   * The hsv helper function returns NaN for hueless and saturation-less values
    * like black, white, grey. We will instead send 0 if either value is NaN for use
    * when calculating handle positioning and coloring on the hue slider and saturation preview.
    *
-   * 0 is used as a fallback value to mirror MDN's handling of hueless and saturationless values:
+   * 0 is used as a fallback value to mirror MDN's handling of hueless and saturation-less values:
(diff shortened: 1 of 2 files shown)


packages/components/src/Form/Inputs/InputColor/utils/stringToSimpleHsv.ts
```diff
@@ -29,5 +29,17 @@ import { SimpleHSV } from '../types'
...
+   * like black, white, grey. We will instead send 0 if either value is NaN for use
+   * when calculating handle positioning and coloring on the hue slider and saturation preview.
+   *
+   * 0 is used as a fallback value to mirror MDN's handling of hueless and saturationless values:
+   * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color
+   */
+  const hue = isNaN(hsvColor.h) ? 0 : hsvColor.h
+  const saturation = isNaN(hsvColor.s) ? 0 : hsvColor.s
+
+  return { h: hue, s: saturation, v: hsvColor.v }
```
lukelooker, line 44: The terse-obsessed nerd in me see this (completely unneeded) refactor to tighten things up a bit here:

```suggestion
  let { h, s, v } = hsv(color)
  /**
   * The hsv helper function returns NaN for hueless and saturation-less values
   * like black, white, grey. We will instead send 0 if either value is NaN for use
   * when calculating handle positioning and coloring on the hue slider and saturation preview.
   *
   * 0 is used as a fallback value to mirror MDN's handling of hueless and saturation-less values:
   * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color
   */
  if(isNaN(h)) h = 0
  if(isNaN(s)) s = 0

  return { h, s, v }
```

lukelooker (approved):
LGTM - I added a comment about an alternate way to write the functional portion of this but I'm not sure my version is any better - just the way my brain wanted to see it. Feel free to ignore and proceed on as-is if you prefer the way you wrote it :)

lukelooker (approved):

PR #2596, opened 2021-07-09
feat: Consistently support props across "Semantic Layout" components as well as `Grid`

lukelooker:
Add new common prop type call `CommonLayoutProps` that supports colors, borders, sizing and positioning


packages/components/src/Icon/Icon.tsx
```diff
@@ -89,7 +86,7 @@ export const Icon = styled(IconLayout).attrs<IconProps>(
     size,
   })
 )<IconProps>`
-  ${sizeSimpleLayoutCSS}
+  ${simpleLayoutCSS}
```
lukelooker, line 89: `LayoutProps` includes `size` so we're good to go without `sizeSimpleLayout`

packages/components/src/Layout/Semantics/Layout/LayoutContext.ts
```diff
@@ -1,42 +0,0 @@
...
-  registerAside: () => void
-  registerFixed: () => void
-}
-
-export const LayoutContext = createContext<LayoutContextProps>({
-  hasAside: false,
-  isFixed: false,
-  registerAside: () => noop,
-  registerFixed: () => noop,
-})
```
lukelooker, line 42: Totally dead file.

packages/components/src/Layout/utils/common.ts
```diff
@@ -24,19 +24,29 @@
...
+import { simpleLayoutCSS, SimpleLayoutProps } from './simple'
+
+export type CommonLayoutProps = SimpleLayoutProps &
+  SemanticBorderProps &
+  ColorProps & {
+    /**
+     * Workaround for Styled Components merge with DOM `color` prop merge issue
+     */
+    color?: any
+  }
```
lukelooker, line 39: Not really intended to be a "move" originally but this type fits pretty well to cover all the bases.

packages/components/src/Layout/utils/common.ts
```diff
@@ -24,19 +24,29 @@
...
+
+export const commonLayoutCSS = css`
+  /**
+   * Rules here should provide convenience styling for Box derived components.
+   * Generally anything here could be overwritten by explicit values set via
+   * Box's prop values. For example a function here that sets 'cursor: pointer'
+   * would be overwritten by an explicit <Box cursor='copy'/>.
+   */
+  ${simpleLayoutCSS}
```
jhardy, line 48: Just thinking about this a bit, we now have `commonLayout`, `simpleLayout`, and `complexLayout`, I think it might be helpful to add some brief documentation for when you might reach for which one and why.

Also just to play 👿 here now that I am looking at it. Do we really need three layout utils?  I guess it comes down to simple is just common with color and border helpers? When would  I reach for simple over common?

packages/components/src/Layout/utils/common.ts
```diff
@@ -24,19 +24,29 @@
...
+
+  ${color}
+  ${borderHelper}
```
jhardy, line 51: Out of curiosity is this documented anywhere?

jhardy (dismissed):
LGTM. Left a few small nits, but I think this is a nice improvement

packages/components/src/Layout/utils/common.ts
lukelooker: I went ahead and added specific documentation the props themselves so when in use they'll show a reasonable doc-string. I'll file a ticket to track adding a "roll-up" documentation page in the future.

packages/components/src/Layout/utils/common.ts
lukelooker: This is a great point. I think the `complex` set should be deprecated in favor of the `common` set. I went ahead and changed things a bit to add `typography` and `borderRadius` support to common and moved `Card*` over.

Changing `Box`, `Flex`, `& `FlexItem` over will be a breaking change so I think looking to do so in 3.x (or another MAJOR release) would be the right call.

lukelooker: I also add formal deprecation text to `complex`

jhardy (approved):
👍

PR #2602, opened 2021-07-12
docs(DataTable): Clarify numeric-width (percentage) column behavior

lukelooker:
Clarifies behavior of `DataTable` columns when a percentage (numeric) width is used.

- [x] 📚 Documentation updated

### DIFF
--- a/www/package.json
+++ b/www/package.json
@@ -10,7 +10,7 @@
   "devDependencies": {
     "@looker/components": "*",
     "@looker/components-theme-editor": "*",
-    "@looker/components-date": "^2.0.0-alpha",
+    "@looker/components-date": "*",
     "@looker/filter-components": "*",
     "@looker/design-tokens": "*",
     "@looker/icons": "*",
--- a/www/src/MDX/Headings.tsx
+++ b/www/src/MDX/Headings.tsx
@@ -30,7 +30,7 @@ import styled from 'styled-components'
 import { maxTextWidth } from './styles'
 
 const generateHeadingAnchor = (children?: ReactNode) => {
-  return (children as string).toLowerCase().replace(/\s/g, '-')
+  return String(children).toLowerCase().replace(/\s/g, '-')
 }
 
 const StyledHeading = styled(Heading).attrs<HeadingProps>(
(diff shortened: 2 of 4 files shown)


www/package.json
```diff
@@ -10,7 +10,7 @@
   "devDependencies": {
     "@looker/components": "*",
     "@looker/components-theme-editor": "*",
-    "@looker/components-date": "^2.0.0-alpha",
+    "@looker/components-date": "*",
```
lukelooker, line 13: Old version, `*` is more universal for this case anyway.

www/src/MDX/Headings.tsx
```diff
@@ -30,7 +30,7 @@ import styled from 'styled-components'
 import { maxTextWidth } from './styles'
 
 const generateHeadingAnchor = (children?: ReactNode) => {
-  return (children as string).toLowerCase().replace(/\s/g, '-')
+  return String(children).toLowerCase().replace(/\s/g, '-')
```
lukelooker, line 33: This will ensure that `children` is treated as a string rather than just _telling_ the type system

www/src/documentation/components/content/data-table.mdx
```diff
@@ -15,6 +15,42 @@ The `<DataTable>` component acts as the main wrapper to other `<DataTable>`-rela
 
 For accessibility `caption` must be specified to describe the table's contents.
 
+## Columns
+
+`DataTableColumn`s are defined as an array of objects that confirm the the `DataTableColumn` interface:
```
laraharrow, line 20: `DataTableColumn`s are defined as an array of objects that confirm the `DataTableColumn` interface:
```suggestion
`DataTableColumn`s are defined as an array of objects that confirms the `DataTableColumn` interface:
```

laraharrow (changes requested):
Looks good just a saw a little typo other then that if test passes LGTM

www/src/documentation/components/content/data-table.mdx
lukelooker: Oops! I'd actually meant to say "that confirms to the". Fixing now.

laraharrow (approved):
LGTM🚢

PR #2604, opened 2021-07-12
refactor: Consistently import styled-system via @looker/design-tokens

lukelooker:
Currently we're somewhat inconsistent about how we import `styled-system` types and helpers. This PR consolidates all utilized imports in @looker/design-tokens (to aid a prospective replacement at some undetermined future point).

PR also removes dependency on `styled-system` from all packages except `design-tokens`


mdodgelooker (approved):
LGTM!

PR #2609, opened 2021-07-13
refactor(Tree): Tree and TreeItem Indentation

ElliotTheHuman:
- Fixed border radius behavior and focus visible styling
- Added Asides to NavList and FieldPicker stories to mimic actual use cases
- Removed right padding from tree items and trees
- Deprecated branchFontWeight
- Deprecated assumeIconAlignment and forceLabelPadding props and removed stories
- Created NavTree and NavTreeItem components
- Resolved bug where tree item props landed on accordion DOM elements
- Deprecated TreeStyle

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


lukelooker (commented):
Ok - quick review of snapshots shows a few issues:

- NavLIst/Mixed Navigation
  - `NavTree` - Text in "selected" `Tree` is no longer using key color (see `NavTree`)
  - `NavTreeItem` - text and icon no longer using key color  (see `My Awesome Item`)

- Border Radius appears to have lost radius and background (maybe just background is making radius invisible since background/foreground match?)

- `Tree: Disabled & Selected` - indicator is now black/tex5 instead of the previous lighter color (not exactly sure of value)

image

- `NavList` and tree-truncate examples are both making me wonder if we want to apply a small amount of padding to the right side of the truncate to prevent it from getting too close to the edge with ellipsis:

image

packages/components/src/NavList/NavList.story.tsx
```diff
@@ -78,46 +78,36 @@ export const MixedNavigation = () => (
...
-        label="Tree w icon-free TreeItems and long title"
+        label="NavTree w icon-free NavTreeItems and long title"
         defaultOpen
       >
-        <Tree label="Folders" defaultOpen assumeIconAlignment>
-          <TreeItem truncate>
+        <NavTree label="Folders" defaultOpen>
+          <NavTreeItem description="description" detail="detail" selected>
+            My Awesome Item
+          </NavTreeItem>
```
ElliotTheHuman, line 102: I don't believe any child tree items in DiscoNav's example have icons so I removed the icon on the child tree item in our example. 

If there are, they we may need to introduce a new prop to prevent some child tree items from misaligning themselves relative to our normal alignment pattern.

ElliotTheHuman:
Alright, I _think_ I've fixed all the issues that popped up via snapshot discrepancies.

`NavList / Mixed Navigation`
- Fixed "selected" text color in a selected `NavTree` not being applied
- Fixed text and icon no longer using key color in `NavTreeItem`

`Tree: Disabled & Selected`
- Fixed indicator not being that lighter greyish color when disabled

`Border Radius`
- Border radius is working afaict, but the `dividers` prop wasn't working which I think resulted in the snapshot discrepancy. I went ahead and fixed that

lukelooker (commented):
Haven't made it all the way through this but wanted to give you feedback sooner than later since my schedule today is goofy. 

I pushed a PR trying to get as many of the bold text examples set back as before so I could see any intentional visual variances. 

Annoyingly some of the screenshots don't seem to be updating (re-running now to see if I can figure out where things went funny).

That said this is looking really good - most of my feedback is nits and/or asks for small additional explanations of some of the thornier pieces here.

lukelooker (approved):
Awesome refactor and A LOT easier to read through now - thanks for taking the time to do this as well as make sure snapshots were table to help make it feel stable :)

PR #2610, opened 2021-07-14
fix(InputChips): Undefined height prop should always default to 'auto'

mdodgelooker:
1. `InputChipsBase` sets `height="auto"` on its internal `InputText` but then spreads incoming props after that
2. Setting `height={undefined}` on `InputChips` (as `@looker/filter-components` `MultiStringInput` effectively does), will overwrite the above-mentioned `height="auto"` and instead invoke the `InputText` default of 36px.
3. If there are more values than fit in one line, the rest will be cut off – scrollable, but still a bad UX.
3. This PR fixes the issue by putting `height` in the props destructure section and sets the default of `'auto'` there.

#### Before fix, you only see the first line of values:
image

#### After fix, height grows to fit contents:
image


laraharrow (approved):
LGTM🚢

PR #2611, opened 2021-07-15
fix(Panel): Avoid focusing elements underneath

mdodgelooker:
- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/Panel/Panel.story.tsx
```diff
@@ -44,6 +44,7 @@ export default {
 const Template: Story<PanelProps> = (args) => (
   <Page hasAside>
     <Aside width="12rem">
+      <Button>Before</Button>
```
mdodgelooker, line 47: Adding a couple focusable elements to test the flow.

packages/components/src/Panel/Panels.tsx
```diff
@@ -24,9 +24,107 @@
 
  */
 
+import { CompatibleHTMLProps } from '@looker/design-tokens'
```
mdodgelooker, line 27: The basic gist of the changes in this file are:
1. When focus lands inside `Panels`, check if it's in the panel, if there is one (or the top panel if there are multiple).
2. If it's not, we can't let focus stay there, because nothing else inside `Panels` would be visible. So we bump focus to:
    a. The (top) panel itself, unless focus was previously inside that panel, in which case move it to:
    b. The first tabbable element before or after `Panels`, depending on where it landed relative to the panel

packages/components/src/Panel/usePanel.tsx
```diff
@@ -78,12 +78,19 @@ export const usePanel = ({
     onClose && onClose()
   }
 
+  const setInitialFocus = useCallback((element: HTMLDivElement | null) => {
+    element?.focus({ preventScroll: true })
+  }, [])
```
mdodgelooker, line 83: This is very unsophisticated. At the very least we should move focus from the button that opened the panel onto something inside the panel, but probably not to the close icon button – so this just focuses the panel itself.

To improve the UX, we could implement something similar to what `useFocusTrap` does, where it picks what to focus by priority – first any input, then any footer button, then the 1st focusable element, then the surface itself.

packages/components/src/Panel/usePanel.tsx
```diff
@@ -78,12 +78,19 @@ export const usePanel = ({
...
+
   const panel = renderDOM && (
     <PanelWindow>
       <PanelSurface
         aria-busy={busy ? true : undefined}
         className={className}
         direction={direction}
+        data-panel
+        tabIndex={-1}
```
mdodgelooker, line 92: Without this, there's a bug when used inside a `List` where if you click somewhere inside the panel but not specifically on a focusable element, focus lands on the `List` and the focus handler from `useKeyboardNav` battles with that in `Panels`, creating an infinite loop.

Also, it provides a "fallback" focusable element other than the close button (see comment above).

packages/components/src/Panel/usePanel.tsx
```diff
@@ -78,12 +78,19 @@ export const usePanel = ({
...
+    element?.focus({ preventScroll: true })
+  }, [])
+
   const panel = renderDOM && (
     <PanelWindow>
       <PanelSurface
         aria-busy={busy ? true : undefined}
         className={className}
         direction={direction}
+        data-panel
```
mdodgelooker, line 91: This is here so that `Panels` can find all its children panels. My first thought was to use an aria role, but the closest I could find was `tabpanel`, which isn't quite right... is it?

mdodgelooker (commented):
@lukelooker here's my idea for the focus issue with `Panel`. Let me know if you think it's on the right track.

packages/components/src/Panel/Panels.tsx
```diff
@@ -24,9 +24,114 @@
...
+        {...props}
+        ref={ref}
+      />
+    )
+  }
+)
+
+PanelsInternal.displayName = 'PanelsInternal'
+
+export const Panels = styled(PanelsInternal)`
```
lukelooker, line 134: Minor nit - I've been moving to wrapping the `*Internal` components inside the `styled` block unless there's a case where the `styled` needs to get props "back" from the React component:

```export const Panels = styled(props) => {
  /* do react stuff here */
})`
  height: 100%;
  position: relative;
  width: 100%;
`

**Review by lukelooker — CHANGES_REQUESTED** [review]
Sorry for the delay getting feedback to you on this! Anyway...

It's funny how hard it is to keep tabs off of things :)

I really wish there was a way to make this "just work" via DOM and/or removing elements but I still haven't come up with anything that works well / doesn't get more complicated than what you've got here. 

Overall this seems the best bet to move forward – the implementation makes sense to me and reads clearly. 

- [ ] I think adding inline comments similar to your PR comment regarding changes would be good to make sure future developers have good context on what's going on and the intent.
- [ ] Needs some test coverage to validate that it suppresses things as expected
- [ ] Does this suppress screen-readers successfully too – I know they can be uniquely aggressive about finding DOM targets at times but sometimes don't seem to trigger "real" focus events)

**Comment by mdodgelooker** [comment]
@lukelooker thanks for the feedback! Navigating via screenreader is definitely still an issue: VoiceOver ctrl+option+left-arrow can get you into the hidden content underneath an open panel. I'm hoping that some combination of aria & role can help with that. 🤞 

Do you think that this PR should tackle "smart" initial focus on opening a `Panel`, similar to what focus trap now does? Or just focus the panel surface for now and add improvements in a future PR?

**Comment by mdodgelooker** [comment]
Actually I'm gonna explore putting `visibility: hidden' on `Panels` and then `visibility: visible` on the active `Panel`. Oddly that works (visually) – we'll see what it does for tab nav and screenreaders.

**Comment by mdodgelooker** [comment]
@lukelooker I think the `visibility: hidden` on `Panels` and overridden with `visibility: visible` on the topmost `Panel` will do the trick. The only challenge is making sure that _only_ the topmost `Panel` has `visibility: visible`, which is a very similar problem to focus trap and scroll lock. So I'm using the `TrapStack` infrastructure here, which sounds a little funny, because we're not really trapping anything. I could rename it to something even more generic...

Still working on the test.

File: packages/components/src/Panel/Panels.tsx [hunk 7]
```diff
@@ -24,9 +24,51 @@
...
+  }
+}
+
+export type PanelsProps = CompatibleHTMLProps<HTMLDivElement>
+
+export const Panels = styled((props: PanelsProps) => {
+  // TrapStackProvider calculates the topmost active Panel and activates
+  // the visibility styles accordingly
+  return (
+    <TrapStackProvider activate={activatePanel} context={PanelsContext}>
```
lukelooker, line 67: This does feel _way_ better :)

I don't mind `TrapStack` here since we're "trapping" the visibility using a stack :)

packages/components/src/Panel/usePanel.tsx
lukelooker: Yeah  reading up on `tabpanel` it doesn't seem quite right for this. I'm cool with your solve here.

packages/components/src/Panel/usePanel.tsx
mdodgelooker: The first paragraph is no longer relevant after switching to the `visibility` strategy, but the 2nd still is.

packages/components/src/utils/useAnimationState.ts
```diff
@@ -40,7 +40,7 @@ interface UseAnimationStateReturn {
   /**
    * className will transition from 'entering` => `entered` => `exiting` => `exited`
    */
-  className: string
+  className: 'entering' | 'entered' | 'exiting' | 'exited'
```
mdodgelooker, line 43: Felt like this needed a bit more type safety.

mdodgelooker:
@lukelooker ready for re-review when you've got time. 🙂

lukelooker (approved):
Works great!

PR #2619, opened 2021-07-19
fix(Filter Packages): main and module fields

mdodgelooker:
Fix main & module fields so they'll work for publishing.


lukelooker (dismissed):
LGTM

lukelooker (approved):

PR #2620, opened 2021-07-20
chore: Fix snapshot infrastructure and specific stories

ElliotTheHuman:
Various snapshots have become out-of-sync with their actual storybook code (possibly due to our current discrepancy threshold not being strict out). 

This PR forcibly updates those out-of-sync snapshots to match their current storybook code.

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


storybook/src/stories.shots.ts
```diff
@@ -82,8 +82,8 @@ const imageSnapshots = () => {
         return {
           customSnapshotIdentifier: story,
           customSnapshotsDir: path.join('../snapshots', ...kind.split('/')),
-          failureThreshold: 0.005,
-          failureThresholdType: 'percent',
+          failureThreshold: 1,
+          failureThresholdType: 'pixel',
```
ElliotTheHuman, line 86: nit: Technically speaking, `failureThresholdType` defaults to 'pixel' so we don't need this here per:
https://github.com/americanexpress/jest-image-snapshot

But I think being explicit is probably best so I'm cool with keeping this line.

packages/components/src/Button/MultiFunctionButton.story.tsx
```diff
@@ -68,9 +80,13 @@ export const ComponentSize = () => {
       alternate={<Button size="large">This is a very large button</Button>}
       swap={change}
     >
-      <ButtonOutline onClick={handleSwap} iconBefore={<Add />}>
+      <CustomButtonOutline
+        onClick={handleSwap}
+        width={400}
```
ElliotTheHuman, line 85: For some reason, setting the width on the starting button doesn't change the width of that starting button / the `MultiFunctionButton`. Basically, this `width={400}` is ignored and the `MultiFunctionButton` is whatever the size of the alternate is.

Not a major blocker for this PR, but I went ahead and created a bug ticket and left it unassigned:
https://buganizer.corp.google.com/issues/194234951

For now, we can probably just set the width on the "This is a very large button" alternate to get what we want here.

packages/components/src/Layout/Space/Space.story.tsx
```diff
@@ -80,49 +80,3 @@ export const SpaceCrush = () => (
...
-      <button>boo</button>
-      <button>boo</button>
-      <button>boo</button>
-      <button>boo</button>
-      <button>boo</button>
-      <button>boo</button>
-      <button>boo</button>
-    </Space>
-  </>
-)
```
ElliotTheHuman, line 128: Went ahead and deleted the snapshot for this.

packages/components/src/Button/MultiFunctionButton.story.tsx
ElliotTheHuman: Went ahead and moved `width={400}` to the alternate "This is a very large button" button instead to get the desired effect for now.

storybook/src/stories.shots.ts
lukelooker: Yeah - that was my thought as well :)

lukelooker (dismissed):
LGTM

lukelooker (approved):
Huzzah!

PR #2626, opened 2021-07-21
fix(Popover): voiceOver error

laraharrow:
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/Overlay/OverlaySurface.tsx
```diff
@@ -75,7 +75,10 @@ const OverlaySurfaceLayout = forwardRef(
 
 OverlaySurfaceLayout.displayName = 'OverlaySurfaceLayout'
 
-export const OverlaySurface = styled(OverlaySurfaceLayout)`
+export const OverlaySurface = styled(OverlaySurfaceLayout).attrs(() => ({
+  'aria-modal': true,
+  role: 'dialog',
+}))`
```
mdodgelooker, line 81: This is good but it won't work unless `props` are spread onto the div – right now only some specific props are passed in. You could change lines 49 & 50 above to:
```
({ children, eventHandlers, placement, ...props }: OverlaySurfaceProps, forwardedRef: Ref<HTMLDivElement>) => {
```
and then change lines 62 & 63 to:
```
{...omitStyledProps(props)}
```
which will make it a little closer to `SurfaceBase` that's used in `Dialog`.

mdodgelooker (changes requested):
Good start! Just need to adjust how props are handled.

packages/components/src/Overlay/OverlaySurface.tsx
laraharrow: 🤦🏽‍♀️ that makes total sense, I can't believe I didn't realize that!

PR #2627, opened 2021-07-21
fix(DataTable): Stop explicitly removing scrollbar from DataTable overflow behavior

lukelooker:
Additional context available in: https://b.corp.google.com/issues/194313414

- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/DataTable/Table.tsx
+++ b/packages/components/src/DataTable/Table.tsx
@@ -226,8 +226,4 @@ const InterimState = styled.div`
 
 export const TableScroll = styled.div`
   overflow-x: auto;
-  scrollbar-width: none; /* Firefox */
-  ::-webkit-scrollbar {
-    display: none; /* Safari and Chrome */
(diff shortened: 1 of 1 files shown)


lukelooker:
@jhardy This PR removes a little bit of CSS that we were using to suppress the visibility of scroll bars on `DataTable` - admittedly I prefer the hidden version but I can also see how it may impact usability / accessibility in its current form.

Thoughts?

ElliotTheHuman (approved):
LGTM if Jared's cool with the look of `DataTable` with horizontal scroll bar

jhardy (approved):
This LGTM. I do agree the scrolls bars are not as nice, but I agree it is an accessibility and usability win.  We can probably look at leveraging some of the `::--webkit-scrollbar` selectors and styling it a bit better.

PR #2631, opened 2021-07-22
fix(IconButton): IconButton label should support same typing as Tooltip content

lukelooker:
IconButton's label element is used to propagate a tooltip applied to the `IconButton`. With that in mind they should share the same typing

Bug: https://buganizer.corp.google.com/issues/194341079

- [ ] 👾 Browsers tested
  - [x] Chrome
  - [ ] Firefox


mdodgelooker (approved):
LGTM!

PR #2635, opened 2021-07-22
fix(Popover): Screen reader accessibility

laraharrow:
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Popover/Popover.test.tsx
+++ b/packages/components/src/Popover/Popover.test.tsx
@@ -61,13 +61,17 @@ describe('Popover', () => {
 
     const openPopover = screen.getByText('Open')
     fireEvent.click(openPopover)
-    const header = screen.queryByText('Header text')
-    expect(header).toBeInTheDocument()
-    expect(header?.closest('div')).toHaveAttribute(
+
+    expect(screen.queryByText('Header text')).toBeInTheDocument()
+    expect(screen.getByRole('dialog')).toBeInTheDocument()
+    expect(screen.getByRole('dialog')).toHaveAttribute(
       'aria-labelledby',
       'a11y-heading'
     )
-    expect(header).toHaveAttribute('id', 'a11y-heading')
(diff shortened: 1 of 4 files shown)


packages/components/src/Popover/Popover.test.tsx
```diff
@@ -61,17 +61,13 @@ describe('Popover', () => {
...
+    expect(header).toBeInTheDocument()
+    expect(header?.closest('div')).toHaveAttribute(
       'aria-labelledby',
       'a11y-heading'
     )
-    expect(screen.getByText('Header text')).toHaveAttribute(
-      'id',
-      'a11y-heading'
-    )
+    expect(header).toHaveAttribute('id', 'a11y-heading')
```
mdodgelooker, line 70: I think `aria-labelledby` should probably move onto `SurfaceComponent` in `usePopover.tsx` so it can stay with `role="dialog"`. Then this test can be reverted.

But the `Select` test is having a similar issue and does need to be updated. The scroll event now should be fired on the div right inside the role="dialog".

mdodgelooker (changes requested):

packages/components/src/Popover/usePopover.tsx
```diff
@@ -251,21 +251,22 @@ export const usePopover = ({
...
           <Flex
             alignItems="flex-start"
-            aria-labelledby={`${id}-heading`}
             borderRadius="inherit"
             flexDirection="column"
             id={id}
             maxHeight={`calc(${verticalSpace - 10}px - 1rem)`}
             overflowY="auto"
             ref={contentContainerRef}
-            role="dialog"
```
lukelooker, line 268: This is the root-cause of the test breakage in `Select`. Why does this need to get moved?

mdodgelooker (approved):
Awesome! 🚢

PR #2636, opened 2021-07-22
test(Tree): Added story and test to cover Tree with detail accessory enabled

ElliotTheHuman:

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Tree/TreeCollection.test.tsx
+++ b/packages/components/src/Tree/TreeCollection.test.tsx
@@ -230,6 +230,43 @@ describe('TreeCollection', () => {
       expect(gouda).toHaveFocus()
     })
 
+    test('pressing arrow down from a detail element with accessory enabled moves user to next item', () => {
+      renderWithTheme(
+        <>
+          <button>My Button</button>
+          <TreeCollection>
+            <Tree
+              defaultOpen
+              label="Cheeses"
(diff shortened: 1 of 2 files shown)


packages/components/src/Tree/TreeCollection.test.tsx
```diff
@@ -230,6 +230,43 @@ describe('TreeCollection', () => {
...
+              defaultOpen
+              label="Cheeses"
+              detail={{
+                content: <button>Accessory Button</button>,
+                options: { accessory: true },
+              }}
+            >
+              <TreeItem>Cheddar</TreeItem>
+            </Tree>
+          </TreeCollection>
```
ElliotTheHuman, line 248: I wanted to reuse the `Accessory` story we have in the `Tree.story` file, but the basic template is a little wacky (i.e. has two `Tree`s with the label "Orders" to mimic what often happens in Looker with the "Orders" explore and the "Orders" view).

Opted to just grow a fresh `Tree` instead.

lukelooker (approved):
I think you should disable the image snapshot and then this will be good to go!

lukelooker (approved):

PR #2638, opened 2021-07-22
fix(DataTable): Make first column sortable again

lukelooker:


packages/components/src/DataTable/Header/DataTableHeaderCell.tsx
```diff
@@ -48,7 +56,7 @@ const DataTableHeaderCellLayout = forwardRef(
       className,
       title,
       titleIcon,
-      id,
+      columnId,
       size,
       sortDirection,
       type,
```
ElliotTheHuman, line 62: Vanilla `id` currently isn't being passed down to the DOM element so we'll need to add that prop assignment to the `FocusableCell` element below (I _think_)

ElliotTheHuman: Fixing this, though, will mean that the first header cell will receive the `{ id: `rowheader-${id}`, role: 'rowheader' }` from the `cloneElement` call in `DataTableRow.tsx`.

I think it makes sense to add an additional condition on `DataTableRow.tsx > Line 82` using the `isHeaderRow` prop, like this:
Screen Shot 2021-07-22 at 3 14 00 PM

ElliotTheHuman: Alternatively, we can just remove vanilla `id` as a prop on `DataTableHeaderCell` but I'm unsure of whether that's a problem or not 🤷 . Considering it isn't being passed down to the DOM element on `main` this seems like a reasonable option instead of the changes listed above.

ElliotTheHuman (changes requested):
There's a minor "bug" where vanilla `id` doesn't get passed down to `DataTableHeaderCell`'s actual DOM element. Whether we want to do that (or not) though is up to you!

packages/components/src/DataTable/Header/DataTableHeaderCell.tsx
lukelooker: It's actually completely fine to just remove support for `id` altogether (there's no need to attach the "Select All" checkbox to a cell as it leverages `aria-label` for it's label.

dbchristopher (dismissed):
This looks good! I was a little bit thrown off at first because I saw the `canSort` attribute, but then noticed other stories toggling this behavior using the `sort` property:

Screen Shot 2021-07-23 at 9 57 05 AM

After a bit of digging I saw that `sort` was just specific to the Template component used for those stories. Not sure if we care to have those storybook controls match up with the real prop names exactly or not 🤷‍♂️  If so it would be a quick update!

lukelooker:
@dbchristopher Renamed `sort` to `canSort` in story example prop.

dbchristopher (approved):

PR #2641, opened 2021-07-23
feat(ComponentsProvider): ThemeCustomizations should support overloading fontSources

lukelooker:
Currently `ThemeCustomizations` supports overloading `fontFamilies` but not `fontSources`. Both would be needed to successfully use a font-face that's not available on locally.


dbchristopher (approved):
Not sure the best way to test this out myself, so I'll trust that it works!

lukelooker:
@dbchristopher Added test coverage for this. Sorry, should've done so out-of-the-gate.

dbchristopher (approved):

PR #2652, opened 2021-07-26
fix(Tree): Controlled isOpen

mdodgelooker:
Check `isOpen` for being defined rather than truthy to get correct accordion props.

Add story and update controlled test to catch this failure.


lukelooker (approved):
🤦  Looks great! Thanks for fixing that so quick! :)

PR #2653, opened 2021-07-28
fix: Export a few private items to allow downstream consumers to leverage without breaking tree-shaking

lukelooker:


jhardy (approved):
LGTM

PR #2661, opened 2021-07-30
feat(design-tokens): Create new unit-based spacing tokens to represent new design system values

jhardy:
This PR adds a new `spacing` option to theme, these represent our new spacing values we want to use on the design system. 

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/design-tokens/src/system/space.ts
```diff
@@ -51,6 +51,28 @@ export type SpacingSizes =
...
+  | 'size7'
+  | 'size8'
+  | 'size9'
+  | 'size10'
+  | 'size11'
+  | 'size12'
+  | 'size13'
+  | 'size14'
+  | 'size15'
+  | 'size16'
```
jhardy, line 72: Interested in your opinion here @lukelooker, I've gone back and forth on naming a few times, but I feel like this is something memorable as well as not too different from our t-shirt sizing concept.

packages/design-tokens/src/theme/theme.ts
```diff
@@ -55,6 +56,7 @@ import {
   sizes,
   shadows,
   space,
+  spacing,
```
jhardy, line 59: adding the new tokens under `spacing` instead of `space` so we can deprecate the old tokens.

packages/design-tokens/src/system/space.ts
lukelooker: For some reason the idea of introducing `size` here raises my hackles a bit – I think perhaps the collision with the idea of `iconSize` and such?

What do you think of referencing `u` (for unit)? I guess I think the predicate here should be as diminutive as possible to encourage scannability of the difference value (`0.5`, `1`, `2`, etc...)

```
`u0.5`,
`u1`,
`u2`,
`u3`,
```

```
<Paragraph mb="size4" mt="size2">...
```
vs.
```
<Paragraph mb="u4" mt="u2">...
```

lukelooker (changes requested):
Arg! I forgot to "Send" my review

lukelooker (approved):
LGTM

PR #2676, opened 2021-08-02
feat(NavTree): Added indicator toggle only behavior

ElliotTheHuman:
## Changes
- Isolated accordion toggle behavior in `NavTree` to only the indicator
- Rearranged `NavTree` structure to avoid indicator from being included within label container (i.e. indicator should NOT trigger linking / onClick behavior of the label)
- Unrelated: Added 8px of spacing to the `Nav` component suite "indent" (note: `NavTree` gets its additional 8px by adding padding to the indicator for a larger click target)
  - This was added to satisfy a feature request regarding larger indicator click targets for top-level `NavTree`s 

## Follow-up PR
- Fix keyboard nav by adding custom `getNextFocus` to `NavList`

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/NavTree/NavTree.tsx
```diff
@@ -24,18 +24,288 @@
...
+
+  // These additional props on indicator produce the "indicator toggle only" behavior
+  const renderedIndicator = cloneElement(indicator, {
+    onClick: onIndicatorClick,
+    onKeyUp: onIndicatorKeyUp,
+    style: {
+      paddingLeft: INDICATOR_SPACER,
+    },
+    tabIndex: -1,
+  })
```
ElliotTheHuman, line 238: This is how we get the toggle behavior (and `tabIndex = -1`) onto the `AccordionIndicator` div since that isn't exposed to `useAccordion2` callers. 

Wondering if we should just find a way to do this within `useAccordion2` itself via an `indicatorToggleOnly` prop, but the `cloneElement` seemed like the least complicated solution.

packages/components/src/NavList/NavList.tsx
```diff
@@ -68,7 +69,10 @@ export const NavList = styled(List).attrs(({ color = 'key' }) => ({ color }))`
 
   & > ${ListItem} {
     ${ListItemContent} {
-      padding-left: ${({ theme }) => `${theme.sizes.medium}`};
+      padding-left: ${({ theme }) => {
+        const iconGutterSize = theme.sizes.medium
+        return `calc(${iconGutterSize} + ${INDICATOR_SPACER})`
+      }};
```
ElliotTheHuman, line 75: I know I could just use `theme.sizes.large` or a combination of `theme.space.X`s to get the `2rem` I need, but I figure since `INDICATOR_SPACER` might need to be changed more than once based on DiscoNav feedback, it's better to just tie all our overrides to it.

lukelooker (changes requested):
Can we split this into two PRs - with Jared OOO this week I'd prefer to ship the "label toggle" behavior sooner and then have the spacing / sizing conversation as a separate,  narrower conversation.

ElliotTheHuman:
Indicator spacing changes reverted in fea9b0a

packages/components/src/NavTree/NavTree.test.tsx
```diff
@@ -39,30 +39,34 @@ describe('NavTree', () => {
...
 
-    const treeLabel = screen.getByText('Parent Tree with Icon')
+    const indicator = screen.getAllByTestId('accordion-indicator')[0]
     screen.getByText('Cheddar')
-    fireEvent.click(treeLabel)
+    fireEvent.click(indicator)
     expect(screen.queryByText('Cheddar')).not.toBeInTheDocument()
+  })
+
+  test('Does not change accordion open state on label click', () => {
```
lukelooker, line 51: I think the implementation (and the test title/behavior by way of that) is slightly incorrect – the `ParentIcon` story doesn't have a link/onClick and therefore clicking the label _should_ open and close the underlying accordion.

lukelooker: I think a test for the `Link` example with a very similar test should be added however.

packages/components/src/NavTree/NavTree.tsx
```diff
@@ -24,18 +24,302 @@
...
+                {/**
+                 * @TODO: Delete labelBackgroundOnly behavior once FieldItem component is completed
+                 */}
+                {hasLabelBackgroundOnly ? (
+                  <TreeItemLabel {...statefulProps}>
+                    {disclosureLabel}
+                  </TreeItemLabel>
+                ) : (
+                  disclosureLabel
+                )}
```
lukelooker, line 296: `NavTree` probably doesn't need this variant?

packages/components/src/NavTree/NavTree.tsx
```diff
@@ -24,18 +24,302 @@
...
+ */
+const NavTreeLayout = ({
+  assumeIconAlignment,
+  border: propsBorder,
+  children,
+  className,
+  dividers,
+  forceLabelPadding,
+  isOpen: propsIsOpen,
+  itemRole = 'none', // By default, Tree's content container should be a 'div'
```
lukelooker, line 78: Can we trim `NavTree` back to a more minimal implementation removing things like `itemRole`?

lukelooker (commented):
Early thoughts commented on here - would like to discuss how we can narrow `NavTree` some.

ElliotTheHuman:
Next steps:
- Make label toggle behavior conditional based on if an `href` is passed in or not
- Tear out unused pieces like `labelBackgroundOnly` and `itemRole` weirdness
- Fix keyboard nav

packages/components/src/NavTree/NavTree.tsx
```diff
@@ -23,19 +23,293 @@
...
+     */
+    if (
+      restProps &&
+      (treeItemInnerPropKeys as ReadonlyArray<string>).includes(propKey)
+    ) {
+      treeItemInnerProps[propKey] = propValue
+    } else {
+      accordionInnerProps[propKey] = propValue
+    }
+  })
```
lukelooker, line 104: Dupe from `Tree`

packages/components/src/NavTree/NavTree.tsx
```diff
@@ -23,19 +23,293 @@
...
+
+  const {
+    density: propsDensity,
```
lukelooker, line 107: Not sure we should support `density` on `NavTree`

packages/components/src/NavTree/NavTree.tsx
```diff
@@ -23,19 +23,293 @@
...
+  const handleWrapperBlur = (event: FocusEvent<HTMLElement>) => {
+    const nextFocusTarget = getNextFocusTarget(event)
+
+    if (
+      nextFocusTarget &&
+      !event.currentTarget.contains(nextFocusTarget as Node)
+    ) {
+      setHovered(false)
+    }
+  }
```
lukelooker, line 136: Extract to util

lukelooker (dismissed):

lukelooker (approved):
💥  Dope, ship it.

PR #2683, opened 2021-08-03
test(DataTable): Utilities Test Coverage

dbchristopher:
Increases test coverage for DataTable utilities and child components. This is accomplished by both adding tests but also refactoring instances where branches of code were functionally unreachable.

https://buganizer.corp.google.com/issues/194403850

## Screenshot

Group

## Followup

I'll create a new branch for DataTable/getNextFocus tests, which is a separate buganizer ticket and will be a decent chunk of work on its own.

### DIFF
--- a/packages/components/src/DataTable/Item/DataTableItem.tsx
+++ b/packages/components/src/DataTable/Item/DataTableItem.tsx
@@ -84,13 +84,6 @@ const DataTableItemLayout: FC<DataTableItemProps> = (props) => {
 
   const handleClick = disabled ? undefined : onClick || undefined
 
-  const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
-    const isEventFromChild = event.currentTarget !== event.target
-    if (event.keyCode === 13 && !isEventFromChild) {
-      event.currentTarget.click()
-    }
-  }
-
   const ItemActions = (actionPrimary || actions) && (
     <ItemTargetGroup>
       {actionPrimary && <ItemTarget>{actionPrimary}</ItemTarget>}
@@ -121,7 +114,6 @@ const DataTableItemLayout: FC<DataTableItemProps> = (props) => {
       id={id}
       onChange={onChange}
       onClick={handleClick}
-      onKeyDown={handleKeyDown}
       ref={ref}
       secondary={ItemActions}
       tabIndex={0}
--- a/packages/components/src/DataTable/Item/DataTableRow.tsx
+++ b/packages/components/src/DataTable/Item/DataTableRow.tsx
@@ -79,16 +79,14 @@ const DataTableRowLayout = forwardRef(
     const sizedChildren = Children.map(children, (child, index) => {
       if (columnsDisplayState && !columnsDisplayState[index]) {
         return null
-      } else if (isValidElement(child)) {
-        const size = getColumnSize(index)
-        const cellProps =
-          index === 0
-            ? { id: `rowheader-${id}`, role: 'rowheader', size }
-            : { size }
-        return cloneElement(child, cellProps)
-      } else {
-        return child
       }
+
+      const size = getColumnSize(index)
+      const cellProps =
+        index === 0
+          ? { id: `rowheader-${id}`, role: 'rowheader', size }
+          : { size }
+      return isValidElement(child) && cloneElement(child, cellProps)
     })
 
     const handleOnClick = (event: React.MouseEvent<HTMLElement>) => {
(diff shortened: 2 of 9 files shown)


packages/components/src/DataTable/Item/DataTableItem.tsx
```diff
@@ -84,13 +84,6 @@ const DataTableItemLayout: FC<DataTableItemProps> = (props) => {
 
   const handleClick = disabled ? undefined : onClick || undefined
 
-  const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
-    const isEventFromChild = event.currentTarget !== event.target
-    if (event.keyCode === 13 && !isEventFromChild) {
-      event.currentTarget.click()
-    }
```
dbchristopher, line 91: I could never identify an edge case where the event originated from somewhere other than the child. I assume this handler was deprecated and this line was unreachable, so I deleted it.

packages/components/src/DataTable/Item/DataTableRow.tsx
```diff
@@ -79,16 +79,14 @@ const DataTableRowLayout = forwardRef(
...
         return null
-      } else if (isValidElement(child)) {
-        const size = getColumnSize(index)
-        const cellProps =
-          index === 0
-            ? { id: `rowheader-${id}`, role: 'rowheader', size }
-            : { size }
-        return cloneElement(child, cellProps)
-      } else {
-        return child
```
dbchristopher, line 90: refactored this to negate the need for that final `else` branch which seemed to be unreachable (everything would pass `isValidElement` check)

packages/components/src/DataTable/utils/sort_utils.test.ts
```diff
@@ -48,172 +48,99 @@ describe('DataTable Sort Utils', () => {
...
-            },
-            Object {
-              "id": 2,
-              "name": "Kobe",
-            },
-          ],
-        },
-      ]
-    `
-    )
```
dbchristopher, line 217: I broke these snapshots apart to be more easily parsed by test case. See [`/sort_utils.test.ts.snap`](https://github.com/looker-open-source/components/pull/2683/files#diff-972f44e1dae0badff5d7f520b579a56fb63f38a8648f392a589103aa62c190c4) for final output

packages/components/src/DataTable/utils/sort_utils.test.ts
```diff
@@ -48,172 +48,99 @@ describe('DataTable Sort Utils', () => {
...
+    [
+      'Sort Name Ascending',
       {
-        id: 'id',
+        id: 'name',
         sortDirection: 'desc',
       },
+    ],
+    [
+      'Sort Date Ascending',
```
dbchristopher, line 126: Added test case to verify the date sorting functionality

packages/components/src/DataTable/utils/useSelectManager.tsx
```diff
@@ -41,7 +41,7 @@ export const useSelectManager = (
     setSelections(
       selections.includes(selectionId)
         ? selections.filter(
-            (itemId) => itemId !== selectionId && possibilities.includes(itemId)
+            (itemId) => possibilities.includes(itemId) && itemId !== selectionId
```
dbchristopher, line 44: Flipped the order of these checks as it would always short circuit on the `itemId !== selectionId` check and therefor never actually check if the `itemId` was included in the `possibilities` array

packages/components/src/DataTable/Item/DataTableItem.tsx
laraharrow: should the `event.keyCode === 13` part stay as it is checking if the enter key is clicked?

laraharrow (approved):
LGTM 🚢

packages/components/src/DataTable/Item/DataTableItem.tsx
mdodgelooker: I think you're right @dbchristopher – you can't focus the row itself in order to "click via enter". Maybe this was from a previous iteration where the row was focusable.

mdodgelooker (approved):
Nice work @dbchristopher !

lukelooker (approved):
Ditto on other's reviews. LGTM :)

packages/components/src/DataTable/Item/DataTableItem.tsx
dbchristopher: awesome thanks for weighing @mdodgelooker, I think that would definitely explain it!

PR #2684, opened 2021-08-03
test: Introduce @storybook/testing-react

lukelooker:
Adds `@storybook/testing-react` test dependency and leverages it where it's useful.

### DIFF
--- a/packages/filter-components/src/Filter/Filter.stories.tsx
+++ b/packages/filter-components/src/Filter/Filter.stories.tsx
@@ -25,9 +25,11 @@
  */
 import React, { useState } from 'react'
 import type { Story } from '@storybook/react/types-6-0'
-
+import { i18nInit } from '../utils/i18n'
 import type { FilterProps, FilterChangeProps } from './types/filter_props'
 import { Filter } from './Filter'
+
+i18nInit()
 
 const Template: Story<FilterProps> = (args) => {
   const [expression, setExpression] = useState(args.expression)
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -25,6 +25,8 @@
  */
 
 import { format } from 'util'
+import { setGlobalConfig } from '@storybook/testing-react'
+import * as globalStorybookConfig from './storybook/.storybook/preview'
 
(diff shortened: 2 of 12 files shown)


lukelooker:
@mdodgelooker Okay, this is _actually_ ready for review now.

packages/filter-components/src/Filter/Filter.stories.tsx
```diff
@@ -25,10 +25,12 @@
  */
 import React, { useState } from 'react'
 import type { Story } from '@storybook/react/types-6-0'
-
+import { i18nInit } from '../utils/i18n'
 import type { FilterProps, FilterChangeProps } from './types/filter_props'
 import { Filter } from './Filter'
 
+i18nInit()
```
lukelooker, line 32: This can't be included in `componentsDecorator` because some tests explicitly test `i18nInit` so we need to make sure it's not loaded "earlier" now that Jest `componentsDecorator` is part of the general Jest setup steps.

mdodgelooker: Trying to think of a cleaner solution for this...

dbchristopher (approved):
This is really nice! I love all the utility we are able to get from storybook. What a useful product. 

Will we need to use `composeStories` every time we render stories in a jest test? Or are there only certain instances where that is required?

lukelooker:
> Will we need to use `composeStories` every time we render stories in a jest test? Or are there only certain instances where that is required?

You don't _have_ to but I think it's probably going to be a best-practice in most cases. I suspect we'll find cases where we just want to pull in the Story component and then mess around with args within the Jest test as we do more cross-cutting work but YMMV.

packages/filter-components/src/Filter/Filter.stories.tsx
mdodgelooker: Can't figure it out. Hopefully this change will be ok on the core product side – I created a ticket to update: https://b.corp.google.com/issues/196231259

mdodgelooker (approved):

PR #2689, opened 2021-08-03
test(Accordion): Add coverage

ElliotTheHuman:
- Added a test to check `console.warn` when `Accordion` is only partially controlled (i.e. only receives one of two controlled open state props)
- Added a test to check legacy composition

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Accordion/Accordion.test.tsx
+++ b/packages/components/src/Accordion/Accordion.test.tsx
@@ -133,13 +133,6 @@ describe('Accordion', () => {
   })
 
   describe('warnings', () => {
-    beforeEach(() => {
-      global.console = {
-        ...globalConsole,
-        warn: jest.fn(),
-      }
-    })
-
     test('warns if isOpen is provided without toggleOpen prop', () => {
       global.console = {
         ...globalConsole,
@@ -154,6 +147,7 @@ describe('Accordion', () => {
 
       // eslint-disable-next-line no-console
       expect(console.warn).toHaveBeenCalled()
+      global.console = globalConsole
     })
   })
 
--- a/packages/components/src/Accordion/Accordion.test.tsx
+++ b/packages/components/src/Accordion/Accordion.test.tsx
@@ -140,6 +140,10 @@ describe('Accordion', () => {
       }
     })
 
+    afterEach(() => {
+      global.console = globalConsole
+    })
+
     test('warns if isOpen is provided without toggleOpen prop', () => {
       renderWithTheme(
         <Accordion isOpen={true} content="My Accordion Content">
@@ -149,7 +153,6 @@ describe('Accordion', () => {
 
       // eslint-disable-next-line no-console
       expect(console.warn).toHaveBeenCalled()
-      global.console = globalConsole
     })
 
     test('warns if children is a falsy value', () => {
@@ -159,7 +162,6 @@ describe('Accordion', () => {
 
       // eslint-disable-next-line no-console
       expect(console.warn).toHaveBeenCalled()
-      global.console = globalConsole
     })
   })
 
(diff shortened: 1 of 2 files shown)


packages/components/src/Accordion/Accordion.test.tsx
```diff
@@ -129,4 +131,43 @@ describe('Accordion', () => {
...
     expect(handleKeyUp).toHaveBeenCalled()
   })
+
+  describe('warnings', () => {
+    beforeEach(() => {
+      global.console = {
+        ...globalConsole,
+        warn: jest.fn(),
+      }
+    })
```
mdodgelooker, line 141: This is duplicate since you've got it in the test and since there's only 1 test in this describe, `beforeEach` is overkill.
```suggestion
```

packages/components/src/Accordion/Accordion.test.tsx
```diff
@@ -129,4 +131,43 @@ describe('Accordion', () => {
...
+      }
+
+      renderWithTheme(
+        <Accordion isOpen={true} content="My Accordion Content">
+          My Accordion Label
+        </Accordion>
+      )
+
+      // eslint-disable-next-line no-console
+      expect(console.warn).toHaveBeenCalled()
```
mdodgelooker, line 156: Reset `global.console`:
```suggestion
      expect(console.warn).toHaveBeenCalled()
      global.console = globalConsole
```

packages/components/src/Accordion/Accordion.tsx
```diff
@@ -87,13 +87,6 @@ const AccordionInternal: FC<AccordionProps> = ({
     )
   }
 
-  if (!label) {
-    // eslint-disable-next-line no-console
-    console.warn(
-      "<Accordion>'s child is falsy (i.e. undefined, null, false, etc). Please give <Accordion> a valid ReactNode child."
-    )
-  }
-
```
mdodgelooker, line 96: Is this no longer applicable? Or just not really worth doing since `Accordion` is deprecated?

mdodgelooker (changes requested):
Just a couple minor tweaks. 👍

packages/components/src/Accordion/Accordion.tsx
ElliotTheHuman: Ah meant to leave a comment here, my bad.

In the function parameters, we rename `children` to `label`. We do this because in `Accordion2` terms, the `children` prop should be the accordion's hidden/shown content, not the disclosure like in the old accordion world. This `label` variable should always be truthy since the `children` prop in `Accordion` is required, so I don't _believe_ we need this warn.

packages/components/src/Accordion/Accordion.test.tsx
ElliotTheHuman: 👍  Good catch, thank you.

packages/components/src/Accordion/Accordion.tsx
ElliotTheHuman: 🤦  Yes, you're totally right. In which case, keeping the warn makes sense, so I'll add another test case to the 'warn' test suite in Accordion.test.tsx

packages/components/src/Accordion/Accordion.test.tsx
ElliotTheHuman: Ended adding this `beforeEach` call back in because I added a new warn test. Instead, I removed the global.console assignment within the first warn test.

packages/components/src/Accordion/Accordion.test.tsx
ElliotTheHuman: Ended up just adding an `afterEach` call to handle this since I added a second warn test.

mdodgelooker (approved):
LGTM!

PR #2693, opened 2021-08-04
test(getNextFocus): keyboard navigation tests

dbchristopher:
Adds tests to verify arrow key navigation within DataTable.

## Screenshot
Screen_Shot_2021-08-04_at_11_28_03_AM

## Caveats

There are a few return statements which exist to satisfy typescript but are functionally unreachable, so I wasn't able to hit 100%.

1. Helper would return null if there wasn't a thead or tbody element in the dom:
Screen Shot 2021-08-04 at 11 24 11 AM

2. `getNextFocus` would return null if there were no tab stops on the page 
Screen Shot 2021-08-04 at 11 24 14 AM


packages/components/src/DataTable/DataTable.test.tsx
```diff
@@ -923,4 +925,140 @@ describe('DataTable', () => {
...
+    test('Tabbing from outside selects the first thead cell ', () => {
+      const { headerCells } = renderAndSelectCells()
+      const tabStops = getTabStops(document.body)
+
+      tabStops[0].focus()
+      fireEvent.keyDown(tabStops[0], {
+        code: 'Tab',
+        key: 'Tab',
+      })
+      // no change to focus:
```
ElliotTheHuman, line 1060: I don't _think_ this comment is relevant to this Tab key test since we land from `tabStops[0]` to `headerCells[0]` but I could be wrong.

ElliotTheHuman (changes requested):
Super small nits / questions. Replacing `lodash/chunk` with native js would be nice, but maybe that could be a follow-up PR and just a TODO comment for this PR.

packages/components/src/DataTable/DataTable.test.tsx
dbchristopher: ah, good catch! yeah that was totally just copy/paste cruft

packages/components/src/DataTable/DataTable.test.tsx
```diff
@@ -923,4 +925,140 @@ describe('DataTable', () => {
     renderWithTheme(<FilterDataTable />)
     expect(screen.getByText('Filter List')).toBeInTheDocument()
   })
+
+  describe('Keyboard Navigation', () => {
+    const renderAndSelectCells = () => {
+      renderWithTheme(dataTableWithGeneratedHeader)
+      const columns = Object.keys(data[0]).length + 1 // add one for the hidden unfocusable table cell prefacing each row
+      const tableCells = chunk(screen.getAllByRole('cell'), columns)
```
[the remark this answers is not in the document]
dbchristopher: Cool, yeah I just checked the You don't Need Lodash recommendation and they just suggested rewriting the helper from scratch:

```js
const chunk = (input, size) => {
  return input.reduce((arr, item, idx) => {
    return idx % size === 0
      ? [...arr, [item]]
      : [...arr.slice(0, -1), [...arr.slice(-1)[0], item]];
  }, []);
};
```

While we do prefer vanilla js when there is native support, I'm pretty sure we're still ok using lodash in instances that would require maintaining our own copies of the function. But I'll double check in the team channel on where we're at with best practices there!

ElliotTheHuman (approved):
LGTM!

PR #2694, opened 2021-08-04
chore: yarn deduplicate

dbchristopher:


ElliotTheHuman (approved):
Looks good 2 me.

lukelooker (changes requested):
Can we hold this for a moment? - I think I see a way to narrow things up.

lukelooker (approved):
Pulled out references to `0.6.0-alpha` since there's now a publicly installable package version and Yarn can rationalize itself that way.

PR #2697, opened 2021-08-05
fix(Tooltip): Reduce re-renders to improve performance

mdodgelooker:
Rearrange the use of hooks in `useTooltip` and `Tooltip` to reduce re-renders. The difference is most noticeable when `styled-components` is not involved.

Here's the new `PerformanceTest` story I added post-performance improvements:
https://user-images.githubusercontent.com/53451193/128405431-a6496964-05ad-4f04-a899-a422ba78512f.mov

And here it is without the improvements:
https://user-images.githubusercontent.com/53451193/128405470-c0aef579-044f-4fd2-a867-172e685cb0c3.mov


packages/components/src/Button/IconButton.tsx
```diff
@@ -95,14 +94,12 @@ const IconButtonComponent = forwardRef(
       onMouseOver: useWrapEvent(onMouseOver, propsOnMouseOver),
     }
 
-    const actualRef = useForkedRef<HTMLButtonElement>(forwardRef, ref)
-
```
mdodgelooker, line 99: `ref` from `useTooltip` is now a no-op and deprecated.

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -106,7 +96,7 @@ export const Tooltip = forwardRef(
         'aria-haspopup': ariaHaspopup,
         // Tooltip
         className: mergeClassNames([className, children.props.className]),
-        ref,
+        ref: forwardedRef,
```
mdodgelooker, line 101: Our solution for when you needed to put a ref on `Tooltip`'s child was to instead put it on `Tooltip` and we'd combine it with the one from `useTooltip` and pass that to the child. But now `ref` from `useTooltip`'s response is deprecated, so there's no good reason I can see for `Tooltip` accepting one. I'm not sure how to mark that as deprecated when using `forwardRef` though... 🤔

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -69,35 +70,24 @@ export const Tooltip = forwardRef(
...
-
-      const mergedHandlers = Object.keys(handlers).reduce(
-        (acc: Partial<typeof handlers>, key) => {
-          return {
-            ...acc,
-            [key]: mergeHandlers(handlers[key], children.props[key]),
-          }
-        },
-        {}
-      )
```
mdodgelooker, line 97: This was causing unnecessary re-renders on `Tooltip`'s child since the handlers were "new" on every render.

packages/components/src/Tooltip/types.ts
```diff
@@ -99,18 +99,15 @@ export interface UseTooltipProps {
...
-  onMouseOver: () => void
-  /**
-   * Used by popper.js to position the OverlaySurface relative to the trigger
-   */
-  ref: (node: HTMLElement | null) => void
-}
+type UseTooltipCallbacks = Required<
+  Pick<
+    CompatibleHTMLProps<HTMLElement>,
+    'onBlur' | 'onFocus' | 'onMouseOut' | 'onMouseOver'
```
mdodgelooker, line 105: Needed these to be required so that `useWrapEvent` would work (in `Tooltip.tsx` above).

packages/components/src/Tooltip/useTooltip.tsx
```diff
@@ -60,41 +60,43 @@ export function useTooltip({
   )
 
   const [surfaceElement, surfaceCallbackRef] = useCallbackRef()
-  const [newTriggerElement, callbackRef] = useCallbackRef()
+  const [newTriggerElement, setTriggerElement] = useState<HTMLElement | null>(
+    null
+  )
```
mdodgelooker, line 68: We're gonna get the trigger element from the mouseover/focus event, so we can just use `useState` instead of `useCallbackRef`, which was the source of 1 re-render.

packages/components/src/Tooltip/useTooltip.tsx
```diff
@@ -60,41 +63,43 @@ export function useTooltip({
...
   // If the triggerElement is passed in props, use that instead of the new element
-  const element =
-    typeof triggerElement === 'undefined' ? newTriggerElement : triggerElement
+  const element = triggerElement ?? newTriggerElement
 
-  const handleOpen = () => {
-    if (!disabled && (!element || !element.dataset.notooltip)) {
-      setIsOpen(true)
-    }
-  }
```
mdodgelooker, line 72: Moved `handleOpen` into the big `useMemo` in the return below.

packages/components/src/Tooltip/useTooltip.tsx
```diff
@@ -114,55 +119,85 @@ export function useTooltip({
...
-          style={style}
-          maxWidth={maxWidth}
-          invert={invert}
-        >
-          <TooltipContent
-            role="tooltip"
-            id={guaranteedId}
-            width={width}
-            textAlign={textAlign}
+  // Memo all the things!
```
mdodgelooker, line 130: This is a pretty huge `useMemo` but when breaking it apart, the dependency arrays just have a lot of overlap, so I figured why not just do one big one.

lukelooker (dismissed):
This changes looks solid to me - I think were on the same page that chasing renders is probably not the ideal way to solve this kind of issue (and won't get us the gains were looking for in-total) but this does seem better when compared to `main`. 

I had one thought about how to mark the `ref` on Tooltip as deprecated (which I think we could simply leave as a future follow-up / tech-debt issue to resolve as well)

mdodgelooker:
@lukelooker thanks for the deprecated ref tip, I think I got it working well enough (I didn't commit this example):
image

lukelooker (approved):
Excellent! :)

PR #2715, opened 2021-08-10
fix(ButtonToggle): Remove background color for focus via click

mdodgelooker:
When a user opens a button toggle filter via click inside a popover, the first item is focused (via the "initial focus" logic of the popover's focus trap) and the `neutralSubtle` override used in the core product is a dark enough grey as to make it look disabled (see screen recording below).

If the user opens the popover via keyboard, the focus ring is shown, which is a more obvious indicator. Activating the "click only" focus style is actually rare and pretty much only happens with the initial focus of the popover. Short of our preferred solution of adding a Done button to the popover that would instead receive initial focus, removing the background color for click-only focus will reduce confusion.

### Issue in the core product:
https://user-images.githubusercontent.com/53451193/128808419-9ff7d0ed-b0a0-48a9-bf1f-f620711a1b1b.mov

### After fix, open popover via click:
image

### Open via keyboard:
image

### DIFF
--- a/packages/components/src/Button/ButtonItem.tsx
+++ b/packages/components/src/Button/ButtonItem.tsx
@@ -98,6 +98,7 @@ const ButtonOuter = styled.button.attrs(({ type = 'button' }) => ({
 }))<ButtonItemProps & FocusVisibleProps>`
   ${focusVisibleCSSWrapper(
     ({ theme }) => css`
+      background: ${({ theme }) => theme.colors.neutralSubtle};
       box-shadow: 0 0 0.5px 1px ${theme.colors.keyFocus};
     `
   )}
@@ -122,7 +123,6 @@ export const ButtonItem = styled(ButtonLayout)`
   ${space}
 
   &:active,
-  &:focus,
   &:hover {
     background: ${({ theme }) => theme.colors.neutralSubtle};
     color: ${({ theme }) => theme.colors.text5};
--- a/packages/components/src/Button/ButtonSet.tsx
+++ b/packages/components/src/Button/ButtonSet.tsx
@@ -31,6 +31,7 @@ import React, {
   ReactNode,
   Ref,
   useCallback,
+  useRef,
   useState,
 } from 'react'
 import styled from 'styled-components'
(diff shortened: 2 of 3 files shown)


packages/components/src/Button/ButtonSet.tsx
```diff
@@ -101,19 +102,31 @@ export const ButtonSetLayout = forwardRef(
...
+          }
+
+          if (height > 0) {
+            getIsWrapping()
           } else {
-            setIsWrapping(false)
+            timeoutRef.current = setTimeout(getIsWrapping, 10)
           }
+        } else if (timeoutRef.current) {
+          clearTimeout(timeoutRef.current)
```
mdodgelooker, line 129: When adding a popover story to `ButtonToggle.story.tsx` I was seeing a squared off focus ring due to a bug in the logic that determines whether the items are wrapping and inner corners should be square vs rounded. A slight delay as the popover animates in is to blame, so I added this check to re-run the logic if the height of the element is 0.

packages/components/src/Button/ButtonItem.tsx
```diff
@@ -122,7 +123,6 @@ export const ButtonItem = styled(ButtonLayout)`
   ${space}
 
   &:active,
-  &:focus,
```
mdodgelooker, line 125: Side note: this doesn't affect `ButtonGroup`, where the background has always been white on focus because of this line: https://github.com/looker-open-source/components/blob/main/packages/components/src/Button/ButtonGroup.tsx#L67

mdodgelooker:
@lukelooker , I chatted it over with @jhardy who felt that `neutralSubtle` wasn't problematic itself, but combined with the darker override in the core product, and the fact that a user opening the popover via click finds the focus style more confusing than helpful, we decided restricting the background color to visible focus for keyboard users was a good compromise. Thoughts?

packages/components/src/Button/ButtonSet.tsx
lukelooker: Neat!

lukelooker (approved):
Works for me! :)

PR #2716, opened 2021-08-10
feat: Add new elevation values

jhardy:
Adds new elevation values and basic documentation 

image


packages/design-tokens/src/elevation/index.ts
```diff
@@ -0,0 +1,40 @@
...
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+import { Elevations } from './types'
+export { Elevations } from './types'
+
+const colorBase = '60, 64, 67'
```
lukelooker, line 30: Should this derive from a theme color? Should shadows "follow" some combination of text and background?

NOTE: This can totally be a conversation for later - just pondered it when I was looking at this.

lukelooker (approved):
I made a couple of small adjustments (restructured additions to `design-tokens` package to follow our newer code layout convention used for `space` and `colors` as well as marking `shadows` deprecated.

I have a follow-up PR to leverage these new values across our components.

packages/design-tokens/src/elevation/index.ts
jhardy: I had the same thought, I think we should eventually derive this from the theme color for sure. I would like to think about how we generate that color, so sounds like a good future TODO

PR #2720, opened 2021-08-11
feat: Ripple effect

mdodgelooker:
Adds hooks and style helpers to support the ripple effect, and a story file to demo it.

This PR adds it to `IconButton`: https://github.com/looker-open-source/components/pull/2721

**Question: should the ripple functions be exported or for internal use only? If the latter, I guess this would be a "chore" PR instead of a "feat".**

- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 – only the background renders, not the foreground ripple. Same with material web components.

### DIFF
--- a/packages/components/src/Ripple/rippleStyle.ts
+++ b/packages/components/src/Ripple/rippleStyle.ts
@@ -25,7 +25,6 @@
  */
 
 import { css, keyframes, Keyframes } from 'styled-components'
-import { RippleAnimationValues, RippleColorProps } from './types'
 
 const rippleRadiusIn: Keyframes = keyframes`
 from {
@@ -58,21 +57,10 @@ to {
 }
 `
 
-export type RippleStyleProps = RippleAnimationValues & RippleColorProps
-
 /**
  * Uses :before as fade-in overlay "background" and :after as rippling overlay
  */
-export const rippleStyle = ({
-  color = 'neutral',
-  rippleOffset,
-  rippleScaleRange,
-  rippleSize,
-}: RippleStyleProps) => css`
-  --ripple-scale-start: ${rippleScaleRange[0] || 1};
-  --ripple-scale-end: ${rippleScaleRange[1] || 1};
-  --ripple-translate: ${rippleOffset};
-
+export const rippleStyle = css`
   outline: none;
   overflow: hidden;
   position: relative;
@@ -80,17 +68,17 @@ export const rippleStyle = ({
 
   &::before,
   &::after {
-    background-color: ${({ theme }) => theme.colors[color]};
+    background-color: var(--ripple-color, #000000);
     border-radius: 50%;
     content: '';
-    height: ${rippleSize};
+    height: var(--ripple-size, 100%);
     left: 0;
     opacity: 0;
     pointer-events: none;
     position: absolute;
     top: 0;
     transform-origin: center center;
-    width: ${rippleSize};
+    width: var(--ripple-size, 100%);
   }
 
   &::before {
--- a/packages/components/src/Ripple/rippleStyle.ts
+++ b/packages/components/src/Ripple/rippleStyle.ts
@@ -1,0 +1,117 @@
+/*
+
+ MIT License
+
+ Copyright (c) 2021 Looker Data Sciences, Inc.
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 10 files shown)


packages/design-tokens/src/color/types/index.ts
```diff
@@ -42,6 +42,7 @@ export type { DerivativeColors } from './derivative'
 export type { CoreColors, IntentColors, SpecifiableColors } from './specifiable'
 export type { BlendColors } from './blends'
 export type {
+  ExtendedStatefulColor,
```
mdodgelooker, line 45: Wanted to use this so all the options for `IconButton`'s `toggleColor` will be reflected in a ripple color.

lukelooker: Now my terrible variable name shall forever be a standard. Huzzah! 😃

packages/components/src/Ripple/useRippleState.ts
```diff
@@ -0,0 +1,111 @@
...
+      // this ensures we don't see the ripple fade-out without rippling-in
+      return state === 'IN' ? 'OUT' : state
+    case 'DONE':
+      return 'OFF'
+  }
+}
+
+const getRippleClassName = (rippling: RippleState) => {
+  if (rippling === 'IN') {
+    return `fg-in`
```
lukelooker, line 58: Doesn't need to support interpolation does it?

```suggestion
    return 'fg-in'
```

packages/components/src/Ripple/useRippleState.ts
```diff
@@ -0,0 +1,111 @@
...
+  } else if (rippling === 'OUT') {
+    return `fg-out`
```
lukelooker, line 60: return 'fg-out'

packages/components/src/Ripple/useRipple.tsx
```diff
@@ -0,0 +1,128 @@
...
+): [number, number] => {
+  // For squares it looks best to start the ripple very small
+  const start = 0.1
+  if (bounded) {
+    // For rectangles it looks better to start at the size of the smaller dimension
+    // which is 1 because of how size is calculated
+    const startBounded = min === max ? start : 1
+    // The ripple needs to spread past all corners, use hypotenuse as the
+    // final diameter, and start at 1 to make the animation less jarring
+    return [startBounded, Math.hypot(min, max) / min]
```
lukelooker, line 74: TIL: `Math.hypot`

packages/components/src/Ripple/useRipple.tsx
```diff
@@ -0,0 +1,128 @@
...
+  // Start small and expand to the full size
+  return [start, 1]
+}
+
+const getRippleOffset = (min: number, max: number, bounded?: boolean) => {
+  if (!bounded || min === max) {
+    return '0, 0'
+  }
+  // If the element is rectangular, adjust the center of the ripple
+  // further along the larger dimension
```
lukelooker, line 85: Why? (not saying it's incorrect, just don't understand why it's not dead-center)

packages/components/src/Ripple/rippleStyle.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+to {
+  opacity: .12;
+}
+`
+
+const rippleOpacityOut: Keyframes = keyframes`
+from {
+  animation-timing-function: linear;
+  opacity: .12;
```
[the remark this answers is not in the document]
lukelooker: I think it's fine for now – we can explore how it interacts with theming and adjust in the future if necessary. :)

lukelooker (approved):
This is really cool – and also possibly the best tested piece of code I've seen out-of-the-gate in our code base :)

Few minor nits and a question or two but nothing that I'd block the PR on.

packages/design-tokens/src/color/types/index.ts
mdodgelooker: 😂

lukelooker (dismissed):

lukelooker (approved):
LGTM!

packages/components/src/Ripple/useRipple.tsx
mdodgelooker: By default it renders at `left: 0; top: 0;` so a shift to the right, i.e. `transform: translate(XXpx, 0);` will get it to the center. I _think_ `translate`, along with `transform-origin: center center`, is better than adjusting the `left` value since then you don't have to worry about the changing `scale` value.

mdodgelooker: But I'll admit I'm mostly going along with what Material did. 😁

PR #2721, opened 2021-08-11
feat(IconButton): Add ripple effect

mdodgelooker:
- Depends on: https://github.com/looker-open-source/components/pull/2720
- Adds ripple effect to `IconButton`
- Removes focus-visible behavior from `IconButton` (ripple replaces the need for this, and also conflicts visually with it)
- Refactors `ButtonBase.tsx` mainly to avoid prematurely adding the ripple effect to all buttons

- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11


packages/components/src/Button/ButtonBase.tsx
```diff
@@ -44,18 +44,13 @@ import {
...
+const buttonCSS = css<ButtonColorProps>`
   ${reset}
   ${maxWidth}
   ${minWidth}
   ${width}
 
-  ${focusVisibleCSSWrapper(({ color, toggleColor }) =>
-    buttonShadow(toggleColor || color)
-  )}
-
```
mdodgelooker, line 58: Moving visible focus out of here and down to a new `ButtonOuterFocusVisible` to be used by the non-icon buttons only.

packages/components/src/Button/ButtonBase.tsx
```diff
@@ -123,16 +121,10 @@ const ButtonLayout = forwardRef(
...
         {iconAfter}
-      </ButtonOuter>
+      </ButtonOuterFocusVisible>
     )
-  }
-)
-
-ButtonLayout.displayName = 'ButtonLayout'
-
-export const GenericButtonBase = styled(ButtonLayout)``
```
mdodgelooker, line 133: This is no longer needed – `IconButton.tsx` pulls in `ButtonOuter` directly and doesn't need anything happening in `ButtonLayout` now that ripple is replacing focus visible.

packages/components/src/Button/IconButton.story.tsx
```diff
@@ -108,6 +108,7 @@ export const ToggleOff = Template.bind({})
 ToggleOff.args = {
   ...Basic.args,
   toggle: false,
+  toggleColor: 'calculation',
```
mdodgelooker, line 111: I wanted to make sure that the ripple color would still be neutral if there was a `toggleColor` but `toggle` is false.

packages/components/src/Button/IconButton.tsx
```diff
@@ -24,22 +24,35 @@
...
 
-const IconButtonComponent = forwardRef(
-  (props: IconButtonProps, ref: Ref<HTMLButtonElement>) => {
+/**
+ * Appears as just an `Icon` but with proper HTML semantics to produce a `button`
+ * DOM element that is properly announced to screen-readers as well as providing
+ * keyboard tooltip support.
+ */
+export const IconButton = styled(
+  forwardRef((props: IconButtonProps, forwardedRef: Ref<HTMLButtonElement>) => {
```
mdodgelooker, line 55: Took this opportunity to update to the new `styled`-wraps-everything pattern here.

packages/components/src/Button/IconButton.tsx
```diff
@@ -87,76 +111,53 @@ const IconButtonComponent = forwardRef(
...
         {tooltip}
-      </GenericButtonBase>
+      </ButtonOuter>
     )
-  }
-)
-
-IconButtonComponent.displayName = 'IconButtonComponent'
-
-const outlineCSS = () => {
```
mdodgelooker, line 121: Moved this to a separate file to help with the `styled`-wraps-everything change.

packages/components/src/Ripple/useRipple.test.tsx
```diff
@@ -116,6 +117,26 @@ describe('useRipple', () => {
...
+  test('theme setting brandAnimation false', () => {
+    renderWithTheme(
+      <ThemeProvider
+        theme={(theme) => ({
+          ...theme,
+          defaults: { ...theme.defaults, brandAnimation: false },
+        })}
+      >
+        <RippleComponent color="key" />
+      </ThemeProvider>
```
mdodgelooker, line 129: This seems like a laborious way to turn off the ripple effect. Should we make it easy, maybe in `ThemeCustomizations`?

packages/components/src/Button/IconButton.story.tsx
```diff
@@ -129,13 +130,12 @@ ToggleColor.args = {
...
 ToggleColorFocused.args = {
   ...ToggleColor.args,
-  margin: 'small',
 }
 
 ToggleColorFocused.parameters = {
   beforeScreenshot: async (page: Page) => {
     const button = await page.$('button')
-    await button?.type(' ')
+    await button?.focus()
```
mdodgelooker, line 138: This was using `type(' ')` to simulate both focus and keyboard entry, both necessary to show the focus visible ring. But I just realized since it was a space key, the ripple animation was getting triggered too, so I updated to just `.focus()`. And also removed the margin since the focus style no longer exceeds the dimensions.

lukelooker (approved):
This looks good to me. The only open question (which can get addressed a little later - but before we cut 2.3) is whether we want to have ripple on or off by default.

packages/design-tokens/src/theme/theme.ts
```diff
@@ -76,14 +76,15 @@ export interface Theme {
...
+    brandAnimation: boolean
     density: DensityRamp
   }
 }
 
 export const theme: DefaultTheme = {
   breakpoints,
   colors,
-  defaults: { density: 0 },
+  defaults: { brandAnimation: true, density: 0 },
```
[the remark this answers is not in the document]
mdodgelooker: Good q @lukelooker – I guess since we could address it upstream in the core product, it comes down to how ready are we for extensions (and wherever else `@looker/components` is used) to start showing the ripple on `IconButton` only (depending on when we get the others done...)?

Another thing I've been wondering about is should we allow disabling the animation via a prop on `IconButton` itself? It wouldn't be hard to do and could be useful for those cases where interesting style override choices may look odd with the animation.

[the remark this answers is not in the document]
jhardy: I don't a strong opinion here, but a couple thoughts/questions:
- Do we want to roll this out as a package of updates to various components? So make it true, only when we have a series of components with `Ripple` implemented?  If so I could see us setting this to `false` for now
- Does setting this to `true` require a major version bump? I know its mostly visual, but it is quite different than before and I could almost argue a large visual change is a breaking change for some down stream consumers. If so I could also see this being set to false until we are ready to bump to 3.0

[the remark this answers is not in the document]
mdodgelooker: @jhardy I think that’s a good idea - keep it false till the whole group is ready. I’ll go ahead and make that update before merging this, unless you object @lukelooker ?

mdodgelooker (commented):
@jhardy and @lukelooker I've changed the default for `brandAnimation` to false, and I made a few tweaks to the stories and tests to account for this. Ready for re-review! 🙏

packages/design-tokens/src/theme/theme.ts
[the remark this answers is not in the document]
lukelooker: I like this plan! 

3.0 for switch - Helltool _can_ opt-in beforehand if we want do so but it won't get "sprung" on anyone that way

lukelooker (approved):
Ship it! :)

PR #2726, opened 2021-08-11
feat(Tabs2): Tabs with improved interface

laraharrow:
- [x] 📚 Documentation updated
- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11

We are a-ok on test coverage:
Screen Shot 2021-08-13 at 8 59 25 AM


packages/components/src/Tabs2/TabList2.tsx
```diff
@@ -0,0 +1,101 @@
...
+`
+
+const distributeCSS = css`
+  display: grid;
+  grid-auto-columns: 1fr;
+  grid-auto-flow: column;
+  ${Tab2} {
+    padding: ${({ theme: { space } }) => `${space.xsmall} ${space.medium}`};
+  }
+`
```
lukelooker, line 82: These should be exported and then imported into `TabList`

packages/components/src/Tabs2/TabList2.tsx
```diff
@@ -0,0 +1,101 @@
...
+  ${fontSize}
+  border-bottom: 1px solid ${(props) => props.theme.colors.ui2};
+  -ms-overflow-style: none; /* Internet Explorer 10+ */
+  overflow-x: auto;
+  scrollbar-width: none; /* Firefox */
+  white-space: nowrap;
+  &::-webkit-scrollbar {
+    display: none; /* Safari and Chrome */
+  }
+  ${({ distribute }) => (distribute ? distributeCSS : defaultLayoutCSS)}
```
lukelooker, line 100: Shared this block between here an "Tab1"

packages/components/src/Tabs2/TabList2.tsx
```diff
@@ -0,0 +1,101 @@
...
+        role="tablist"
+        {...navProps}
+      >
+        {children}
+      </div>
+    )
+  }
+)
+
+TabList2Layout.displayName = 'TabList2Layout'
```
lukelooker, line 64: Collapse this "into" `TabList`

packages/components/src/Tabs2/TabList2.tsx
```diff
@@ -0,0 +1,101 @@
...
+} from '@looker/design-tokens'
+import styled, { css } from 'styled-components'
+import { useArrowKeyNav } from '../utils'
+import { Tab2 } from './Tab2'
+
+export interface TabList2Props extends PaddingProps, FontSizeProps {
+  children: JSX.Element[]
+  className?: string
+  distribute?: boolean
+}
```
lukelooker, line 44: Make this a `type`

packages/components/src/Tabs2/Tabs2.tsx
```diff
@@ -0,0 +1,93 @@
...
+      label={label}
+      key={index}
+      selected={id === tabId}
+      onSelect={() => handleTabChange(id || label)}
+    >
+      {label}
+    </Tab2>
+  ))
+
+  const currentTab = tabs.find((tab) => tab.id === tabId)
```
lukelooker, line 81: Explain all the magic here-in

packages/components/src/Tabs2/types.ts
```diff
@@ -0,0 +1,72 @@
...
+  TypographyProps,
+} from '@looker/design-tokens'
+
+export type Tab2Props = Omit<CompatibleHTMLProps<HTMLButtonElement>, 'type'> &
+  LayoutProps &
+  PaddingProps &
+  TypographyProps & {
+    label: string
+    onSelect?: () => void
+    selected?: boolean
```
lukelooker, line 41: Add JSDoc

packages/components/src/Tabs2/Tabs2.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+ */
+
+import React, { Children, useEffect, useState } from 'react'
+import { Tab2 } from './Tab2'
+import { TabList2 } from './TabList2'
+import { TabPanels2 } from './TabPanels2'
+import { Tabs2Props, TabStack } from './types'
+
+export const Tabs2 = ({
```
lukelooker, line 33: Add `JSDoc` explanation of `Tabs2`

lukelooker (changes requested):
- [x] Mark `Tab` & `Tabs` as `@deprecated` (see `Accordion`)
- [x] Update `tabs.mdx` to use `Tabs2`

packages/components/src/Tabs/Tab.tsx
```diff
@@ -56,100 +46,45 @@ export interface TabProps
...
-    >
-      {children}
-    </TabStyle>
-  )
-})
-
-TabJSX.displayName = 'TabJSX'
-
-export const Tab = styled(TabJSX).attrs(
+export const Tab = styled(
```
laraharrow, line 49: We are updating the file to use `Tab2` instead.
`Tab` will be deprecated and this file will eventually be removed so some of the code got removed from here and added to the new version `Tab2`.

packages/components/src/Tabs2/TabList2.tsx
laraharrow: moving the style to `TabList` creates the same breaking that we were getting with having `Tab2Style` in a separated file. Basically somehow passing css style between the components is breaking the style.

packages/components/src/Tabs2/TabList2.tsx
laraharrow: read comment above.

packages/components/src/Tabs/TabList.tsx
```diff
@@ -26,16 +26,10 @@
...
-  PaddingProps,
-  padding,
-  reset,
-} from '@looker/design-tokens'
-import styled, { css } from 'styled-components'
+import { FontSizeProps, PaddingProps } from '@looker/design-tokens'
+import styled from 'styled-components'
 import { useArrowKeyNav } from '../utils'
-import { Tab } from '.'
+import { TabList2CSS } from '../Tabs2'
```
laraharrow, line 32: Adding the style via import is breaking the component.

packages/components/src/Tabs2/Tabs2.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+import 'jest-styled-components'
+import '@testing-library/jest-dom/extend-expect'
+import { renderWithTheme } from '@looker/components-test-utils'
+import React from 'react'
+import { composeStories } from '@storybook/testing-react'
+import { fireEvent, screen } from '@testing-library/react'
+import * as stories from './Tabs2.story'
+import { Controlled } from './Tabs2.story'
```
laraharrow, line 34: seems like the * as stories doesn't import the stories that are not set as Template so I had to import the controlled version separated.

packages/components/src/Tabs2/Tabs2.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    if (
+      // check if the defaultTabId is passed and if not set to display the first Tab available.
+      !defaultTabId &&
+      draftTabs.length > 0 &&
+      !draftTabs.find((tab) => tab.id === defaultTabId)
+    ) {
+      // select the first `Tab` that is not disabled
+      setCurrentTabId(
+        draftTabs[draftTabs.findIndex((tab) => !tab.disabled === true)].id
```
laraharrow, line 76: I added a little bit of new logic to the setting of the `Tab` to take in account if the `draftTabs[0]` is set as disabled

packages/components/src/Tabs2/index.ts
```diff
@@ -0,0 +1,30 @@
...
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+export * from './Tab2'
+export { TabList2CSS } from './TabList2'
```
laraharrow, line 28: this may not be necessary?

packages/components/src/Tabs/TabList.tsx
```diff
@@ -45,47 +45,14 @@ export interface TabListProps extends PaddingProps, FontSizeProps {
...
-    )
-  }
-)
-
-TabListLayout.displayName = 'TabListLayout'
+/**
+ * @deprecated Use `Tabs2` and `Tab2` instead
+ */
 
 const defaultLayoutCSS = css`
```
laraharrow, line 52: check comment bellow on why the style is not imported from `TabList2`

www/src/documentation/components/content/tabs.mdx
```diff
@@ -115,43 +120,20 @@ Using the distribute prop on `TabList` will equally distributed each `Tab` in th
...
+        <Tab2 id="dogs" label="Dogs">
+          Cats are way better than dogs. Go to other tab
+        </Tab2>
+      </Tabs2>
     </>
   )
 }
 ```
-
-## Keyboard Navigation
```
laraharrow, line 136: I was going to update the code so the new `Tabs2` component has better relationship between `active` and `focus` but as I looked into the [figma](https://www.figma.com/file/zd6MjUXIIstvG7gTJWaX2Y/Looker-Components?node-id=6740%3A0) file, looks like we will soon be doing updates to this component based on the GM2 roadmap. So I held off on that so the work can be done on a separated PR.

laraharrow:
There is a problem with the snapshots. After making the changes to the area content of the `Tab` the snapshots seem to not be matching anymore but this is the first time we are logging them so I can't even see the diff other then on the `image snapshots` tab I see that the height dimensions are smaller then before and that as I understand is what we want.

laraharrow:
We are a-ok on test coverage:
Screen Shot 2021-08-13 at 8 59 25 AM

www/src/documentation/components/content/tabs.mdx
```diff
@@ -1,110 +1,115 @@
...
-The order in which elements are placed inside `TabList` and `TabPanels` defines the relationship between them. The first `Tab` is related to the first `TabPanel` and so on.
+<MessageBar canDismiss={false} intent="inform" mb="medium">
+  Tabs has been deprecated and replaced with
+  <Code fontWeight="bold" fontSize="inherit">
+    Tabs2
+  </Code> please update any existing instances to use the new improved version. Deprecation
+  of `Tabs` is planned for @looker/components 3.0
+</MessageBar>
+
+`<Tabs2>` expects to receive one or more `Tab2` components. `Tabs2` manages the visibility of the single `content` based on its internal `Tab2` content.
```
lukelooker, line 16:
```suggestion
`Tabs2` contains one or more `Tab2` components. `Tabs2`, when left uncontrolled, manages the visible tab as well as switching between tabs.
```

www/src/documentation/components/content/tabs.mdx
```diff
@@ -1,110 +1,115 @@
...
+
+`Tab2` expects to have `label` that will be the value displayed on the `Tab` itself. The value passed as children will be displayed as its content area.
```
lukelooker, line 18:
```suggestion
Each `Tab2` must have a `label`. The value passed as children will be displayed as its content area. `id` is optional (`label` will be used in its place if not supplied) but `id` is highly recommended when using `Tabs2` in a controlled state.
```

www/src/documentation/components/content/tabs.mdx
```diff
@@ -1,110 +1,115 @@
...
+  <Tab2 label="1">
+    Really long tab title that should help demonstrate side-scroll
+  </Tab2>
+  <Tab2 label="2">Yet-another-tab-title to get things width enough</Tab2>
+  <Tab2 label="3">
+    Really, you must be obsessed with these long tab titles?
+  </Tab2>
+  <Tab2 label="4">
+    Okay, okay, we get it you want to see scrolling in action...
+  </Tab2>
```
lukelooker, line 47: `label` and `children` need to get inverted here so that the `Tabs2` has really long labels, not content :)

www/src/documentation/components/content/tabs.mdx
```diff
@@ -1,110 +1,115 @@
...
+</Tabs2>
 ```
 
 ## Disabled Tab(s)
 
-To create a disabled Tab add the `disabled` prop. NOTE: A `TabPanel` must still exist for disabled tabs to allow for tabs to keep the index positions in sync.
+To create a disabled Tab add the `disabled` prop to each individual `Tab2`.
```
lukelooker, line 53:
```suggestion
To display a tab that can't be activated add the `disabled` prop.
```

www/src/documentation/components/content/tabs.mdx
```diff
@@ -1,110 +1,115 @@
...
+    Rebecca Purple{' '}
+    <div
+      style={{
+        backgroundColor: 'RebeccaPurple',
+        height: '350px',
+      }}
+    />
+  </Tab2>
+  <Tab2 label="Coral">
+    {' '}
```
lukelooker, line 86:
```suggestion
```

www/src/documentation/components/content/tabs.mdx
```diff
@@ -1,110 +1,115 @@
...
+    <div
+      style={{
+        backgroundColor: 'coral',
+        height: '350px',
+      }}
+    />
+  </Tab2>
+  <Tab2 label="Forest Green">
+    {' '}
```
lukelooker, line 95:
```suggestion
```

www/src/documentation/components/content/tabs.mdx
```diff
@@ -1,110 +1,115 @@
...
+    <div
+      style={{
+        backgroundColor: 'ForestGreen',
+        height: '350px',
+      }}
+    />
+  </Tab2>
+  <Tab2 label="Cornflower Blue">
+    {' '}
```
lukelooker, line 104:
```suggestion
```

www/src/documentation/components/content/tabs.mdx
```diff
@@ -115,43 +120,20 @@ Using the distribute prop on `TabList` will equally distributed each `Tab` in th
...
-        <TabPanels>
-          <TabPanel>A</TabPanel>
-          <TabPanel>B</TabPanel>
-        </TabPanels>
-      </Tabs>
+      <Space my="small">
+        <Button onClick={() => setTab('cats')}>check out cats</Button>
+        <Button onClick={() => setTab('dogs')}>check out dogs</Button>
+      </Space>
+      <Divider />
```
lukelooker, line 127: For the purposes of user-facing demonstrations i prefer to see "vanilla" HTML as it reduces the amount of context the reader needs about our component suite. 

```suggestion
  <button onClick={() => setTab(0)}>Go to A</button>
  <button onClick={() => setTab(1)}>Go to B</button>
```

NOTE: We _do_ use styled components within Storybook when we're taking image snapshots because we've seen slight rendering variations between browsers with vanilla components.


Also, above let's change: 

```
const [currentTabIndex, setTab] = useState(0)
```

to 
```
const [tabId, setTabId] = useState(0)
```

In keeping with the new `tabId` nomenclature.

lukelooker (changes requested):

laraharrow:
Updates done after the review!
I also updated the snapshots to integrate the test based on having the first `Tab` as `disabled`. The error on the snapshots remain thou.
@lukelooker any luck on that?

lukelooker (approved):
Change looks good to me but I'd like someone who was less "hands-on" to do another review as well. :)

packages/components/src/Tabs2/types.ts
```diff
@@ -0,0 +1,103 @@
...
+  onTabChange?: (tabId: string) => void
+}
+
+type Uncontrolled = {
+  defaultTabId?: never
+  onTabChange?: never
+  tabId?: string
+}
+
+export type Tabs2Props = (Controlled | Uncontrolled) & {
```
dbchristopher, line 94: It looks like the intent here is to prevent users from using a `tabId` without also passing an `onTabChange` prop. That doesn't seem to work as expected! I can totally set `<Tabs2 tabId={currentTabId}>` or `<Tabs2 onTabChange={() => null}>` without typescript complaining. 

I took a bit to play around with the right type definitions, and I think I came up with a structure that requires you to do *both*  `tabId` and `onTabChange` OR do `defaultTabId`, but not any other combination. I think it should be structured something like this:

```suggestion
type Controlled = {
  /**
   * Controlled: which tab to show now
   */
  tabId: string
  /**
   * Callback called when tabId changes
   */
  onTabChange: (tabId: string) => void
  defaultTabId?: never
}

type Uncontrolled = {
  /**
   * Which tab to show on load
   */
  defaultTabId: string
  tabId?: never
  onTabChange?: never
}

type Default = {
  defaultTabId?: never
  tabId?: never
  onTabChange?: never
}

export type Tabs2Props = (Controlled | Uncontrolled | Default) & {
```

CC'ing @lukelooker to double check my logic there

dbchristopher (commented):
This is all really great @laraharrow! The only real bug I found is with the types where it was meant to be enforce controlled vs uncontrolled prop combinations. Otherwise there was just some tiny typos and some possible (but really minor) css improvements. Great work!

packages/components/src/Tabs2/index.ts
```diff
@@ -0,0 +1,29 @@
...
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+export * from './Tab2'
+export * from './Tabs2'
```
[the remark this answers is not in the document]
laraharrow: the second option.  Right now all the `Tabs` work is set as `@depricated` and in the future the plan is to make `Tabs2` the official component's `Tabs`.

packages/components/src/Tabs/Tabs.story.tsx
```diff
@@ -60,6 +60,19 @@ const Template: Story<DemoProps> = ({ tabCount, tabPrefix, ...args }) => {
...
+
+export const Distributed = Template.bind({})
+Distributed.args = {
+  distribute: true,
+  tabCount: 3,
+  tabPrefix: 'My Awesome Tab',
+}
+
+export const Scrolling = Template.bind({})
+Scrolling.args = {
```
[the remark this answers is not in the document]
laraharrow: good idea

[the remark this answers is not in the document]
laraharrow: good idea!

packages/components/src/Tabs2/Tab2.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  &:active {
+    border-bottom-color: ${({ selected, theme }) =>
+      selected ? theme.colors.key : theme.colors.text2};
+  }
+
+  &:active,
+  &:hover {
+    border-bottom-color: transparent;
+  }
```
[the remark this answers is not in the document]
laraharrow: good point

packages/components/src/Tabs2/Tab2.tsx
```diff
@@ -0,0 +1,135 @@
...
+  &:hover {
+    border-bottom-color: ${({ selected, theme }) =>
+      selected ? theme.colors.key : theme.colors.ui3};
+  }
+
+  &:disabled {
+    border-bottom-color: transparent;
+    color: ${({ theme }) => theme.colors.text1};
+    cursor: default;
+  }
```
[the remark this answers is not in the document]
laraharrow: I had that test there on one of the clean ups it got removed, adding it again, good catch

packages/components/src/Tabs2/TabList2.tsx
```diff
@@ -0,0 +1,89 @@
...
+import React, { forwardRef, Ref } from 'react'
+import { fontSize, padding, reset } from '@looker/design-tokens'
+import styled, { css } from 'styled-components'
+import { useArrowKeyNav } from '../utils'
+import { TabList2Props } from './types'
+
+const defaultLayoutCSS = css`
+  button {
+    min-width: 3rem;
+  }
```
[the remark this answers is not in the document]
laraharrow: Not really because that value gets removed if the prop `distribute` is passed so it needs to be a separated set of rules that gets toggled based on the value of  `distribute`

packages/components/src/Tabs2/Tabs2.story.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const Template: Story<Partial<Tabs2Props>> = (args) => (
+  <Box height="8rem">
+    <Tabs2 {...args}>
+      <Tab2 id="cats" label="Cats">
+        Here's awesome story about cats
+      </Tab2>
+      <Tab2 id="dogs" label="Dogs">
+        Cats are way better than dogs. Go to other tab
+      </Tab2>
```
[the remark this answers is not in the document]
laraharrow: This is Lukes work 😅
Also Truffle doesn't know how to read so please don't tell him!

packages/components/src/Tabs2/types.ts
laraharrow: Oh man I always get so confused on the props that should and should not be a part of controlled or uncontrolled. 
You are right what I had didn't quite work , but I think both `Uncontrolled` and `Controlled` should have the possibility of passing `defaultTabId` and to be honest the `Default` option gets me even more confused. Seems like what you tried and what I tried is still not quite the right solution, I'm getting a TS error with ether variaton.

`
  Type '{ children: Element[]; defaultTabId?: string | undefined; tabId?: string | undefined; onTabChange?: ((tabId: string) => void) | undefined; distributed?: boolean | undefined; } | { ...; }' is not assignable to type 'IntrinsicAttributes & Tabs2Props'.
    Type '{ children: Element[]; defaultTabId?: string | undefined; tabId?: string | undefined; onTabChange?: ((tabId: string) => void) | undefined; distributed?: boolean | undefined; }' is not assignable to type 'IntrinsicAttributes & Tabs2Props'.
      Type '{ children: Element[]; defaultTabId?: string | undefined; tabId?: string | undefined; onTabChange?: ((tabId: string) => void) | undefined; distributed?: boolean | undefined; }' is not assignable to type 'Uncontrolled'.
      Types of property 'defaultTabId' are incompatible.
      Type 'string | undefined' is not assignable to type 'undefined'.
      Type 'string' is not assignable to type 'undefined'.
  `

dbchristopher: Ugh yeah this is really hard to type! Where is that error coming from?

About `defaultTabId` vs `tabId`, I'm pretty sure the best practice would be that you can't use them both simultaneously. My thinking is that if the values conflict, which one should the component implement? 

I looked up some articles about doing `value` vs `defaultValue` in forms, and they seem to confirm that the pattern should be mutually exclusive

https://scriptverse.academy/tutorials/reactjs-defaultvalue-value.html

Screen Shot 2021-08-17 at 4 32 16 PM

dbchristopher: >  to be honest the Default option gets me even more confused.

Yeah I totally understand that! I added it to account for the use case where they don't pass either `tabId` or `defaultTabId`. The thinking is that there are actually three possible scenarios here:

1. User can pass BOTH `tabId` and `onTabChange` (to require both you have to remove the optional `?` setting from those props)
2. User can pass ONLY `defaultTabId`
3. The user passes none of the above options. <-- that's the case that I added the `Default` option to describe. I agree that it seems like it might not be necessary, but it seemed to help in my local tests 🤷‍♂️

packages/components/src/Tabs2/Tabs2.story.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export const Controlled = () => {
+  const [currentTabId, setTabId] = useState('cats')
+
+  return (
+    <>
+      <button onClick={() => setTabId('cats')}>Switch to Cats</button>
+      <button onClick={() => setTabId('dogs')}>Switch to Dogs</button>
+      <Tabs2 tabId={currentTabId} onTabChange={setTabId}>
+        <Tab2 id="cats" label="Cats">
```
roninjin10, line 67: Awesome!

roninjin10 (commented):
Addresses https://yaqs.corp.google.com/eng/q/4348128683202969600?from_email&action=view-question nicely!   Thanks!

you-did-good-job-meme

PR #2727, opened 2021-08-12
fix: Replace usage of t-shirt size space with units

lukelooker:


lukelooker:
@jhardy - Missed a few spots somehow when I did the big update pass.

jhardy (approved):
👍

PR #2728, opened 2021-08-12
refactor(density): Consolidate default density in theme

lukelooker:
Refactors all locations that initialize `density` as `0` to instead leverage `theme.defaults.density`

This refactor will allow for easy theme-based override of density in applications.


ElliotTheHuman (approved):
LGTM

PR #2729, opened 2021-08-12
feat(Density): Add new helpers for changing density default

lukelooker:
Adds new `<Density />` helper as well as documentation for its usage

image

- [ ] 📚 Documentation updated


packages/components/src/Density/Density.tsx
```diff
@@ -0,0 +1,54 @@
...
+ SOFTWARE.
+
+ */
+
+import React, { FC, useContext } from 'react'
+import { DensityRamp } from '@looker/design-tokens'
+import { ThemeContext, ThemeProvider } from 'styled-components'
+
+export type DensityProps = {
+  density: DensityRamp
```
jhardy, line 32: I am not sure why, but `<Density density={-1} />` feels a bit strange, a couple thoughts I had.

```
<Density adjust={-1} />
<Density scale={-1} />
<Density set={-1} />
```

 I will say I think `density` prop will probably be easier to remember and maybe that is more important than how it reads

packages/components/src/Density/Density.tsx
```diff
@@ -0,0 +1,54 @@
...
+  theme.defaults.density = density
+
+  return <ThemeProvider theme={theme}>{children}</ThemeProvider>
+}
+
+export const Density1: FC = (props) => <Density density={1} {...props} />
+export const Density0: FC = (props) => <Density density={0} {...props} />
+export const DensityN1: FC = (props) => <Density density={-1} {...props} />
+export const DensityN2: FC = (props) => <Density density={-2} {...props} />
+export const DensityN3: FC = (props) => <Density density={-3} {...props} />
```
jhardy, line 54: I kind of dislike the way the `DensityN*` reads, maybe I am getting hung up on the visual appearance more than the utility, which I think could be nice. 

I do wonder if perhaps only export the helpers and not the main `Density` component. I'm thinking from a developer standpoint if I see `<Density density={*}>` and `<Density*>` across various places in a code base is there a case where I get confused about what the correct style to use is  and if they function differently?

packages/components/src/Density/Density.tsx
lukelooker: I had the same misgiving on the prop name. I actually like the way `scale` looks / reads.

packages/components/src/Density/Density.tsx
lukelooker: I tried `DensityNegativeX` instead as well as adding JSDoc blocks that explain that the helper functions are just aliases. 

I think the base `Density` component makes sense because there may be spots where folks want to specify the `scale` dynamically (imagine a "zoom" feature or something of that nature) and the code-mechanics of changing component names would be annoying compared to simply changing the property value.

www/src/MDX/Pre/allComponents.ts
```diff
@@ -56,6 +56,12 @@ import {
   ConfirmLayout,
   CopyToClipboard,
   doDataTableSort,
+  Density,
+  Density0,
+  Density1,
+  DensityN1,
+  DensityN2,
+  DensityN3,
```
jhardy, line 64:
```suggestion
  DensityNegative1,
  DensityNegative2,
  DensityNegative3,
```

www/src/MDX/Pre/allComponents.ts
```diff
@@ -265,6 +271,12 @@ export const allComponents = {
   DataTableItem,
   DateFormat,
   DateTimeFormat,
+  Density,
+  Density0,
+  Density1,
+  DensityN1,
+  DensityN2,
+  DensityN3,
```
jhardy, line 279:
```suggestion
  DensityNegative1,
  DensityNegative2,
  DensityNegative3,
```

www/src/documentation/foundations/density.mdx
```diff
@@ -20,3 +20,27 @@ A couple key things to keep in mind when using density are:
...
+<Density1>
+  <ListItem>+1 density</ListItem>
+</Density1>
+<Density0>
+  <ListItem>0 density (default)</ListItem>
+</Density0>
+<DensityNegative3>
+  <ListItem>-3 Density</ListItem>
+</DensityNegative3>
+```
```
jhardy, line 46: Getting an error on this page
image

jhardy (changes requested):
Small issue with mdx all component exports

jhardy (approved):
👍

PR #2731, opened 2021-08-12
feat(LkFieldTree): Add LkField suite of components

ElliotTheHuman:
## Follow-up Work
- Tests for `LkFieldTree`

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/Tree/types.ts
```diff
@@ -97,9 +97,8 @@ export type TreeProps = ControlledLoosely &
      */
     dividers?: boolean
     /**
-     * Use itemRole to set the type of wrapper element for the label
+     * Use itemRole to set the type of wrapper element for the label. By default the label's wrapper element is a <button>.
      * Note: Unlike TreeItem, 'button' is not an allowed value since click event handlers are passed to the nested <AccordionDisclosure> div
-     * @default none
      */
```
lukelooker, line 102: Should this get marked as `deprecated`? Can we switch to just `button` or `link` depending on presence of `href`?

ElliotTheHuman: We totally could! I can mark this prop as deprecated on this PR, but did we want to include the auto-switching behavior on this PR as well? I think separate PR might be better because that would require story and test changes, so it'll probably add a sizable chunk to the line count.

ElliotTheHuman: Actually, maybe it's better to mark this deprecated only after getting the auto-switch behavior. Thoughts?

lukelooker (changes requested):
- [ ] Needs test coverage :)

ElliotTheHuman:
Tests are in! Note test additions increased line count by a bunch.

packages/components/src/LkFieldTree/LkFieldItemLabel.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+import styled from 'styled-components'
+import {
+  listItemBackgroundColor,
+  ListItemBackgroundColorProps,
+} from '../ListItem/utils'
+
+/**
+ * @TODO: Delete TreeItemLabel once FieldItem component is completed
+ */
```
lukelooker, line 35: Is this true? Wouldn't `LkFieldItemLabel` survive?

packages/components/src/LkFieldTree/stories/FieldPicker.story.tsx
```diff
@@ -26,12 +26,20 @@
...
+import { FieldPickerItem } from './FieldPickerItem'
 
-const BorderRadiusOverrideTree = styled(Tree)`
+export default {
+  argTypes,
+  component: LkFieldTree,
+  title: 'LkFieldTree',
+}
+
+const BorderRadiusOverrideTree = styled(LkFieldTree)`
```
lukelooker, line 42: Were you thinking of building this in to LkFieldTree?

packages/components/src/LkFieldTree/types.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { GenericClickProps, ListItemProps, TreeProps } from '..'
+import { ControlledLoosely } from '../Accordion2/controlTypes'
+
+export type LkFieldTreeProps = ControlledLoosely &
+  GenericClickProps<HTMLElement> &
+  Pick<
+    TreeProps,
+    | 'color'
+    | 'density'
```
lukelooker, line 35: Does it need `density` support?

packages/components/src/LkFieldTree/types.ts
```diff
@@ -0,0 +1,47 @@
...
+    | 'disabled'
+    | 'hovered'
+    | 'icon'
+    | 'selected'
+    | 'label'
+  >
+
+export type LkFieldItemProps = Omit<
+  ListItemProps,
+  'density' | 'description' | 'itemRole' | 'truncate'
```
lukelooker, line 46: Ditto for `density`?

packages/components/src/Tree/types.ts
lukelooker: I'm cool with that.

packages/components/src/LkFieldTree/LkFieldItemLabel.tsx
ElliotTheHuman: :doh: Yep this comment only applies to stock `Tree`'s label. Deleted.

packages/components/src/LkFieldTree/stories/FieldPicker.story.tsx
ElliotTheHuman: I was. More specifically, I was hoping to bake all style overrides from Field Picker in one-go in a follow up PR. If it's preferable to bring the border radius behavior right now though I can definitely do that.

lukelooker: Separately is totally fine - just checking in :)

packages/components/src/LkFieldTree/types.ts
ElliotTheHuman: I _think_ we settled on forcing `LkField` components to be at density = -3. I'm not super opposed to allowing density, _but_ it may make lead to some more tech debt when implementing the item padding refactor work (i.e. we may need to add a new attribute `heightWithDescription` to `ListItemDimensions`).

Happy to talk this over though and clarify why allowing multiple densities would lead to a new item dimension property.

packages/components/src/LkFieldTree/types.ts
lukelooker: I was backwards here.

lukelooker (approved):
LGTM

PR #2732, opened 2021-08-12
refactor: Ripple effect

mdodgelooker:
Adjust the hook & style functions so they can both be applied to the same component instead of the style needing to be applied to a child component.

### DIFF
--- a/packages/components/src/Ripple/Ripple.story.tsx
+++ b/packages/components/src/Ripple/Ripple.story.tsx
@@ -26,58 +26,45 @@
 
 import React from 'react'
 import { Story } from '@storybook/react/types-6-0'
-import { shouldForwardProp } from '@looker/design-tokens'
 import styled from 'styled-components'
 import { simpleLayoutCSS, SimpleLayoutProps } from '../Layout'
-import {
-  useRipple,
-  useRippleHandlers,
-  UseRippleProps,
-  RippleColorProps,
-  rippleStyle,
-  RippleStyleProps,
-} from './'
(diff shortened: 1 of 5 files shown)


packages/components/src/Ripple/Ripple.story.tsx
```diff
@@ -26,58 +26,45 @@
...
+        {...rippleHandlers}
+        {...props}
+      >
+        click me
+      </div>
+    )
+  }
+)<SimpleLayoutProps>`
+  ${simpleLayoutCSS}
+  ${rippleStyle}
```
mdodgelooker, line 62: Now instead of needing a nested `RippleStyle` component, `rippleStyle` can be applied to the same component that calls the `useRipple` hook.

packages/components/src/Ripple/rippleStyle.ts
```diff
@@ -58,39 +57,28 @@ to {
...
-export const rippleStyle = ({
-  color = 'neutral',
-  rippleOffset,
-  rippleScaleRange,
-  rippleSize,
-}: RippleStyleProps) => css`
-  --ripple-scale-start: ${rippleScaleRange[0] || 1};
-  --ripple-scale-end: ${rippleScaleRange[1] || 1};
-  --ripple-translate: ${rippleOffset};
-
```
mdodgelooker, line 75: Moved all these to just be applied to inline style from the `useRipple` return.

lukelooker (approved):
Really nice clean-up. :)

PR #2735, opened 2021-08-13
fix(Elevation): add elevation level of plus0 to match spec

jhardy:
Add elevation level `plus0` to match spec
image

### DIFF
--- a/packages/design-tokens/src/elevation/index.ts
+++ b/packages/design-tokens/src/elevation/index.ts
@@ -28,13 +28,14 @@ import { Elevations } from './types'
 export { Elevations } from './types'
 
 const colorBase = '60, 64, 67'
-const color1 = `rgba(${colorBase}, .30)`
-const color2 = `rgba(${colorBase}, .15)`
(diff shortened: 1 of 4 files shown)


packages/design-tokens/src/elevation/index.ts
```diff
@@ -28,13 +28,14 @@ import { Elevations } from './types'
 export { Elevations } from './types'
 
 const colorBase = '60, 64, 67'
-const color1 = `rgba(${colorBase}, .30)`
-const color2 = `rgba(${colorBase}, .15)`
+const baseShadowColor = `rgba(${colorBase}, .30)`
+const ambientShadowColor = `rgba(${colorBase}, .15)`
```
jhardy, line 32: I gave these a bit more descriptive names.

lukelooker: OOoohhh! (❤️ )

lukelooker (approved):
Looks great!

PR #2748, opened 2021-08-16
refactor(ListItem): Padding and height refactor

ElliotTheHuman:
- Moved padding and height styling from ListItemContent container to ListItemIcon and ListItemLabel partitions
- Fixed a minor issue where the theme provider's density wasn't considered when calculating density in `List`


packages/components/src/ListItem/ListItemContent.tsx
```diff
@@ -130,14 +118,10 @@ export const ListItemContent = styled(ListItemContentInternal)`
...
-    Remove itemRole-specific space CSS when FieldItem is completed
-  */
-  padding-bottom: ${({ itemRole }) => (itemRole === 'none' ? 0 : undefined)};
-  padding-top: ${({ itemRole }) => (itemRole === 'none' ? 0 : undefined)};
-  ${({ density, itemRole }) =>
-    itemRole === 'none' && itemRoleNoneHeight(density)}
+
+  padding: 0; /* Remove default top and bottom padding */
+  ${({ density }) => listItemPaddingX(density)}
+
```
lukelooker, line 124: That's a lot simpler! :)

packages/components/src/ListItem/ListItemIcon.tsx
```diff
@@ -57,29 +57,31 @@ export const ListItemIcon: StyledComponent<
...
   align-self: ${({ alignStart }) => (alignStart ? 'flex-start' : 'center')};
   display: flex;
   margin-right: ${({ iconGap, theme }) => theme.space[iconGap]};
-  ${colorHelper}
+  ${({ density, theme }) => listItemPaddingY(density || theme.defaults.density)}
 
-  & > svg,
-  ${StyledIconBase}, ${IconPlaceholder} {
+  ${colorHelper} & > svg,
+    ${StyledIconBase}, ${IconPlaceholder} {
```
lukelooker, line 81: I think this is a mistake?

```suggestion
  ${colorHelper} 
  
  & > svg,
    ${StyledIconBase}, ${IconPlaceholder} {
```

packages/components/src/ListItem/ListItemIcon.tsx
```diff
@@ -24,31 +24,31 @@
...
+import {
+  listItemDimensions,
+  listItemIconColor,
+  listItemPaddingY,
+} from './utils'
+import { ListItemDimensions } from './types'
 
 export type ListItemIconProps = DensityProp & {
   color?: string
+  description?: ReactNode
```
lukelooker, line 45: Why is this needed here? (not seeing any new use within the component at least)

packages/components/src/ListItem/utils/createListItemPartitions.tsx
```diff
@@ -60,6 +60,7 @@ export const createListItemPartitions = ({
     children: icon,
     color,
     density,
+    description, // description is passed to icon because the icon determines the item height, which is calculated based on description
```
lukelooker, line 63: I don't see it's in use within the `ListItemIcon`. I'd prefer if rather than sending the actual ReactNode we instead send a boolean - `hasDescription` or similar.

lukelooker:
For the sake of shrinking the diff here in might make sense to pull out the density fixes from the padding change.

packages/components/src/ListItem/ListItemIcon.tsx
ElliotTheHuman: Definitely. Good catch!

packages/components/src/ListItem/ListItemIcon.tsx
ElliotTheHuman: Doh. A previous iteration of this refactor had Icon determining the height of the item, and thus it needed to know if a description as needed.

Jared and I instead decided to make the `ListItemLabel` the ultimately decider of the height (note: Icon still gets padding to make sure it gets the right vertical alignment relative to the label text in both the description and no-description cases).

packages/components/src/ListItem/utils/createListItemPartitions.tsx
ElliotTheHuman: See above comment

lukelooker (approved):
LGTM! :)

PR #2749, opened 2021-08-17
refactor(Tabs2): Implementation clean-ups

laraharrow:
- [x] 📚 Documentation updated
- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11


packages/components/src/Tabs2/types.ts
```diff
@@ -71,31 +71,38 @@ type TabStackMember = Tab2Props & {
...
 type Controlled = {
-  /**
-   * Which tab to show on load
-   */
-  defaultTabId?: string
   /**
    * Controlled: which tab to show now
    */
-  tabId?: string
+  tabId: string
```
laraharrow, line 77: @lukelooker and @dbchristopher is this the correct behavior here, I was getting an error on the stories here.

Type '{ children: Element[]; defaultTabId?: string | undefined; tabId?: string | undefined; onTabChange?: ((tabId: string) => void) | undefined; distributed?: boolean | undefined; } | { ...; }' is not assignable to type 'IntrinsicAttributes & Tabs2Props'. Type '{ children: Element[]; defaultTabId?: string | undefined; tabId?: string | undefined; onTabChange?: ((tabId: string) => void) | undefined; distributed?: boolean | undefined; }' is not assignable to type 'IntrinsicAttributes & Tabs2Props'. Type '{ children: Element[]; defaultTabId?: string | undefined; tabId?: string | undefined; onTabChange?: ((tabId: string) => void) | undefined; distributed?: boolean | undefined; }' is not assignable to type 'Uncontrolled'. Types of property 'defaultTabId' are incompatible. Type 'string | undefined' is not assignable to type 'undefined'. Type 'string' is not assignable to type 'undefined'.

packages/components/src/Tabs2/Tabs2.test.tsx
```diff
@@ -68,34 +66,16 @@ describe('Tabs2', () => {
...
+    renderWithTheme(<Basic />)
+    expect(
+      screen.queryByText("Here's awesome story about cats")
+    ).toBeInTheDocument()
+    expect(screen.queryByText('Humans tab is disabled')).not.toBeInTheDocument()
+    fireEvent.click(screen.getByText('Human'))
+    expect(
+      screen.queryByText("Here's awesome story about cats")
+    ).toBeInTheDocument()
+    expect(screen.queryByText('Humans tab is disabled')).not.toBeInTheDocument()
```
dbchristopher, line 78: nice cleanup 👍

packages/components/src/Tabs2/Tabs2.test.tsx
```diff
@@ -32,14 +32,12 @@ import { composeStories } from '@storybook/testing-react'
 import { fireEvent, screen } from '@testing-library/react'
 import * as stories from './Tabs2.story'
 import { Controlled } from './Tabs2.story'
-import { Tabs2 } from './Tabs2'
```
dbchristopher, line 35: Now that we've set up some restricted prop combinations in typescript, we should validate that it validates props correctly. This would be a great use for `@ts-expect-error` which I found out about fairly recently. I was just playing around with it, and this seems to work:

```ts
  test('validates controlled vs uncontrolled prop combinations', () => {
    renderWithTheme(
      // @ts-expect-error: onTabChange required when tabId is set
      <Tabs2 tabId="3">
        <Tab2 id="tab" label="Tab">
          A single Tab
        </Tab2>
      </Tabs2>
    )

    renderWithTheme(
      // @ts-expect-error: tabId required when onTabChange is set
      <Tabs2 onTabChange="3">
        <Tab2 id="tab" label="Tab">
          A single Tab
        </Tab2>
      </Tabs2>
    )

    renderWithTheme(
      // @ts-expect-error: you can't set both tabId and defaultTabId simultaneously
      <Tabs2 tabId="3" defaultTabId="5">
        <Tab2 id="tab" label="Tab">
          A single Tab
        </Tab2>
      </Tabs2>
    )
  })
```

`@ts-expect-error` is cool because typescript will alert you if you do something like `<Tabs2 tabId="3" defaultTabId="5">` and it actually passes

dbchristopher (changes requested):
The changes look good to me! Looks like snapshots need to be rebuilt, probably because of the additional `disabled` tab added to the story template.

I also suggested a new test we could add to validate the restricted prop combinations

packages/components/src/Tabs2/Tabs2.story.tsx
```diff
@@ -44,6 +44,9 @@ const Template: Story<Partial<Tabs2Props>> = (args) => (
         Cats are way better than dogs. Go to other tab
       </Tab2>
       <Tab2 label="Fish">Are kinda smelly</Tab2>
+      <Tab2 disabled id="human" label="Human">
+        Humans tab is disabled
+      </Tab2>
```
lukelooker, line 49: I don't think this adds much of value to tabs (`disabled` tabs is an edge case we've never actually seen used in the real world) so I'd prefer not to include it in the global example set.

Might be sense to add another story (with storyshots disabled) to show a disabled tab in action so that folks reviewing the stories can see that in a limited scenario.

packages/components/src/Tabs2/Tabs2.test.tsx
lukelooker: 👍

packages/components/src/Tabs2/Tabs2.test.tsx
```diff
@@ -32,14 +32,12 @@ import { composeStories } from '@storybook/testing-react'
...
-import { Tab2 } from './Tab2'
 const { Basic, DefaultTab, Distributed } = composeStories(stories)
 
 describe('Tabs2', () => {
   test('basic', () => {
     renderWithTheme(<Basic />)
-    expect(screen.getAllByRole('tab')).toHaveLength(3)
+    expect(screen.getAllByRole('tab')).toHaveLength(4)
```
lukelooker, line 40:
```suggestion
    expect(screen.getAllByRole('tab')).toHaveLength(3)
```

packages/components/src/Tabs2/Tabs2.test.tsx
```diff
@@ -68,34 +66,16 @@ describe('Tabs2', () => {
...
 
   test('disabled', () => {
-    renderWithTheme(
-      <Tabs2>
-        <Tab2 label="First">1</Tab2>
-        <Tab2 disabled label="Second">
-          2
-        </Tab2>
-      </Tabs2>
-    )
```
lukelooker, line 78: I'd made this change so that Storybook didn't feature a disabled tab by default. 

You could copy this into Storybook as a `Disabled` story and then use that here but as I noted above I'd prefer not to have a `disabled` tab included in the general example of Tabs.

packages/components/src/Tabs2/Tabs2.story.tsx
```diff
@@ -77,3 +80,20 @@ export const Controlled = () => {
...
+  return (
+    <Tabs2 distributed>
+      {tabs.map((value, index) => (
+        <Tab2 label={`Hello World ${index}`} key={index}>
+          This is {value} {index}
+        </Tab2>
+      ))}
+    </Tabs2>
+  )
+}
```
lukelooker, line 95: 👍

lukelooker (changes requested):
Only issue I see is that I'd prefer not have the "Basic" example include a `disabled` tab as that's not a pattern we commonly see (or particularly recommend).

See comments inline for further context.

packages/components/src/Tabs2/Tabs2.test.tsx
laraharrow: the `@ts-expect-error` its a great tool!
thanks for sharing I didn't know about it!

packages/components/src/Tabs2/Tabs2.story.tsx
```diff
@@ -77,3 +77,41 @@ export const Controlled = () => {
...
+      <Tab2 disabled id="human" label="Human">
+        Humans tab is disabled
+      </Tab2>
+    </Tabs2>
+  )
+}
+
+Disabled.parameters = {
+  storyshots: { disable: true },
+}
```
dbchristopher, line 117: I think we would want to capture one snapshot that includes the disabled state, right?

packages/components/src/Tabs2/Tabs2.story.tsx
```diff
@@ -58,9 +58,9 @@ DefaultTab.args = { defaultTabId: 'dogs' }
 
 export const Controlled = () => {
   const [currentTabId, setTabId] = useState('cats')
-
   return (
     <>
+      <p>The current selected tab is: {currentTabId}</p>
```
laraharrow, line 63: I added this as an example for Bryn based on what he wanted from ticket: https://buganizer.corp.google.com/issues/183439880

lukelooker (commented):
Looks good to me. I'll let Daniel have the last word / approval :)

dbchristopher (approved):

PR #2750, opened 2021-08-18
fix(TreeItem): Link-related prop dispersion bug

ElliotTheHuman:

- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/NavTree/NavTreeItem.tsx
+++ b/packages/components/src/NavTree/NavTreeItem.tsx
@@ -84,6 +84,7 @@ export const NavTreeItem = styled((props: NavTreeItemProps) => {
       depth={depth}
       iconGap={iconGap}
       indicatorGap={indicatorGap}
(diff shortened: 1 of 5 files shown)


packages/components/src/NavTree/NavTreeItem.tsx
```diff
@@ -84,6 +84,7 @@ export const NavTreeItem = styled((props: NavTreeItemProps) => {
       depth={depth}
       iconGap={iconGap}
       indicatorGap={indicatorGap}
+      itemRole={props.href ? 'link' : props.itemRole}
```
ElliotTheHuman, line 87: We do an auto change to 'link' mode when an `href` is passed into `NavTree`. I figure it probably makes sense to d the same with `NavTreeItem`.

mdodgelooker (approved):

PR #2752, opened 2021-08-18
feat(Checkbox): add ripple effect

laraharrow:
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/Form/Inputs/Checkbox/Checkbox.tsx
```diff
@@ -94,14 +120,11 @@ const CheckboxLayout = forwardRef(
...
-
-export const Checkbox = styled(CheckboxLayout)`
+  })
+)<CheckboxProps>`
   ${reset}
   ${space}
+  ${rippleStyle}
   height: 1rem;
   position: relative;
   width: 1rem;
```
mdodgelooker, line 130: I think if you update this to the following it should work. That makes it look really big to me, but that's what's in the spec as "Density 0": https://www.figma.com/file/ADmrXwJUbKG0eVAhco8X10/LGM2-Transition?node-id=1470%3A561
```suggestion
  height: ${({ theme: { space } }) => space.u12};
  padding: ${({ theme: { space } }) => space.u3};
  position: relative;
  width: ${({ theme: { space } }) => space.u12};
```

packages/components/src/Form/Inputs/Checkbox/Checkbox.tsx
```diff
@@ -46,19 +52,35 @@ export interface CheckboxProps
...
+      style,
       validationType,
       ...restProps
     } = props
+
+    const {
+      callbacks,
+      className: rippleClassName,
+      style: rippleStyle,
+    } = useRipple({ color: checked ? 'key' : 'neutral' })
```
mdodgelooker, line 72: Actually it looks like the real value you need to determine color is `isChecked` so you'll have to move this down below line 74.

mdodgelooker (commented):
Looks nice! I think you just have to adjust the height & width as well as the padding to make it work. I have a feeling making the checkbox as big as it is in the spec will cause some interesting issues though – something to discuss during code review.

packages/components/src/Form/Inputs/Checkbox/Checkbox.test.tsx
```diff
@@ -26,16 +26,31 @@
 
 import React from 'react'
 import { renderWithTheme } from '@looker/components-test-utils'
-import { fireEvent, screen } from '@testing-library/react'
-import { Checkbox, CheckboxProps } from './Checkbox'
+import { act, fireEvent, screen } from '@testing-library/react'
+import { composeStories } from '@storybook/testing-react'
+import { CheckboxProps } from './Checkbox'
+import * as stories from './Checkbox.story'
+
```
laraharrow, line 33: Updated tests to follow the new stories pattern

packages/components/src/Form/Inputs/Checkbox/Checkbox.test.tsx
```diff
@@ -81,18 +96,49 @@ describe('Checkbox', () => {
...
 
   test('Supports aria-describedby', () => {
-    renderWithTheme(<Checkbox aria-describedby="some-id" id="checkboxID" />)
+    renderWithTheme(<Basic aria-describedby="some-id" id="checkboxID" />)
     expect(screen.getByRole('checkbox')).toHaveAttribute(
       'aria-describedby',
       'some-id'
     )
   })
+
```
laraharrow, line 113: This test is not passing. I can see all the info checked here on storybook, but the tests itself are not passing
Tabs2-controlled

packages/components/src/Form/Inputs/Checkbox/Checkbox.tsx
```diff
@@ -46,19 +52,35 @@ export interface CheckboxProps
   validationType?: ValidationType
 }
 
-const CheckboxLayout = forwardRef(
-  (props: CheckboxProps, ref: Ref<HTMLInputElement>) => {
+export const Checkbox = styled(
```
laraharrow, line 55: Updated the new styled-wraps-everything pattern here.

packages/components/src/Form/Inputs/Checkbox/Checkbox.tsx
```diff
@@ -94,17 +120,15 @@ const CheckboxLayout = forwardRef(
         </FauxCheckbox>
```
jhardy, line 120: One thing I noticed in storybook is this FauxCheckbox is really large 
image

Spec calls for this to be 18px by 18px

laraharrow: yes I had changed the height width and padding to support the LGM2 design and so we could see the ripple better after chatting with Meredith. But after our convo I'm changing things back ti its original sizes.

PR #2753, opened 2021-08-19
feat: Rename @looker/eslint-config to @looker/eslint-config-oss

lukelooker:
Rename `@looker/eslint-config` to `@looker/eslint-config-oss` to reduce confusion within internal product about the "scope" of these presets as well as avoiding naming conflict.


yarn.lock
```diff
@@ -2528,6 +2448,35 @@
...
+    eslint-plugin-prettier "3.4.0"
+    eslint-plugin-promise "^5.1.0"
+    eslint-plugin-react "^7.23.2"
+    eslint-plugin-react-hooks "^4.2.0"
+    eslint-plugin-sort-keys-fix "^1.1.1"
+    eslint-plugin-standard "5.0.0"
+    eslint-plugin-testing-library "4.1.1"
+    prettier "^2.3.1"
+    typescript "4.3.4"
+
```
mdodgelooker, line 2479: This is here now b/c the filter packages are pulling in `@looker/eslint-config` not `@looker/eslint-config-oss`. Is there a down side to updating those?

mdodgelooker (commented):
Looks good but I'm just curious about the filter packages.

yarn.lock
lukelooker: I don't want to update them here because our repo isn't the "source of truth" for those. We'll do a bit of a loop-de-loop:

1. Publish new package
2. Consume new package in source-of-truth
3. Re-sync filer-* packages into this monorepo

All of this drama should be a one-time exercise as we move towards the consolidated monorepo.

mdodgelooker: Sounds good!

mdodgelooker (approved):

PR #2754, opened 2021-08-19
feat(SelectMulti & InputChips): Support value formatting

mdodgelooker:
Adds a `formatInputValues` prop to `InputChips` and `SelectMulti` (which contains `InputChips`). 

The default is `value.trim()`, to match the current behavior, which trims whitespace from inputted values before validating them – `false` will leave the whitespace intact.

### DIFF
--- a/packages/components/src/Form/Inputs/InputChips/InputChips.tsx
+++ b/packages/components/src/Form/Inputs/InputChips/InputChips.tsx
@@ -48,8 +48,8 @@ export interface InputChipsValidationProps {
    */
   validate?: (value: string) => boolean
   /**
-   * Format each value entered, before validation, defaults to `value.trim()`,
-   * set to `false` to avoid trimming whitespace
+   * Callback to format each value entered, before validation.
+   * Defaults to `value.trim()`, set to `false` to avoid trimming whitespace.
    */
   formatInputValue?: FormatInputValue
   /**
--- a/packages/components/src/Form/Inputs/InputChips/InputChips.tsx
+++ b/packages/components/src/Form/Inputs/InputChips/InputChips.tsx
@@ -41,16 +41,17 @@ import { InputChipsBase } from './InputChipsBase'
  * but also allows (validated) user inputs to be stored as 'chips' (see the Chip element)
  */
 
+export type FormatInputValue = ((value: string) => string) | false
 export interface InputChipsValidationProps {
   /**
    * for checking each value before converting to a chip
    */
   validate?: (value: string) => boolean
   /**
-   * Trim whitespace from entered values (before validation)
-   * @default true
+   * Format each value entered, before validation, defaults to `value.trim()`,
+   * set to `false` to avoid trimming whitespace
    */
-  trimInputValues?: boolean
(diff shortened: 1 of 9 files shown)


packages/components/src/Form/Inputs/InputChips/InputChips.tsx
```diff
@@ -121,133 +127,132 @@ export function validateValues(
...
-
-    const [uncontrolledValue, setUncontrolledValue] = useState('')
-    const inputValue = isControlled
-      ? controlledInputValue || ''
-      : uncontrolledValue
-
-    const setInputValue = (
-      val: string,
-      event?: FormEvent<HTMLInputElement>
+export const InputChips = styled(
```
mdodgelooker, line 130: "Hide whitespace changes" is helpful with this section. Just implementing our new "wrap it in a styled" pattern while I'm in here.

packages/components/src/Form/Inputs/InputChips/InputChips.tsx
```diff
@@ -63,7 +68,7 @@ export interface InputChipsValidationProps {
   onDuplicate?: (values: string[]) => void
 }
 
-export function splitInputValue(inputValue: string) {
+export const splitInputValue = (inputValue: string) => {
```
mdodgelooker, line 71: Updating to our standard of arrow functions over function keyword.

www/src/documentation/components/forms/select-multi.mdx
```diff
@@ -88,7 +89,7 @@ callbacks to handle invalid values.
     setMessage('')
   }
   return (
-    <SpaceVertical>
+    <SpaceVertical align="stretch">
```
mdodgelooker, line 92: The input was looking awkwardly small without this.

lukelooker (dismissed):
Code looks good and implementation works as-described.

Out of curiosity... do you think there's a use case for `trimInputValues={false`} - I wonder if eventually we should have a `inputFormatCallback` that would let developers specify any formatting callback they want which would default to trimming but could be set as `undefined` to disable or given a more elaborate scheme.

☝️  Would definitely be a _future_ way to address this issue - just pondering if there's a scenario where you wouldn't want the value trimmed.

mdodgelooker:
> Out of curiosity... do you think there's a use case for `trimInputValues={false`} - I wonder if eventually we should have a `inputFormatCallback` that would let developers specify any formatting callback they want which would default to trimming but could be set as `undefined` to disable or given a more elaborate scheme.

That is an excellent question / idea. The ask is to be able to turn off the trimming that we currently do: https://buganizer.corp.google.com/issues/194944432

But I do like the idea of making it more open-ended as a function, defaulting to `trim` but able to be disabled with `false`. Let me give that a quick shot – stay tuned.

mdodgelooker:
@lukelooker I made the prop a function instead and I definitely like this better. I can think of use cases for it, like converting all values to uppercase, sanitizing or escaping certain characters, etc. How does it look to you?

lukelooker (approved):
I really like the flexibility of this solution. LGTM!

PR #2755, opened 2021-08-19
feat: Update filter packages

mdodgelooker:
Sync filter packages


packages/filter-expressions/src/utils/string/escape_leading_and_trailing_whitespaces.ts
```diff
@@ -0,0 +1,65 @@
...
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ */
+
+/**
+ * This function is ripped from legacy (filters_serice.js) and escapes leading
+ * and trailing whitespaces with a caret for the given string. Trailing whitespaces
+ * are doubly escaped.
+ *
```
lukelooker, line 31: Might be good to sanitize this a bit since OSS folk won't have any context here.

lukelooker (approved):
Noticed one comment that probably could get sanitized a bit (doesn't really make any sense in the OSS context).

Are there any notable features or bug fixes that should get included in the CHANGELOG - if so could those get put in the squash-commit body?

packages/filter-expressions/src/utils/string/escape_leading_and_trailing_whitespaces.ts
mdodgelooker: Good call. I've got a core-product-side PR just starting CI now with this change and a few others.

mdodgelooker:
> Are there any notable features or bug fixes that should get included in the CHANGELOG - if so could those get put in the squash-commit body?
Yes, one major feature/bugfix and a couple minor bugfixes. I'll gather everything of note for the commit body once this is ready to merge. I think there may still be a snapshot issue that I need to figure out.

mdodgelooker:
@lukelooker this is ready for re-review. I figured out the snapshot issue. It was legit – there was a major feature (bugfix?) recently that renders an advanced filter if the expression is incompatible with the selected "control" filter in the config object. In this case, the story had `'slider'` in the config but the expression was `'[0,20],>30'`. I updated the story here and in the source-of-truth-cleanup PR (which is also passing CI).

lukelooker (approved):
Looks good! :)

PR #2757, opened 2021-08-20
refactor: Rename all test files to use *.spec.* file names

lukelooker:
Also alters Jest configuration to ONLY load *.spec.* files to discourage the addition of files using the previous naming convention.

Change was fully automated using the following shell commands:

```
find . -name '*.test.tsx' -exec bash -c ' mv $0 ${0/\test/spec}' {} \;
find . -name '*.test.ts' -exec bash -c ' mv $0 ${0/\test/spec}' {} \;
```


ElliotTheHuman (approved):
LGTM!

PR #2758, opened 2021-08-20
test: Enable @typescript-eslint/consistent-type-imports

lukelooker:
- Enables `@typescript-eslint/consistent-type-imports' in `@looker/eslint-config-oss`
- Applies the result of `yarn lint:es --fix` to all Typescript source in repository

NOTE: This PR is _big_ so review on a per-commit basis:

image

1. Enables the rule itself and bumps up ESLint versions for consistent support with the core-product monorepo
2. Auto-fixes
3. Fixes the auto-generated code Icon components


mdodgelooker (approved):
LGTM!

PR #2764, opened 2021-08-22
fix(PopoverContent): Properly absorb `p` value (also `DialogContent`)

lukelooker:
Fixes the way `p` is absorbed into the `PopoverContent` and `DialogContent` so that it overrides any default values properly.

Before the fix: `<PopoverContent p="xxlarge">` the `xxlarge` padding wouldn't be applied at all because the component internally had more specific selectors/properties `px` & `py`. 

This changes alter the implementation of `PopoverContent` & `DialogContent` to consistent ingest padding values so that externally specified properties act in a predictable manner.


jhardy (approved):
LGTM

mdodgelooker (approved):
LGTM

PR #2765, opened 2021-08-23
fix(SelectOptions): remove Checkbox component

laraharrow:

- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Form/Inputs/Checkbox/FauxCheckbox.tsx
+++ b/packages/components/src/Form/Inputs/Checkbox/FauxCheckbox.tsx
@@ -26,10 +26,17 @@
 
 import styled from 'styled-components'
 
-export const FauxCheckbox = styled.div`
+export type FauxCheckboxProps = {
+  isSelected?: boolean
+}
+
+export const FauxCheckbox = styled.div<FauxCheckboxProps>`
   align-items: center;
   background: currentColor;
+  background-color: ${({ isSelected, theme }) =>
+    isSelected && theme.colors.key};
   border: solid 1px ${({ theme }) => theme.colors.ui2};
+  border-color: ${({ isSelected, theme }) => isSelected && theme.colors.key};
   border-radius: ${({ theme }) => theme.radii.small};
   color: ${({ theme }) => theme.colors.keyText};
   display: flex;
--- a/packages/components/src/Form/Inputs/Combobox/ComboboxMultiOption.tsx
+++ b/packages/components/src/Form/Inputs/Combobox/ComboboxMultiOption.tsx
@@ -31,7 +31,8 @@ import type { Ref } from 'react'
 import React, { forwardRef } from 'react'
 import styled from 'styled-components'
 import { useForkedRef } from '../../../utils'
-import { Checkbox } from '../Checkbox'
+import { FauxCheckbox } from '../Checkbox/FauxCheckbox'
+import { CheckMark } from '../Checkbox/CheckMark'
 import type { ComboboxMultiContextProps } from './ComboboxContext'
 import { ComboboxMultiContext } from './ComboboxContext'
 import {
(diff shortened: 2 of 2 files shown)


packages/components/src/Form/Inputs/Checkbox/FauxCheckbox.tsx
```diff
@@ -26,10 +26,17 @@
 
 import styled from 'styled-components'
 
-export const FauxCheckbox = styled.div`
+export type FauxCheckboxProps = {
```
laraharrow, line 29: This was added to support the use of this component on ComboboxMultiOption.

packages/components/src/Form/Inputs/Combobox/ComboboxMultiOption.tsx
```diff
@@ -40,89 +41,94 @@ import {
...
-      indicator,
-      highlightText = true,
-      scrollIntoView,
-      ...props
-    }: ComboboxOptionProps,
-    forwardedRef: Ref<HTMLLIElement>
-  ) => {
-    const { label, value } = props
+export const ComboboxMultiOption = styled(
+  // eslint-disable-next-line react/display-name
```
laraharrow, line 51: this was added to fix the TS error of display-name request

packages/components/src/Form/Inputs/Combobox/ComboboxMultiOption.tsx
```diff
@@ -40,89 +41,94 @@ import {
...
-ComboboxMultiOptionInternal.displayName = 'ComboboxMultiOptionInternal'
-
-export const ComboboxMultiOption = styled(ComboboxMultiOptionInternal).attrs(
+          <ComboboxOptionIndicator
+            indicator={indicator}
+            isActive={isActive}
+            isSelected={isSelected}
+            isMulti={true}
+          >
+            <FauxCheckbox isSelected={isSelected}>
```
laraharrow, line 100: This component didn't need the complexity of Checkbox, just the displayed element. This change makes it so Changes to Checkbox won't affect this component.

packages/components/src/Form/Inputs/Checkbox/FauxCheckbox.tsx
```diff
@@ -26,10 +26,17 @@
...
+  isSelected?: boolean
+}
```
laraharrow, line 31: The changes added here is to support the transition of `ComboboxMultiOption` to not using the Checkbox component. The component and its complexities are irrelevant to `ComboboxMultiOption` all it needs is a place holder in the shape of a checkbox and `FauxCheckbox` is the perfect solution for that.

packages/components/src/Form/Inputs/Combobox/ComboboxMultiOption.tsx
```diff
@@ -46,68 +47,66 @@ import { useOptionEvents } from './utils/useOptionEvents'
...
-      children,
-      indicator,
-      highlightText = true,
-      scrollIntoView,
-      ...props
-    }: ComboboxOptionProps,
-    forwardedRef: Ref<HTMLLIElement>
-  ) => {
-    const { label, value } = props
+export const ComboboxMultiOption = styled(
```
laraharrow, line 50: updated code to follow the new pattern on `styled`. While doing that a TS error was coming up, display-name was missing. The comment on line 51 is to mitigate it. Open to new ideas here - Couldn't figure out why the error was being trigged.

packages/components/src/Form/Inputs/Combobox/ComboboxMultiOption.tsx
```diff
@@ -46,68 +47,66 @@ import { useOptionEvents } from './utils/useOptionEvents'
...
-ComboboxMultiOptionInternal.displayName = 'ComboboxMultiOptionInternal'
-
-export const ComboboxMultiOption = styled(ComboboxMultiOptionInternal).attrs(
+          <ComboboxOptionIndicator
+            indicator={indicator}
+            isActive={isActive}
+            isSelected={isSelected}
+            isMulti={true}
+          >
+            <FauxCheckbox isSelected={isSelected}>
```
laraharrow, line 100: Simplifying this work by removing Checkbox from here. FauxCheckbox renders as a checkbox but without the complexity of the component.

packages/components/src/Form/Inputs/Combobox/ComboboxMultiOption.tsx
mdodgelooker: That is so weird because `ButtonBase` uses `forwardRef` too but no error in that file. I'm fine with the comment if you are, @lukelooker.

packages/components/src/Form/Inputs/Checkbox/FauxCheckbox.tsx
```diff
@@ -26,10 +26,17 @@
...
+
+export const FauxCheckbox = styled.div<FauxCheckboxProps>`
   align-items: center;
   background: currentColor;
+  background-color: ${({ isSelected, theme }) =>
+    isSelected && theme.colors.key};
   border: solid 1px ${({ theme }) => theme.colors.ui2};
+  border-color: ${({ isSelected, theme }) => isSelected && theme.colors.key};
```
mdodgelooker, line 39: This works great but now we have these styles duplicated both here and in `Checkbox.tsx`: https://github.com/looker-open-source/components/blob/main/packages/components/src/Form/Inputs/Checkbox/Checkbox.tsx#L127
which might be confusing if someone needs to update them later.
What do you think about removing them from `Checkbox.tsx` and adding the `isSelected` prop there instead?

mdodgelooker (changes requested):
This looks great! I think it might be good to remove the duplicated styles in `Checkbox.tsx` though.

packages/components/src/Form/Inputs/Checkbox/FauxCheckbox.tsx
laraharrow: This is not just repeated code. inside `Checkbox`, the style is referent to `FauxCheckbox` and `input` removing it created a domino effect on a few different components (FieldCheckbox, FieldCheckboxGroup). My thought here is that even thou the styles are very similar they are actually accomplishing different things. As in `Checkbox` has a bigger scope and removing the style from it will change other styled of components using `Checkbox` that is not our intention.

packages/components/src/Form/Inputs/Combobox/ComboboxMultiOption.tsx
laraharrow: just trying following the same pattern for Radio and got the same TS error FYI

packages/components/src/Form/Inputs/Checkbox/FauxCheckbox.tsx
mdodgelooker: You're right, the changes I'm thinking of would cause some breakage. I made a PR to see what breaks: https://github.com/looker-open-source/components/pull/2771
(Sorry, I accidentally pushed that change to your branch instead but then reverted it.)

I'm not quite sure why it breaks the `FieldCheckboxGroup` snapshot but I don't think it's worth the time – what you've done already works just fine. For future reference my reasoning was to:
1. Reduce "contextual overrides" i.e. referencing `FauxCheckbox` inside the style block of another component `Checkbox`, which reduces our ability to control a component from inside the component itself
2. Have a single source of truth for the purple bg & border color styles, in case they ever need to be modified

mdodgelooker (approved):
LGTM!

PR #2770, opened 2021-08-24
fix(IconButton): Make hover/focus background consistent with shape

mdodgelooker:
Implements a "complex default" for `IconButton`'s `shape` prop and make the new ripple styles (including the new hover / focus ring shown even when ripple is turned off) respect that default.


lukelooker (approved):
Looks great. Assuming snapshots concur this is Good-to-go :)

PR #2772, opened 2021-08-24
fix(IconButton): IconButton should not shrink within a flex container

lukelooker:
Prevent `IconButton` from getting squashed within flex layouts.


mdodgelooker (approved):
Perfect! I love that we were even squishing our own icon in `FieldChips`. 😜

PR #2775, opened 2021-08-24
feat(Checkbox): add ripple effect

laraharrow:

- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


laraharrow:
As I was testing the stories to make sure everything is working as expected after adding the ripple effect. I notice that there are are a few stories that seems wrong:
1. **Disabled Check** has the mixed value as oppose of having the check
2. **Disabled Mixed** is not disabled and it has the check as oppose of the mixed value when clicked
3. **ReadOnly** has mixed value as oppose of having the check
4. **ReadOnly Checked** is not checked and it is clickable
5. **ReadOnly Mixed** it has the check value and its clickable

checkbox story

laraharrow:
changes to the snapshots:
all snapshots now will have padding around so the ripple is visible.

mdodgelooker (approved):

PR #2785, opened 2021-08-27
fix(useAccordion2): Ignore changes to defaultOpen

lukelooker:
Currently `useAccordion2` (and by proxy `Accordion2` & `Accordion`) responds to mutations of the `defaultOpen` property. Instead it should be a one-and-done property – changes should be ignored.

### DIFF
--- a/packages/components/src/Accordion2/Accordion2.spec.tsx
+++ b/packages/components/src/Accordion2/Accordion2.spec.tsx
@@ -111,4 +111,26 @@ describe('Accordion2', () => {
       <Accordion2 isOpen {...defaultProps} />
     )
   })
+
+  test('defaultOpen mutation should be ignored', () => {
+    const Example = () => {
+      const [defaultOpen, setDefaultOpen] = React.useState(true)
+
+      return (
+        <>
+          <button onClick={() => setDefaultOpen(!defaultOpen)}>Toggle</button>
+          <Accordion2 defaultOpen={defaultOpen} label="Hello">
+            World
+          </Accordion2>
+          )
+        </>
+      )
+    }
+
+    renderWithTheme(<Example />)
+    expect(screen.getByText('World')).toBeInTheDocument()
+    // Accordion should ignore change
+    fireEvent.click(screen.getByText('Toggle'))
+    expect(screen.getByText('World')).toBeInTheDocument()
+  })
 })
--- a/packages/components/src/Accordion2/useAccordion2.tsx
+++ b/packages/components/src/Accordion2/useAccordion2.tsx
@@ -24,7 +24,7 @@
 
  */
 
-import React from 'react'
+import React, { useEffect } from 'react'
 import { useClickable, useToggle, useWrapEvent } from '../utils'
 import { Accordion2Content } from './Accordion2Content'
 import { accordionDefaults, accordionLeftDefaults } from './accordionDefaults'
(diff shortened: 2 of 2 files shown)


packages/components/src/Accordion2/Accordion2.spec.tsx
```diff
@@ -111,4 +111,25 @@ describe('Accordion2', () => {
...
+          </Accordion2>
+          )
+        </>
+      )
+    }
+
+    renderWithTheme(<Example />)
+    expect(screen.getByText('World')).toBeInTheDocument()
+    fireEvent.click(screen.getByText('Toggle'))
+    expect(screen.queryByText('World')).not.toBeInTheDocument()
```
ElliotTheHuman, line 133: Dumb question, but we want `defaultOpen` to NOT change the open state after initial render right?

In which case, we'd want this line of the test to also be, I think?
```
expect(screen.getByText('World')).toBeInTheDocument()
```

packages/components/src/Accordion2/Accordion2.spec.tsx
```diff
@@ -111,4 +111,25 @@ describe('Accordion2', () => {
...
+    const Example = () => {
+      const [defaultOpen, setDefaultOpen] = React.useState(true)
+
+      return (
+        <>
+          <button onClick={() => setDefaultOpen(!defaultOpen)}>Toggle</button>
+          <Accordion2 defaultOpen={defaultOpen} label="Hello">
+            World
+          </Accordion2>
+          )
```
ElliotTheHuman, line 125: I don't think this closing parenthesis has a corresponding open one. Should be able to just delete it.

packages/components/src/Accordion2/useAccordion2.tsx
```diff
@@ -59,7 +59,13 @@ export const useAccordion2 = ({
   toggleOpen: propsToggleOpen,
   ...props
 }: Accordion2Props) => {
-  const { value, toggle } = useToggle(!!defaultOpen)
+  const { change, value, toggle } = useToggle(false)
+  useEffect(() => {
+    if (defaultOpen !== undefined) {
+      change(defaultOpen)
+    }
+  })
```
ElliotTheHuman, line 67: Should be able to get the desired "one and done" behavior by passing an empty array in as the second argument of this `useEffect`.

ElliotTheHuman (changes requested):

packages/components/src/Accordion2/useAccordion2.tsx
lukelooker: Doh!

packages/components/src/Accordion2/Accordion2.spec.tsx
lukelooker: Done! :)

ElliotTheHuman (approved):
LGTM!!

ElliotTheHuman (changes requested):
Looking at the image snapshot / storybook errors, I think the issue is that the `Tree` / `Accordion` will always initially render as closed, but after that initial render the `useEffect` will fire and then change the state to open. So in snapshot land, you get this weird effect where the accordions / trees start as closed but quickly switch to open (eye tests in storybook support this).

Ideally, we'd just want that initial render to know that it should be open, so maybe a viable option would be to have the existing `defaultOpen` implementation, and then add a state piece that checked for the initial render. If the initial render happened, then use the added state to ignore the `defaultOpen` on subsequent rerenders.

ElliotTheHuman (approved):
LGTM

PR #2795, opened 2021-08-31
fix(Tab2): `Tabs` should be output with `type="button"` to prevent accidental form submission

lukelooker:
Addresses a similar issue with `ListItemContent` as well.

### DIFF
--- a/packages/components/src/ListItem/ListItemContent.tsx
+++ b/packages/components/src/ListItem/ListItemContent.tsx
@@ -40,7 +40,9 @@ import type {
 } from './types'
 import { listItemBackgroundColor, listItemPaddingX } from './utils'
 
-const Button = styled.button`
+const Button = styled.button.attrs(({ type = 'button' }) => ({
+  type,
+}))`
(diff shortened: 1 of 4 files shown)


packages/components/src/ListItem/ListItemContent.tsx
```diff
@@ -40,7 +40,9 @@ import type {
 } from './types'
 import { listItemBackgroundColor, listItemPaddingX } from './utils'
 
-const Button = styled.button`
+const Button = styled.button.attrs(({ type = 'button' }) => ({
+  type,
+}))`
```
ElliotTheHuman, line 45: Didn't know that the default `type` for `button` was "submit" 🤷 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

Good call.

ElliotTheHuman (dismissed):
LGTM

lukelooker:
Added test coverage

ElliotTheHuman (approved):
👍 👍

PR #2796, opened 2021-09-01
feat(Text inputs): Google Material style updates

mdodgelooker:
This grew into a bit of a beast as I went through the various inputs trying to not only match the spec but have a bit more consistency in the code for ease of maintenance. The main changes are:

- Border styles
- Spacing
- Icon size & alignment
  - `Select` error icon is now to the right of the arrow icon
- Font color

Again there's a massive number of snapshot diffs. Apologies and thank you!


packages/components-date/src/InputDateRange/InputDateRange.tsx
```diff
@@ -38,6 +38,8 @@ import { useTranslation } from 'react-i18next'
 import type { ValidationType } from '@looker/components'
 import {
   inputCSS,
+  inputHeight,
+  InputTextContent,
```
mdodgelooker, line 42: The input height was slightly short and the error icon placement was off, so I made some adjustments to more closely follow `InputText`.

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -320,10 +322,6 @@ export const InputText = styled(InputTextLayout).attrs<InputTextProps>(
   ${space}
   ${inputCSS}
 
-  input {
-    font-family: inherit;
-  }
-
```
mdodgelooker, line 326: Moved this to `innerInputStyle` so it can be applied in `InputDateRange` and elsewhere.

packages/components/src/Form/Inputs/Select/SelectInputIcon.tsx
```diff
@@ -25,14 +25,16 @@
...
 import { ComboboxContext } from '../Combobox'
 import type { SelectOptionObject } from './types'
 
 export function getOptionIcon(value: string, options: SelectOptionObject[]) {
   if (value && options) {
     const option = options.find((opt) => opt.value === value)
-    return option?.icon ? <Icon color="text1" icon={option.icon} /> : null
+    return option?.icon ? (
+      <InputTextContent pl="u2">{option.icon}</InputTextContent>
+    ) : null
```
mdodgelooker, line 37: Makes `Select` icon consistent with `InputText` icon.

packages/components/src/Form/Inputs/TextArea/TextArea.tsx
```diff
@@ -95,22 +95,29 @@ export const TextArea = styled(TextAreaLayout).attrs<TextAreaProps>(
   width: 100%;
 
   ${ErrorIcon} {
+    pointer-events: none;
     position: absolute;
-    right: 0.5rem;
-    top: 0.5rem;
+    right: calc(${({ theme }) => theme.space.u3} + 1px);
+    top: calc(${({ theme }) => theme.space.u3} / 2);
```
mdodgelooker, line 101: Trying to get the error icon placement to match `InputText`, etc. Unfortunately we can't use the same wrapper layout styles since we rely on the native textarea resize, which needs to be in the bottom right corner.

packages/filter-components/src/Filter/components/AdvancedFilter/components/MidInputLabel/MidInputLabel.tsx
```diff
@@ -42,7 +42,7 @@ export const MidInputLabel = styled(Text).attrs<TextProps>(
   align-items: center;
   align-self: center;
   background: ${({ theme: { colors } }) => colors.background};
-  border: 1px solid ${({ theme: { colors } }) => colors.ui2};
+  border: 1px solid ${({ theme: { colors } }) => colors.ui4};
```
mdodgelooker, line 45: Needs to be updated in the core product when this lands there. (I added a ticket.)

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -267,40 +261,45 @@ export const InputTextContent = styled.div<SpaceProps>`
...
   ${StyledIconBase} {
     color: ${({ theme }) => theme.colors.text1};
-    ${InputIconSize}
   }
 
   ${ErrorIcon} {
     color: ${({ theme }) => theme.colors.critical};
   }
+  svg {
+    ${InputIconSize}
```
mdodgelooker, line 270: Moved this here so it will also apply to custom artwork (e.g. in `Select`).

packages/components-date/src/InputDateRange/InputDateRange.tsx
```diff
@@ -503,21 +507,22 @@ const InputTextGroupWrapper = styled.div<InputTextGroupWrapperProps>`
...
-    height: ${({ theme }) => theme.lineHeights.large};
-
+    margin: ${({ theme }) => theme.space.u05} 0;
     &:focus-within {
       background: ${({ theme }) => theme.colors.keyAccent};
     }
+    input,
+    span {
+      padding: 0 ${({ theme }) => theme.space.u2};
+    }
```
dbchristopher, line 525: Possible design nit: it looks like the selected date range appears overtop of the focus ring when I tab to the calendar input. It's kinda cool actually! But I just wanted to double check if that was intentional.
Screen Shot 2021-09-08 at 2 24 00 PM

mdodgelooker:
@lukelooker for quick reference the link to the spec is in the latest comment on the ticket.

dbchristopher (approved):
This looks great! Left a comment about a small thing with the calendar. Happy to re-approve if that was not intentional styling and you wanted to fix it

packages/components-date/src/InputDateRange/InputDateRange.tsx
mdodgelooker: Oh good catch! Not intentional. I'll see if I can fix that.

packages/components/src/Form/Inputs/innerInputStyle.ts
```diff
@@ -29,8 +29,9 @@ import { css } from 'styled-components'
 export const innerInputStyle = css`
   background: transparent;
   border: none;
-  caret-color: ${({ theme }) => theme.colors.text5};
+  caret-color: ${({ theme }) => theme.colors.key};
```
lukelooker, line 32: This change is intentional? (just checking, not disagreeing)

lukelooker (approved):
I think you might win "most snapshots changed in a single PR" with this one :)

Besides the `DateRange` issue that Daniel mention the code and snapshot changes look good :)

packages/components/src/Form/Inputs/innerInputStyle.ts
mdodgelooker: Yep, it's a super subtle thing that Material does (right @jhardy?) – the cursor is `key`, or `critical` for error state.

packages/components-date/src/InputDateRange/InputDateRange.tsx
mdodgelooker: Calendar has a transparent border to leave space for the focus ring, which turns it key color. I just made it thicker (1px -> 2px) instead of adding the inset box shadow like the inputs have. Hopefully that's ok @jhardy ?

packages/components/src/Form/Inputs/innerInputStyle.ts
jhardy: +1 to what @mdodgelooker says, its the correct behavior

lukelooker: 👍

lukelooker (approved):
🚢

packages/components-date/src/InputDateRange/InputDateRange.tsx
jhardy: The 2px feels a bit heavy to me. But I am ok with it if it makes this change easier. I'd like to revisit our date controls a bit in the future and we could address it then.

PR #2797, opened 2021-09-01
refactor: Deprecate `prebuild` step, check-in generated `@looker/icons` source

lukelooker:
In an effort to simplify the build steps for the library this PR removes the `prebuild` step (only leveraged by the `@looker/icons` package). To do so:

- Altered the package configuration to allow the generated `.tsx` files in `packages/icons/src` to be recognized by Git (no longer included in `.gitignore`
- Added lint check to verify that `packages/icons/src` reflects the output of `yarn workspace @looker/icons build`
- Added root-`package.json` lint check to run any package-specific lint scripts

### Lint Error Example

lint error

### Code Review

NOTE: This PR is most easily reviewed on a per-commit basis:

image


packages/icons/bin/lint.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Do a fresh build
+yarn workspace @looker/icons build > /dev/null
+
+# Check if any changes are present in src
+
+if [ -z "$(git status src --short)" ]; then
+  echo "All files in packages/icons/src committed. 👍"
```
ElliotTheHuman, line 9: So if `git status src --short` doesn't output anything / character length is 0, then the icons package is unchanged and lint won't error out, correct?

package.json
```diff
@@ -22,7 +22,7 @@
     "yarn": ">=1.14.0"
   },
   "scripts": {
-    "prebuild": "yarn clean && yarn lerna run prebuild  --stream",
```
ElliotTheHuman, line 25: Just to make sure I'm following, `yarn lerna run prebuild --stream` would trigger `yarn clean:icons` and `yarn pretest` I imagine since we're removing all 3? And those aren't necessary anymore because of the changes to the icons package?

ElliotTheHuman (approved):
LGTM

package.json
lukelooker: So `prebuild` is called _before_ `build` (Yarn does this automatically calling `pre{whateverScriptName` here if it exists before executing `whateverScriptName` - this also works for `post{whateverScriptName}` BTW

`yarn clean` calls the `clean` script below which _used_ to include `clean:icons` but since we now keep the generated source we don't want or need to remove those any longer.

`yarn pretest` is no longer need because there's nothing that needs to be done before the checked-out source is ready to roll (except running `yarn install` but that prerequisite already existed / remains unchanged)

packages/icons/bin/lint.sh
lukelooker: Yep!

PR #2801, opened 2021-09-02
test: Add CI check to validate that yarn.lock file reflects latest package.json and has been deduplicated

lukelooker:
- Verify `yarn.lock` status within CI scripts
- Add yarn-deduplicate a devDependency

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,14 +24,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: c-hive/gha-yarn-cache@v1
       - run: yarn install
-      - run: yarn lint:ts
-        name: 'Typescript'
-      - run: yarn lint:es
-        name: 'ESLint'
-      - run: yarn lint:css
-        name: 'Stylelint'
-      - run: yarn lint:circular
-        name: 'Madge'
+      - run: yarn lint
   test:
     name: "Tests: Jest + RTL"
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -24,14 +24,7 @@ jobs:
...
       - run: yarn install
-      - run: yarn lint:ts
-        name: 'Typescript'
-      - run: yarn lint:es
-        name: 'ESLint'
-      - run: yarn lint:css
-        name: 'Stylelint'
-      - run: yarn lint:circular
-        name: 'Madge'
+      - run: yarn lint
```
lukelooker, line 27: Breaking apart the lint jobs makes the output easier to view in CI but also means that it's easy for new lint jobs to not be "picked up" by our CI if it's not added in two places.

Consolidating command to reduce duplication seems preferable than the introducing the possibility of lint checks failing to be called.

package.json
```diff
@@ -36,7 +36,7 @@
     "clean:gatsby": "rm -Rf www/.cache www/public",
     "clean:coverage": "rm -rf coverage",
     "develop": "echo -e 'Use `yarn gatsby` to start Gatsby \nUse `yarn storybook` to start Storybook'",
-    "deduplicate": "npx yarn-deduplicate -s fewer yarn.lock && yarn",
+    "deduplicate": "yarn-deduplicate -s fewer yarn.lock",
```
lukelooker, line 39: Add `yarn-deduplicate` as a devDependency since it's now part of our CI run.

mdodgelooker (approved):
LGTM

PR #2802, opened 2021-09-02
test: Remove unneeded eslint-disable react/display-name

lukelooker:
I took at look at the cases where we've had to disable `react/display-name` – in the case of `Checkbox` it appears to be working properly without disabling the rule.


mdodgelooker (approved):
Weird. I was definitely experiencing the error here before, while reviewing that PR – but now it looks good. 👍

PR #2805, opened 2021-09-03
feat(Checkbox & Radio): Google Material styling updates

mdodgelooker:
### Changes to Checkbox & Radio
- Updated border styles for resting, checked, hover/focus & error
- Slightly increased size to 18px, per spec
  - Simplified styles to reflect that this size doesn't change with density (once we implement that)
- Added support for the ripple/halo to be larger than element dimensions (see comments for more detail)

**This PR has a TON of updated snapshots. Please let me know if you see any pixels awry!** 🙏 

- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11


packages/components/src/Form/Fields/FieldInline.tsx
```diff
@@ -95,7 +95,7 @@ const FieldDescription = styled(Paragraph)`
 const InputArea = styled.div`
   grid-column: 1;
   grid-row: 1;
-  padding-right: ${({ theme: { space } }) => space.u2};
+  padding-right: ${({ theme }) => theme.space.u1};
```
mdodgelooker, line 98: Per design spec for -6 density (the _only_ density for now).

packages/components/src/Form/Inputs/Checkbox/CheckMark.tsx
```diff
@@ -27,14 +27,14 @@
 import React from 'react'
 export const CheckMark = () => (
   <svg
-    width="16"
-    height="16"
-    viewBox="0 0 16 16"
+    width="10"
+    height="10"
+    viewBox="0 0 10 10"
```
mdodgelooker, line 32: Crop whitespace around edges to make this work with simplified styles in `FauxCheckbox`. At any density, the dimensions of the checkbox itself (18 x 18) do not change, so we no longer need `position: absolute` and `width: 100%` that were in there to allow for scaling.

packages/components/src/Form/Inputs/Checkbox/CheckMarkMixed.tsx
```diff
@@ -31,9 +31,9 @@ export const CheckMarkMixed = () => {
   const { t } = useTranslation('CheckMarkMixed')
   return (
     <svg
-      width="16"
-      height="16"
-      viewBox="0 0 16 16"
+      width="10"
+      height="10"
+      viewBox="0 0 10 10"
```
mdodgelooker, line 36: See comment above.

packages/components/src/Form/Inputs/Checkbox/Checkbox.tsx
```diff
@@ -43,7 +43,6 @@ import {
 import type { InputProps } from '../InputProps'
 import { pickInputProps } from '../InputProps'
 import type { ValidationType } from '../../ValidationMessage'
-import { inputTextValidation } from '../InputText'
```
mdodgelooker, line 46: Most of `inputTextValidation` either no longer applies to `Checkbox` and `Radio` or never did. No point in sharing.

packages/components/src/Form/Inputs/Checkbox/Checkbox.tsx
```diff
@@ -122,7 +124,7 @@ export const Checkbox = styled(
           onChange={noop} // suppress read-only error as we rely on click rather than change event here
           ref={ref}
         />
-        <FauxCheckbox>
+        <FauxCheckbox isSelected={!!isChecked}>
```
mdodgelooker, line 127: Leverage `isSelected` prop to avoid duplication (see comment below).

packages/components/src/Form/Inputs/Checkbox/Checkbox.tsx
```diff
@@ -149,28 +153,30 @@ export const Checkbox = styled(
...
-  }
-
-  input + ${FauxCheckbox} {
-    ${inputTextValidation}
-  }
-
-  input:checked + ${FauxCheckbox} {
-    background-color: ${({ theme }) => theme.colors.key};
-    border-color: ${({ theme }) => theme.colors.key};
-  }
```
mdodgelooker, line 161: These styles are duplicated in `FauxCheckbox` via the `isSelected` prop (used in `ComboboxMultiOption`).

packages/components/src/Form/Inputs/Checkbox/FauxCheckbox.tsx
```diff
@@ -31,17 +31,15 @@ export type FauxCheckboxProps = {
...
   color: ${({ theme }) => theme.colors.keyText};
-  display: flex;
-  height: 100%;
-  justify-content: center;
-  position: relative;
-  width: 100%;
+  padding: ${({ theme }) => theme.space.u05};
+  svg {
+    display: block;
+  }
```
mdodgelooker, line 44: Resolves [extra spacing below the `svg`](https://stackoverflow.com/questions/24626908/how-to-get-rid-of-extra-space-below-svg-in-div-element) now that it no longer has `position: absolute`.

packages/components/src/Form/Inputs/Combobox/ComboboxMultiOption.tsx
```diff
@@ -98,7 +98,7 @@ export const ComboboxMultiOption = styled(
             isMulti={true}
           >
             <FauxCheckbox isSelected={isSelected}>
-              {isSelected && <CheckMark />}
+              <CheckMark />
```
mdodgelooker, line 101: `CheckMark` is white on a white bg when `isSelected` is false, so no need to not render it. Due to simplified styles (see above), `FauxCheckbox` collapses without it.

packages/components/src/Form/Inputs/Combobox/ComboboxMultiOption.tsx
```diff
@@ -124,9 +124,4 @@ export const ComboboxMultiOption = styled(
   })
 )`
   ${comboboxOptionStyle}
-  ${FauxCheckbox} {
-    height: ${({ theme: { space } }) => space.u4};
-    margin-top: ${({ theme: { space } }) => space.u05};
-    width: ${({ theme: { space } }) => space.u4};
-  }
```
mdodgelooker, line 131: Due to simplified styles in `FauxCheckbox`, these overrides are no longer needed.

packages/components/src/Form/Inputs/Radio/Radio.tsx
```diff
@@ -63,6 +63,9 @@ export const Radio = styled(
         props.checked === true,
         validationType === 'error'
       ),
+      // Only define size for density -6,
+      // to make the halo slightly bigger than the container
+      size: 1.167,
```
mdodgelooker, line 68: See comment about changes in `useRipple` below.

packages/components/src/Ripple/rippleStyle.ts
```diff
@@ -63,7 +63,7 @@ to {
  */
 export const rippleStyle = css`
   outline: none;
-  overflow: hidden;
+  overflow: var(--ripple-overflow);
```
mdodgelooker, line 66: See comment about changes in `useRipple` below.

packages/components/src/Ripple/useRipple.tsx
```diff
@@ -37,13 +37,21 @@ export type UseRippleProps = {
...
    * @default neutral
    */
   color?: ExtendedStatefulColor
+  /**
+   * Decimal multiplier, e.g. 1.5.
+   * Use for unbounded ripple that needs to extend past element dimensions,
+   * don't use with overflow: hidden
+   * @default 1
+   */
+  size?: number
```
mdodgelooker, line 54: At -6 density for `Checkbox` and `Radio`, the size of the halo is slightly larger than the dimensions of the element. Adding a `size` makes this possible.

image

packages/components/src/Form/Inputs/Checkbox/Checkbox.tsx
```diff
@@ -78,6 +77,9 @@ export const Checkbox = styled(
       style: rippleStyle,
     } = useRipple({
       color: inputRippleColor(isChecked !== false, validationType === 'error'),
+      // Only define size for density -6,
+      // to make the halo slightly bigger than the container
+      size: 1.167,
```
mdodgelooker, line 82: At -6 density for `Checkbox` and `Radio`, the size of the ripple/halo is slightly larger than the dimensions of the element. Adding a `size` prop to `useRipple` makes this possible.

The 1.167 number is awkward, but if we use `theme.space.u7` or even 28 instead, `useRipple` would have to measure the element, which is expensive, to come up with 1.167 number on its own.

image

packages/components/src/Form/Fields/FieldToggleSwitch/FieldToggleSwitch.tsx
```diff
@@ -55,6 +55,7 @@ const FieldToggleSwitchLayout = forwardRef(
           id={id}
           validationType={validationMessage && validationMessage.type}
           ref={ref}
+          mr="u1"
```
mdodgelooker, line 58: `ToggleSwitch` needs a bit more space between itself and the label, since it doesn't have the built-in space that `Checkbox` and `Radio` have.

mdodgelooker:
Hi @lukelooker and @jhardy I ended breaking out the "LGM2 border etc" changes on `Radio` & `Checkbox` and wanted to land them first to remove a teensy dependency on `InputText` styles (`inputTextValidation`, specifically). Those changes are coming up next. 🙂 

Please don't hesitate to for 1-pixel, hair-splitting changes if you notice them. I want this to be perfect!

packages/components/src/Form/Inputs/Checkbox/Checkbox.tsx
lukelooker: This makes sense to me but I'm wonder if we can the explain where the number comes from (or pull it in from a `const MAGIC_NUMBER` defined somewhere & explained somewhere centrally)

lukelooker (commented):
Minor nit on explaining the ratio number `1.167` and then one screen question...

On the "SelectionOptions` screenshot the checkbox seems to ride a little high (rather than dead-on vertically centered):

image

It looks like this was true before-hand but the thickened border makes it a smidge more prominent. I'm fine with leaving it as is but happened to notice with the snapshot update.

mdodgelooker:
Thanks for the feedback @lukelooker and @jhardy – ready for re-review. 🙂

lukelooker (approved):
Looks great to me. I'd like Jared to weigh-in as well since he's got the design-eye on this but from a code change perspective I'm happy with this.

packages/components/src/Form/Inputs/Checkbox/CheckMark.tsx
```diff
@@ -27,18 +27,15 @@
...
     xmlns="http://www.w3.org/2000/svg"
   >
     <path
-      d="M4 8L7 11L12 4"
-      stroke="currentColor"
-      strokeWidth="2"
-      strokeLinecap="round"
-      strokeLinejoin="round"
+      d="M5.78626 9.15812L2.20407 5.58594L0.805634 6.99219L5.78626 11.9881L14.7863 2.98812L13.3763 1.57812L5.78626 9.15812Z"
+      fill="white"
```
jhardy, line 38: Just one thought on the updated SVG, should fill be `currentColor` and set to the theme `background` color? I am thinking of cases in dark mode where we don't wan't the checkbox to be white?

jhardy:
This looks really great and I think its ready to go. I just had one small question about the check color and theming.

packages/components/src/Form/Inputs/Checkbox/CheckMark.tsx
mdodgelooker: Ack good catch! Updating...

lukelooker (approved):

PR #2815, opened 2021-09-08
fix(Code): `Code` component should not have a default fontSize, instead inherit via CSS cascade like `Span`

lukelooker:


mdodgelooker (approved):
LGTM!

PR #2817, opened 2021-09-09
chore(deps): Update to Typescript 4.4

lukelooker:


packages/eslint-config-oss/package.json
```diff
@@ -8,9 +8,8 @@
   },
   "sideEffects": false,
   "dependencies": {
-    "@types/vfile-message": "^2.0.0",
```
mdodgelooker, line 11: This is no longer needed?

packages/filter-components/package.json
```diff
@@ -27,12 +27,6 @@
   "@looker/prettier-config": "^1.5.0",
   "@looker/stylelint-config": "^1.5.0",
   "@testing-library/react": "11.1.1",
-  "@types/i18next": "13.0.0",
-  "@types/lodash": "4.14.106",
-  "@types/react": "*",
-  "@types/react-test-renderer": "16.9.1",
-  "@types/styled-components": "*",
-  "@types/styled-system": "*",
```
mdodgelooker, line 35: Can these also be moved to the root package.json in the core product repo? (`react-test-renderer` can just go away completely – I'll make a ticket for that.)

mdodgelooker (approved):
Couple questions, but no objections! 👍

packages/eslint-config-oss/package.json
lukelooker: Apparently not - `vfile-message` package is a transient dependency of Gatsby but regardless the `@types/vfile-message` package is just a "stub" as noted on the package page: https://www.npmjs.com/package/@types/vfile-message (the package has types built-in)

packages/filter-components/package.json
lukelooker: They're actually already there - we'll just want to repeat this change at the source of truth at you noted :)

PR #2821, opened 2021-09-09
chore: Rename all .story.* files to .stories.*

lukelooker:
File rename accomplished with:
`find . -name '*.story.tsx' -exec bash -c ' mv $0 ${0/\story/stories}' {} \;`

Then bulk find & replace for `.story` import and minor config file tweaks.


storybook/.storybook/preview-head.html
```diff
@@ -0,0 +1,10 @@
+<style>
+  .sb-show-main, .sbdocs-wrapper {
+    padding: 0 !important; /* stylelint-disable-line declaration-no-important */
+  }
+  .sbdocs-title {
+    display: none;
+  }
+</style>
+<link rel="preconnect" href="https://fonts.gstatic.com">
+<link href="https://fonts.googleapis.com/css2?family=Roboto+Mono:wght@400;500;600;700&family=Roboto:ital,wght@0,400;0,500;0,700;1,400;1,500;1,700&display=swap" rel="stylesheet">
```
ElliotTheHuman, line 10: Dumb question, but does this file get "loaded" in anywhere / mentioned in a JS file? Or is there just storybook magic that automatically picks up that we have a `preview-head.html` and loads in the relevant fonts / additional styles?

ElliotTheHuman (approved):
LGTM, but seeing a conflict. Happy to re-thumbs up once that's resolved.

storybook/.storybook/preview-head.html
lukelooker: Just "magic" - https://storybook.js.org/docs/react/configure/story-rendering#adding-to-head

ElliotTheHuman (approved):
🎉 🎉 🎉

PR #2829, opened 2021-09-14
docs: Move all demos to separate (unpublished) package

lukelooker:
Moving all demo code used within Gatsby-published web site into a separate (unpublished) package. 

Eventually many of these should be converted into stories but this helps "contain" the demo code into a single, consolidated location.


mdodgelooker (approved):
LGTM!

mdodgelooker:
PS: "More" in the PR title is a type but I don't think it matters much b/c it won't go into the changelog.

bryans99:
If this is not published any suggestions for getting this into the dev portal?

lukelooker:
My thought was to copy it over the same as the MDX files are currently (although admittedly I haven't taken a close look at how that's done). Let's chat if that doesn't seem like a good fit and we can iterate further if needed.

bryans99:
I dont think it would work with the current mdx process but your suggestion does fit in with some of the things I was thinking. The current process clones the components repo into a temp directory, creates nodes using the gatsby-source-filesystem for all MDX files in www/src/documentation. I was NOT able to get it to play nice with tsx files, hence I manually copied them across.

Having said that, this should make the copy process easier but I think the copy process needs to happen before the gatsby process starts. A couple of things come to mind

1. add lerna (or yarn workspaces?) to dev portal and create a sync job the copies from components to portal dev. This is similar to what the marketplace backend does for syncing types from helltool
2. move the clone of the repo outside of gatsby and make a task that runs first. The clone then copies the files into the appropriate package in the dev portal file structure. The package would be git ignored.

Anyway, this is now a job for the dev rel team. @jkaster

PR #2831, opened 2021-09-14
fix(IconButton): Remove useLayoutEffect to avoid SSR warning

mdodgelooker:
Using `ReactDOMServer.renderToString()` on content containing an `IconButton` generates a warning for `useLayoutEffect`: https://github.com/facebook/react/issues/14927

The `useLayoutEffect` hook comes into `IconButton` via `useRipple` -> `useMeasuredElement` -> `useResize`, but actually the last two hooks in that chain aren't necessary. I moved them to a separate hook `useBoundedRipple` that can be used for rectangular elements like `Button` and `ListItem` where measurement is necessary.

I did look into the [`useIsomorphicLayoutEffect`](https://medium.com/@alexandereardon/uselayouteffect-and-ssr-192986cdcf7a) workaround but it wouldn't work in the specific situation that inspired this ticket, because `window` is still defined there.


packages/components/src/Button/IconButton.spec.tsx
```diff
@@ -314,13 +316,21 @@ describe('IconButton', () => {
...
       expect(button).toHaveStyle({
         '--ripple-color': '#71767a',
-        '--ripple-scale-end': '12.041594578792294',
-        '--ripple-scale-start': '12.041594578792294',
-        '--ripple-size': '30px',
-        '--ripple-translate': '165px, 0',
+        '--ripple-scale-end': '1.414',
+        '--ripple-scale-start': '1.414',
+        '--ripple-size': '100%',
+        '--ripple-translate': '0, 0',
```
mdodgelooker, line 322: A square `IconButton` now uses the `size` prop instead of a measurement to make the ripple extend to the corners.

packages/components/src/Ripple/types.ts
```diff
@@ -24,9 +24,65 @@
 
  */
 
+import type { ExtendedStatefulColor } from '@looker/design-tokens'
```
mdodgelooker, line 27: Moved types from `useRipple` into this file.

packages/components/src/Ripple/useBoundedRipple.ts
```diff
@@ -0,0 +1,35 @@
...
+import { useMeasuredElement, useCallbackRef } from '../utils'
+import { useRipple } from './useRipple'
+import type { UseBoundedRippleProps } from './types'
+
+export const useBoundedRipple = (props: UseBoundedRippleProps) => {
+  const [element, ref] = useCallbackRef()
+  const [{ height, width }] = useMeasuredElement(element)
+  const result = useRipple({ ...props, bounded: true, height, width })
+  return { ...result, ref }
+}
```
mdodgelooker, line 35: Currently not used but will be for `Button`, `ListItem`, etc.

packages/components/src/Ripple/useBoundedRipple.ts
```diff
@@ -0,0 +1,35 @@
...
+ SOFTWARE.
+
+ */
+import { useMeasuredElement, useCallbackRef } from '../utils'
+import { useRipple } from './useRipple'
+import type { UseBoundedRippleProps } from './types'
+
+export const useBoundedRipple = (props: UseBoundedRippleProps) => {
+  const [element, ref] = useCallbackRef()
+  const [{ height, width }] = useMeasuredElement(element)
```
lukelooker, line 32: With an eye to the (not-too-distant) future do you think `useMeasuredElement` could be refactored to be more "defensive"? Basically just check if `window` exists and skip calculation if not?

I went ahead and logged a ticket to keep track of this requirement for that work: https://b.corp.google.com/issues/199953481 :)

lukelooker (approved):
Nice fix! :)

packages/components/src/Ripple/useBoundedRipple.ts
mdodgelooker: I think so. But in order to make that change, we should set up a test app that uses SSR. It sounds like [neither `useEffect` nor `useLayoutEffect` should be used with SSR](https://github.com/facebook/react/issues/14927#issuecomment-572720368) so I'm curious to see what falls apart with our components.

PR #2833, opened 2021-09-15
feat(FieldTextArea): Google Material label style

mdodgelooker:
**I've tagged this with `NO-MERGE` since currently the theme setting `externalLabel` is false so you can see this in action as well as the snapshot diffs – but I need to change it to true before merging.**

I've picked `FieldTextArea` as the most straightforward to implement the floating label but I've already got a bunch of other updated fields waiting on other branches, coming soon to a PR near you!

### Question
- Is the `detail` implementation ok? The placement is [inspired by Material](https://material-components.github.io/material-web/demos/textfield/)

- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11


packages/components/src/Form/Fields/FieldSelect/FieldSelect.stories.tsx
```diff
@@ -66,12 +66,24 @@ Basic.args = {
   placeholder: 'Placeholder',
 }
 
+export const Value = Template.bind({})
+Value.args = {
+  ...Basic.args,
+  value: 'gouda',
+}
+
```
mdodgelooker, line 74: Adding a few more stories here to cover the different label positions.

packages/components/src/Form/Fields/FieldTextArea/FieldTextArea.stories.tsx
```diff
@@ -43,9 +46,15 @@ const Template: Story<FieldTextAreaProps> = (args) => (
 export const Basic = Template.bind({})
 Basic.args = { label: 'Text Area' }
 
+export const DefaultValue = Template.bind({})
+DefaultValue.args = { ...Basic.args, defaultValue: 'Default value' }
+
```
mdodgelooker, line 51: More stories to cover the label positions.

packages/components/src/Form/Inputs/Combobox/utils/useBlur.ts
```diff
@@ -72,7 +72,7 @@ export function useBlur<
     const popoverCurrent = listRef ? listRef.current : null
     if (popoverCurrent) {
       const focusInList =
-        popoverCurrent && popoverCurrent.contains(nextFocusTarget as Node)
```
mdodgelooker, line 75: I slightly changed the `getNextFocusTarget` util so you don't have to add `as Node` wherever you use it.

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -237,10 +237,14 @@ export const inputTextFocus = css`
 
 export const inputTextDisabled = css`
   cursor: default;
-  opacity: 0.25;
+  opacity: 0.4;
```
mdodgelooker, line 240: The `0.25` value was from when the borders were darker `ui4`. Now that they're `ui3`, `0.4` is more Material-y.

packages/components/src/Form/Inputs/InputText/InputText.tsx
```diff
@@ -237,10 +237,14 @@ export const inputTextFocus = css`
...
   &:hover {
     border-color: ${({ theme }) => theme.colors.ui3};
   }
+  /* FloatingLabelField handles opacity */
+  [data-disabled='true'] & {
+    opacity: 1;
+  }
```
mdodgelooker, line 247: `FloatingLabelField` also does `opacity: 0.4` for disabled, so that the label can also be faded (as Material does). This bit avoids a double fade with a parent selector `data-disabled="true"`, which `FloatingLabelField` applies to the wrapper.

packages/components/src/utils/useSyncedState.ts
```diff
@@ -0,0 +1,47 @@
...
+  const [state, setState] = useState(prop)
+  const isMountedRef = useRef(false)
+  useEffect(() => {
+    if (isMountedRef.current) {
+      setState(prop)
+    }
+    isMountedRef.current = true
+  }, [prop])
+  return [state, setState]
+}
```
mdodgelooker, line 47: A hook that syncs a state value to a prop value whenever the latter changes. This is useful for `useFloatingLabel` to be able cover controlled and uncontrolled components (`hasValue` prop & state are synced).

packages/components/src/Form/Fields/Field/Field.tsx
```diff
@@ -90,93 +61,38 @@ export const omitFieldProps = (props: FieldProps) => omit(props, fieldPropKeys)
...
-    <Label htmlFor={ariaLabelOnly ? undefined : id} id={`labelledby-${id}`}>
-      {label}
-      {required && <RequiredStar />}
-    </Label>
-  )
-
-  return (
+export const Field = styled(
+  ({
+    autoResize,
```
mdodgelooker, line 66: Need to destruture this and apply to `InputArea` since that component moved to its own file.

packages/components/src/Form/Fields/Field/useFloatingLabel.ts
```diff
@@ -0,0 +1,103 @@
...
+}: UseFloatingLabelProps) => {
+  const [isFocused, setIsFocused] = useState(false)
+  const [hasValue, setHasValue] = useSyncedState(propsHasValue)
+
+  const theme = useContext(ThemeContext)
+  // Limitations of style/CSSProperties type
+  // https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors
+  const style = {
+    ['--label-translate' as any]: `${labelOffset}, ${theme.space.u4}`,
+  }
```
mdodgelooker, line 76: This will be useful for the `before` prop when `FieldText` is converted – since the width of that string/component can be variable, it will measure itself and send the value to `labelOffset`.

packages/design-tokens/src/theme/theme.ts
```diff
@@ -81,13 +81,14 @@ export interface Theme {
...
     density: DensityRamp
+    externalLabel: boolean
   }
 }
 
 export const theme: DefaultTheme = {
   breakpoints,
   colors,
-  defaults: { brandAnimation: false, density: 0 },
+  defaults: { brandAnimation: false, density: 0, externalLabel: false },
```
mdodgelooker, line 91: I'll change this to `true` and update snapshots before merging.

packages/components-date/src/FieldTime/FieldTime.tsx
```diff
@@ -65,6 +66,4 @@ const FieldTimeComponent = forwardRef(
 
 FieldTimeComponent.displayName = 'FieldTimeComponent'
 
-export const FieldTime = styled(FieldTimeComponent)`
-  width: 100%;
-`
```
mdodgelooker, line 70: This was previously working because `className` was being applied to both the field and the input, and it broke when I added `className` to `fieldPropKeys` (see below). I fixed it by moving the width to a prop on `InputTime`.

jhardy (commented):
From a design standpoint, I think this looks great. I had just a few comments, but none are blockers. I'd like to get @lukelooker take on the code, but design has a 👍 from me.

packages/components/src/Form/Fields/Field/FloatingLabelField.tsx
```diff
@@ -0,0 +1,154 @@
...
+              {detail}
+            </FieldDetail>
+          )}
+        </Space>
+      </div>
+    )
+  }
+)`
+  display: ${({ autoResize }) => (autoResize ? 'inline-block' : 'block')};
+  opacity: ${({ disabled }) => (disabled ? '0.4' : '1')};
```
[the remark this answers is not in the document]
mdodgelooker: Good idea 👍

packages/components/src/Form/Fields/Field/types.ts
```diff
@@ -53,8 +81,40 @@ export interface FieldBaseProps {
...
+  inline?: boolean
   /**
    * Holds the type of validation (error, warning, etc.) and corresponding message.
    */
   validationMessage?: ValidationMessageProps
 }
+
+export type HideLabelProps = {
+  /**
+   * Hide label on Field
```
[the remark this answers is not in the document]
mdodgelooker: It wraps it in `VisuallyHidden`. I'll update the comment.

lukelooker (commented):
If you want to flip the default "back" and re-run snapshots I'm ready to review this it's former state.

Looks good as-is.

packages/components/src/Form/Fields/Field/HelperText.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Paragraph } from '../../../Text'
+import { ValidationMessage } from '../../ValidationMessage'
+import type { FieldBaseProps } from './types'
+
+export const HelperText = styled(
+  ({ className, description, id, validationMessage }: FieldBaseProps) => {
+    return (
+      <SpaceVertical
+        pt={description || validationMessage ? 'u2' : 'none'}
+        gap="u1"
```
mdodgelooker, line 39: @lukelooker the remaining snapshot diffs after setting `externalLabel` to true are from using `u1` here per @jhardy.

lukelooker (approved):
Looks great. Ship it! :)

PR #2835, opened 2021-09-16
fix(InputText): Darken disabled opacity

mdodgelooker:
Darkens disabled opacity from `0.25` to `0.4`. The former value was added when the resting border color was `ui4` but now that it's `ui3`, the disabled state is too light.


lukelooker (approved):
LGTM

PR #2842, opened 2021-09-20
feat(Dialog): add methods onAfterClose and OnAfterOpen

laragoogle:
Update Dialog to have a method available before the component opens and after it closes.

- [x] 📚 Documentation updated
- [x] 👾 Browsers tested
  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Dialog/Dialog.spec.tsx
+++ b/packages/components/src/Dialog/Dialog.spec.tsx
@@ -28,6 +28,7 @@ import 'jest-styled-components'
 import React, { useState } from 'react'
 import { renderWithTheme } from '@looker/components-test-utils'
 import {
+  act,
   screen,
   fireEvent,
   waitForElementToBeRemoved,
@@ -184,6 +185,32 @@ describe('Dialog', () => {
     const doneButton = screen.getByText('Done Reading')
     fireEvent.click(doneButton)
     await waitForElementToBeRemoved(() => screen.getByText(/We the People/))
+  })
+
+  test('props onAfterClose and onAfterOpen are called on appropriated time', async () => {
+    jest.useFakeTimers()
+
+    const onAfterClose = jest.fn()
+    const onAfterOpen = jest.fn()
+
+    renderWithTheme(
+      <Dialog
+        onAfterClose={onAfterClose}
+        onAfterOpen={onAfterOpen}
+        content={<SimpleContent />}
+      >
+        <a>Open Dialog</a>
+      </Dialog>
+    )
+
+    fireEvent.click(screen.getByText('Open Dialog'))
+    act(() => {
+      jest.runOnlyPendingTimers()
+    })
+    expect(onAfterOpen).toBeCalled()
+    fireEvent.click(screen.getByText('Done'))
+    await waitForElementToBeRemoved(() => screen.getByText('Dialog content'))
+    expect(onAfterClose).toBeCalled()
   })
 
   test('onClose callback', () => {
--- a/packages/components/src/Dialog/stories/Dialog.stories.tsx
+++ b/packages/components/src/Dialog/stories/Dialog.stories.tsx
@@ -72,6 +72,22 @@ Open.parameters = {
   docs: { disable: true },
 }
 
+export const AfterCloseAfterOpenMethods = Template.bind({})
+AfterCloseAfterOpenMethods.args = {
+  ...Basic.args,
+  onAfterClose: () => {
+    // eslint-disable-next-line no-console
+    console.log('Close')
+  },
+  onAfterOpen: () => {
+    // eslint-disable-next-line no-console
+    console.log('Open')
+  },
+}
+AfterCloseAfterOpenMethods.parameters = {
+  storyshots: { disable: true },
+}
+
 export const MediumContent = Template.bind({})
 MediumContent.args = {
   content: <DialogMediumContent />,
--- a/packages/components/src/Dialog/stories/Dialog.stories.tsx
+++ b/packages/components/src/Dialog/stories/Dialog.stories.tsx
@@ -72,8 +72,8 @@ Open.parameters = {
   docs: { disable: true },
 }
 
-export const AfterCloseAfterOpenMethods = Template.bind({})
-AfterCloseAfterOpenMethods.args = {
+export const AnimationCallbacks = Template.bind({})
+AnimationCallbacks.args = {
   ...Basic.args,
   onAfterClose: () => {
     // eslint-disable-next-line no-console
@@ -84,7 +84,7 @@ AfterCloseAfterOpenMethods.args = {
     console.log('Open')
   },
 }
-AfterCloseAfterOpenMethods.parameters = {
+AnimationCallbacks.parameters = {
   storyshots: { disable: true },
 }
 
--- a/packages/components/src/Dialog/useDialog.tsx
+++ b/packages/components/src/Dialog/useDialog.tsx
@@ -165,15 +165,22 @@ export const useDialog = ({
       ? controlledIsOpen || false
       : uncontrolledIsOpen
 
-  const { busy, className, renderDOM } = useAnimationState(
+  const { busy, className, renderDOM, transitionState } = useAnimationState(
     isOpen,
     defaultOpen ? 'none' : undefined
   )
-
+  // eslint-disable-next-line no-console
+  console.log('transitionState: ', transitionState)
+  // eslint-disable-next-line no-console
+  console.log('className: ', className)
   useEffect(() => {
-    if (className === 'exiting' && onAfterClose) onAfterClose()
-    if (className === 'entered' && onAfterOpen) onAfterOpen()
-  }, [className, onAfterClose, onAfterOpen])
+    if (transitionState === 'entered' && onAfterOpen) onAfterOpen()
+
+    if (transitionState === 'exited' && onAfterClose) {
+      onAfterClose()
+    }
+  }, [transitionState, onAfterClose, onAfterOpen])
+
   const setOpen =
     isControlled && controlledSetOpen
       ? controlledSetOpen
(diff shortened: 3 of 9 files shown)


packages/components/src/Dialog/useDialog.tsx
```diff
@@ -159,6 +171,10 @@ export const useDialog = ({
     defaultOpen ? 'none' : undefined
   )
 
+  useEffect(() => {
+    if (className === 'exiting' && onAfterClose) onAfterClose()
```
laragoogle, line 175: I'm using `exiting` as oppose to `exited` because that value is actually used before `Dialog` is opened as well as after its closed.

lukelooker: 1. I wouldn't tie to the `className` here since we may (at some future point) want to separate the styles attached from the callbacks. Instead I'd extend `useAnimationState` to return an `animationState` that could use here.
2. The `animationState` should get fixed up if it's incorrect currently - we really only want to call `onAfterClose` immediately after the animation has finished the sequence was originally conceived to be:

```Entering => Entered => Exiting => Exited```

**Review by lukelooker — CHANGES_REQUESTED** [review]

File: packages/components/src/Dialog/useDialog.tsx (hunk 1 above)
Comment (laragoogle) [reply packages/components/src/Dialog/useDialog.tsx:175]: 1. `[className](https://github.com/looker-open-source/components/blob/214d0232febb3f8e44ad144722ae13b1fbaa0673/packages/components/src/utils/useAnimationState.ts#L44)` is how the state is passed from `useAnimationState` as far as I can tell.
2. I think the status make sense thou, because when `Dialog` has not been opened it is for all intents and purposes `exited`. Basically we could create a new value to be the default value of `state` meaning when the component is rendered that is what it would be.

File: packages/components/src/Dialog/useDialog.tsx [hunk 2]
```diff
@@ -159,6 +170,20 @@ export const useDialog = ({
...
+
+    if (className === 'exited' && onAfterClose) {
+      if (hasBeenOpenRef.current) {
+        onAfterClose()
+      }
+      hasBeenOpenRef.current = false
+    } else {
+      hasBeenOpenRef.current = true
+    }
+  }, [className, onAfterClose, onAfterOpen])
```
lukelooker, line 185: While className does reflect the state today I'd prefer not to directly tie those together because that may change at future point. For instance in this particular refactor we may want to be able to add another "state" for the callbacks but leave the className behavior as-is to save having to refactor the CSS.

`hasBeenOpenRef.current = false` is a clever workaround but really this could done more cleanly by returning a some kind of state value from the `useAnimation` hook that encapsulates that behavior (on/off) completely within the useAnimation hook rather than having to do inference here via the `useRef`

packages/components/src/Drawer/stories/Drawer.stories.tsx
```diff
@@ -83,6 +83,22 @@ Width.parameters = {
   docs: { disable: true },
 }
 
+export const AfterCloseAfterOpenMethods = Template.bind({})
```
lukelooker, line 86: Ditto for renaming above.

lukelooker (changes requested):

packages/components/src/Dialog/useDialog.tsx
mdodgelooker: I think `useRef` will be necessary in some form because you need to capture when the animation state changes to `'exited'` from one of the other values:
1) Initial render, also `'exited'` – though you might be able around this by creating a new animation state value `'initial'`...
2) `onAfterOpen` or `onAfterClose` props change while animation state remains `'exited'`.

But actually I think it might be good to move both this whole new chunk of code into `useAnimationState` so it can be easily accessed by other components like `Panel`. Thoughts?

packages/components/src/Dialog/stories/Dialog.stories.tsx
```diff
@@ -72,6 +72,22 @@ Open.parameters = {
   docs: { disable: true },
 }
 
+export const AnimationCallbacks = Template.bind({})
+AnimationCallbacks.args = {
+  ...Basic.args,
+  onAfterClose: () => {
+    // eslint-disable-next-line no-console
```
laragoogle, line 79: After all of this I think we still have to use useRef. as I'm testing the component on render this is what comes back from the consologs:
Screen Shot 2021-09-23 at 11 59 57 AM

mdodgelooker: Ah yes, `useAnimationState` is acting funny because of `undefined`. I'll add a comment over there but it may be a moot point if you decide to go back to `useRef`.

The issue with changed animation callbacks getting called at inappropriate times is a concern. If you add this story, you can see it in action:
```

export const ChangingCallbacks = () => {
  const { value, toggle } = useToggle()
  const { value: value2, toggle: toggle2 } = useToggle()
  const onAfterOpen = useCallback(() => {
    alert(value ? 'The changed onAfterOpen' : 'The original onAfterOpen')
  }, [value])
  const onAfterClose = useCallback(() => {
    alert(value2 ? 'The changed onAfterClose' : 'The original onAfterClose')
  }, [value2])
  return (
    <Space>
      <Dialog
        onAfterOpen={onAfterOpen}
        onAfterClose={onAfterClose}
        content={
          <DialogLayout>
            <FieldToggleSwitch
              label="Change onAfterOpen"
              on={value}
              onChange={toggle}
            />
          </DialogLayout>
        }
      >
        <Button>Open Dialog</Button>
      </Dialog>
      <FieldToggleSwitch
        label="Change onAfterClose"
        on={value2}
        onChange={toggle2}
      />
    </Space>
  )
}
```

packages/components/src/utils/useAnimationState.ts
```diff
@@ -105,9 +109,11 @@ export const useAnimationState = (
     }
   }, [isOpen, timingEnter, timingExit, state])
 
+  const definedState = state === undefined ? 'exited' : state
```
mdodgelooker, line 112: There's a bunch of logic looking at `state` in the `useEffect` above, but it would need to look at `definedState` now.

mdodgelooker:
Also, you might notice that `onAfterOpen` seems to get called before the animation. This is due to using CSS transition instead of animation and I think we should address it separately so I made a new ticket.

laragoogle:
@lukelooker what are your thoughts on how to move forward here?
Should we go back to the first approach and use `useRef` or there is a better option here with some refactoring?
@mdodgelooker thanks for the story, that was helpful to see. Also what are your thoughts on what we should do here?

mdodgelooker:
My vote would be to go back the the first version with `useRef` BUT to implement it inside of `useAnimationState` instead of in `useDialog`. That way we can eventually add the callbacks to `Panel` as well.

laragoogle:
@mdodgelooker I think that is what we have right now, I agree with you so I went back to the version using `useRef` but now also using a new prop `transitionState`. Let me know if you see any other change I should do.

lukelooker (dismissed):
This looks great. 

Test coverage is showing a reduction on coverage for `useAnimationState` hook because there are no tests that leverage `onAfterOpen` & `onAfterClose`. Can you add a test for `Dialog` that shows those working so we can get that coverage number back up?

packages/components/src/Dialog/Dialog.spec.tsx
```diff
@@ -186,6 +186,27 @@ describe('Dialog', () => {
     await waitForElementToBeRemoved(() => screen.getByText(/We the People/))
   })
 
+  test('props onAfterClose and onAfterOpen are called on appropriated time', async () => {
```
laragoogle, line 189: Ok we have test for the new functions. expect `onAfterOpen` doesn't seem to be called. I wrote the same dialog example here as a story and that works as we would expect. So I'm not sure why its not recognizing the call here.
The other point is that even with this test `coverage` is telling me that the file is not completely tested.
Screen Shot 2021-09-28 at 10 03 51 AM

packages/components/src/Dialog/Dialog.spec.tsx
```diff
@@ -186,6 +186,27 @@ describe('Dialog', () => {
...
+        onAfterClose={onAfterClose}
+        onAfterOpen={onAfterOpen}
+        content={<SimpleContent />}
+      >
+        <a>Open Dialog</a>
+      </Dialog>
+    )
+
+    fireEvent.click(screen.getByText('Open Dialog'))
+    expect(onAfterOpen).toBeCalled()
```
mdodgelooker, line 204: Since there's a delay before this is called, you'll need to use jest timers. Copy over the 1st & last line from the test below and put the `act` bit right before this line. 
https://github.com/looker-open-source/components/blob/main/packages/components/src/CopyToClipboard/CopyToClipboard.spec.tsx#L74-L85

mdodgelooker (changes requested):

packages/components/src/utils/useAnimationState.ts
```diff
@@ -105,6 +110,21 @@ export const useAnimationState = (
...
+    if (state === 'exited' && onAfterExited) {
+      if (hasBeenOpenRef.current) {
+        onAfterExited()
+      }
+      hasBeenOpenRef.current = false
+    } else {
+      hasBeenOpenRef.current = true
+    }
+  }, [state, onAfterExited, onAfterEntered])
+
```
mdodgelooker, line 127: Looks like the "changing callbacks" issue demonstrated [here](https://github.com/looker-open-source/components/pull/2842#discussion_r715142617) is still an issue for `onAfterOpen`. Just need a slight tweak to fix it:
```suggestion
  const previousStateRef = useRef(state)

  useEffect(() => {
    if (state === 'entered' && previousStateRef.current !== 'entered') {
      onAfterEntered?.()
    }

    if (state === 'exited' && previousStateRef.current !== 'exited') {
      onAfterExited?.()
    }
    previousStateRef.current = state
  }, [state, onAfterExited, onAfterEntered])

```

mdodgelooker (approved):
Awesome! 🚢 it!

PR #2845, opened 2021-09-21
feat: Support specification of `brandAnimation` and `density` via `themeCustomizations`

lukelooker:
Introduce support for `themeDefaults` to be specified via `themeCustomizations` so that consumers have a consistent interface to use for overriding these settings going forward.


mdodgelooker (approved):
LGTM! Thanks for this!

PR #2847, opened 2021-09-21
fix(InputColor): Disabled swatch vs input style mismatch

mdodgelooker:


packages/components/src/Form/Inputs/InputColor/Swatch/Swatch.tsx
```diff
@@ -82,14 +76,11 @@ export const Swatch = styled.div
   ${inputCSS}
   ${width}
   ${height}
-  background-color: ${({ color, disabled }) =>
-    disabled ? disabledSwatchColor(color) : color};
```
mdodgelooker, line 86: `disabledSwatchColor` is no longer needed since `inputTextDisabled` does an opacity fade now.

lukelooker (approved):
LGTM

PR #2848, opened 2021-09-22
feat(Date Components): Accept date-fns locale object

mdodgelooker:
## Deprecated
- `Locales` enum
- `dateFnLocaleMap` map
- `dateStringLocale` prop
- `localization` prop

## Added
- `locale` prop, accepts `date-fns` type `Locale` – replaces `dateStringLocale` and `localization` props
- `dateStringFormat` prop, supports all `date-fns` format patterns – replaces a certain use of `dateStringLocale` (i.e. how we currently use `dateStringFormat={Locales.Japanese}` in filter-components just to get the YYYY-MM-DD format
- `firstDayOfWeek` prop, in case it needs to be different from what the `date-fns` locale has set, (filter-components does)

## Next steps

1. With **v3.0** we can remove deprecated props & exports. Removing `dateFnLocaleMap` specifically will improve bundle size since we will no longer be importing any of the locales from `date-fns` besides `en`.
2. Update `@looker/filter-components` with the new props above
3. Both filter packages: stop exporting every locale's translations (except `en`) and instead support deep importing / dynamic importing those resources.
4. Explore ways for packages (`components-date`, `filter-components`, etc) to store their specific locale resources in the `ComponentsProvider` context.


mdodgelooker:
Tagging @ElliotTheHuman too because you're also working on a date thing at the moment. 🙂

packages/components-date/src/DateTimeFormat/DateTimeFormat.tsx
```diff
@@ -26,60 +26,45 @@
...
-type DateFormats = 'short' | 'medium' | 'long' | 'full'
-
-export interface DateTimeFormatProps {
+export type DateTimeFormatProps = {
   children?: Date
-  format?: DateFormats
-  locale?: LocaleCodes
+  format?: DateFormats | string
+  // TODO delete Locales when enum is removed
+  locale?: Locale | Locales
```
lukelooker, line 39: Should this get marked as `@deprecated` ?

packages/components-date/src/DateTimeFormat/DateTimeFormat.tsx
```diff
@@ -26,60 +26,45 @@
...
-export interface ExtendedDateTimeFormatOptions
-  extends Intl.DateTimeFormatOptions {
-  dateStyle?: DateFormats
-  timeStyle?: DateFormats
+// TODO delete when Locales enum is removed
+const localeIsString = (
+  locale: DateTimeFormatProps['locale']
+): locale is Locales => {
+  return typeof locale === 'string'
 }
```
lukelooker, line 50: Mark as `@deprecated`?

lukelooker (approved):
Awesome update! The scheme makes a lot more sense after this refactor :)

packages/components-date/src/DateTimeFormat/DateTimeFormat.tsx
mdodgelooker: The prop will remain, it'll just get a narrower type after 3.0. Is there a way to mark that? Or should I create a new prop "localeConfig" so I can mark the old one as deprecated?

lukelooker: Oh. Sorry, I misread it – I don't think there's a good way to mark it as "partially deprecated" 

I think it's fine as-is for now.

mdodgelooker: I just added a note in the JSDoc, if anyone reads those. 🙂 Rebased and needing another ✅

lukelooker (approved):

lukelooker:
@mdodgelooker You got bitten by my prettier update. Need to rebase and update. Sorry!

lukelooker (approved):

PR #2849, opened 2021-09-22
fix(FieldInline): Improve accessibility via composition improvements

laragoogle:
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11

### DIFF
--- a/packages/components/src/Form/Fields/Field/FieldInline.tsx
+++ b/packages/components/src/Form/Fields/Field/FieldInline.tsx
@@ -44,13 +44,6 @@ interface FieldInlinePropsInternal extends FieldBaseProps {
   id: string
 }
 
-const amendedChildren = (children: ReactNode, id: string) =>
-  isValidElement(children)
-    ? React.cloneElement(children, {
-        'aria-describedby': `${id}-description`,
-      })
-    : children
-
 const FieldInlineLayout: FC<FieldInlinePropsInternal> = ({
   className,
   children,
@@ -61,23 +54,27 @@ const FieldInlineLayout: FC<FieldInlinePropsInternal> = ({
   required,
   validationMessage,
 }) => {
+  const describedbyId = `${id}-description`
+
+  const inputWithAriaDescribed = isValidElement(children)
+    ? React.cloneElement(children, {
+        'aria-describedby': describedbyId,
+      })
+    : children
+
   return (
     <div className={className}>
-      <InputArea>{amendedChildren(children, id)}</InputArea>
-      <Label htmlFor={id} as="span">
+      <InputArea>{inputWithAriaDescribed}</InputArea>
+      <Label htmlFor={id}>
         <Truncate>{label}</Truncate>
         {required && <RequiredStar />}
       </Label>
-      {detail && <FieldDetail id={`${id}-description`}>{detail}</FieldDetail>}
+      {detail && <FieldDetail>{detail}</FieldDetail>}
       <MessageArea>
         {description && (
-          <FieldDescription id={`${id}-description`}>
-            {description}
-          </FieldDescription>
+          <FieldDescription id={describedbyId}>{description}</FieldDescription>
         )}
-        {validationMessage ? (
-          <ValidationMessage {...validationMessage} />
-        ) : null}
+        {validationMessage && <ValidationMessage {...validationMessage} />}
       </MessageArea>
     </div>
   )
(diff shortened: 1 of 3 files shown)


packages/components/src/Form/Fields/FieldInline.tsx
```diff
@@ -54,20 +54,22 @@ const FieldInlineLayout: FC<FieldInlinePropsInternal> = ({
...
+    <div className={className}>
+      <InputArea aria-describedby={id}>{children}</InputArea>
+      <Label id={id} as="span">
         <Truncate>{label}</Truncate>
         {required && <RequiredStar />}
       </Label>
-      {detail && <FieldDetail>{detail}</FieldDetail>}
-      <MessageArea id={`${id}-describedby`}>
-        {description && <FieldDescription>{description}</FieldDescription>}
+      {detail && <FieldDetail id={id}>{detail}</FieldDetail>}
```
lukelooker, line 63: Not sure how to attach the detail to the input but `id={id}` will definitely bust things.

```suggestion
      {detail && <FieldDetail id={`${id}-detail`}>{detail}</FieldDetail>}
```

lukelooker (changes requested):
Noted issues inline.

Might need some test coverage additions perhaps too?

laragoogle:
I think the code is good here, but I'm having a problem with the tests, seems to be unrelated to the work on this PR.

lukelooker (changes requested):
Pushed a small commit - 81fafaead95ce104a9f7f22c858d9a7e2e30da07 to address feedback. 

- Since `amendedChildren` is really only used within the component I'd just assign it as a const within the function/component itself.
- Build `aria-describedby` ID string once and re-use as `const` to prevent typoes
- Remove `as="span"` (this fixes all but 3 or 4 test breakage issues)
- Only assign `describedbyId` as an `id` in one place (`FieldDescription` ✅ , `FieldDetail` 🚫 )

Once tests are fixed up I think this should be good-to-go.

packages/components/src/Form/Fields/Field/FieldInline.tsx
```diff
@@ -53,21 +54,29 @@ const FieldInlineLayout: FC<FieldInlinePropsInternal> = ({
...
         <Truncate>{label}</Truncate>
         {required && <RequiredStar />}
       </Label>
       {detail && <FieldDetail>{detail}</FieldDetail>}
-      <MessageArea id={`${id}-describedby`}>
-        {description && <FieldDescription>{description}</FieldDescription>}
-        {validationMessage ? (
-          <ValidationMessage {...validationMessage} />
-        ) : null}
+      <MessageArea>
```
lukelooker, line 73: ARIA id needs to be assigned here to keep the "validationMessage" tied to the input for screen-reader announcement.

```suggestion
      <MessageArea id={`${id}-describedby`}>
```

packages/components/src/Form/Fields/FieldCheckbox/FieldCheckbox.spec.tsx
```diff
@@ -30,46 +30,67 @@ import { renderWithTheme } from '@looker/components-test-utils'
...
 
-  const input = screen.getByDisplayValue('example')
-  const id = input.getAttribute('aria-describedby')
-  expect(id).toBeDefined()
-  expect(id).toEqual('describedby-test')
-  expect(container).toHaveTextContent(errorMessage)
+    expect(screen.getByText('describe something here.')).toHaveAttribute(
+      'id',
+      'id-description'
+    )
```
lukelooker, line 94: This connects the description text nicely but the validationMessage is not because the ARIA id was moved. Fix commit incoming.

packages/components/src/Form/Fields/FieldToggleSwitch/FieldToggleSwitch.spec.tsx
```diff
@@ -50,7 +50,7 @@ describe('FieldToggleSwitch', () => {
     const input = screen.getByDisplayValue('example')
     const id = input.getAttribute('aria-describedby')
     expect(id).toBeDefined()
-    expect(id).toEqual('describedby-test')
+    expect(id).toEqual('test-description')
```
lukelooker, line 53: In the interest of reducing downstream snapshot changes I think we should go back to the old format (either's fine, not changing means less "noise" in snapshots and changeset here). Update incoming.

```suggestion
    expect(id).toEqual('describedby-test')
```

lukelooker (changes requested):

laragoogle:
Thanks Luke!

lukelooker (approved):

PR #2851, opened 2021-09-23
test: Improve CI job time

lukelooker:
- Corrects caching for `node_nodules` in CI accelerating each job (30s - 1m savings)
- Moves Jest artifact upload to non-blocking `coverage` task (~2m savings)
- Removes Jest coverage check for core run - now run on separate coverage job (~4m savings)
- `image-snapshots` no longer re-runs on failure (~3m)

## Before

image

## After

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "@types/jest-image-snapshot": "^4.3.1",
     "@types/lodash": "^4.14.172",
     "@types/node": "16.9.0",
+    "@types/puppeteer": "^5.4.4",
     "@types/react": "^17.0.9",
     "@types/react-dom": "^17.0.9",
     "@types/styled-components": "5.1.14",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -79,6 +79,7 @@
     "@types/jest-image-snapshot": "^4.3.1",
     "@types/lodash": "^4.14.172",
     "@types/node": "16.9.0",
+    "@types/puppeteer": "^5.4.4",
```
lukelooker, line 82: This is implicit dependency right now and gets pulled in via a non-hoisted `node_modules` within one of the packages. With cache restoration and not explicitly running `yarn install` for every job it's revealed that this may not end up installed. 

Adding here to make _sure_ it's in place.

mdodgelooker (approved):
Nice!

PR #2853, opened 2021-09-24
fix(MultiFunctionButton): update code to actually support component to…

laragoogle:
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/Button/MultiFunctionButton.tsx
```diff
@@ -76,11 +76,24 @@ export const MultiFunctionButton = forwardRef(
     useEffect(() => {
       const a = aRef.current
       const b = bRef.current
+
```
laragoogle, line 79: Somehow using `ref` was just returning the width value of the alternate but not the children. I did a work around to collect that value from the props or children and alternate, this seems to solve the problem.
Because the stories were actually not behaving correctly both are updated in this pr.

packages/components/src/Button/MultiFunctionButton.tsx
```diff
@@ -76,11 +76,24 @@ export const MultiFunctionButton = forwardRef(
...
       if (a && b) {
+        const aWidth: any = children.props.width
+          ? children.props.width
+          : a.offsetWidth
+        const bWidth: any = alternate.props.width
+          ? alternate.props.width
+          : a.offsetWidth
```
lukelooker, line 86: `any` here is hiding a whole bunch of potential pitfalls.There are a lot of cases where this won't work:

- button doesn't have an explicitly specified width (this is the default actually - it's dynamically determined based on content length)
- button width is a string instead of number e.g.: `10rem` - there's no way for JS to understand that as a numeric comparison

packages/components/src/Button/MultiFunctionButton.tsx
lukelooker: Image snapshots were correct before (smaller of the two buttons _should_ get stretched to match width of the largest), this change breaks that.

lukelooker (changes requested):
Fix isn't working as intended.

The ref does seem to be the underlying issue but the proposed fix doesn't work in many scenarios and causes a new regression to the display.

PR #2868, opened 2021-09-27
feat(FieldSelect, FieldChips, FieldTime, FieldTimeSelect): Google Material label style

mdodgelooker:
- Updates 4 fields to use `FloatingLabelField`:
  - `FieldSelect`
  - `FieldChips`
    - There's no "uncontrolled" version so `hasValue` is derived simply from `props.values`
    - `checkValueOnBlur` is set to `false`, meaning `FloatingLabelField` doesn't need to check for a value in the internal input on blur event, since that is automatically converted to a chip on blur (via `onChange`)
  - `FieldTime`
    - Custom `checkValueOnBlur` since there are actually 3 internal inputs to check
  - `FieldTimeSelect`
- Wrapped all the story templates in `ExtendedComponentsThemeProvider` so that `externalLabel` can be toggled in the controls
- Added a few explicitly `externalLabel: false` stories for each component


packages/components-date/src/FieldTime/FieldTime.stories.tsx
```diff
@@ -38,31 +44,51 @@ export default {
...
+  >
+    <FieldTime {...args} />
+  </ExtendComponentsThemeProvider>
+)
 
 export const Basic = Template.bind({})
-Basic.args = { defaultValue: '14:34', format: '12h', label: 'Label' }
+Basic.args = {
+  defaultValue: '14:34',
+  externalLabel: true,
```
lukelooker, line 61: Rather than adding this here you could add it as a default value to the template (see comment above)

packages/components-date/src/FieldTime/FieldTime.stories.tsx
```diff
@@ -38,31 +44,51 @@ export default {
   title: 'Date / FieldTime',
 }
 
-const Template: Story<FieldTimeProps> = (args) => <FieldTime {...args} />
+const Template: Story<FieldTimeProps & { externalLabel: boolean }> = ({
+  externalLabel,
```
lukelooker, line 48: Doing this would remove the need to make any changes to existing stories.

```suggestion
  externalLabel = true,
```

packages/components-date/src/FieldTimeSelect/FieldTimeSelect.stories.tsx
```diff
@@ -36,26 +40,42 @@ export default {
   title: 'Date / FieldTimeSelect',
 }
 
-const Template: Story<FieldTimeSelectProps> = (args) => (
-  <FieldTimeSelect {...args} />
+const Template: Story<FieldTimeSelectProps & { externalLabel: boolean }> = ({
+  externalLabel,
```
lukelooker, line 44: See proposed revision to `FieldTime` 

```suggestion
  externalLabel = true,
```

lukelooker (changes requested):
Using `externalLabel = true` when deconstructing properties in the stories for each component would remove the need to change existing stories (see inline comment)

The only other thing that struck me as odd is the behavior of the "Placeholder" popping up before I pick a value a-la: 

image

Perhaps this is Material-correct and just read's funny because the placeholder text is "Placeholder" instead of something like "Select a cheese"? I could see a scenario where the label stayed in place until an actual value is selected.

FieldChips looks smooshed as we've discussed. I look forward to `density` :)

packages/components-date/src/FieldTime/FieldTime.stories.tsx
mdodgelooker: If I remove `externalLabel: true` here then it disappears from the storybook controls. I bet there's a different way to add it back though? 
image

mdodgelooker:
@lukelooker I see your point about the placeholder. With text fields, Material seems to allow placeholder but barely mention it in docs: https://material.angular.io/components/form-field/overview#form-field-appearance

But it doesn't seem to exist at all for selects. But the label does move up on focus for selects. Do we want to stop just supporting `placeholder` on `Select`? Or at least I can remove it from our examples...

jhardy:
My guess is that we have quite a few uses of `placeholder`'s in select, I would be hesitant to remove them.

mdodgelooker:
@lukelooker I was trippin about `externalLabel` in the stories – it does show in the controls whether or not you have it in the story args.

As for the `Select` placeholder... it seems like we need to move the label up immediately on focus to be material-y, but disabling `placeholder` might be too risky. Maybe we just discourage its use in the docs (once we turn `externalLabel` off by default in 3.0) or say to only use it with `externalLabe={true}`? Any other ideas?

lukelooker:
> @lukelooker I was trippin about `externalLabel` in the stories – it does show in the controls whether or not you have it in the story args.

👍 

> As for the `Select` placeholder... it seems like we need to move the label up immediately on focus to be material-y, but disabling `placeholder` might be too risky. Maybe we just discourage its use in the docs (once we turn `externalLabel` off by default in 3.0) or say to only use it with `externalLabe={true}`? Any other ideas?

I'm fine with moving ahead as-is – I suspect we'll see people downplay the use of placeholders in designs with this new design and I do think it'll be less disconcerting with _real_ placeholder / label text. :)

lukelooker:
Happy to re-review when the stories are cleaned up (use `externalLabel = true`in template instead of modifying all stories) :)

mdodgelooker:
@lukelooker I think the externalLabel fixes are all in.

lukelooker (approved):
LGTM! :)

(Sorry, missed the update commit)

PR #2869, opened 2021-09-28
feat(FieldSelectMulti): Google Material label style

mdodgelooker:
Use `FloatingLabelField` in `FieldSelectMulti`.

I saved this for its own PR because I somehow thought I would have to deprecate `defaultValues`, but I was wrong. 🙂


laragoogle:
when out of focus the `label` is coming back down in the input and covering the chip
0a204a2e-86df-4209-8acf-ba2bc648633c

packages/components/src/Form/Fields/FieldSelectMulti/FieldSelectMulti.stories.tsx
```diff
@@ -69,12 +79,22 @@ export const Basic = Template.bind({})
...
   description: 'this is the description',
   detail: '5/50',
+  externalLabel: true,
   isFilterable: true,
   label: 'Label',
   options: selectOptions,
   placeholder: 'Search fruits',
 }
 
+export const FloatingLabel = Template.bind({})
```
laragoogle, line 89: this example seems to have a problem, see gif on comments.

packages/components/src/Form/Fields/FieldSelectMulti/FieldSelectMulti.stories.tsx
```diff
@@ -69,12 +79,22 @@ export const Basic = Template.bind({})
...
+FloatingLabel.args = { ...Basic.args, externalLabel: false }
+
+export const Values = Template.bind({})
+Values.args = { ...Basic.args, defaultValues: ['Apples', 'Oranges'] }
+
+export const ValuesFloatingLabel = Template.bind({})
```
laragoogle, line 95: this one seems to have the correct behavior when values are added to the input and it is out of focus.

laragoogle (changes requested):

packages/components/src/Form/Fields/FieldSelectMulti/FieldSelectMulti.stories.tsx
mdodgelooker: Whoops, thanks for catching that. Fix is pushed!

laragoogle:
Hey @mdodgelooker  I'm still seeing the same issue on stories.

laragoogle (approved):
LGTM🚢

PR #2870, opened 2021-09-28
fix(Popover): preventDefault on closing click

mdodgelooker:
`Popover` applies an `onClick` to its child that opens the popover and also calls `e.preventDefault` to avoid unintended actions, like a button submitting a form.

The "closing" click is  more complicated. Once open, a click listener is added to the document,  so that a click anywhere will close the popover. If that click happens to be on the trigger element, `event.stopPropagation` is called and then `return`, which avoids calling the original "opener" `onClick` and re-opening the popover, so `e.preventDefault` doesn't get called. This PR fixes that.

### DIFF
--- a/packages/components/src/Popover/Popover.spec.tsx
+++ b/packages/components/src/Popover/Popover.spec.tsx
@@ -72,6 +72,8 @@ describe('Popover', () => {
       'id',
       'a11y-heading'
     )
+
+    fireEvent.click(document)
   })
 
   test('cloneElement style opens and closes', () => {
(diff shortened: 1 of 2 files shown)


packages/components/src/Popover/Popover.spec.tsx
```diff
@@ -72,6 +72,8 @@ describe('Popover', () => {
       'id',
       'a11y-heading'
     )
+
+    fireEvent.click(document)
```
mdodgelooker, line 76: Unrelated but fixes some noise (`act` warning).

packages/components/src/Popover/Popover.spec.tsx
```diff
@@ -114,24 +116,23 @@ describe('Popover', () => {
...
+
     renderWithTheme(
-      <Popover content={SimpleContent}>
-        <button>Test</button>
-      </Popover>
+      <form onSubmit={mockFormSubmit}>
+        <Popover content={SimpleContent}>
+          <button>Test</button>
+        </Popover>
+      </form>
```
mdodgelooker, line 128: A form with an `onSubmit` and a `button` wrapped in `Popover` were the original inspiration for the ticket.

lukelooker (approved):
LGTM

PR #2873, opened 2021-09-28
feat(FieldDateRange): add label animation to component

laragoogle:
- [ ] 📚 Documentation updated
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


laragoogle:
Here is the first run at it!
1. Should I add documentation on the changes to gatsby?
2. when I check the test coverage, looks looks like its not covered at all, but then the  file it self shows as 100% covered.
95c743db-2dc9-4af0-957e-3d2bc79d7e45

packages/components-date/src/FieldDateRange/FieldDateRange.stories.tsx
```diff
@@ -43,9 +47,14 @@ export default {
   title: 'Date / FieldDateRange',
 }
 
-const Template: Story<FieldInputDateRangeProps> = (args) => (
-  <FieldDateRange {...args} />
-)
+const Template: Story<FieldInputDateRangeProps & { externalLabel: boolean }> =
+  ({ externalLabel, ...args }) => (
```
mdodgelooker, line 51: Set the default for `externalLabel` to true here, then add one or two stories that explicitly have `externalLabel: false`. That way the existing snapshots don't change but we have some new ones to capture the floating label style.

```suggestion
  ({ externalLabel = true, ...args }) => (
```

mdodgelooker (changes requested):
No need to add documentation. We'll do that in a separate PR once all the fields are updated. And I wouldn't worry about testing, I think coverage is good.

A couple issues I saw:
1) Can you use the `.label-down` class to hide the `HyphenWrapper` via CSS? 
image
2) It looks like a custom function for `checkValueOnBlur` will be necessary, because the label moves down if you only clear the "from" date. [Check out `FieldTime` for an example](https://github.com/looker-open-source/components/pull/2868/files#diff-0a1d6048ab4ccaedfbed4956df1f05438e34ea844e90eb8506182006a94b74c3R46)
c8c83ba2-01fb-43d3-9b92-72aee1068e5d
.

packages/components-date/src/FieldDateRange/FieldDateRange.tsx
```diff
@@ -40,17 +40,27 @@ import { InputDateRange } from '../InputDateRange'
...
   (props: FieldInputDateRangeProps, ref: Ref<HTMLInputElement>) => {
     const validationMessage = useFormContext(props)
     const id = useID(props.id)
+
     return (
-      <Field
+      <FloatingLabelField
         {...pickFieldProps(props)}
+        checkValueOnBlur={checkValueOnBlur}
+        hasValue={!!props.defaultValue}
```
mdodgelooker, line 63: This works for uncontrolled but not for controlled. If you add the story below, you'll see the label initially in the wrong position.
image
```js
export const Controlled = () => {
  const [range, setRange] = useState<FieldInputDateRangeProps['value']>({
    from: new Date('May 18, 2020'),
    to: new Date('May 21, 2020'),
  })
  return (
    <ExtendComponentsThemeProvider
      themeCustomizations={{ defaults: { externalLabel: false } }}
    >
      <FieldDateRange
        externalLabel={false}
        label="Controlled"
        value={range}
        onChange={setRange}
      />
    </ExtendComponentsThemeProvider>
  )
}
```

mdodgelooker (changes requested):
Getting close! `hasValue` needs an update.

packages/components-date/src/FieldDateRange/FieldDateRange.tsx
laragoogle: good catch, fixed!

mdodgelooker (approved):
LGTM!

PR #2876, opened 2021-09-30
test: Update prettier config

lukelooker:
Updates prettier config to match the core products configuration (moving code between will be more straightforward with a single, common configuration)

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,5 +1,6 @@
 {
-  "$schema": "http://json.schemastore.org/prettierrc",
   "semi": false,
-  "singleQuote": true
+  "singleQuote": true,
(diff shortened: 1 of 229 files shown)


.prettierrc.json
```diff
@@ -1,5 +1,6 @@
 {
-  "$schema": "http://json.schemastore.org/prettierrc",
   "semi": false,
-  "singleQuote": true
+  "singleQuote": true,
+  "trailingComma": "es5",
+  "arrowParens": "avoid"
```
lukelooker, line 5: The actual change. 

Everything else it auto-fix output.

dbchristopher (approved):

PR #2877, opened 2021-09-30
feat(FieldDate): add label animation to component

laragoogle:
- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components-date/src/FieldDate/FieldDate.tsx
```diff
@@ -38,15 +38,19 @@ import {
...
 
 const FieldDateComponent = forwardRef(
   (props: FieldDateProps, ref: Ref<HTMLInputElement>) => {
     const validationMessage = useFormContext(props)
     const id = useID(props.id)
     return (
-      <Field
+      <FloatingLabelField
         {...pickFieldProps(props)}
+        hasValue={!!props.defaultValue}
```
mdodgelooker, line 53: If you put an initial date value in the controlled example the label will start in the wrong position. I think you can just use the `getHasValue` helper here like in https://github.com/looker-open-source/components/blob/main/packages/components/src/Form/Fields/FieldTextArea/FieldTextArea.tsx#L49
```suggestion
        hasValue={getHasValue(props)}
```

mdodgelooker (changes requested):

packages/components-date/src/FieldDate/FieldDate.stories.tsx
```diff
@@ -106,6 +125,24 @@ Controlled.parameters = {
...
+      <FieldDate
+        externalLabel={false}
+        label="Controlled"
+        value={today}
+        onChange={setToday}
+      />
+    </ExtendComponentsThemeProvider>
+  )
+}
+
```
mdodgelooker, line 145: Can you disable storyshots on this? I just realized that the snapshot will fail each time the month changes because without a starting value, the calendar defaults to "this month"
```suggestion
ControlledFloatingLabel.parameters = {
  storyshots: { disable: true },
}
```

mdodgelooker (changes requested):

packages/components-date/src/FieldDate/FieldDate.tsx
laragoogle: `getHasValue` was complaining about `props`  it seems to work fine with `!!props.defaultValue || !!props.value`

packages/components-date/src/FieldDate/FieldDate.stories.tsx
laragoogle: good point!

laragoogle: I also removed the snapshot for `FloatingLabelNoDefaultValue`, because this one shouldn't have a default value every time its renders the calendar displayed will refer to the month we are in.

mdodgelooker (approved):
LGTM! 🛳

PR #2878, opened 2021-09-30
fix(Dialog): Animation delay on open

mdodgelooker:
Previously, `DialogSurface`, `DrawerSurface`, and `Backlog` were using CSS transition to animate in and out. The problem with this is that the transition _starts_ when the className changes from `'entering'` to `'entered'` which is 300ms after the click to open. The dialog renders immediately on click, then sits hidden for 300ms before animating in.

We want the animation to begin immediately on render, so we have to use CSS animation instead. `PanelSurface` was already doing this.

- [x] 👾 Browsers tested
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11


lukelooker (approved):
Looks great! :)

PR #2879, opened 2021-09-30
fix(Menu): Nested menu parent item background and click

mdodgelooker:
### A couple fixes on a nested menu parent item:
1. The nested menu opens on hover but a user could easily assume they need to click to open it. If they do this, it opens briefly when they hover then quickly closes when they click. I thought this was a regression but apparently we did this intentionally without thinking of this scenario.
2. Awhile back the `aria-selected` prop was moved to `ListItemContent` but the associated background style was not.

### DIFF
--- a/packages/components/src/ListItem/ListItemWrapper.tsx
+++ b/packages/components/src/ListItem/ListItemWrapper.tsx
@@ -43,13 +43,6 @@ export const ListItemWrapper = styled.li.attrs(
   outline: none;
   text-decoration: none;
 
-  /**
-    Styling for items that have nested menus
-   */
-  &[aria-expanded='true'] {
-    background: ${({ theme: { colors } }) => colors.ui1};
-  }
-
   &[disabled] {
     & > * {
       cursor: not-allowed;
--- a/packages/components/src/ListItem/types.ts
+++ b/packages/components/src/ListItem/types.ts
@@ -93,7 +93,7 @@ export type ListItemStatefulProps = {
   /**
    * Indicates an open nested menu
    */
-  'aria-expanded'?: boolean
+  'aria-expanded'?: boolean | 'true' | 'false'
   /**
    * If true, the ListItem will have a "disabled" presentation.
    * @default false
--- a/packages/components/src/ListItem/types.ts
+++ b/packages/components/src/ListItem/types.ts
@@ -91,6 +91,11 @@ export type ListItemColorProp = {
 
 export type ListItemStatefulProps = {
   /**
+   * Indicates an open nested menu
+   * @private
+   */
+  'aria-expanded'?: boolean
+  /**
    * If true, the ListItem will have a "disabled" presentation.
    * @default false
    */
--- a/packages/components/src/ListItem/types.ts
+++ b/packages/components/src/ListItem/types.ts
@@ -91,10 +91,6 @@ export type ListItemColorProp = {
 
 export type ListItemStatefulProps = {
   /**
-   * Indicates an open nested menu
-   */
-  'aria-expanded'?: boolean | 'true' | 'false'
-  /**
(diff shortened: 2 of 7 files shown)


packages/components/src/ListItem/ListItemWrapper.tsx
```diff
@@ -43,13 +43,6 @@ export const ListItemWrapper = styled.li.attrs(
   outline: none;
   text-decoration: none;
 
-  /**
-    Styling for items that have nested menus
-   */
-  &[aria-expanded='true'] {
-    background: ${({ theme: { colors } }) => colors.ui1};
-  }
-
```
mdodgelooker, line 52: Moving this into `listItemBackgroundColor`.

packages/components/src/Menu/Menu.spec.tsx
```diff
@@ -341,10 +341,10 @@ describe('<Menu />', () => {
       fireEvent.click(screen.getByText('French'))
       expect(screen.getByText('Camembert')).toBeVisible()
 
-      // If the the nestedMenu was already opened via hover, click closes both Menus
```
mdodgelooker, line 344: This test tells me that the close on click was intentional, but in retrospect it's super annoying – if you don't already know that the nested menu opens on hover. Whoops.

packages/components/src/Menu/useNestedMenu.tsx
```diff
@@ -111,7 +111,7 @@ export const useNestedMenu = ({
     onClick: useWrapEvent((e: MouseEvent<HTMLLIElement>) => {
       // If there's an onClick in MenuItem props, that wins
       // Otherwise preventDefault to keep the parent Menu open
-      if (nestedMenu && !onClick && !isOpen) {
```
mdodgelooker, line 114: We actually want to `preventDefault` whether or not the nested menu is already open.

packages/components/src/ListItem/types.ts
```diff
@@ -90,6 +90,10 @@ export type ListItemColorProp = {
 }
 
 export type ListItemStatefulProps = {
+  /**
+   * Indicates an open nested menu
+   */
+  'aria-expanded'?: boolean
```
mdodgelooker, line 96: Need this here so `listItemBackgroundColor` can use it.

mdodgelooker:
Oops, I messed up `Tree` with my `aria-expanded` fix. Stay tuned...

lukelooker (commented):
Code looks good but test / lint is unhappy:

image

`Booleanish` is a new one :)

packages/components/src/ListItem/ListItemWrapper.tsx
mdodgelooker: Scratch that, moved it to `MenuItem`! 😜

packages/components/src/Menu/MenuItem.tsx
```diff
@@ -41,86 +41,91 @@ export interface MenuItemProps
...
-    const { closeParentMenu } = useContext(NestedMenuContext)
-
-    const handleOnClick = (event: MouseEvent<HTMLLIElement>) => {
-      // nestedMenuOnClick wraps onClick from props
-      nestedMenuOnClick(event)
-      // Close the Menu unless event has preventDefault
-      if (!event.defaultPrevented) {
-        closeModal?.()
-        closeParentMenu?.()
+export const MenuItem = styled(
```
mdodgelooker, line 44: Just updating to the new wrap-it-in-a-styled pattern, plus the `aria-expanded` style down below.

mdodgelooker:
Ok sorry about that. I just re-did the `aria-expanded` style fix but this time without breaking lint or snapshots. 🙂

lukelooker (approved):
LGTM!

PR #2880, opened 2021-09-30
fix(FieldTime): Floating label position with partial value

mdodgelooker:
My original logic for `checkValueOnBlur` was wrong. I was checking if _all_ of the 3 inputs had a value but it should be if _any_ of them have a value.


lukelooker (dismissed):
👍

lukelooker (approved):
Fixed prettier out-of-date.

PR #2885, opened 2021-10-01
fix(Menu): Error when nesting 3 levels deep

mdodgelooker:
Though it wasn't specifically designed for it, nesting menus 3 levels deep mostly works, except for these issues when clicking on an item in the 3rd level menu:

1. There's an error from focus trap
2. Only the 2nd and 3rd menus close, not the 1st

This PR addresses both by addressing issue 2. 

Issue 1 is due to a complicated collision between how React bubbles events up through a portal, how we handle focus & blur events in `useArrowKeyNav`, and focus trap. Calling `stopPropagation` could fix this but might also have unforeseen consequences. So closing all the menus is the simpler solution.

### DIFF
--- a/packages/components/src/Menu/Menu.spec.tsx
+++ b/packages/components/src/Menu/Menu.spec.tsx
@@ -413,6 +413,45 @@ describe('<Menu />', () => {
       expect(onClickMock).toHaveBeenCalledTimes(1)
       expect(child).toBeVisible()
       expect(parent).toBeVisible()
+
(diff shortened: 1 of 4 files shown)


lukelooker (approved):
Works great. :)

Just because I had to know I tried adding a 4th level and that also appears to work properly. :)

mdodgelooker:
> Just because I had to know I tried adding a 4th level and that also appears to work properly. :)

Haha woohoo! 🎉

PR #2887, opened 2021-10-01
test: Lint configuration tweaks and fixes

lukelooker:
- Move type definitions closer to their use cases (within package source)
- Remove `eslintrc.js` within packages/components
  - Replace with eslintConfig within package.json
  - Effectively disables i18next/no-literal-string at this time (ticket to track)
- Fix use of generic `a target="_blank"` tags that was missing `noopener`


packages/components-date/tsconfig.build.json
```diff
@@ -4,5 +4,5 @@
     "rootDir": "./src",
     "outDir": "./lib"
   },
-  "include": ["../../types", "src/**/*"]
+  "include": ["src/**/*"]
```
lukelooker, line 7: Added types to each package so that packages are fully self-contained.

packages/components/package.json
```diff
@@ -47,6 +47,27 @@
...
+    "root": true,
+    "extends": [
+     "@looker/eslint-config-oss"
+    ],
+    "overrides": [
+      {
+        "files": ["*.tsx", "*.ts"],
+        "rules": {
+          "@typescript-eslint/no-explicit-any": "off",
+          "i18next/no-literal-string": "off"
```
lukelooker, line 60: This needs to get re-enabled in the future.

Ticket: https://b.corp.google.com/issues/201815746

packages/components/src/DataTable/DataTable.spec.tsx
```diff
@@ -107,7 +107,11 @@ const data = [
     calories: 103,
     id: 3,
     name: (
-      <a href="https://components.looker.com/" target="_blank" rel="noreferrer">
+      <a
+        href="https://components.looker.com/"
+        target="_blank"
+        rel="noopener noreferrer"
```
lukelooker, line 113: Lint fixes (core product has lint which catches the missing `noopener`)

mdodgelooker (approved):
LGTM!

PR #2888, opened 2021-10-02
fix(MultiFunctionButton): fix width bug, update test and story

laragoogle:

- [ ] 👾 Browsers tested
  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11


packages/components/src/Button/MultiFunctionButton.tsx
```diff
@@ -77,10 +77,21 @@ export const MultiFunctionButton = forwardRef(
...
-        setContainerWidth(Math.max(a.offsetWidth, b.offsetWidth, 0))
+        setContainerWidth(Math.max(aWidth, bWidth, 0))
       }
-    }, [containerHeight, containerWidth])
+    }, [
+      alternate.props.width,
+      children.props.width,
+      containerHeight,
+      containerWidth,
+    ])
```
mdodgelooker, line 94: This helps but it wouldn't capture a styled override example (see above). That might be admittedly rare, but actually I think you can fix the issue more simply with styles. `a.offsetWidth` is affected by the width of the container, which starts at 0, due to `display: flex` so it reflects a "squished" value. Meanwhile, the alternate has `position: absolute` so it does not get squished.

If you remove the above, then add the following to the style block, that should do it:
```
  > * {
    flex-shrink: 0;
  }
  ```

File: packages/components/src/Button/MultiFunctionButton.stories.tsx [hunk 2]
```diff
@@ -87,7 +87,11 @@ export const ComponentSize = () => {
       }
       swap={change}
     >
-      <CustomButtonOutline onClick={handleSwap} iconBefore={<Add />}>
+      <CustomButtonOutline
+        onClick={handleSwap}
+        iconBefore={<Add />}
+        width={700}
+      >
```
mdodgelooker, line 94: I tried adding an example with width set via styled block rather than a prop, and it still has the issue. I think a CSS fix might work better (see my comment below).
```
const ButtonA = styled(Button)`
  width: 400px;
`
const ButtonB = styled(Button)`
  width: 150px;
`

export const ComponentSize2 = () => {
  const [change, setChange] = useState(false)

  const handleSwap = () => {
    setChange(true)
  }
  return (
    <MultiFunctionButton
      alternate={<ButtonB size="large">Alternate</ButtonB>}
      swap={change}
    >
      <ButtonA onClick={handleSwap}>Initial</ButtonA>
    </MultiFunctionButton>
  )
}
```

mdodgelooker (changes requested):

mdodgelooker (approved):
LGTM!

PR #2901, opened 2021-10-04
fix: Restore deep-imports for @styled-icons imports (only in built artifacts)

lukelooker:
Partial revert of "refactor: Import icons from package boundary (remove deep imports)"

Re-introduces lint to prevent "shallow" (package entry-point) imports within production artifacts (`.ts` / `.tsx` files that are shipped in the built-JS artifacts) but allows internal usage (stories, tests) to continue using better import paths.

This should be reversible once the core-product has addressed remaining tree-shaking woes.


.eslintrc.js
```diff
@@ -24,9 +24,6 @@
 
  */
 
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const ignoreProps = require('./packages/eslint-config-oss/src/i18nIgnoredProps.js')
-
```
lukelooker, line 29: Unused import. Unrelated to this PR.

mdodgelooker (approved):
👍

PR #2903, opened 2021-10-04
chore: Clean-up dependency versions

lukelooker:
Clean-up disparate package versions and use root package.json to act as "central" location for package versions


mdodgelooker (commented):
Should `filter-demo` get upgraded to styled-components 5.3.1?
https://github.com/looker-open-source/components/blob/main/apps/filter-demo/package.json#L29

mdodgelooker (approved):
👍

PR #2904, opened 2021-10-05
chore: Jest mock consistency

mdodgelooker:
1. Match the `requestAnimationFrame` mock in jest setup with that in the core product.
2. Fix `paddingSpy` in `useScrollLock.spec.tsx` – not sure why it was working before but the core product wasn't having it.

I removed all existing instances of `jest.spyOn(window, 'requestAnimationFrame')`, since they are now redundant, except for two in `Combobox.spec` and `Select.spec`. Those will take some time to unwind so I want to save them for a follow-up (I created a ticket). In most cases an additional `runTimers()` call was necessary since the `requestAnimationFrame` mock uses a `setTimeout`.

### DIFF
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -44,6 +44,9 @@ globalAny.IntersectionObserver = observeMock
 // js-dom doesn't do scrollIntoView
 Element.prototype.scrollIntoView = jest.fn()
 
+// js-dom doesn't do requestAnimationFrame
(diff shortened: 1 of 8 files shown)


lukelooker (approved):
Very nice :)

mdodgelooker:
Oops sorry @lukelooker I thought I was sneaking that 2nd commit in before you saw this.

lukelooker (approved):

PR #2905, opened 2021-10-05
fix: Export type for SpecifiableTextColors

lukelooker:
Core product uses deep-import to pull in `SpecifiableTextColors` type currently. This will be problematic once code is shifted to core monorepo (and there's not really a good argument _not_ to go ahead and export it anyway) :)


laragoogle (approved):
LGTM🚢

PR #2908, opened 2021-10-05
refactor: Disable Typescript suppressImplicitAnyIndexErrors option and correct errors

lukelooker:


packages/design-tokens/src/utils/paddingDefaultsHelper.ts
```diff
@@ -66,12 +66,14 @@ export const paddingDefaultsHelper = (
     pt: py ? undefined : pt,
     px,
     py,
-  }
+  } as const
+
+  type responseKey = keyof typeof response
```
mdodgelooker, line 71: Are we strict about capitalizing type alias?
```suggestion
  type ResponseKey = keyof typeof response
```

mdodgelooker (commented):
Just a couple questions but overall looks good

packages/design-tokens/src/utils/paddingDefaultsHelper.ts
lukelooker: No but we probably should be :)

mdodgelooker (approved):
LGTM!

PR #2932, opened 2022-01-10
chore: Vis/Filter Demo

mdodgelooker:
Adds a new demo app for vis + filters.


guyellis:
While pairing with @mdodgelooker we agreed to close this.

PR #2933, opened 2022-01-11
Potential changes to Query

mdodgelooker:
Exploring a possible way to avoid duplicate calls to get query metadata in the Vis/Filter Demo app.

(There's currently a bug with this approach where if you try to re-submit the original query ID after filtering it, you still just get the filtered one.)

### DIFF
--- a/apps/vis-filter-demo/src/VisFilterDemo.tsx
+++ b/apps/vis-filter-demo/src/VisFilterDemo.tsx
@@ -134,9 +134,11 @@ const VisFilterDemoInternal: FC = () => {
         </Paragraph>
       )}
       {queryId && <Divider my="medium" />}
-      <Query query={queryId} sdk={core40SDK}>
-        <Visualization />
-      </Query>
+      {queryDetails && (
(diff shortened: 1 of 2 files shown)


dbchristopher (commented):
Thanks for sharing this @mdodgelooker! It totally makes sense that the vis components should be more portable -- that they could easily receive a query result from an external source. But to be honest, the only reason `<Query />` needs to be involved at all at this point is because of these few lines where we do some light data/config normalizing before adding that to the QueryContext:

https://github.com/looker-open-source/components/blob/78db77d8d2c8ed0758267f7a9cd7dbebcabe1c17/packages/visualizations-adapters/src/Query/Query.tsx#L351-L360

I would suggest that the `cleanup -> QueryContext` step should be extracted to another layer, and maybe we could come up with a Query component that could unify both filter and vis component systems? Or at least Query could be used if you don't currently have another request/response system set up

mdodgelooker:
Thanks for taking a look @dbchristopher, I love your ideas! Let's discuss in the components app dev sync this afternoon.

mdodgelooker:
No longer needed.

PR #2941, opened 2022-02-08
docs: Updates for 3.0 and more

mdodgelooker:
Updates docs for:

- Removed `Tree` props
- `themeCustomizations` including `brandAnimation` and `externalLabel`
- Other recently-added features


www/src/documentation/components/actions/link.mdx
```diff
@@ -16,3 +16,25 @@ The `<Link />` component renders an `<a>` tag that accepts and `href` property.
   </li>
 </UnorderedList>
 ```
+
+## Underline
+
+By default, the underline appears on hover, focus, active, and visited states. To disable the underline always, set the `underline` prop to `false`. To enable it always, set it to `true`.
```
heyelic, line 22: I would suggest rewording the second sentence a little.

To always disable the underline, set the `underline` prop to `false`. To always enable it, set the prop to `true`.

www/src/documentation/develop/theme.mdx
```diff
@@ -50,66 +50,75 @@ You can use the theme's properties within your custom components:
...
     </ComponentsProvider>
   )
 }
 ```
 
-## Nesting ComponentsProvider (ThemeProviders)
+The following colors are available for customization, the first three being the most likely to need to be customized:
+
+- key: Key color is applied strategically across the UI. Used for default action buttons, toggle switches, interactive component accents
+- text: Used for text
```
heyelic, line 74: The background specifies that it's a default. I think that makes sense for the text color as well. Something like "Default text color" is short and descriptive.

www/src/documentation/components/filters/filter-token.mdx
```diff
@@ -0,0 +1,19 @@
+---
+title: FilterToken
+github: filter-components/src/Token/FilterToken.tsx
+storybook: true
+storybookPath: Filters-FilterToken
+---
+
+<ImportMessage packageName="filter-components">Filter</ImportMessage>
```
ElliotTheHuman, line 8: Noticed that the `<ImportMessage>` ends up getting rendered as a floating string right above the paragraph below. Wasn't sure if it was intentional / supposed to have some margin underneath.
Screen Shot 2022-02-08 at 9 59 21 AM

www/src/documentation/develop/theme.mdx
```diff
@@ -50,66 +50,75 @@ You can use the theme's properties within your custom components:
 }
 ```
 
-## Extending the Theme
+## Customizing the Theme
 
-You can extend the theme provided by Looker Components by creating a copy and then providing that version of theme to the ThemeProvider.
+You can extend the theme provided by Looker Components by passing colors, fonts, and more into the `themeCustomizations` prop.
```
ElliotTheHuman, line 55: nit: Might be nice to remove the word "extend" from the opening sentence. Happy to keep it around though if preferable.
```suggestion
You can customize the theme provided by Looker Components by passing colors, fonts, and more into the `themeCustomizations` prop.
```

www/src/documentation/develop/theme.mdx
```diff
@@ -50,66 +50,75 @@ You can use the theme's properties within your custom components:
...
-      <PrettyButton>Pink is so Pretty!</PrettyButton>
+    <ComponentsProvider themeCustomizations={{ colors: { key: 'hotpink' } }}>
+      <Button>Pink is so Pretty!</Button>
     </ComponentsProvider>
   )
 }
 ```
 
-## Nesting ComponentsProvider (ThemeProviders)
+The following colors are available for customization, the first three being the most likely to need to be customized:
```
ElliotTheHuman, line 71: Super nit: Could be useful to give some context as to why we think these colors require customizing.
```suggestion
The following colors are available for customization. The first three are most likely to require customization since they are heavily used throughout the Looker Components library:
```

www/src/documentation/develop/theme.mdx
```diff
@@ -50,66 +50,75 @@ You can use the theme's properties within your custom components:
...
+
+- key: Key color is applied strategically across the UI. Used for default action buttons, toggle switches, interactive component accents
+- text: Used for text
+- background: Default page background used for application background
+
+These colors are less likely to need to be customized:
```
ElliotTheHuman, line 77: Makes this line a little more consistent with my last suggestion. Feel free to keep as is.
```suggestion
These colors are less likely to require customization:
```

www/src/documentation/develop/theme.mdx
```diff
@@ -50,66 +50,75 @@ You can use the theme's properties within your custom components:
...
+
+- pageBackground: Used as page background in some themes pageBackground and background have a subtle difference, wherein the pageBackground is closer to `ui1` but blends are calculated from a more "absolute" background color (white or black)
```
ElliotTheHuman, line 79:
```suggestion
- pageBackground: Used as page background in some themes. pageBackground and background have a subtle difference, wherein the pageBackground is closer to `ui1` but blends are calculated from a more "absolute" background color (white or black)
```

www/src/documentation/develop/theme.mdx
```diff
@@ -50,66 +50,75 @@ You can use the theme's properties within your custom components:
...
 
-`ComponentsProvider` as a Styled Components' ThemeProvider acts as a React Context and in so doing is able to support the standard behavior of nested contexts - the theme defined nearest to the component will be the version used.
+The following font family options may be customized (use the `fontSources` property to specify the sources):
 
-NOTE: You'll also see that `PrettyButton` in the example simply falls back to `Button`'s default style when `myCustomAttribute` isn't assigned.
+- body: Used for body text
+- brand: Used for headings
+- code: Used for code snippets
+
+Additionaly properties to customize:
```
ElliotTheHuman, line 97:
```suggestion
Additional properties to customize:
```

www/src/documentation/develop/theme.mdx
```diff
@@ -50,66 +50,75 @@ You can use the theme's properties within your custom components:
...
+
+- brandAnimation: The Google Material ripple effect for pressable elements
+- externalLabel: True sets the field label style to external, false sets it to the Google Material floating label style
+
+## Nesting ComponentsProvider (ThemeProviders)
+
+`ExtendComponentsThemeProvider` as a Styled Components' ThemeProvider acts as a React Context and in so doing is able to support the standard behavior of nested contexts - the theme defined nearest to the component will be the version used.
```
ElliotTheHuman, line 104: I _think_ that's supposed to be an "is" but I could be totally off. Wanted to point it out.
```suggestion
`ExtendComponentsThemeProvider` is a Styled Components' ThemeProvider that acts as a React Context and in so doing is able to support the standard behavior of nested contexts - the theme defined nearest to the component will be the version used.
```

ElliotTheHuman (changes requested):
Had a couple nits, but nothing crazy. Let me know if any of my suggestions don't make sense / are better left as is.

www/src/documentation/components/actions/link.mdx
mdodgelooker: ✅

www/src/documentation/components/filters/filter-token.mdx
mdodgelooker: Ah whoops! Something must've changed at some point and now `ImportMessage` is not in scope for the mdx files. So I added imports to each one that uses it.

www/src/documentation/develop/theme.mdx
mdodgelooker: ✅

www/src/documentation/components/content/tree.mdx
```diff
@@ -82,62 +82,6 @@ Use the `disabled` prop to present your `Tree` in a disabled state. Note that us
 </TreeCollection>
 ```
 
-## Changing Background Behavior
```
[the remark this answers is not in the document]
mdodgelooker: 👍

ElliotTheHuman (approved):
🥳 🥳 🥳

PR #2959, opened 2022-05-23
chore: Release 05/23/2022

dbchristopher:


package.json
```diff
@@ -39,7 +39,8 @@
     "deduplicate": "yarn-deduplicate -s fewer yarn.lock",
     "playground": "yarn workspace playground develop",
     "gatsby": "yarn workspace www start",
-    "fix": "yarn lint:es --fix",
+    "fix": "yarn lint:es --fix && yarn prettify-fix",
+    "prettify-fix": "yarn prettier --config='.prettierrc.json' packages/* --write",
```
dbchristopher, line 43: I found myself needing to manually run `yarn prettier` to be able to get the precommit hooks to run. I figured that should probably just be an included step in the `yarn fix` command

packages/components/package.json
```diff
@@ -56,6 +58,15 @@
...
     "overrides": [
+      {
+        "files": [
+          "*.mdx",
+          "*.md"
+        ],
+        "rules": {
+          "import/no-webpack-loader-syntax": "off"
+        }
+      },
```
dbchristopher, line 69: @treshugart I disabled that webpack loader rule for mdx files. It is weird that it didn't get flagged in our source repo, but I can duplicate this change over there for good measure

treshugart (approved):

package.json
treshugart: Should this be duplicated in our source repo?

dbchristopher: No need! The root package.json already has a `prettify-fix` script

PR #3002, opened 2025-03-26
fix(components-data): ensure all IDs are strings instead of numbers

### LINKED ISSUE #2997: Support for new endpoint query ID behavior
We recently received the notice regarding changes to how [query IDs are handled](https://cloud.google.com/looker/docs/best-practices/query-id-update-instructions). It appears that the SDK uses numeric query IDs for calls to the affected endpoints in some instances, such as https://github.com/looker-open-source/components/blob/bff6103cf5b83046d216e0d009e66e13ce2d5701/packages/components-data/src/hooks/useQueryMetadata.ts#L71

Are there plans to release an updated SDK that's compatible with these changes?

Thanks!

pjmsmithgoogle:
Since all IDs in API 4.0 are strings and query IDs now cannot be numeric, this change removes any number casting for IDs and sets the type for all dashboard and query IDs to `string` from `number`.

Fixes #2997

### DIFF
--- a/packages/components-data/src/hooks/useExplore.spec.tsx
+++ b/packages/components-data/src/hooks/useExplore.spec.tsx
@@ -39,7 +39,7 @@ type TestComponentProps = {
   id?: string;
 };
 
-const TestComponent = ({ id = 1 }: TestComponentProps) => {
+const TestComponent = ({ id = '1' }: TestComponentProps) => {
   const response = useExplore(id);
   dataContainerListener(response);
   return null;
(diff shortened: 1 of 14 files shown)


jkaster (dismissed):
LGTM

aschott-looker (approved):
TY!

packages/components-data/src/hooks/useExplore.spec.tsx
```diff
@@ -36,7 +36,7 @@ import { useExplore } from './useExplore';
 const dataContainerListener = jest.fn();
 
 type TestComponentProps = {
-  id?: number;
+  id?: string;
 };
 
 const TestComponent = ({ id = 1 }: TestComponentProps) => {
```
pjmsmithgoogle, line 42: Whoops, missed a number.

jkaster (approved):

pjmsmithgoogle:
This fix will get copied over eventually so need to merge this CL.

